PR #39, opened 2015-03-09
New file format

iyer-arvind:
This branches changes the file format form npy to hdf5. The reader is still capable of accepting the files in the old format. The current version provides the parallel feature of HDF5 when available and suitable, however this can be over-ridden by the "serial-h5 = true" option under [soln-output]. The "format = pyfrs-file" option under the same section, is no longer used.

### DIFF
--- a/pyfr/integrators/__init__.py
+++ b/pyfr/integrators/__init__.py
@@ -4,7 +4,7 @@ import re
 
 from pyfr.integrators.controllers import BaseController
 from pyfr.integrators.steppers import BaseStepper
-from pyfr.integrators.writers import BaseWriter
+from pyfr.integrators.writers import H5Writer
 from pyfr.util import subclass_where
 
 
@@ -12,11 +12,11 @@ def get_integrator(backend, systemcls, rallocs, mesh, initsoln, cfg):
     # Look-up the controller, stepper and writer names
     c = cfg.get('solver-time-integrator', 'controller')
     s = cfg.get('solver-time-integrator', 'scheme')
-    w = cfg.get('soln-output', 'format')
+    w = 'HDF5'
 
     controller = subclass_where(BaseController, controller_name=c)
     stepper = subclass_where(BaseStepper, stepper_name=s)
-    writer = subclass_where(BaseWriter, writer_name=w)
+    writer = H5Writer
 
     # Determine the integrator name
     name = '_'.join([c, s, w])
--- a/pyfr/integrators/writers.py
+++ b/pyfr/integrators/writers.py
@@ -1,17 +1,16 @@
 # -*- coding: utf-8 -*-
 
-from abc import abstractmethod
 import itertools as it
 import os
 
+import h5py
 import numpy as np
 
 from pyfr.integrators.base import BaseIntegrator
 from pyfr.mpiutil import get_comm_rank_root
-from pyfr.util import rm
 
 
-class BaseWriter(BaseIntegrator):
+class H5Writer(BaseIntegrator):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
@@ -22,13 +21,67 @@ class BaseWriter(BaseIntegrator):
         # Output counter (incremented each time output() is called)
         self.nout = 0
 
+        # MPI info
+        comm, rank, root = get_comm_rank_root()
+
+        # Get the type and shape of each element in the partition
+        etypes, shapes = self.system.ele_types, self.system.ele_shapes
+
+        # Gather this information onto the root rank
+        eleinfo = comm.gather(zip(etypes, shapes), root=root)
+
+        # Deciding if parallel
+        parallel = h5py.get_config().mpi
+        parallel &= h5py.version.version_tuple[:2] >= (2, 5)
+        parallel &= not self.cfg.getbool('soln-output', 'serial-h5', False)
+
+        if parallel:
+            self._write = self._write_parallel
+
+            if rank == root:
+                sollist = []
+                for mrank, meleinfo in enumerate(eleinfo):
+                    prank = self.rallocs.mprankmap[mrank]
+                    sollist.extend(
+                        (self._get_name_for_soln(etype, prank), dims)
+                        for etype, dims in meleinfo
+                    )
+
+            else:
+                sollist = None
+
+            self.sollist = comm.bcast(sollist, root=root)
+        else:
+            self._write = self._write_serial
+
+            if rank == root:
+                self._mpi_rbufs = mpi_rbufs = []
+                self._mpi_rreqs = mpi_rreqs = []
+                self._mpi_names = mpi_names = []
+                self._loc_names = loc_names = []
+
+                for mrank, meleinfo in enumerate(eleinfo):
+                    prank = self.rallocs.mprankmap[mrank]
+                    for tag, (etype, dims) in enumerate(meleinfo):
+                        name = self._get_name_for_soln(etype, prank)
+
+                        if mrank == root:
+                            loc_names.append(name)
+                        else:
+                            rbuf = np.empty(dims, dtype=self.backend.fpdtype)
+                            rreq = comm.Recv_init(rbuf, mrank, tag)
+
+                            mpi_rbufs.append(rbuf)
+                            mpi_rreqs.append(rreq)
+                            mpi_names.append(name)
+
     def output(self, solnmap, stats):
         comm, rank, root = get_comm_rank_root()
 
         # Convert the config and stats objects to strings
         if rank == root:
-            metadata = dict(config=self.cfg.tostr().encode(),
-                            stats=stats.tostr().encode(),
+            metadata = dict(config=self.cfg.tostr(),
+                            stats=stats.tostr(),
                             mesh_uuid=self._mesh_uuid)
         else:
             metadata = None
@@ -41,10 +94,6 @@ class BaseWriter(BaseIntegrator):
 
         # Increment the output number
         self.nout += 1
-
-    @abstractmethod
-    def _write(self, path, solnmap, metadata):
-        pass
 
     def _get_output_path(self):
         # Substitute %(t) and %(n) for the current time and output number
@@ -60,44 +109,34 @@ class BaseWriter(BaseIntegrator):
         prank = prank or self.rallocs.prank
         return 'soln_{}_p{}'.format(etype, prank)
 
-
-class FileWriter(BaseWriter):
-    writer_name = 'pyfrs-file'
-
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-
-        # MPI info
+    def _write_parallel(self, path, solnmap, metadata):
         comm, rank, root = get_comm_rank_root()
 
-        # Get the type and shape of each element in the partition
-        etypes, shapes = self._system.ele_types, self._system.ele_shapes
+        with h5py.File(path, 'w', driver='mpio', comm=comm) as h5file:
+            smap = {}
+            for s, shape in self.sollist:
+                smap[s] = h5file.create_dataset(
+                    s, shape, dtype=self.backend.fpdtype
+                )
 
-        # Gather this information onto the root rank
-        eleinfo = comm.gather(zip(etypes, shapes), root=root)
+            for e, sol in solnmap.items():
+                s = self._get_name_for_soln(e, self.rallocs.prank)
+                smap[s][:] = sol
 
-        if rank == root:
-            self._mpi_rbufs = mpi_rbufs = []
-            self._mpi_rreqs = mpi_rreqs = []
-            self._mpi_names = mpi_names = []
-            self._loc_names = loc_names = []
+            # Metadata information has to be transferred to all the ranks
+            if rank == root:
+                mmap = [(k, len(v.encode()))
+                        for k, v in metadata.items()]
+            else:
+                mmap = None
 
-            for mrank, meleinfo in enumerate(eleinfo):
-                prank = self.rallocs.mprankmap[mrank]
-                for tag, (etype, dims) in enumerate(meleinfo):
-                    name = self._get_name_for_soln(etype, prank)
+            mmap = comm.bcast(mmap, root=root)
+            for name, size in mmap:
+                d = h5file.create_dataset(name, (), dtype='S{}'.format(size))
+                if rank == root:
+                    d.write_direct(np.array(metadata[name]).astype('S'))
 
-                    if mrank == root:
-                        loc_names.append(name)
-                    else:
-                        rbuf = np.empty(dims, dtype=self.backend.fpdtype)
-                        rreq = comm.Recv_init(rbuf, mrank, tag)
-
-                        mpi_rbufs.append(rbuf)
-                        mpi_rreqs.append(rreq)
-                        mpi_names.append(name)
-
-    def _write(self, path, solnmap, metadata):
+    def _write_serial(self, path, solnmap, metadata):
         from mpi4py import MPI
 
         comm, rank, root = get_comm_rank_root()
@@ -117,31 +156,6 @@ class FileWriter(BaseWriter):
             # Create the output dictionary
             outdict = dict(zip(names, solns), **metadata)
 
-            with open(path, 'wb') as f:
-                np.savez(f, **outdict)
-
-
-class DirWriter(BaseWriter):
-    writer_name = 'pyfrs-dir'
-
-    def _write(self, path, solnmap, metadata):
-        comm, rank, root = get_comm_rank_root()
-
-        # Create the output directory and save the config/status files
-        if rank == root:
-            if os.path.exists(path):
-                rm(path)
-
-            os.mkdir(path)
-
-            # Write out our metadata
-            for name, data in metadata.items():
-                np.save(os.path.join(path, name), data)
-
-        # Wait for this to complete
-        comm.barrier()
-
-        # Save the solutions
-        for etype, buf in solnmap.items():
-            solnpath = os.path.join(path, self._get_name_for_soln(etype))
-            np.save(solnpath, buf)
+            with h5py.File(path, 'w') as h5file:
+                for k, v in outdict.items():
+                    h5file.create_dataset(k, data=v)
--- a/pyfr/partitioners/base.py
+++ b/pyfr/partitioners/base.py
@@ -42,7 +42,7 @@ class BasePartitioner(object):
                     spts[en].append(mesh['spt_{0}_p{1}'.format(en, i)])
 
         def offset_con(con, pr):
-            con = con.copy()
+            con = con.copy().astype('U4,i4,i1,i1')
 
             for en, pn in pinf.items():
                 if pn[pr] > 0:
@@ -103,7 +103,7 @@ class BasePartitioner(object):
 
         # Sort by the left hand side
         idx = np.lexsort([con['f0'][0], con['f1'][0]])
-        con = con[:,idx]
+        con = con[:, idx]
 
         # Left and right hand side element types/indicies
         lhs, rhs = con[['f0', 'f1']]
@@ -138,7 +138,7 @@ class BasePartitioner(object):
         # Partition the shape points
         spt_px = defaultdict(list)
         for (etype, eidxg), part in zip(vetimap, vparts):
-            spt_px[etype, part].append(spt_p0[etype][:,eidxg,:])
+            spt_px[etype, part].append(spt_p0[etype][:, eidxg, :])
 
         # Stack
         return {'spt_{0}_p{1}'.format(*k): np.array(v).swapaxes(0, 1)
@@ -154,7 +154,7 @@ class BasePartitioner(object):
         # Partition the solutions
         soln_px = defaultdict(list)
         for (etype, eidxg), part in zip(vetimap, vparts):
-            soln_px[etype, part].append(soln_p0[etype][...,eidxg])
+            soln_px[etype, part].append(soln_p0[etype][..., eidxg])
 
         # Stack
         return {'soln_{0}_p{1}'.format(*k): np.dstack(v)
--- a/pyfr/readers/base.py
+++ b/pyfr/readers/base.py
@@ -1,10 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from abc import ABCMeta, abstractmethod
-import re
 import uuid
-
-import numpy as np
 
 
 class BaseReader(object, metaclass=ABCMeta):
--- a/pyfr/readers/native.py
+++ b/pyfr/readers/native.py
@@ -3,15 +3,16 @@
 """Allows for interopability between .pyfr{m, s}-{file, dir} archive formats
 
 """
-from collections import Mapping, OrderedDict
 import errno
 import os
 import re
 
+from collections import Mapping, OrderedDict
+import h5py
 import numpy as np
 
 from pyfr.shapes import BaseShape
-from pyfr.util import subclasses, lazyprop
+from pyfr.util import lazyprop, subclasses
 
 
 class PyFRBaseReader(Mapping):
@@ -136,8 +137,13 @@ class PyFRDirReader(PyFRBaseReader):
 
     def __getitem__(self, aname):
         try:
-            return np.load(os.path.join(self.fname, aname + '.npy'),
-                           mmap_mode='r')
+            ret = np.load(os.path.join(self.fname, aname + '.npy'),
+                          mmap_mode='r')
+
+            ret = ret.item() if ret.dtype.kind in 'SU' else ret
+
+            return ret.decode() if isinstance(ret, bytes) else ret
+
         except IOError as e:
             if e.errno == errno.ENOENT:
                 raise KeyError
@@ -157,13 +163,39 @@ class PyFRFileReader(PyFRBaseReader):
         self._npf = np.load(fname)
 
     def __getitem__(self, aname):
-        return self._npf[aname]
+        ret = self._npf[aname]
+
+        ret = ret.item() if ret.dtype.kind in 'SU' else ret
+
+        return ret.decode() if isinstance(ret, bytes) else ret
 
     def __iter__(self):
         return iter(self._npf.files)
 
     def __len__(self):
         return len(self._npf.files)
+
+
+class PyFRH5Reader(PyFRBaseReader):
+    """Does everything that file reader does but with a h5py file"""
+    def __init__(self, fname):
+        self._file = h5py.File(fname, 'r')
+
+    def __getitem__(self, aname):
+        ret = self._file[aname]
+
+        if ret.shape == ():
+            ret = ret.value
+        else:
+            ret = np.array(ret)
+
+        return ret.decode() if isinstance(ret, bytes) else ret
+
+    def __iter__(self):
+        return iter(self._file)
+
+    def __len__(self):
+        return len(self._file)
 
 
 def read_pyfr_data(fname):
@@ -207,4 +239,7 @@ def read_pyfr_data(fname):
     if os.path.isdir(fname):
         return PyFRDirReader(fname)
     else:
-        return PyFRFileReader(fname)
+        try:
+            return PyFRFileReader(fname)
+        except OSError:
+            return PyFRH5Reader(fname)
--- a/pyfr/scripts/mesh.py
+++ b/pyfr/scripts/mesh.py
@@ -4,10 +4,10 @@ from argparse import FileType
 import itertools as it
 import os
 
-import numpy as np
+import h5py
 
-from pyfr.partitioners import BasePartitioner, get_partitioner_by_name
-from pyfr.readers import BaseReader, get_reader_by_name, get_reader_by_extn
+from pyfr.partitioners import BasePartitioner, get_partitioner
+from pyfr.readers import BaseReader, get_reader_by_extn, get_reader_by_name
 from pyfr.readers.native import read_pyfr_data
 from pyfr.util import subclasses
 
@@ -19,7 +19,7 @@ def add_args(ap):
     ap_convert = sp.add_parser('convert', help='convert --help')
     ap_convert.add_argument('inmesh', type=FileType('r'),
                             help='input mesh file')
-    ap_convert.add_argument('outmesh', type=FileType('wb'),
+    ap_convert.add_argument('outmesh', type=str,
                             help='output PyFR mesh file')
     types = sorted(cls.name for cls in subclasses(BaseReader))
     ap_convert.add_argument('-t', dest='type', choices=types,
@@ -56,7 +56,9 @@ def process_convert(args):
     mesh = reader.to_pyfrm()
 
     # Save to disk
-    np.savez(args.outmesh, **mesh)
+    with h5py.File(args.outmesh, 'w') as msh5:
+        for k, v in mesh.items():
+            msh5.create_dataset(k, data=v)
 
 
 def process_partition(args):
@@ -75,11 +77,11 @@ def process_partition(args):
 
     # Create the partitioner
     if args.partitioner:
-        part = get_partitioner_by_name(args.partitioner, pwts, opts)
+        part = get_partitioner(args.partitioner, pwts, opts)
     else:
         for name in sorted(cls.name for cls in subclasses(BasePartitioner)):
             try:
-                part = get_partitioner_by_name(name, pwts)
+                part = get_partitioner(name, pwts)
                 break
             except RuntimeError:
                 pass
@@ -102,5 +104,7 @@ def process_partition(args):
         path = os.path.join(args.outd, os.path.basename(path.rstrip('/')))
 
         # Open and save
-        with open(path, 'wb') as f:
-            np.savez(f, **data)
+        # Save to disk
+        with h5py.File(path, 'w') as msh5:
+            for k, v in data.items():
+                msh5.create_dataset(k, data=v)
--- a/pyfr/writers/base.py
+++ b/pyfr/writers/base.py
@@ -33,4 +33,4 @@ class BaseWriter(object):
                                (args.solnf, args.meshf))
 
         # Load config file
-        self.cfg = Inifile(self.soln['config'].item().decode())
+        self.cfg = Inifile(self.soln['config'])
--- a/setup.py
+++ b/setup.py
@@ -31,6 +31,7 @@ modules = [
     'pyfr.backends.openmp',
     'pyfr.backends.openmp.kernels',
     'pyfr.integrators',
+    'pyfr.plugins',
     'pyfr.quadrules',
     'pyfr.readers',
     'pyfr.partitioners',
@@ -87,6 +88,7 @@ data_files = [
 
 # Hard dependencies
 install_requires = [
+    'h5py >= 2.4',
     'mako',
     'mpi4py >= 1.3',
     'mpmath >= 0.18',
--- a/doc/ext/tikz.py
+++ b/doc/ext/tikz.py
@@ -145,7 +145,7 @@ def render_tikz(self,tikz,libs='',stringsubst=False):
     if stringsubst:
         tikz = tikz % {'wd': curdir}
     latex += DOC_BODY % tikz
-    if isinstance(latex, unicode):
+    if isinstance(latex, str):
         latex = latex.encode('utf-8')
 
     if not hasattr(self.builder, '_tikz_tempdir'):
@@ -163,7 +163,7 @@ def render_tikz(self,tikz,libs='',stringsubst=False):
         try:
             p = Popen(['pdflatex', '--interaction=nonstopmode', 'tikz.tex'],
                       stdout=PIPE, stderr=PIPE)
-        except OSError, err:
+        except OSError as err:
             if err.errno != ENOENT:   # No such file or directory
                 raise
             self.builder.warn('LaTeX command cannot be run')
@@ -187,7 +187,7 @@ def render_tikz(self,tikz,libs='',stringsubst=False):
     try:
         p = Popen(['pdftoppm', '-r', '120', 'tikz.pdf', 'tikz'],
                   stdout=PIPE, stderr=PIPE)
-    except OSError, e:
+    except OSError as e:
         if e.errno != ENOENT:   # No such file or directory
             raise
         self.builder.warn('pdftoppm command cannot be run')
@@ -210,7 +210,7 @@ def render_tikz(self,tikz,libs='',stringsubst=False):
             p1 = Popen(['convert', '-trim'] + convert_args +
                        ['tikz-1.ppm', outfn],
                        stdout=PIPE, stderr=PIPE)
-        except OSError, e:
+        except OSError as e:
             if e.errno != ENOENT:   # No such file or directory
                 raise
             self.builder.warn('convert command cannot be run')
@@ -229,7 +229,7 @@ def render_tikz(self,tikz,libs='',stringsubst=False):
     elif self.builder.config.tikz_proc_suite == 'Netpbm':
         try:
             p1 = Popen(['pnmcrop', 'tikz-1.ppm'], stdout=PIPE, stderr=PIPE)
-        except OSError, err:
+        except OSError as err:
             if err.errno != ENOENT:   # No such file or directory
                 raise
             self.builder.warn('pnmcrop command cannot be run:')
@@ -245,7 +245,7 @@ def render_tikz(self,tikz,libs='',stringsubst=False):
         try:
             p2 = Popen(['pnmtopng'] + pnm_args, stdin=p1.stdout,
                        stdout=PIPE, stderr=PIPE)
-        except OSError, err:
+        except OSError as err:
             if err.errno != ENOENT:   # No such file or directory
                 raise
             self.builder.warn('pnmtopng command cannot be run:')
@@ -282,7 +282,7 @@ def html_visit_tikzinline(self,node):
     libs = libs.replace(' ', '').replace('\t', '').strip(', ')
     try:
         fname = render_tikz(self,node['tikz'],libs);
-    except TikzExtError, exc:
+    except TikzExtError as exc:
         info = str(exc)[str(exc).find('!'):-1]
         sm = nodes.system_message(info, type='WARNING', level=2,
                                   backrefs=[], source=node['tikz'])
@@ -304,7 +304,7 @@ def html_visit_tikz(self,node):
 
     try:
         fname = render_tikz(self,node['tikz'],libs,node['stringsubst'])
-    except TikzExtError, exc:
+    except TikzExtError as exc:
         info = str(exc)[str(exc).find('!'):-1]
         sm = nodes.system_message(info, type='WARNING', level=2,
                                   backrefs=[], source=node['tikz'])
--- a/doc/src/user_guide.rst
+++ b/doc/src/user_guide.rst
@@ -244,19 +244,20 @@ Parameterises the OpenMP backend with
 
     *string*
 
-2. ``cblas-st`` --- path to shared single-threaded C BLAS library
+2. ``cblas`` --- path to shared C BLAS library
 
     *string*
 
-3. ``cblas-mt`` --- path to shared multi-threaded C BLAS library
+3. ``cblas-type`` --- type of BLAS library
 
-    *string*
+    ``serial`` | ``parallel``
 
 Example::
 
     [backend-openmp]
     cc = gcc
-    cblas-mt = example/path/libBLAS.dylib
+    cblas= example/path/libBLAS.dylib
+    cblas-type = parallel
 
 [constants]
 ^^^^^^^^^^^
@@ -655,20 +656,17 @@ Example::
 
 Parameterises the output with
 
-1. ``format`` --- format of the outputs:
 
-    ``pyfrs-file`` | ``pyfrs-dir``
-
-2. ``basedir`` --- relative path to directory where outputs will be
+1. ``basedir`` --- relative path to directory where outputs will be
    written
 
     *string*
 
-3. ``basename`` --- pattern of output names
+2. ``basename`` --- pattern of output names
 
     *string*
 
-4. ``times`` --- times at which outputs will be dumped
+3. ``times`` --- times at which outputs will be dumped
 
     ``range(`` *float* ``,`` *float* ``,`` *int* ``)``
 
@@ -700,6 +698,42 @@ Parameterises an exponential solution filter with
 4. ``cutoff`` --- cutoff frequency below which no filtering is applied:
 
     *int*
+
+[soln-plugin-nancheck]
+^^^^^^^^^^^^^^^^^^^^^^
+
+Periodically checks the solution for NaN values
+
+1. ``freq`` --- frequency at which the check is performed:
+
+    *int*
+
+[soln-plugin-sampler]
+^^^^^^^^^^^^^^^^^^^^^
+
+Periodically samples specific points in the volume and writes them out
+to a CSV file.
+
+1. ``freq`` --- frequency at which to sample:
+
+    *int*
+
+2. ``samp-pts`` --- list of points to sample:
+
+    ``[(x, y), (x, y), ...]`` | ``[(x, y, z), (x, y, z), ...]``
+
+3. ``format`` --- output variable format:
+
+    ``primitive`` | ``conservative``
+
+4. ``file`` --- output file path; should the file already exist it
+   will be appended to:
+
+    *string*
+
+5. ``header`` --- if to output a header row or not:
+
+    *boolean*
 
 [soln-bcs-name]
 ^^^^^^^^^^^^^^^
--- a/examples/couette_flow_2d/couette_flow_2d.ini
+++ b/examples/couette_flow_2d/couette_flow_2d.ini
@@ -40,8 +40,10 @@ soln-pts = williams-shunn
 [solver-elements-quad]
 soln-pts = gauss-legendre
 
+[solver-plugin-nancheck]
+freq = 50
+
 [soln-output]
-format = pyfrs-file
 basedir = .
 basename = couette_flow_2d_%(t).2f
 times = range(0, 4, 41)
--- a/examples/euler_vortex_2d/euler_vortex_2d.ini
+++ b/examples/euler_vortex_2d/euler_vortex_2d.ini
@@ -33,7 +33,6 @@ flux-pts = gauss-legendre
 soln-pts = gauss-legendre
 
 [soln-output]
-format = pyfrs-file
 basedir = .
 basename = euler_vortex_2d-%(t).1f
 times = range(0.0, 100, 11)
--- a/examples/euler_vortex_2d/euler_vortex_2d.ini
+++ b/examples/euler_vortex_2d/euler_vortex_2d.ini
@@ -32,6 +32,9 @@ flux-pts = gauss-legendre
 [solver-elements-quad]
 soln-pts = gauss-legendre
 
+[solver-plugin-nancheck]
+freq = 50
+
 [soln-output]
 basedir = .
 basename = euler_vortex_2d-%(t).1f
--- a/pyfr/backends/base/generator.py
+++ b/pyfr/backends/base/generator.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-from abc import ABCMeta, abstractmethod, abstractproperty
+from abc import ABCMeta, abstractmethod
 import re
 
 import numpy as np
--- a/pyfr/backends/cuda/cublas.py
+++ b/pyfr/backends/cuda/cublas.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-from ctypes import CDLL, POINTER, c_int, c_double, c_float, c_void_p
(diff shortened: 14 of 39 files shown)


pyfr/integrators/writers.py
```diff
@@ -10,6 +10,8 @@
 from pyfr.mpiutil import get_comm_rank_root
 from pyfr.util import rm
 
+import h5py
```
FreddieWitherden: Import in the wrong place.  We (with one or two exceptions) go Python imports, other package imports, PyFR imports with there being one line break in between sections.  Import names are sorted alphabetically.  See other PyFR as a reference.

pyfr/integrators/writers.py
```diff
@@ -61,43 +63,106 @@ def _get_name_for_soln(self, etype, prank=None):
...
 
 
-class FileWriter(BaseWriter):
-    writer_name = 'pyfrs-file'
+class H5Writer(BaseWriter):
+    writer_name = 'pyfrs-h5-file'
 
     def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+        super(H5Writer, self).__init__(*args, **kwargs)
```
FreddieWitherden: Outdated use of super; use `super()`.

pyfr/integrators/writers.py
```diff
@@ -61,43 +63,106 @@ def _get_name_for_soln(self, etype, prank=None):
...
 
         # MPI info
         comm, rank, root = get_comm_rank_root()
 
         # Get the type and shape of each element in the partition
         etypes, shapes = self._system.ele_types, self._system.ele_shapes
-
```
FreddieWitherden: Random newline change.

pyfr/integrators/writers.py
```diff
@@ -61,43 +63,106 @@ def _get_name_for_soln(self, etype, prank=None):
...
-                        rbuf = np.empty(dims, dtype=self.backend.fpdtype)
-                        rreq = comm.Recv_init(rbuf, mrank, tag)
-
-                        mpi_rbufs.append(rbuf)
-                        mpi_rreqs.append(rreq)
-                        mpi_names.append(name)
+        # Deciding if parallel
+        parallel = h5py.get_config().mpi
+        parallel &= h5py.version.version_tuple[0] >= 2
+        parallel &= h5py.version.version_tuple[1] >= 5
```
FreddieWitherden: Use tuple comparison.

pyfr/integrators/writers.py
```diff
@@ -61,43 +63,106 @@ def _get_name_for_soln(self, etype, prank=None):
...
+        parallel &= not self.cfg.getbool('soln-output', 'serial-h5', False)
+
+        if(parallel):
```
FreddieWitherden: Not PEP8.  Please re-read.

pyfr/integrators/writers.py
```diff
@@ -61,43 +63,106 @@ def _get_name_for_soln(self, etype, prank=None):
...
+            self._write = self.__writeParallel
+        else:
+            self._write = self.__writeSerial
+
+        if (parallel):
```
FreddieWitherden: Not PEP8 (and not even consistent with the previous `if` statement format!).  Please re-read PEP8.

pyfr/integrators/writers.py
```diff
@@ -61,43 +63,106 @@ def _get_name_for_soln(self, etype, prank=None):
...
+            if rank == root:
+                solList = []
+                for mrank, meleinfo in enumerate(eleinfo):
+                    prank = self.rallocs.mprankmap[mrank]
+                    solList.extend(
+                            [
+                                (self._get_name_for_soln(etype, prank), dims)
+                                for (etype, dims) in meleinfo
```
FreddieWitherden: Unnecessary parenthesis.

pyfr/integrators/writers.py
```diff
@@ -61,43 +63,106 @@ def _get_name_for_soln(self, etype, prank=None):
...
+        else:
+            self._write = self.__writeSerial
+
+        if (parallel):
+            if rank == root:
+                solList = []
+                for mrank, meleinfo in enumerate(eleinfo):
+                    prank = self.rallocs.mprankmap[mrank]
+                    solList.extend(
+                            [
```
FreddieWitherden: Unnecessary list; `.extend` supports any iterable.

pyfr/integrators/writers.py
```diff
@@ -61,43 +63,106 @@ def _get_name_for_soln(self, etype, prank=None):
...
+                            rreq = comm.Recv_init(rbuf, mrank, tag)
+
+                            mpi_rbufs.append(rbuf)
+                            mpi_rreqs.append(rreq)
+                            mpi_names.append(name)
 
     def _write(self, path, solnmap, metadata):
+        raise Exception
+
+    def __writeParallel(self, path, solnmap, metadata):
```
FreddieWitherden: Unnecessary use of `__` (super-private) member.

pyfr/integrators/writers.py
```diff
@@ -61,43 +63,106 @@ def _get_name_for_soln(self, etype, prank=None):
...
+        mMap = comm.bcast(mMap, root=root)
+        for name, shape, dtype in mMap:
+            d = _file.create_dataset(name, shape, dtype=dtype)
+            if(rank == root):
+                if(isinstance(metadata[m], str)):
+                    d.write_direct(np.array(metadata[name]).astype('S'))
+
+        _file.close()
+
+    def __writeSerial(self, path, solnmap, metadata):
```
FreddieWitherden: Unnecessary use of `__` (super-private) member.  The above three functions can be simplified by an assignment in `__init__`:

```
if parallel:
    self._writer = self._writer_parallel
else:
    self._writer = self._writer_serial
```

pyfr/integrators/writers.py
```diff
@@ -117,31 +182,7 @@ def _write(self, path, solnmap, metadata):
...
-                np.save(os.path.join(path, name), data)
-
-        # Wait for this to complete
-        comm.barrier()
-
-        # Save the solutions
-        for etype, buf in solnmap.items():
-            solnpath = os.path.join(path, self._get_name_for_soln(etype))
-            np.save(solnpath, buf)
+            _file = h5py.File(path, "w")
```
FreddieWitherden: Random use of `_`.  Use of double quotes.  (PyFR uses single quotes except when the string itself contains a single quote.)

pyfr/integrators/writers.py
iyer-arvind: Fixed

pyfr/readers/native.py
```diff
@@ -165,6 +165,21 @@ def __iter__(self):
     def __len__(self):
         return len(self._npf.files)
 
+class PyFRH5Reader(PyFRBaseReader):
+    """Does everything that file reader does but with a h5py file"""
+    def __init__(self,fname):
+        import h5py
```
FreddieWitherden: Local import for a globally required package.

pyfr/readers/native.py
```diff
@@ -165,6 +165,21 @@ def __iter__(self):
...
+        self._file = h5py.File(fname,'r')
+
+    def __getitem__(self,aname):
+        return np.array(self._file[aname])
+
+    def __iter__(self):
+        return iter(self._file.keys())
```
FreddieWitherden: Probably unnecessary use of `.keys()`.

pyfr/readers/native.py
```diff
@@ -165,6 +165,21 @@ def __iter__(self):
...
+
+    def __len__(self):
+        return len(self._file.keys())
```
FreddieWitherden: Probably unnecessary use of `.keys()`.

pyfr/readers/native.py
```diff
@@ -165,6 +165,21 @@ def __iter__(self):
     def __len__(self):
         return len(self._npf.files)
```
FreddieWitherden: Missing newline.  Please re-read PEP8.

pyfr/readers/native.py
```diff
@@ -207,4 +222,7 @@ def read_pyfr_data(fname):
     if os.path.isdir(fname):
         return PyFRDirReader(fname)
     else:
-        return PyFRFileReader(fname)
+        try:
+            return PyFRFileReader(fname)
+        except OSError as e:
```
FreddieWitherden: Unused variable `e`.

pyfr/scripts/mesh.py
```diff
@@ -56,7 +56,10 @@ def process_convert(args):
     mesh = reader.to_pyfrm()
 
     # Save to disk
-    np.savez(args.outmesh, **mesh)
+    import h5py
```
FreddieWitherden: Local import.

pyfr/scripts/mesh.py
```diff
@@ -56,7 +56,10 @@ def process_convert(args):
...
+    msh5=h5py.File(args.outmesh,'w')
```
FreddieWitherden: Missing space between `=` and after `,`.  Please re-read PEP8.

pyfr/scripts/mesh.py
```diff
@@ -56,7 +56,10 @@ def process_convert(args):
...
+    for k in mesh:
+        msh5.create_dataset(k,data=mesh[k])
```
FreddieWitherden: Missing space after `,`.  Please re-read PEP8.

pyfr/scripts/mesh.py
```diff
@@ -102,5 +105,8 @@ def process_partition(args):
         path = os.path.join(args.outd, os.path.basename(path.rstrip('/')))
 
         # Open and save
-        with open(path, 'wb') as f:
-            np.savez(f, **data)
+        # Save to disk
+        import h5py
```
FreddieWitherden: Local import.

pyfr/scripts/mesh.py
```diff
@@ -102,5 +105,8 @@ def process_partition(args):
...
+        msh5=h5py.File(path,'w')
```
FreddieWitherden: Missing space between `=` and after `,`.  Please re-read PEP8.

pyfr/scripts/mesh.py
```diff
@@ -102,5 +105,8 @@ def process_partition(args):
...
+        for k in data:
+            msh5.create_dataset(k,data=data[k])
```
FreddieWitherden: Missing space after `,`.  Please re-read PEP8.

pyfr/writers/base.py
```diff
@@ -33,4 +33,8 @@ def __init__(self, args):
                                (args.solnf, args.meshf))
 
         # Load config file
-        self.cfg = Inifile(self.soln['config'].item().decode())
+        self.cfg = Inifile(
+                (self.soln['config'].item()
+                    if type(self.soln['config'].item()) is str
+                    else self.soln['config'].item().decode())
+                          )
```
FreddieWitherden: Split out to simplify.  Grab `self.soln['config'].item()` into a variable and go from there.  Use of `type` for instance checking; please see Python documentation.

setup.py
```diff
@@ -92,6 +92,7 @@
     'mpmath >= 0.18',
     'numpy >= 1.8',
     'pytools >= 2014.3'
+    'h5py >= 2.4'
```
FreddieWitherden: List should be sorted alphabetically.

pyfr/integrators/writers.py
iyer-arvind: Fixed

pyfr/integrators/writers.py
iyer-arvind: Fixed

pyfr/integrators/__init__.py
```diff
@@ -12,11 +12,11 @@ def get_integrator(backend, systemcls, rallocs, mesh, initsoln, cfg):
     # Look-up the controller, stepper and writer names
     c = cfg.get('solver-time-integrator', 'controller')
     s = cfg.get('solver-time-integrator', 'scheme')
-    w = cfg.get('soln-output', 'format')
+    w = "HDF5"
```
FreddieWitherden: Use of double quoted string for a string that does not contain a single quote.

pyfr/integrators/writers.py
iyer-arvind: Fixed

pyfr/integrators/writers.py
iyer-arvind: Fixed

pyfr/integrators/writers.py
iyer-arvind: Fixed

pyfr/integrators/writers.py
```diff
@@ -61,43 +63,106 @@ def _get_name_for_soln(self, etype, prank=None):
...
 
     def _write(self, path, solnmap, metadata):
+        raise Exception
+
+    def __writeParallel(self, path, solnmap, metadata):
+        from mpi4py import MPI
+
+        comm, rank, root = get_comm_rank_root()
+
+        _file = h5py.File(path, "w", driver='mpio', comm=comm)
```
FreddieWitherden: Unnecessary `_`.  Aliasing things like `file` and `type` is well accepted.

pyfr/integrators/writers.py
iyer-arvind: Fixed

pyfr/integrators/writers.py
```diff
@@ -61,43 +63,106 @@ def _get_name_for_soln(self, etype, prank=None):
...
+        sMap = {}
+        for s, shape in self.solList:
+            sMap[s] = _file.create_dataset(s, shape, dtype='float64')
```
FreddieWitherden: Should the data type not depend on what precision PyFR is running at?  Otherwise we risk wasting storage/bandwidth for single precision simulations.

pyfr/integrators/writers.py
```diff
@@ -61,43 +63,106 @@ def _get_name_for_soln(self, etype, prank=None):
...
+
+        for e in solnmap:
+            s = self._get_name_for_soln(e, self.rallocs.prank)
+            sMap[s][:] = solnmap[e]
+
+        if(rank == root):
```
FreddieWitherden: Not PEP8.  Please re-read.

FreddieWitherden:
Sorry if I've just hammered you're inbox!

setup.py
iyer-arvind: Missing ',' after pytools, doesn't install!

pyfr/readers/native.py
iyer-arvind: H5File object is not iterable, afaik.

FreddieWitherden: Just checked:
```
In [10]: f.keys()
Out[10]: [u'a', u'b']

In [11]: list(f)
Out[11]: [u'a', u'b']
```
Still might need the `iter()`, though, but the `.keys()` can go.

FreddieWitherden:
Other notes.  For the configuration file `serial-h5` is probably not the best descriptor.  Where possible we should seek to avoid boolean options in lieu of more flexible alternatives,  e.g, `io-type = serial` and `io-type = parallel`.

The next thing to note is that if the user explicitly asks for parallel I/O and we can not provide this then we should raise an exception.  Otherwise we risk serious — and silent — performance degradation.

Further, be careful with using the default value parameter in the Inifile API.  Doing `self.cfg.get('soln-output', 'io-type', 'serial') will *set* the value if it is not present.  So the next time the simulation is restarted it will run with serial I/O (even if the default for the platform happens to be parallel).  This 'set if not present' behaviour is unfortunate and something I am looking to get rid of in the future.  But until then we need to work around it.

iyer-arvind:
The current system has the default parallel when
1) parallel hdf5 and p5py install present
2) h5py is above or equal to 2.5

In this case, the user can opt to have a serial output, as an over ride
There is NO override to make it parallel if we deem the installation
unfit for parallel.

Hence there is no question of io-type being serial or parallel.

Perhaps 'force-serial-h5' may be a better descriptor.



On Mon, Mar 9, 2015 at 1:42 PM, Freddie Witherden <notifications@github.com>
wrote:

> Other notes. For the configuration file serial-h5 is probably not the
> best descriptor. Where possible we should seek to avoid boolean options in
> lieu of more flexible alternatives, e.g, io-type = serial and io-type =
> parallel.
>
> The next thing to note is that if the user explicitly asks for parallel
> I/O and we can not provide this then we should raise an exception.
> Otherwise we risk serious — and silent — performance degradation.
>
> Further, be careful with using the default value parameter in the Inifile
> API. Doing `self.cfg.get('soln-output', 'io-type', 'serial') will *set*
> the value if it is not present. So the next time the simulation is
> restarted it will run with serial I/O (even if the default for the platform
> happens to be parallel). This 'set if not present' behaviour is unfortunate
> and something I am looking to get rid of in the future. But until then we
> need to work around it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/vincentlab/PyFR/pull/39#issuecomment-77854882>.
>



-- 
================================
Dr. Iyer Arvind Sundaram

The reasonable man adapts himself to the world; the unreasonable one
persists in trying to adapt the world to himself.

Therefore all progress depends on the unreasonable man

- George Bernard Shaw.

pyfr/integrators/writers.py
```diff
@@ -76,28 +76,90 @@ def __init__(self, *args, **kwargs):
...
+
+        if parallel:
+            if rank == root:
+                solList = []
+                for mrank, meleinfo in enumerate(eleinfo):
+                    prank = self.rallocs.mprankmap[mrank]
+                    solList.extend(
+                                (self._get_name_for_soln(etype, prank), dims)
+                                for (etype, dims) in meleinfo
+                            )
```
FreddieWitherden: Extra indentation.

pyfr/integrators/writers.py
```diff
@@ -76,28 +76,90 @@ def __init__(self, *args, **kwargs):
...
+                            loc_names.append(name)
+                        else:
+                            rbuf = np.empty(dims, dtype=self.backend.fpdtype)
+                            rreq = comm.Recv_init(rbuf, mrank, tag)
+
+                            mpi_rbufs.append(rbuf)
+                            mpi_rreqs.append(rreq)
+                            mpi_names.append(name)
 
     def _write(self, path, solnmap, metadata):
```
FreddieWitherden: No need for this.  Ditch the abstract BaseWriter class to simplify.

pyfr/integrators/writers.py
```diff
@@ -76,28 +76,90 @@ def __init__(self, *args, **kwargs):
...
+        parallel &= not self.cfg.getbool('soln-output', 'serial-h5', False)
+
+        if parallel:
+            self._write = self._writeParallel
+        else:
+            self._write = self._writeSerial
+
+        if parallel:
+            if rank == root:
+                solList = []
```
FreddieWitherden: Camel case, see PEP8.

pyfr/integrators/writers.py
```diff
@@ -76,28 +76,90 @@ def __init__(self, *args, **kwargs):
...
     def _write(self, path, solnmap, metadata):
+        raise Exception
+
+    def _writeParallel(self, path, solnmap, metadata):
+        from mpi4py import MPI
+
+        comm, rank, root = get_comm_rank_root()
+
+        h5File = h5py.File(path, 'w', driver='mpio', comm=comm)
+        sMap = {}
```
FreddieWitherden: Camel case see PEP8.

pyfr/integrators/writers.py
```diff
@@ -76,28 +76,90 @@ def __init__(self, *args, **kwargs):
...
+        for s, shape in self.solList:
+            sMap[s] = h5File.create_dataset(s, shape, dtype=self.backend.fpdtype)
+
+        for e in solnmap:
+            s = self._get_name_for_soln(e, self.rallocs.prank)
+            sMap[s][:] = solnmap[e]
+
+        if rank == root:
+            mMap = []
+            for m in metadata:
```
FreddieWitherden: Can we try and simplify this?  h5py, unlike NumPy, has a sane means of saving strings and making them 'just work' without the need for funky data types.

pyfr/writers/base.py
```diff
@@ -31,6 +31,7 @@ def __init__(self, args):
         if self.mesh['mesh_uuid'] != self.soln['mesh_uuid']:
             raise RuntimeError('Solution "%s" was not computed on mesh "%s"' %
                                (args.solnf, args.meshf))
-
```
FreddieWitherden: Random deletion of newline.

pyfr/integrators/writers.py
iyer-arvind: Yes we could. But as we decidedm that step we will take in the next
version.
This version is for exact npy in hdf5 format. Will not affect the
performance in any way

On Mon, Mar 9, 2015 at 1:54 PM, Freddie Witherden <notifications@github.com>
wrote:

> In pyfr/integrators/writers.py
> <https://github.com/vincentlab/PyFR/pull/39#discussion_r26036419>:
>
> > +        from mpi4py import MPI
> > +
> > +        comm, rank, root = get_comm_rank_root()
> > +
> > +        h5File = h5py.File(path, 'w', driver='mpio', comm=comm)
> > +        sMap = {}
> > +        for s, shape in self.solList:
> > +            sMap[s] = h5File.create_dataset(s, shape, dtype=self.backend.fpdtype)
> > +
> > +        for e in solnmap:
> > +            s = self._get_name_for_soln(e, self.rallocs.prank)
> > +            sMap[s][:] = solnmap[e]
> > +
> > +        if rank == root:
> > +            mMap = []
> > +            for m in metadata:
>
> Can we try and simplify this? h5py, unlike NumPy, has a sane means of
> saving strings and making them 'just work' without the need for funky data
> types.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/vincentlab/PyFR/pull/39/files#r26036419>.
>



-- 
================================
Dr. Iyer Arvind Sundaram

The reasonable man adapts himself to the world; the unreasonable one
persists in trying to adapt the world to himself.

Therefore all progress depends on the unreasonable man

- George Bernard Shaw.

FreddieWitherden: If we can use the string handling code in h5py then we should.  (Under the assumption that it does not complicate reading.)  The format is still the same, we are just exploiting the fact that HDF5 has a sane way of dealing with strings.

iyer-arvind: I am not clear what you wish to change it to.

If you are referring to saving them as attributes, then there is a need to
move
the info to all the ranks and make all of them write - dirty and unnecessary

If you are referring to variable length strings, then parallel
does not support it yet.


Please let me know, if you meant anything else.



On Mon, Mar 9, 2015 at 2:01 PM, Freddie Witherden <notifications@github.com>
wrote:

> In pyfr/integrators/writers.py
> <https://github.com/vincentlab/PyFR/pull/39#discussion_r26036998>:
>
> > +        from mpi4py import MPI
> > +
> > +        comm, rank, root = get_comm_rank_root()
> > +
> > +        h5File = h5py.File(path, 'w', driver='mpio', comm=comm)
> > +        sMap = {}
> > +        for s, shape in self.solList:
> > +            sMap[s] = h5File.create_dataset(s, shape, dtype=self.backend.fpdtype)
> > +
> > +        for e in solnmap:
> > +            s = self._get_name_for_soln(e, self.rallocs.prank)
> > +            sMap[s][:] = solnmap[e]
> > +
> > +        if rank == root:
> > +            mMap = []
> > +            for m in metadata:
>
> If we can use the string handling code in h5py then we should. (Under the
> assumption that it does not complicate reading.) The format is still the
> same, we are just exploiting the fact that HDF5 has a sane way of dealing
> with strings.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/vincentlab/PyFR/pull/39/files#r26036998>.
>



-- 
================================
Dr. Iyer Arvind Sundaram

The reasonable man adapts himself to the world; the unreasonable one
persists in trying to adapt the world to himself.

Therefore all progress depends on the unreasonable man

- George Bernard Shaw.

FreddieWitherden: I was wondering if we could use variable length strings with parallel I/O.  As this is not possible a variation on our current solution seems to be most appropriate.  I'll add some coments on how we can clean this up.

pyfr/integrators/writers.py
```diff
@@ -76,28 +76,90 @@ def __init__(self, *args, **kwargs):
...
+        sMap = {}
+        for s, shape in self.solList:
+            sMap[s] = h5File.create_dataset(s, shape, dtype=self.backend.fpdtype)
+
+        for e in solnmap:
+            s = self._get_name_for_soln(e, self.rallocs.prank)
+            sMap[s][:] = solnmap[e]
+
+        if rank == root:
+            mMap = []
```
FreddieWitherden: `mmap = [(k, (), 'S{}'.format(len(v.encode()))
        for k, v in metadata.items()]`
Will need to break across two lines.

iyer-arvind: I need to raise an exception when the metadata is not a string,
and just that it is not yet implemented. Can incorportate another
elif and prevent this exception as and when needed.

For example, perhaps its better in future that the stats
can be a dict, which needs to be pickled.

Better explicit exception than implicit assumption of string.


On Mon, Mar 9, 2015 at 2:22 PM, Freddie Witherden <notifications@github.com>
wrote:

> In pyfr/integrators/writers.py
> <https://github.com/vincentlab/PyFR/pull/39#discussion_r26038657>:
>
> > +    def _writeParallel(self, path, solnmap, metadata):
> > +        from mpi4py import MPI
> > +
> > +        comm, rank, root = get_comm_rank_root()
> > +
> > +        h5File = h5py.File(path, 'w', driver='mpio', comm=comm)
> > +        sMap = {}
> > +        for s, shape in self.solList:
> > +            sMap[s] = h5File.create_dataset(s, shape, dtype=self.backend.fpdtype)
> > +
> > +        for e in solnmap:
> > +            s = self._get_name_for_soln(e, self.rallocs.prank)
> > +            sMap[s][:] = solnmap[e]
> > +
> > +        if rank == root:
> > +            mMap = []
>
> mmap = [(k, (), 'S{}'.format(len(v.encode()))
> for k, v in metadata.items()]
> Will need to break across two lines.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/vincentlab/PyFR/pull/39/files#r26038657>.
>



-- 
================================
Dr. Iyer Arvind Sundaram

The reasonable man adapts himself to the world; the unreasonable one
persists in trying to adapt the world to himself.

Therefore all progress depends on the unreasonable man

- George Bernard Shaw.

FreddieWitherden: If we decide in the future that stats should be a dict then whoever makes this change will also update the format.  Until then we're adding code to cover a hypothetical.

More formally: anything except strings are not just not implemented, but not supported as part of the current file format.  Moreover, you are almost certain to get an exception at the moment if you pass anything string like as it will lack an `.encode` method and will not allow itself to be cast to a NumPy `Sxx` so it does not buy as anything.

These checks really do not buy you anything other than extra lines of code.  Hence why PyFR seldom bothers with them.

iyer-arvind: ​Ok if you are restricting only string=>string mapping for this version
then fine!

I would have any day preferred to dump all this info into a single pickle
called _metadata,
no assumptions, just that picklable, but for the requirement that this
commit be identical to legacy files.



On Mon, Mar 9, 2015 at 2:43 PM, Freddie Witherden <notifications@github.com>
wrote:

> In pyfr/integrators/writers.py
> <https://github.com/vincentlab/PyFR/pull/39#discussion_r26040562>:
>
> > +    def _writeParallel(self, path, solnmap, metadata):
> > +        from mpi4py import MPI
> > +
> > +        comm, rank, root = get_comm_rank_root()
> > +
> > +        h5File = h5py.File(path, 'w', driver='mpio', comm=comm)
> > +        sMap = {}
> > +        for s, shape in self.solList:
> > +            sMap[s] = h5File.create_dataset(s, shape, dtype=self.backend.fpdtype)
> > +
> > +        for e in solnmap:
> > +            s = self._get_name_for_soln(e, self.rallocs.prank)
> > +            sMap[s][:] = solnmap[e]
> > +
> > +        if rank == root:
> > +            mMap = []
>
> If we decide in the future that stats should be a dict then whoever makes
> this change will also update the format. Until then we're adding code to
> cover a hypothetical.
>
> More formally: anything except strings are not just not implemented, but
> not supported as part of the current file format. Moreover, you are almost
> certain to get an exception at the moment if you pass anything string like
> as it will lack an .encode method and will not allow itself to be cast to
> a NumPy Sxx so it does not buy as anything.
>
> These checks really do not buy you anything other than extra lines of
> code. Hence why PyFR seldom bothers with them.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/vincentlab/PyFR/pull/39/files#r26040562>.
>



-- 
================================
Dr. Iyer Arvind Sundaram

The reasonable man adapts himself to the world; the unreasonable one
persists in trying to adapt the world to himself.

Therefore all progress depends on the unreasonable man

- George Bernard Shaw.

FreddieWitherden: The very first, unreleased, version of PyFR did exactly this.  Take each metadata time, call `pickle.dumps(...)` on it and save this.  There were two reasons for moving away from this to our current format.  The first was to make it easier (longer term) for other people to poke around our file format.  Pickle is Python specific.  A string-ified Inifile is easier for people to interact with.  The second reason is left as an exercise for the reader, but is relevant for PyFR in the cloud.

RcktMan77:
Just a quick tip to pass along with regards to pep8:

You can make yourself a shell alias like pep8pyfr that you can run before committing your branch that might look like:

```
alias pep8pyfr='pep8 $(pwd) --exclude=<a comma separated list of directories and files to ignore>'
```

If you run this alias from the PyFR root directory in your local branch before you commit and make a pull request, then you should be able to avoid any pep8 conflicts.

iyer-arvind:
Hi,

Thanks a lot Zach, this is quite useful.

I did use this tool, but the problem is there are quite a few other places
in the code which are not exactly PEP8 and there are many changes requested
are beyond what pep8 tool picks up.

Perhaps using in combination with git may make sure that the incremental
change is at least PEP8.

Maybe we require pep8+pyfr tool.

Regards
Arvind

Just a quick tip to pass along with regards to pep8:

You can make yourself a shell alias like pep8pyfr that you can run before
committing your branch that might look like:

alias pep8pyfr='pep8 $(pwd) --exclude=<a comma separated list of
directories and files to ignore>'

If you run this alias from the PyFR root directory in your local branch
before you commit and make a pull request, then you should be able to avoid
any pep8 conflicts.

—
Reply to this email directly or view it on GitHub
<https://github.com/vincentlab/PyFR/pull/39#issuecomment-77939280>.

FreddieWitherden:
The pylint tool can also sometimes be useful.  Although it tends to generate a huge number of false positives it is very good at finding unused/unnecessary imports.

FreddieWitherden:
About to get on a plane but some quick comments:

 - use the `with` statement when opening files for writing;
 - when iterating over a dictionary use `.items()` as opposed to iterating over the keys and doing `mydict[k]` everywhere`;
 - remove the now useless `BaseWriter` class; if we ever get another writer we can look into reintroducing it (this avoids the silliness of implementing the `_write` method only to have it throw a `NotImplementedError`)

pyfr/integrators/writers.py
```diff
@@ -1,17 +1,16 @@
 # -*- coding: utf-8 -*-
 
-from abc import abstractmethod
 import itertools as it
 import os
-
+import h5py
+from mpi4py import MPI
```
FreddieWitherden: Do not globally import mpi4py.  (It is one of the few exceptions to the rule about avoiding local imports.)

pyfr/integrators/writers.py
```diff
@@ -22,13 +21,67 @@ def __init__(self, *args, **kwargs):
...
+        # Gather this information onto the root rank
+        eleinfo = comm.gather(zip(etypes, shapes), root=root)
+
+        # Deciding if parallel
+        parallel = h5py.get_config().mpi
+        parallel &= h5py.version.version_tuple[:2] >= (2, 5)
+        parallel &= not self.cfg.getbool('soln-output', 'serial-h5', False)
+
+        if parallel:
+            self._write = self._writeParallel
```
FreddieWitherden: PEP8; rename to `_write_parallel`.

pyfr/integrators/writers.py
```diff
@@ -22,13 +21,67 @@ def __init__(self, *args, **kwargs):
...
+        else:
+            self._write = self._writeSerial
```
FreddieWitherden: PEP8; rename to `_write_serial`.

pyfr/integrators/writers.py
```diff
@@ -22,13 +21,67 @@ def __init__(self, *args, **kwargs):
...
+                sollist = []
+                for mrank, meleinfo in enumerate(eleinfo):
+                    prank = self.rallocs.mprankmap[mrank]
+                    sollist.extend(
+                                (self._get_name_for_soln(etype, prank), dims)
+                                for (etype, dims) in meleinfo
+                            )
+            else:
+                sollist = None
+            self.sollist = comm.bcast(sollist, root=root)
```
FreddieWitherden: Add newline above, remove newline below.

pyfr/integrators/writers.py
```diff
@@ -60,46 +109,33 @@ def _get_name_for_soln(self, etype, prank=None):
...
-                        mpi_names.append(name)
-
-    def _write(self, path, solnmap, metadata):
-        from mpi4py import MPI
-
+        with h5py.File(path, 'w', driver='mpio', comm=comm) as h5file:
+            smap = {}
+            for s, shape in self.sollist:
+                smap[s] = h5file.create_dataset(
+                        s, shape, dtype=self.backend.fpdtype)
```
FreddieWitherden: Odd line breaking; nowhere else in PyFR does this convention appear.  We stick to either

```
x = self.foo(bar, baz.
             spam, eggs)
```
             
```
x = self.foo(
        bar, baz, spam, eggs
    )
```

pyfr/integrators/writers.py
```diff
@@ -60,46 +109,33 @@ def _get_name_for_soln(self, etype, prank=None):
...
+            # Metadata information has to be transferred to all the ranks
+            if rank == root:
+                mmap = [(k, len(v.encode()))
+                        for k, v in metadata.items()]
+            else:
+                mmap = None
+
+            mmap = comm.bcast(mmap, root=root)
+            for name, size in mmap:
+                d = h5file.create_dataset(name, (), dtype="S{}".format(size))
```
FreddieWitherden: Use of double quotes.

pyfr/integrators/writers.py
```diff
@@ -117,31 +153,6 @@ def _write(self, path, solnmap, metadata):
...
-        # Wait for this to complete
-        comm.barrier()
-
-        # Save the solutions
-        for etype, buf in solnmap.items():
-            solnpath = os.path.join(path, self._get_name_for_soln(etype))
-            np.save(solnpath, buf)
+            with h5py.File(path, 'w') as h5file:
+                for k, v in outdict.items():
+                    h5file.create_dataset(k, data=outdict[k])
```
FreddieWitherden: No idea what is going on here.  We're using `.items()` to get the key and the value but then completely ignoring the value and instead doing `outdict[k]`.

pyfr/integrators/writers.py
iyer-arvind: Why is this so?

pyfr/partitioners/base.py
```diff
@@ -241,7 +241,7 @@ def partition(self, mesh):
             newmesh = mesh
 
         # Generate a new UUID for the mesh
-        newmesh['mesh_uuid'] = newuuid = str(uuid.uuid4())
+        newmesh['mesh_uuid'] = newuuid = uuid.uuid4().hex.encode()
```
FreddieWitherden: Not entirely sure what is going on here.  The UUID is now a bytes instead of a string.  Thus far everything has in the file format has either been an array (float/double) or a string (which we sometimes need to kick a bit).

pyfr/integrators/writers.py
iyer-arvind: Mistake

pyfr/partitioners/base.py
iyer-arvind: True.

pyfr/integrators/writers.py
FreddieWitherden: As importing mpi4py can — if you are not very careful — result in MPI_Init being called.  This gets us into all kinds of trouble as when running a simulation we need to ensure that we prefork first.  When post-processing we do not want MPI at all.  Hence, we always import MPI locally to avoid the possibility of this.

pyfr/scripts/mesh.py
```diff
@@ -56,7 +56,9 @@ def process_convert(args):
     mesh = reader.to_pyfrm()
 
     # Save to disk
-    np.savez(args.outmesh, **mesh)
+    msh5 = h5py.File(args.outmesh, 'w')
```
FreddieWitherden: See previous comment with regards to the `with` statement.

pyfr/scripts/mesh.py
```diff
@@ -102,5 +104,7 @@ def process_partition(args):
         path = os.path.join(args.outd, os.path.basename(path.rstrip('/')))
 
         # Open and save
-        with open(path, 'wb') as f:
-            np.savez(f, **data)
+        # Save to disk
+        msh5 = h5py.File(path, 'w')
```
FreddieWitherden: `with`.

FreddieWitherden:
Regarding the `__getitem__` blocks.  I agree that handing strings here is the right place to do it.  However, the current implementations leave a lot to be desired.

Firstly, from a semantic point of view I note that the pyfr-dir reader will invoke `.item()` on any arrays which are strings (`S`).  However, the pyfr-file reader will also do it on unicode arrays.  Why is this exactly?  Is there a reason why a common implementation can not be used for both files and dirs?  In which case I would suggest that we factor out the code into a function.

Secondly, the pyfr-file  code is awful.  We end up reading each item from disk *twice*.  There is no good reason for this.

Thirdly, the formatting is ghastly.  I note the `if(foo)` type statement appearing despite having pointed this out several times already as being non-standard.  Splitting inline `if` statements across multiple lines is also probably not the best idea (it becomes harder to follow than a normal block). Something along the lines of:

```
    def __getitem__(self, aname):
        arr = self._npf[aname]
        
        if arr.dtype.kind == 'S':
            return arr.item().decode()
        else:
            return arr
```
is easier to follow and only reads the file from disk once.  It is of course predicated on the assumption that we do not have any `U` types in our files.  The code is easy to modify if this is the case, however.

iyer-arvind:
Well the twice reading problem is well taken, will update it, but as I mentioned, I have encountered the case of a 'U' written by a previous code. If 2 lines is not many, we may well have it.

iyer-arvind:
Regarding the _ghastly_ if() statements, no one is doing it to be an irritant. Just that it does not strike me in the eye as badly as it does to you, Freddie. I am neither defending it, not asking it to be with (), not suggesting that PEP8 is wrong. Its just that change of style needs time. In the meanwhile I have only the lint tools to rely on, pep8 tool and flake8, neither of which picked it up.

FreddieWitherden:
If `U` strings can find their way into one of the PyFR file formats then they can probably find their way into the other.  My suggestion then, if this makes sense to you, is to have a function that can handle both `S` and `U` types (perhaps `_process_item`) and then have `__getitem__` call this function.  Should save code and work for all types.

iyer-arvind:
The 'U' strings will not appear henceforth, and anyway in the future versions, we wont need to have the npy readers anyway. In the furture, with HDF5, there will be no need for UIDS, and all string data can be attributes, so wont need the conversion from arrays anyway.

FreddieWitherden:
With regards to the pep8 checkers and pylint.  While they are useful in catching some classes of problem there is a lot they miss.  They are also idiosyncratic in some places.  Unfortunately there is really no substitute for experience and review by a human.

Without a doubt the best way to pick up a good Python style is by reading good Python code.  This is one of the reasons why I am so anal about formatting and consistency in PyFR: every fragment should be an example of best practice.  If you have a style question (say `if` statement formatting, or how to break long lines) you should be able to look at *any* line in PyFR where such a statement occurs and have your answer.  Having inconsistent formatting, or using different approaches in different parts of the code to solve the same problem is a great way to confuse beginners.

Another great place to find good 'Pythonic' Python is in a book ("Learning Python" by Lutz is my preference).  Looking through the code examples you'll never encounter `if(something)` but always `if something`.  Same thing goes for line-broken inline if statements.  After working through a couple of chapters you should have no trouble identifying Python code that 'looks off'.

FreddieWitherden:
One more thing before I forget.  Can you verify/confirm that opening/reading an HDF5 file (in serial) using h5py + HDF5 compiled with MPI support does not result in MPI_Init being called?

If MPI_Init is getting called then it is not a huge issue but some minor plumbing will be required to work around the issue.

iyer-arvind:
In h5py, there is the presence of these lines: 

```python
if MPI:
    from mpi4py.mpi_c cimport MPI_Comm, MPI_Info, MPI_Comm_dup, MPI_Info_dup, \
    MPI_Comm_free, MPI_Info_free
```
I am not sure if these initialize the mpi. 

However the following test gives more hint.
```python
>>> import h5py
>>> from mpi4py import rc
>>> rc.initialize = False
>>> from mpi4py import MPI
>>> MPI.Is_initialized()
True
```
Where as

```python
>>> from mpi4py import rc
>>> rc.initialize = False
>>> from mpi4py import MPI
>>> MPI.Is_initialized()
>>> False
>>> import h5py
>>> MPI.Is_initialized()
>>> False
```
Which shows that h5py initializes MPI if simply put. However, if we have 

``` python
rc.initialize = False
```
import h5py does not initialize MPI. But will it may affect the initialization where we need it?

perhaps a solution is 

```python
from mpi4py import rc
initialize_mpi = rc.initialize 
rc.initialize = False
import h5py
rc.initialize = initialize_mpi
del initialize_mpi
```

Kindly suggest

FreddieWitherden:
None of the solutions available to us are great but the best of a bad bunch appears to be along the lines of:

```
import mpi4py.rc
mpi4py.rc.initialize = False
```

and putting this in sim.py.  This is what we did for an earlier version of PyFR.  It isn't great, but we know it works.  The import should of course be 'high up' in the list; certainly above any of the third party imports in 'sim.py'.  (This is a necessary exception to the sort imports alphabetically rule!)  Then in `_process_common` after `from mpi4py import MPI` do `MPI.Initialized()`.  I suggest we still look to always keep `MPI` imported locally as it will make it easier to money-patch later on.  (A long term goal is to eliminate the dependency on mpi4py if one just wishes to run with a single rank.)

pyfr/integrators/writers.py
```diff
@@ -1,17 +1,17 @@
 # -*- coding: utf-8 -*-
 
-from abc import abstractmethod
 import itertools as it
 import os
 
+import h5py
+
```
FreddieWitherden: Remove newline.

pyfr/integrators/writers.py
```diff
@@ -22,13 +22,67 @@ def __init__(self, *args, **kwargs):
...
+        # Gather this information onto the root rank
+        eleinfo = comm.gather(zip(etypes, shapes), root=root)
+
+        # Deciding if parallel
+        parallel = h5py.get_config().mpi
+        parallel &= h5py.version.version_tuple[:2] >= (2, 5)
+        parallel &= not self.cfg.getbool('soln-output', 'serial-h5', False)
+
+        if parallel:
```
FreddieWitherden: Merge if/else block with that on line 44 (the condition `if parallel` is the same for both blocks).

pyfr/integrators/writers.py
```diff
@@ -60,44 +110,34 @@ def _get_name_for_soln(self, etype, prank=None):
...
-                        mpi_rbufs.append(rbuf)
-                        mpi_rreqs.append(rreq)
-                        mpi_names.append(name)
-
-    def _write(self, path, solnmap, metadata):
+        with h5py.File(path, 'w', driver='mpio', comm=comm) as h5file:
+            smap = {}
+            for s, shape in self.sollist:
+                smap[s] = h5file.create_dataset(
+                                 s, shape, dtype=self.backend.fpdtype
```
FreddieWitherden: Extra indentation.  Should be one level deeper than `smap[s]`.  The file `pyfr.solvers.baseadvec.elements` is a good reference here.

pyfr/partitioners/base.py
```diff
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 
-from collections import Counter, defaultdict, namedtuple
 import re
 import uuid
+from collections import Counter, defaultdict, namedtuple
```
FreddieWitherden: Revert reordering.

pyfr/partitioners/base.py
```diff
@@ -241,7 +241,7 @@ def partition(self, mesh):
             newmesh = mesh
 
         # Generate a new UUID for the mesh
-        newmesh['mesh_uuid'] = newuuid = str(uuid.uuid4())
+        newmesh['mesh_uuid'] = newuuid = uuid.uuid4().hex
```
FreddieWitherden: Revert change to how we stringify UUIDs.

pyfr/readers/base.py
```diff
@@ -17,6 +17,6 @@ def to_pyfrm(self):
         mesh = self._to_raw_pyfrm()
 
         # Add metadata
-        mesh['mesh_uuid'] = str(uuid.uuid4())
+        mesh['mesh_uuid'] = uuid.uuid4().hex
```
FreddieWitherden: Revert change to how we stringify UUIDs.

pyfr/readers/native.py
```diff
@@ -3,15 +3,16 @@
...
 
 """
-from collections import Mapping, OrderedDict
 import errno
 import os
 import re
 
+import h5py
 import numpy as np
+from collections import Mapping, OrderedDict
```
FreddieWitherden: Revert reordering of imports.

pyfr/readers/native.py
```diff
@@ -166,6 +182,32 @@ def __len__(self):
...
+    """Does everything that file reader does but with a h5py file"""
+    def __init__(self, fname):
+        self._file = h5py.File(fname, 'r')
+
+    def __getitem__(self, aname):
+        ret = self._file[aname]
+
+        if self._file[aname].shape == ():
+            ret = ret.value
+
```
FreddieWitherden: Extra newline.

pyfr/scripts/mesh.py
```diff
@@ -1,13 +1,13 @@
 # -*- coding: utf-8 -*-
 
-from argparse import FileType
 import itertools as it
 import os
+from argparse import FileType
```
FreddieWitherden: Revert reordering.

pyfr/readers/native.py
```diff
@@ -166,6 +182,32 @@ def __len__(self):
...
+        else:
+            ret = np.array(ret)
+
+        if isinstance(ret, bytes):
+            return ret.decode()
+
+        return ret
```
FreddieWitherden: Put this into an `else` block or, if you prefer merge into a single line `return ret.decode() if isinstance(ret, bytes) else ret`.

pyfr/readers/native.py
```diff
@@ -157,7 +166,14 @@ def __init__(self, fname):
...
     def __getitem__(self, aname):
-        return self._npf[aname]
+        ret = self._npf[aname]
+
+        ret = ret.item() if ret.dtype.kind in 'SU' else ret
+
+        if isinstance(ret, bytes):
+            return ret.decode()
+
+        return ret
```
FreddieWitherden: See comment below regarding formatting.

pyfr/readers/native.py
```diff
@@ -136,8 +137,16 @@ def __init__(self, fname):
 
     def __getitem__(self, aname):
         try:
-            return np.load(os.path.join(self.fname, aname + '.npy'),
-                           mmap_mode='r')
+            ret = np.load(
+                    os.path.join(self.fname, aname + '.npy'),
+                    mmap_mode='r')
```
FreddieWitherden:
```
            ret = np.load(os.path.join(self.fname, aname + '.npy'),
                          mmap_mode='r')
```

pyfr/readers/native.py
```diff
@@ -136,8 +137,16 @@ def __init__(self, fname):
...
+
+            ret = ret.item() if ret.dtype.kind in 'SU' else ret
+
+            if isinstance(ret, bytes):
+                return ret.decode()
+
+            return ret
```
FreddieWitherden: See comments below regarding formatting.

pyfr/integrators/writers.py
```diff
@@ -22,13 +22,67 @@ def __init__(self, *args, **kwargs):
...
+            self._write = self._write_serial
+
+        if parallel:
+            if rank == root:
+                sollist = []
+                for mrank, meleinfo in enumerate(eleinfo):
+                    prank = self.rallocs.mprankmap[mrank]
+                    sollist.extend(
+                                (self._get_name_for_soln(etype, prank), dims)
+                                for (etype, dims) in meleinfo
```
FreddieWitherden: Extra parenthesis.

pyfr/integrators/writers.py
```diff
@@ -22,13 +22,67 @@ def __init__(self, *args, **kwargs):
...
+        else:
+            self._write = self._write_serial
+
+        if parallel:
+            if rank == root:
+                sollist = []
+                for mrank, meleinfo in enumerate(eleinfo):
+                    prank = self.rallocs.mprankmap[mrank]
+                    sollist.extend(
+                                (self._get_name_for_soln(etype, prank), dims)
```
FreddieWitherden: Extra indentation, should be one level deeper than `sollist...`.

pyfr/integrators/writers.py
```diff
@@ -22,13 +22,67 @@ def __init__(self, *args, **kwargs):
...
+                                for (etype, dims) in meleinfo
+                            )
```
FreddieWitherden: Extra indentation, should be at the same level as `sollist...`.

pyfr/readers/base.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 
-from abc import ABCMeta, abstractmethod
 import uuid
+from abc import ABCMeta, abstractmethod
```
FreddieWitherden: Revert reordering.

pyfr/integrators/writers.py
```diff
@@ -22,13 +21,65 @@ def __init__(self, *args, **kwargs):
...
+        eleinfo = comm.gather(zip(etypes, shapes), root=root)
+
+        # Deciding if parallel
+        parallel = h5py.get_config().mpi
+        parallel &= h5py.version.version_tuple[:2] >= (2, 5)
+        parallel &= not self.cfg.getbool('soln-output', 'serial-h5', False)
+
+        if parallel:
+            self._write = self._write_parallel
+            if rank == root:
```
FreddieWitherden: Add a newline.

pyfr/integrators/writers.py
```diff
@@ -22,13 +21,65 @@ def __init__(self, *args, **kwargs):
...
+                for mrank, meleinfo in enumerate(eleinfo):
+                    prank = self.rallocs.mprankmap[mrank]
+                    sollist.extend(
+                        (self._get_name_for_soln(etype, prank), dims)
+                        for etype, dims in meleinfo
+                    )
+
+            else:
+                sollist = None
+            self.sollist = comm.bcast(sollist, root=root)
```
FreddieWitherden: Add a newline.

pyfr/integrators/writers.py
```diff
@@ -22,13 +21,65 @@ def __init__(self, *args, **kwargs):
...
+
+        else:
+            self._write = self._write_serial
+            if rank == root:
```
FreddieWitherden: Add a newline.

pyfr/readers/native.py
```diff
@@ -136,8 +137,13 @@ def __init__(self, fname):
...
     def __getitem__(self, aname):
         try:
-            return np.load(os.path.join(self.fname, aname + '.npy'),
-                           mmap_mode='r')
+            ret = np.load(os.path.join(self.fname, aname + '.npy'),
+                          mmap_mode='r')
+
+            ret = ret.item() if ret.dtype.kind in 'SU' else ret
+
+            return ret.decode if isinstance(ret, bytes) else ret
```
FreddieWitherden: I think you need to call ret.decode() rather than return the method itself.

pyfr/readers/native.py
```diff
@@ -166,6 +176,29 @@ def __len__(self):
...
 
+class PyFRH5Reader(PyFRBaseReader):
+    """Does everything that file reader does but with a h5py file"""
+    def __init__(self, fname):
+        self._file = h5py.File(fname, 'r')
+
+    def __getitem__(self, aname):
+        ret = self._file[aname]
+
+        if self._file[aname].shape == ():
```
FreddieWitherden: Probably want to use `ret` here.  While h5py is probably not stupid enough to read the entire file it is certainly going to be more work than using `ret`.

FreddieWitherden:
Fold into a single commit and we should be just about ready to merge.

PR #43, opened 2015-03-31
Update User Guide.

pv101:


FreddieWitherden:
Also remove the documentation for the `format` key.

pv101:
I think I did this? Line 676 removed? Have pack and unpack now died? Shall I remove them from post-p docs?

FreddieWitherden:
I see the change on line 676 but it is also documented immediately above:

```
1. ``format`` --- format of the outputs:

    ``pyfrs-file`` | ``pyfrs-dir``
```

FreddieWitherden:
Pack and unpack have now died as they are not longer necessary with the new file format.  I have removed the documentation I could find in the user guide.  I think everything should merge fine, though.

PR #45, opened 2015-04-17
Features/pvtu

iyer-arvind:
This implements a parallel VTU writer.


pyfr/writers/paraview.py
```diff
@@ -63,6 +65,75 @@ def write_out(self):
         write_s('\n</AppendedData>\n</VTKFile>')
 
 
+class ParallelParaviewWriter(BaseWriter):
+    """Wrapper for writing serial .vtu Paraview files"""
```
FreddieWitherden: Sigh.

pyfr/writers/paraview.py
```diff
@@ -63,6 +65,75 @@ def write_out(self):
...
+    # Supported file types and extensions
+    name = 'paraview'
```
FreddieWitherden: Name must be unique across subclasses (I'll comment on a better way to handle this later).

pyfr/writers/paraview.py
```diff
@@ -63,6 +65,75 @@ def write_out(self):
...
+        basename = os.path.splitext(self.outf.name)[0]
+
+        write_s('<?xml version="1.0" ?>\n<VTKFile '
+                'byte_order="LittleEndian" type="PUnstructuredGrid" '
+                'version="0.1">\n<PUnstructuredGrid>\n')
+
+        for pi, (mk, sk) in enumerate(zip(self.mesh_inf, self.soln_inf)):
+            pfn = basename+'_part{}.vtu'.format(pi)
+            with open(pfn, 'wb') as fh:
+                write_sp = lambda s: fh.write(s.encode('utf-8'))
```
FreddieWitherden: Why are we redefining this every time the loop runs?

iyer-arvind: fh gets redifined

pyfr/writers/paraview.py
```diff
@@ -63,6 +65,75 @@ def write_out(self):
...
+        write_s = lambda s: self.outf.write(s.encode('utf-8'))
+
+        basename = os.path.splitext(self.outf.name)[0]
+
+        write_s('<?xml version="1.0" ?>\n<VTKFile '
+                'byte_order="LittleEndian" type="PUnstructuredGrid" '
+                'version="0.1">\n<PUnstructuredGrid>\n')
+
+        for pi, (mk, sk) in enumerate(zip(self.mesh_inf, self.soln_inf)):
+            pfn = basename+'_part{}.vtu'.format(pi)
```
FreddieWitherden: Use of '+' for string building is usually not appropriate (PEP8).  Use `'{0}_part{1}'.format(basename, pi)`.

pyfr/writers/paraview.py
```diff
@@ -303,12 +376,19 @@ def _write_vtu_header(args, vtuf, m_inf, s_inf, off):
     nnodes = len(nodes)*m_inf[1][1]
 
     # Standard template for vtk DataArray string
-    darray = '<DataArray Name="%s" type="%s" NumberOfComponents="%s" '\
-             'format="appended" offset="%d"/>\n'
+    if parallel:
+        darray = '<'+pp('DataArray')+' Name="%s" type="%s" '\
```
FreddieWitherden: Unnecessary use of `pp`.

pyfr/writers/paraview.py
```diff
@@ -303,12 +376,19 @@ def _write_vtu_header(args, vtuf, m_inf, s_inf, off):
...
+            'NumberOfComponents="%s"/>\n'
+    else:
+        darray = '<'+pp('DataArray')+' Name="%s" type="%s" '\
```
FreddieWitherden: Unnecessary use of `pp`.

pyfr/writers/paraview.py
```diff
@@ -290,6 +361,8 @@ def _write_vtu_header(args, vtuf, m_inf, s_inf, off):
 
     write_s = lambda s: vtuf.write(s.encode('utf-8'))
 
+    pp = (lambda x: 'P'+x) if parallel else (lambda x: x)
```
FreddieWitherden: Consider defining a variable like `prefix = 'P' if parallel else ''` and then just using that where needed: 'stuff...{prefix}...'.format(..., prefix=prefix)`, also avoids the use of '+' for string building.

pyfr/writers/paraview.py
```diff
@@ -63,6 +65,75 @@ def write_out(self):
...
+                # Write .vtu file header
+                write_sp('<?xml version="1.0" ?>\n<VTKFile '
+                         'byte_order="LittleEndian" type="UnstructuredGrid" '
+                         'version="0.1">\n<UnstructuredGrid>\n')
+                # Initialise offset (in bytes) to end of appended data
+                off = 0
+
+                # Write data description header.  A vtk "piece" is used for each
+                # element in a partition.
+                off += _write_vtu_header(self.args, fh, self.mesh_inf[mk],
```
FreddieWitherden: I might be missing something here but it looks like: We define `off` to be zero.  Then pass it (zero) to `_write_vtu_header` incrementing `off` by whatever it returns.  However, it does not look as if we use this incremented version of off.  Would `_write_vtu_header(self.args, fh, self.mesh_inf[mk], self.soln_inf[sk], off=0)` not work and save us defining/incrementing the variable?

iyer-arvind: True. It was silly.

FreddieWitherden:
With the caveat that I have not slept in 36 hours and am jet lagged...

From what I can gather this will output one file per element type per partition.  However, often when we have mixed element meshes there is an asymmetry in terms of # of elements of a given type in a given partition.  This will lead to poor load balancing and a non-obvious number of parallel outputs.  (As it depends on how the partitioner felt like distributing elements.)

I think we therefore will want the ability to have multiple element types inside of a single parallel part.

It would also be better to merge the two classes into a single write that supports two extensions: `['.vtu', '.pvtu']` and then looks at the output file name to decide which to do.  It would eliminate some code duplication and the reuse of the `paraview` writer name.

pyfr/writers/paraview.py
```diff
@@ -22,45 +24,67 @@ class ParaviewWriter(BaseWriter):
...
-
-        # Write .vtu file footer
-        write_s('\n</AppendedData>\n</VTKFile>')
+            j, el, part = mk.split('_')
+            if parallel:
+                pfn = '{}_{}.vtu'.format(basename, part)
+            else:
+                pfn = self.outf
+
+            parts.setdefault(pfn, [])
```
FreddieWitherden: Looks like you want `defaultdict` from `collections`.

pyfr/writers/paraview.py
```diff
@@ -22,45 +24,67 @@ class ParaviewWriter(BaseWriter):
...
-        # Write data "piece"wise
+        parts = {}
         for mk, sk in zip(self.mesh_inf, self.soln_inf):
-            _write_vtu_data(self.args, self.outf, self.cfg, self.mesh[mk],
-                            self.mesh_inf[mk], self.soln[sk],
-                            self.soln_inf[sk])
-
-        # Write .vtu file footer
-        write_s('\n</AppendedData>\n</VTKFile>')
+            j, el, part = mk.split('_')
```
FreddieWitherden: `mk.split('_')[1:]` avoids the need for a dummy assignment, but see below.

pyfr/writers/paraview.py
```diff
@@ -22,45 +24,67 @@ class ParaviewWriter(BaseWriter):
...
+            if parallel:
+                pfn = '{}_{}.vtu'.format(basename, part)
+            else:
+                pfn = self.outf
+
+            parts.setdefault(pfn, [])
+            parts[pfn].append((el, mk, sk))
```
FreddieWitherden: As far as I can tell we never actually use `el` when we later iterate over the list (lines 57 and 67).   If so we can save some effort by dropping it.

pyfr/writers/paraview.py
```diff
@@ -22,45 +24,67 @@ class ParaviewWriter(BaseWriter):
...
+            with open(pfn, 'wb') as fh:
+                write_sp = lambda s: fh.write(s.encode('utf-8'))
+                # Write .vtu file header
+                write_sp('<?xml version="1.0" ?>\n<VTKFile '
+                         'byte_order="LittleEndian" '
+                         'type="UnstructuredGrid" '
+                         'version="0.1">\n<UnstructuredGrid>\n')
+                off = 0
+                for el, mk, sk in misil:
+
```
FreddieWitherden: Newline

pyfr/writers/paraview.py
```diff
@@ -22,45 +24,67 @@ class ParaviewWriter(BaseWriter):
...
+                # Write end/start of header/data sections
+                write_sp('</UnstructuredGrid>\n'
+                         '<AppendedData encoding="raw">\n_')
+                for el, mk, sk in misil:
+                    # Write data "piece"wise
+                    _write_vtu_data(self.args, fh, self.cfg, self.mesh[mk],
+                                    self.mesh_inf[mk], self.soln[sk],
+                                    self.soln_inf[sk])
+
+                    # Write .vtu file footer
```
FreddieWitherden: Indentation.

pyfr/writers/paraview.py
```diff
@@ -22,45 +24,67 @@ class ParaviewWriter(BaseWriter):
...
+                write_sp('\n</AppendedData>\n</VTKFile>')
+        if parallel:
+            with open(self.outf, 'wb') as fh:
+                write_s = lambda s: fh.write(s.encode('utf-8'))
+                write_s('<?xml version="1.0" ?>\n<VTKFile '
+                        'byte_order="LittleEndian" type="PUnstructuredGrid" '
+                        'version="0.1">\n<PUnstructuredGrid>\n')
+                el, mk, sk = (next(iter(parts.items()))[1][0])
```
FreddieWitherden: Excess parenthesis.  Also we're never interested in the key so `.values()` should be used instead of items, also saves the `[1]`.

pyfr/writers/paraview.py
```diff
@@ -22,45 +24,67 @@ class ParaviewWriter(BaseWriter):
...
+                _write_vtu_header(self.args, fh, self.mesh_inf[mk],
+                                  self.soln_inf[sk], 0, True)
+                for pfn, misil in parts.items():
```
FreddieWitherden: We don't use `misil`, so no need for `.items()` can just iterate directly over `parts`.

pyfr/writers/paraview.py
```diff
@@ -318,21 +351,30 @@ def _write_vtu_header(args, vtuf, m_inf, s_inf, off):
...
-    offs[[1,5,6,7]] *= flt[1]*npts
+    offs[[1, 5, 6, 7]] *= flt[1]*npts
 
     # Write vtk DaraArray headers
     for i in range(len(nams)):
-        write_s(darray % (nams[i], typs[i], ncom[i],
-                          sum(offs[:i+1]) + i*4 + off))
+        if parallel:
+            write_s(darray % (nams[i], typs[i], ncom[i]))
+
```
FreddieWitherden: Newline.

iyer-arvind:
Implemented the parallel writer for pvtu. With help of @Aerojspark , found that parallel writer writes much faster than the serial .pvtu writer. The .pvtu file export can be done without mpirun, (single rank) or mpirun -np <number of partitions>, not any other number of ranks. '.vtu' writer _has_ to be serial.

FreddieWitherden:
Can you grab a Python profile of where the time is being spent in the writer when it is run serially compared to when it is with MPI?  When I have previously looked into this a good chunk of time has been spent either: recomputing the operator matrices (which we, foolishly, do not cache) and the matrix multiplies themselves in the case when numpy has not been linked against a decent BLAS.

The current MPI implementation is awkward, however.  A better solution would be to take the code you had before (pre-MPI) and then do:

```
size = comm.size

for pfn, misil in it.islice(parts.items(), rank, None, size):
   ...
```

along with a `rank == root` check somewhere to decide if to output the supplementary file or not.  This gives some degree of flexibility of using as many MPI ranks as you feel like for the problem rather than having to (arbitrarily) match the number of partitions.

Although, again, I would strongly caution against going towards MPI until the low hanging fruit has been addressed.  It makes things such as dynamic repartitioning (so going from N solution partitions to M vis partitions without hitting disk) somewhat more complicated for, perhaps, only a minor improvement over properly tuned serial code.

iyer-arvind:
Actually I could do that, but I dont think there is a need for profiling. It is quite clear that perhaps 10-20% of the time is taken in deciding the location of points, 50% in calculating the solution(which is a numpy.dot!) and remaining in writing. I also know that the first 2 portions can be parallellized, and not the third part, because its an IO. Using a decent BLAS can speed up the first two parts significantly. 

The question is not just that. This is a precursor to more elaborate post processing. This involves calculation of gradients too, and combination of calculations with it. Which I think could get a bit elaborate. Parallel implementation also allows for a future addition of corrected gradients, whether we need it or not needs to be seen. 

Secondly, the parallel implementation does not mean serial implementation of PVTU writer will not work. Firing with a single thread will work exactly as the previous commit. Parallel writer is a feature add.

Thirdly, the question of the N to M partitions, this can anyway be done on the solution file first and then the M partitions can be post processed - Does not prevent anything.

FreddieWitherden:
>Actually I could do that, but I dont think there is a need for profiling. It is quite clear that perhaps 10-20% of the time is taken in deciding the location of points, 50% in calculating the solution(which is a numpy.dot!) and remaining in writing. I also know that the first 2 portions can be parallellized, and not the third part, because its an IO. Using a decent BLAS can speed up the first two parts significantly. 

The first part is not too dependent on the availability of a decent BLAS library.  However, small speed ups are possible by installing `gmpy2`.  The second part is indeed heavily dependent on the BLAS library used by `np.dot`.  The difference between multi-threaded OpenBLAS and the reference BLAS that comes with Numpy can be a factor of 20-30.  All comparisons need to be done within such a context.

>The question is not just that. This is a precursor to more elaborate post processing. This involves calculation of gradients too, and combination of calculations with it. Which I think could get a bit elaborate. Parallel implementation also allows for a future addition of corrected gradients, whether we need it or not needs to be seen. 

It is best not to try and pre-empt what we might need in the future.  Uncorrected gradients, and simple metrics thereof, just fall back onto `np.dot` and memory-bandwidth bound numpy kernels for which the advantage of MPI is not obvious.   When we have such routines, have profiled them, have optimised their serial performance and subsequently decided that this is not sufficient is the time when we should consider making the post processing routines parallel.  Currently, we do not even know how easily amenable to paralleisation these hypothetical corrected gradient routines are!

>Secondly, the parallel implementation does not mean serial implementation of PVTU writer will not work. Firing with a single thread will work exactly as the previous commit. Parallel writer is a feature add.

Exactly, it is a feature add.  As a rule of thumb [all features start from -100](http://blogs.msdn.com/b/ericgu/archive/2004/01/12/57985.aspx).  A feature such as this implies a testing, support, and maintenance burden.  It is important to appreciate this relative to the functionality offered by the feature.

Currently post-processing has two levels of parallelism.  Firstly, at the highest level you can batch process solution files independently of on another.  (And with post processing we're after throughput as opposed to latency.)  This is what we often do when making animations; use GNU parallel/job system to process a bunch of solutions at once.  It confers a benefit for both regular and PVTU files.  Secondly, at the BLAS, level a multi-threaded library should allow postp to saturate most of the cores on a node.  This benefits everyone with no input from the user (so long as numpy has been installed correctly).

What is being proposed is to add a third level of parallelism that requires a degree of input from the user (they need to do `mpirun`), and only benefits the case of writing out PVTU files.  It does so with the added complication of brining MPI into post processing.  (This *does* have consequences, especially on clusters, which need to be given careful consideration.)

Given this, for the feature to be given serious consideration at this stage, I'll need to be convinced that it that it can offer significant improvements over a serial solution (possibly with operator matrix caching) combined with a decent BLAS library.  If so then we can take input on how important the latency is in practise (rather than just post processing more solutions at once) and go from there.

>Thirdly, the question of the N to M partitions, this can anyway be done on the solution file first and then the M partitions can be post processed - Does not prevent anything.
This is *exactly* what we want to get away from.  We do not want to have to write out a repartitioned mesh and solution to disk.  Rather, we want to load the current solution, repartition in-memory, and then post-process this in memory repartitioned solution.  The effect is to save us a lot of needless I/O (as I/O is the one thing that is not scaling at the moment --- ask any sysadmin how to take down a cluster and they'll say hit the parallel file system).  However, with an MPI enabled PVTU writer some extra work is needed to broadcast this repartitioned solution out to the other ranks.

iyer-arvind:
Ok! Reverted back!

pyfr/writers/paraview.py
```diff
@@ -22,45 +26,66 @@ class ParaviewWriter(BaseWriter):
...
+            if parallel:
+                pfn = '{}_{}.vtu'.format(basename, part)
+            else:
+                pfn = self.outf
+
+            parts[pfn].append((mk, sk))
+
+        for pfn, misil in parts.items():
+            with open(pfn, 'wb') as fh:
+                write_sp = lambda s: fh.write(s.encode('utf-8'))
```
FreddieWitherden: I think I might have asked this before, but I am still not exactly sure why we're redefining this each iteration.

pyfr/writers/paraview.py
```diff
@@ -22,45 +26,66 @@ class ParaviewWriter(BaseWriter):
...
-
-        # Write data description header.  A vtk "piece" is used for each
-        # element in a partition.
-        for mk, sk in zip(self.mesh_inf, self.soln_inf):
-            off += _write_vtu_header(self.args, self.outf, self.mesh_inf[mk],
-                                     self.soln_inf[sk], off)
-
-        # Write end/start of header/data sections
-        write_s('</UnstructuredGrid>\n<AppendedData encoding="raw">\n_')
+        basename = os.path.splitext(self.outf)[0]
```
FreddieWitherden: Variable name is possibly misleading as it is not a basename (in the sense of `os.path.basename`).

pyfr/writers/paraview.py
```diff
@@ -22,45 +26,66 @@ class ParaviewWriter(BaseWriter):
...
+            with open(self.outf, 'wb') as fh:
+                write_s = lambda s: fh.write(s.encode('utf-8'))
+                write_s('<?xml version="1.0" ?>\n<VTKFile '
+                        'byte_order="LittleEndian" type="PUnstructuredGrid" '
+                        'version="0.1">\n<PUnstructuredGrid>\n')
+                mk, sk = next(iter(parts.values()))[0]
+                _write_vtu_header(self.args, fh, self.mesh_inf[mk],
+                                  self.soln_inf[sk], 0, True)
+                for pfn in parts:
+                    relpath = os.path.relpath(pfn, os.path.dirname(self.outf))
```
FreddieWitherden: Is this not always just the base name (in the `os.path.basename` sense) of `pfn`?

pyfr/writers/paraview.py
iyer-arvind: Correct me if I am wrong, pfn changes every iteration, so fh changes, hence write_sp changes.

pyfr/writers/paraview.py
iyer-arvind: No, not if you are writing the pvtu files into another directory. In fact, it may always be just '.'

pyfr/writers/paraview.py
FreddieWitherden: The fact `fh` is changing does not matter; Python is lexically scoped with closed variables being bound at call time.  As a simple example consider the following:

```
l = [lambda x: x + i for i in range(10)]
print(','.join(str(f(0)) for f in l)
```

Indeed, `i` does not even need to exist when we define the lambda:

```
f = lambda x: x + i

i = 5
print(str(f(0)))
```

Hence why I am wondering why we keep redefining.  Can we not just do `write_fh = lambda s: fh.write(s.encode('utf-8'))` at the top?

pyfr/writers/paraview.py
FreddieWitherden: Can you explain this to me in a bit more detail?

Lets say that `self.outf = 'foo/bar/baz.vtu` then by line 37 `basename = 'foo/bar/baz'`.  Then `pfn = 'foo/bar/baz_0.vtu` by line 44.  For line 86 I get `relpath = os.path.relpath(pfn, os.path.dirname(self.outf)) = os.path.relpath('foo/bar/baz_0.vtu', os.path.dirname('foo/bar/baz.vtu')) = 'baz_0.vtu'`.  By my count `os.path.basename(pfn) = os.path.basename('foo/bar/baz_0.vtu') = 'baz_0.vtu'`.

iyer-arvind: We need to open the path of foo/bar/baz_0.vtu to write, but make the entry baz_0.vtu alone in the pvtu file. As long as the pvtu, and the vtu file are going to be written in the same /foo/bar/ directory, just extracting the not-dir part of the file path should work.

FreddieWitherden: From my reading of the current code will the vtu files not always end up in the same directory as the pvtu file?  (And I think that this is sensible and probably what we want.)

FreddieWitherden:
I think we're almost ready to merge.

Can you fold the request down into a single commit; shorten the first line of the commit message, and undo the changes to the user guide (as we're now going for the `soln-` as opposed to `solver-` prefix)?

pyfr/writers/paraview.py
```diff
@@ -22,45 +26,68 @@ class ParaviewWriter(BaseWriter):
...
-
-        # Write data description header.  A vtk "piece" is used for each
-        # element in a partition.
-        for mk, sk in zip(self.mesh_inf, self.soln_inf):
-            off += _write_vtu_header(self.args, self.outf, self.mesh_inf[mk],
-                                     self.soln_inf[sk], off)
-
-        # Write end/start of header/data sections
-        write_s('</UnstructuredGrid>\n<AppendedData encoding="raw">\n_')
+        filerootname = os.path.splitext(self.outf)[0]
```
FreddieWitherden: Merge with line 29.

pyfr/writers/paraview.py
```diff
@@ -303,12 +332,19 @@ def _write_vtu_header(args, vtuf, m_inf, s_inf, off):
     nnodes = len(nodes)*m_inf[1][1]
 
     # Standard template for vtk DataArray string
-    darray = '<DataArray Name="%s" type="%s" NumberOfComponents="%s" '\
-             'format="appended" offset="%d"/>\n'
+    if parallel:
+        darray = '<PDataArray Name="%s" type="%s" '\
```
FreddieWitherden: Avoid '\' for line continuation.

PR #49, opened 2015-05-04
Updates to User Guide.

pv101:


doc/src/user_guide.rst
```diff
@@ -660,6 +664,14 @@ Parameterises an exponential solution filter with
 
     *int*
 
+Example::
+
+    [solver-filter]
```
FreddieWitherden: I think this should be `soln`.

pv101:
Thanks, should be fixed now.

PR #51, opened 2015-05-12
Implementation of space and time dependent bc.

iyer-arvind:
This implements the framework for a spatially and temporally
varying boundary condition. As an extension to the implementation
the sub-in-frv, sup-in-fa are now capable of taking in equations
in terms of x, y, z and t. Also, for the ease of usage, the
constants defined in the constants section are also available
in the equations.

### DIFF
--- a/pyfr/solvers/base/elements.py
+++ b/pyfr/solvers/base/elements.py
@@ -55,8 +55,8 @@ class BaseElements(object, metaclass=ABCMeta):
 
         # Apply the operator to the mesh elements and reshape
         plocfpts = np.dot(plocop, eles.reshape(nspts, -1))
-        plocfpts = plocfpts.reshape(self.nfpts, neles, ndims)
-        plocfpts = plocfpts.transpose(1, 2, 0).tolist()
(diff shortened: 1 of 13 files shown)


FreddieWitherden:
I've taken a look over the implementation and it seems to be needlessly complicated compared to what we do for space and time dependent source terms.  Here we manage with just a single kernel that can handle all cases (with/without source terms).  Further, the handling of constants is markedly inferior to what we do for source terms (where the substitution is done at the expression level).

Some of the Python and C code is also rather crude.  Given this I do not think that it is worth doing a line by line review.  My recommendation would be to look how space dependencies are handled when it comes to source terms in the volume and use this is a model to architect the design.

iyer-arvind:
I will do the regex subs instead of #defines. But do you really want me to upload plocs for all the boundaries and modify all the BCs to ignore plocs when not needed?

FreddieWitherden:
Again, see how we handle this with source terms.  Here if you source term does not depend on `xyz` you do not pay the cost.

PR #63, opened 2015-07-01
Add a residual plugin

FreddieWitherden:
This adds a plugin for computing the residual.  It can be invoked by

```
[soln-plugin-residual]
nsteps = 50
file = resid.csv
header = true
```

The request also changes the existing plugins to use `nacptsteps` as opposed to `nsteps` when deciding if to run or not.  This results in more consistent output when an error estimating RK scheme is used.

### DIFF
--- a/pyfr/plugins/__init__.py
+++ b/pyfr/plugins/__init__.py
@@ -3,6 +3,7 @@
 from pyfr.plugins.base import BasePlugin
 from pyfr.plugins.fluidforce import FluidForcePlugin
 from pyfr.plugins.nancheck import NaNCheckPlugin
+from pyfr.plugins.residual import ResidualPlugin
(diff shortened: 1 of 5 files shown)


Aerojspark:
I have tested this plugin for 2-D aerofoil and 3-D wedge cases and it works well. I think we may mention the normalisation by dt in user guide because many steady FVM solvers (especially with local time stepping) use just residual itself.

residual

FreddieWitherden:
The normalisation by `dt` is what takes you from a difference between two solutions (in units of the solution) to something which has the same units as div f -- the residual.  Lets check with Peter to see what makes the most sense.

pv101:
I chatted with Jin Seok about this yesterday. As I understand it the plugin now plots div(F) - which I think is the most sensible thing to plot. So from my perspective fine to accept if everyone else is happy.

PR #67, opened 2015-07-17
Implementation of a spatially varying viscosity.

iyer-arvind:
This implements a spatially varying viscosity. A spatially varying
viscosity is unphysical, but useful to define a sponge region to mop up
the fine turbulent structures before exiting a riemann invariant
boundary. A flow field with smaller gradients is less prone to
spurious reflections from such a boundary.


iyer-arvind:
Will reimplement it and file a fresh pull request after other changes are brought in as discussed.

PR #69, opened 2015-07-24
This implements an averaging plugin.

iyer-arvind:
This implements an averaging plugin which performs windowed averaging
if the solution every 'nsteps' iterations. The output is written to a
separate file with _avg.pyfrs ending. The writing is skipped every
'writeskip' iterations with respect to the writing of the solutions.
This makes use of the existing infrastructure of file writing, hence the
serial-h5 options under soln-output influences this plugin writer too.

### DIFF
--- a/pyfr/integrators/writers.py
+++ b/pyfr/integrators/writers.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 
+from collections import OrderedDict
 import itertools as it
 import os
 
@@ -28,51 +29,45 @@ class H5Writer(BaseIntegrator):
         etypes, shapes = self.system.ele_types, self.system.ele_shapes
 
         # Gather this information onto the root rank
-        eleinfo = comm.gather(zip(etypes, shapes), root=root)
+        eleinfo = comm.gather(tuple(zip(etypes, shapes)), root=root)
 
         # Deciding if parallel
         parallel = (h5py.get_config().mpi and
                     h5py.version.version_tuple >= (2, 5) and
                     not self.cfg.getbool('soln-output', 'serial-h5', False))
 
+        if rank == root:
+            sollist = OrderedDict()
+            for mrank, meleinfo in enumerate(eleinfo):
+                prank = self.rallocs.mprankmap[mrank]
+                for etype, dims in meleinfo:
+                    sollist[etype, prank] = \
+                            (self._get_name_for_soln(etype, prank), dims)
+
+        else:
+            sollist = None
+
+        self.sollist = comm.bcast(sollist, root=root)
+
         if parallel:
             self._write = self._write_parallel
 
-            if rank == root:
-                sollist = []
-                for mrank, meleinfo in enumerate(eleinfo):
-                    prank = self.rallocs.mprankmap[mrank]
-                    sollist.extend(
-                        (self._get_name_for_soln(etype, prank), dims)
-                        for etype, dims in meleinfo
-                    )
-            else:
-                sollist = None
-
-            self.sollist = comm.bcast(sollist, root=root)
         else:
             self._write = self._write_serial
 
             if rank == root:
                 self._mpi_rbufs = mpi_rbufs = []
                 self._mpi_rreqs = mpi_rreqs = []
-                self._mpi_names = mpi_names = []
-                self._loc_names = loc_names = []
 
                 for mrank, meleinfo in enumerate(eleinfo):
                     prank = self.rallocs.mprankmap[mrank]
                     for tag, (etype, dims) in enumerate(meleinfo):
-                        name = self._get_name_for_soln(etype, prank)
-
-                        if mrank == root:
-                            loc_names.append(name)
-                        else:
+                        if mrank != root:
                             rbuf = np.empty(dims, dtype=self.backend.fpdtype)
                             rreq = comm.Recv_init(rbuf, mrank, tag)
 
                             mpi_rbufs.append(rbuf)
                             mpi_rreqs.append(rreq)
-                            mpi_names.append(name)
 
     def output(self, solnmap, stats):
         comm, rank, root = get_comm_rank_root()
@@ -89,7 +84,7 @@ class H5Writer(BaseIntegrator):
         path = self._get_output_path()
 
         # Delegate to _write to do the actual outputting
-        self._write(path, solnmap, metadata)
+        self._write(path, solnmap, self.sollist, metadata)
 
         # Increment the output number
         self.nout += 1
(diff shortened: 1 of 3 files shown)


pyfr/integrators/writers.py
```diff
@@ -28,51 +29,45 @@ def __init__(self, *args, **kwargs):
         etypes, shapes = self.system.ele_types, self.system.ele_shapes
 
         # Gather this information onto the root rank
-        eleinfo = comm.gather(zip(etypes, shapes), root=root)
+        eleinfo = comm.gather(tuple(zip(etypes, shapes)), root=root)
```
FreddieWitherden: Why are we wrapping this in `tuple`?

pyfr/integrators/writers.py
```diff
@@ -28,51 +29,45 @@ def __init__(self, *args, **kwargs):
...
         parallel = (h5py.get_config().mpi and
                     h5py.version.version_tuple >= (2, 5) and
                     not self.cfg.getbool('soln-output', 'serial-h5', False))
 
+        if rank == root:
+            sollist = OrderedDict()
+            for mrank, meleinfo in enumerate(eleinfo):
+                prank = self.rallocs.mprankmap[mrank]
+                for etype, dims in meleinfo:
+                    sollist[etype, prank] = \
```
FreddieWitherden: Use of `\` to continue line.

pyfr/plugins/average.py
```diff
@@ -0,0 +1,63 @@
...
+
+class AveragePlugin(BasePlugin):
+    name = 'average'
+    systems = ['*']
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.nsteps = self.cfg.getint(self.cfgsect, 'nsteps')
+        self.writeskip = self.cfg.getint(self.cfgsect, 'writeskip')
+        self.tout = list(args[0].tout)
```
FreddieWitherden: If you need to use the arguments declare them.  `args[0]` gives no indication about what the argument is.

pyfr/plugins/average.py
```diff
@@ -0,0 +1,63 @@
...
+        self.solnprev = None
+        self.tprev = None
+        self.dtprev = 0
+        self.prevwrite = args[0].tcurr
+        self.avglist = OrderedDict()
+        for (e, p), (name, shape) in args[0].sollist.items():
+            self.avglist[e + '_avg', p] = (name + '_avg', shape)
+
+    def __call__(self, intg):
+        if not len(self.tout):
```
FreddieWitherden: Use of `len` to see if a list of empty.

pyfr/plugins/average.py
```diff
@@ -0,0 +1,63 @@
...
+            return
+
+        if (intg.nacptsteps % self.nsteps == 0) or\
```
FreddieWitherden: Use of `\` for line continuation.

pyfr/plugins/average.py
```diff
@@ -0,0 +1,63 @@
...
+           (intg.tcurr >= self.tout[0]):
+            # if this is not the first iteration, we can integrate
```
FreddieWitherden: Comments should start with capital letters.

pyfr/plugins/average.py
```diff
@@ -0,0 +1,63 @@
...
+            if self.tprev is not None:
+                dt = intg.tcurr - self.tprev
+                for avg, soln in zip(self.averages, self.solnprev):
+                    avg += 0.5*(self.dtprev+dt)*soln
```
FreddieWitherden: Whitespace around `+`.

pyfr/plugins/average.py
```diff
@@ -0,0 +1,63 @@
...
+
+                self.dtprev = dt
+                # if its time to write
```
FreddieWitherden: Comments should start with capital letters.

pyfr/plugins/average.py
```diff
@@ -0,0 +1,63 @@
...
+                if intg.tcurr >= self.tout[0]:
+                    for avg, soln in zip(self.averages, intg.soln):
+                        avg += 0.5*dt*soln
+                    avgnames = list(e + '_avg' for e in intg.system.ele_types)
```
FreddieWitherden: Why is this a list?  And if it does need to be a list why is list comprehension not used?

pyfr/plugins/average.py
```diff
@@ -0,0 +1,63 @@
...
+                    path = intg._get_output_path().\
+                        replace('.pyfrs', '_avg.pyfrs')
+
+                    avgmap = OrderedDict(zip(avgnames, self.averages))
+                    metadata = {'tstart': str(self.prevwrite),
+                                'tend': str(self. tout[0])}
+
+                    intg._write(path, avgmap, self.avglist, metadata)
```
FreddieWitherden: Use of private method from external class.

pyfr/plugins/average.py
```diff
@@ -0,0 +1,63 @@
...
+                if intg.tcurr >= self.tout[0]:
+                    for avg, soln in zip(self.averages, intg.soln):
+                        avg += 0.5*dt*soln
+                    avgnames = list(e + '_avg' for e in intg.system.ele_types)
+                    path = intg._get_output_path().\
+                        replace('.pyfrs', '_avg.pyfrs')
+
+                    avgmap = OrderedDict(zip(avgnames, self.averages))
+                    metadata = {'tstart': str(self.prevwrite),
+                                'tend': str(self. tout[0])}
```
FreddieWitherden: Whitespace.

pyfr/plugins/average.py
```diff
@@ -0,0 +1,63 @@
...
+from pyfr.plugins.base import BasePlugin
+
+
+class AveragePlugin(BasePlugin):
+    name = 'average'
+    systems = ['*']
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.nsteps = self.cfg.getint(self.cfgsect, 'nsteps')
```
FreddieWitherden: Newline after `super()`.

pyfr/plugins/average.py
```diff
@@ -0,0 +1,63 @@
...
+                for avg, soln in zip(self.averages, self.solnprev):
+                    avg += 0.5*(self.dtprev+dt)*soln
+
+                self.dtprev = dt
+                # if its time to write
+                if intg.tcurr >= self.tout[0]:
+                    for avg, soln in zip(self.averages, intg.soln):
+                        avg += 0.5*dt*soln
+                    avgnames = list(e + '_avg' for e in intg.system.ele_types)
+                    path = intg._get_output_path().\
```
FreddieWitherden: Use of `\` for line continuation.

pyfr/integrators/writers.py
iyer-arvind: zip returns an iterator, which vanishes on one read. We need it twice

pyfr/plugins/average.py
iyer-arvind: Should I make _write to write then?

FreddieWitherden: Not that simple.  The write method has a lot of internal semantics attached to it, especially with regards to things like metadata.  It really is not a suitable method to expose.

iyer-arvind: I understand the comment and concern. You are right that _write to write would be a quick but dirty fix. My suggestion would be to liberate H5Writer from the BaseIntegrator parent class, and pass it the system(for information of types), rallocs, mesh (for uuid) and some other info it may need, and have it as an instance. Then use a separate instance in whichever plugin needs it. I suspect that we may have more plugins in future to need the IO framework, and it would be nice to have it independent of the Integrator.

FreddieWitherden: I agree that we need a means for plugins to perform I/O.  However, separating out the writer is probably not a good solution.  Generally you want to write out solutions at specific times.  The integrator, however, needs to be aware of these times so that it can take suitable time steps.

What we need is a means for plugins to firstly declare that they want to perform I/O; and when they want to perform it.  The integrator can then keep track of this (as it currently does for regular I/O).  Next a method should be added to the plugins which is called by the Integrator when it wants them to do I/O.  As part of this we should seek to completely decouple the time dumping from the Integrator.  Or, in other words, time dumping should be performed by a plugin.

iyer-arvind:
I am implementing Freddie's suggestion of how the plugins should do io, will submit a fresh pull request after that.

FreddieWitherden:
Okay.  Just as an aside if I were writing this I would expect to end up rewriting/redesigning my solution two to three times before getting something I was happy with.  So don't just submit the first thing that works.  Try other solutions and see which is the most flexible and easiest to implement.

PR #70, opened 2015-08-13
Improve filtering for tensor product elements.

FreddieWitherden:
This resolves an issue around filtering tensor product elements that could result in the filtering being more aggressive than necessary.

### DIFF
--- a/pyfr/shapes.py
+++ b/pyfr/shapes.py
@@ -140,7 +140,7 @@ class BaseShape(object):
     def m11(self):
         ub = self.ubasis
 
-        n = ub.order
+        n = max(ub.degrees)
         ncut = self.cfg.getint('soln-filter', 'cutoff')
         order = self.cfg.getint('soln-filter', 'order')
         alpha = self.cfg.getfloat('soln-filter', 'alpha')
@@ -148,7 +148,7 @@ class BaseShape(object):
         A = np.ones(self.nupts)
         for i, d in enumerate(ub.degrees):
             if d >= ncut < n:
-                A[i] = exp(-alpha*(float(d - ncut)/(n - ncut))**order)
+                A[i] = exp(-alpha*((d - ncut)/(n - ncut))**order)
 
         return np.linalg.solve(ub.vdm, A[:,None]*ub.vdm).T
 


pv101:
Sorry for delay on this my side. To what extent has it been tested?

FreddieWitherden:
I have manually looked at the filter coefficients and they now appear to be correct.  I've also washed out an Euler vortex on a quad mesh.  However, some more testing would be nice if anyone has any cases which depend on filtering and use either quads/hexes.

Aerojspark:
I have tested this pull request with my NACA0021 simulation. Previously I used nsteps=50, cutoff=1, order=16 and alpha=36.0 for P3 simulation and it worked well. However, the simulation was crashed with this pull request even when I reduced nsteps as 1. I can adjust other coefficient but it may be too diffusive. I am wondering that we may not remove x^3 or y^3 modal coefficient for P3 quad/hex even with cutoff=0 filter.

I found that Gassner also used modal filter for DG quad/hexa mesh but he applied 1D filter matrix in a tensor product fashion. I think we may need to test this pull request a little bit more.

FreddieWitherden:
The crux of the filter is:

    A[i] = exp(-alpha*((d - ncut)/(n - ncut))**order)

and this request modifies the definition of `n` such that it will in general be larger (3 times larger in the case of hexes).  Hence `n - ncut` is larger for the same cutoff and so `1/(n - ncut)` is smaller.  This reduces the effective strength of the filter for the same values of `alpha` and `order`.

I would therefore be very surprised if you did not have to change the coefficients to achieve a similar result.

FreddieWitherden:
Are we okay to merge?

pv101:
I think it is fine to merge this now. Out of interest, why has the 'float' gone? Is this because we are Python 3? Or some other reason?

FreddieWitherden:
Yes, the numerator and denominator are both integers and in Python 2 `int/int => int` and so it is necessary to cast one of them to a float.  This is no longer required in Python 3 where `int/int => float`.

pv101:
OK, thought so. Please merge.

PR #71, opened 2015-09-01
This implements the solution writer as a plugin.

iyer-arvind:
The implementation of the solution writer as a plugin comes with the following changes:
 - The integrator is now a merge of the stepper and controller alone and not of the writer,
 - The hdf5 writer is shifted to the root directory,
 - The plugins can now set a 'dt', the interval after which the integrator is guaranteed to stop,
 - The t0 is renamed to 'tstart' in the solver-time-integrator section of the ini file and is accompanied by the 'tend', which determines when the solver stops.

### DIFF
--- a/examples/couette_flow_2d/couette_flow_2d.ini
+++ b/examples/couette_flow_2d/couette_flow_2d.ini
@@ -23,7 +23,8 @@ order = 2
 [solver-time-integrator]
 scheme = rk4
 controller = none
-t0 = 0.0
+tstart = 0.0
+tend = 4
 dt = 0.00004
 
 [solver-interfaces]
@@ -43,10 +44,10 @@ soln-pts = gauss-legendre
 [soln-plugin-nancheck]
 nsteps = 50
 
-[soln-output]
+[soln-plugin-solnwriter]
+dt-out = 0.1
 basedir = .
-basename = couette_flow_2d_%(t).2f
-times = range(0, 4, 41)
+basename = couette_flow_2d-%(t).2f
 
 [soln-bcs-bcwallupper]
 type = no-slp-isot-wall
--- a/pyfr/h5writer.py
+++ b/pyfr/h5writer.py
@@ -6,50 +6,51 @@ import os
 import h5py
 import numpy as np
 
-from pyfr.integrators.base import BaseIntegrator
 from pyfr.mpiutil import get_comm_rank_root
 
 
-class H5Writer(BaseIntegrator):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+class H5Writer(object):
+    def __init__(self, intg, basedir, basename, prefix):
+        # Base output directory and file name and prefix for the data-set.
+        self._basedir = basedir
+        self._basename = basename
+        self._prefix = prefix
 
-        # Base output directory and file name
-        self._basedir = self.cfg.getpath('soln-output', 'basedir', '.')
-        self._basename = self.cfg.get('soln-output', 'basename', raw=True)
-
-        # Output counter (incremented each time output() is called)
+        # Output counter (incremented each time write() is called)
         self.nout = 0
+
+        # Copy the float type
+        self.fpdtype = intg.backend.fpdtype
 
         # MPI info
         comm, rank, root = get_comm_rank_root()
 
         # Get the type and shape of each element in the partition
-        etypes, shapes = self.system.ele_types, self.system.ele_shapes
+        etypes, shapes = intg.system.ele_types, intg.system.ele_shapes
 
-        # Gather this information onto the root rank
-        eleinfo = comm.gather(zip(etypes, shapes), root=root)
+        # Gather this information and distribute
+        eleinfo = comm.allgather(zip(etypes, shapes))
 
         # Deciding if parallel
         parallel = (h5py.get_config().mpi and
                     h5py.version.version_tuple >= (2, 5) and
-                    not self.cfg.getbool('soln-output', 'serial-h5', False))
+                    'PYFR_FORCE_SERIAL_HDF5' not in os.environ)
 
         if parallel:
             self._write = self._write_parallel
+            self._loc_names = loc_names = []
+            self._global_shape_list = []
 
-            if rank == root:
-                sollist = []
-                for mrank, meleinfo in enumerate(eleinfo):
-                    prank = self.rallocs.mprankmap[mrank]
-                    sollist.extend(
-                        (self._get_name_for_soln(etype, prank), dims)
-                        for etype, dims in meleinfo
-                    )
-            else:
-                sollist = None
+            for mrank, meleinfo in enumerate(eleinfo):
+                prank = intg.rallocs.mprankmap[mrank]
 
-            self.sollist = comm.bcast(sollist, root=root)
+                # Loop over all element types across all ranks
+                for etype, shape in meleinfo:
+                    name = self._get_name_for_data(etype, prank)
+                    self._global_shape_list.append((name, shape))
+
+                    if rank == mrank:
+                        loc_names.append(name)
         else:
             self._write = self._write_serial
 
@@ -60,43 +61,33 @@ class H5Writer(BaseIntegrator):
                 self._loc_names = loc_names = []
 
                 for mrank, meleinfo in enumerate(eleinfo):
-                    prank = self.rallocs.mprankmap[mrank]
-                    for tag, (etype, dims) in enumerate(meleinfo):
-                        name = self._get_name_for_soln(etype, prank)
+                    prank = intg.rallocs.mprankmap[mrank]
+                    for tag, (etype, shape) in enumerate(meleinfo):
+                        name = self._get_name_for_data(etype, prank)
 
                         if mrank == root:
                             loc_names.append(name)
                         else:
-                            rbuf = np.empty(dims, dtype=self.backend.fpdtype)
+                            rbuf = np.empty(shape, dtype=self.fpdtype)
                             rreq = comm.Recv_init(rbuf, mrank, tag)
 
                             mpi_rbufs.append(rbuf)
                             mpi_rreqs.append(rreq)
                             mpi_names.append(name)
 
-    def output(self, solnmap, stats):
-        comm, rank, root = get_comm_rank_root()
-
-        # Convert the config and stats objects to strings
-        if rank == root:
-            metadata = dict(config=self.cfg.tostr(),
-                            stats=stats.tostr(),
-                            mesh_uuid=self._mesh_uuid)
-        else:
-            metadata = None
-
+    def write(self, data, metadata, tcurr):
         # Determine the output path
-        path = self._get_output_path()
+        path = self._get_output_path(tcurr)
 
         # Delegate to _write to do the actual outputting
-        self._write(path, solnmap, metadata)
+        self._write(path, data, metadata)
 
         # Increment the output number
         self.nout += 1
 
-    def _get_output_path(self):
+    def _get_output_path(self, tcurr):
         # Substitute %(t) and %(n) for the current time and output number
-        fname = self._basename % dict(t=self.tcurr, n=self.nout)
+        fname = self._basename % dict(t=tcurr, n=self.nout)
 
         # Append the '.pyfrs' extension
         if not fname.endswith('.pyfrs'):
@@ -104,23 +95,21 @@ class H5Writer(BaseIntegrator):
 
         return os.path.join(self._basedir, fname)
 
-    def _get_name_for_soln(self, etype, prank=None):
-        prank = prank or self.rallocs.prank
-        return 'soln_{}_p{}'.format(etype, prank)
+    def _get_name_for_data(self, etype, prank):
+        return '{}_{}_p{}'.format(self._prefix, etype, prank)
 
-    def _write_parallel(self, path, solnmap, metadata):
+    def _write_parallel(self, path, data, metadata):
         comm, rank, root = get_comm_rank_root()
 
         with h5py.File(path, 'w', driver='mpio', comm=comm) as h5file:
-            smap = {}
-            for name, shape in self.sollist:
-                smap[name] = h5file.create_dataset(
-                    name, shape, dtype=self.backend.fpdtype
+            dmap = {}
+            for name, shape in self._global_shape_list:
+                dmap[name] = h5file.create_dataset(
+                    name, shape, dtype=self.fpdtype
                 )
 
-            for e, sol in solnmap.items():
-                s = self._get_name_for_soln(e, self.rallocs.prank)
-                smap[s][:] = sol
+            for s, dat in zip(self._loc_names, data):
+                dmap[s][:] = dat
 
             # Metadata information has to be transferred to all the ranks
             if rank == root:
@@ -135,29 +124,29 @@ class H5Writer(BaseIntegrator):
                 if rank == root:
                     d.write_direct(np.array(metadata[name], dtype='S'))
 
-    def _write_serial(self, path, solnmap, metadata):
+    def _write_serial(self, path, data, metadata):
         from mpi4py import MPI
 
         comm, rank, root = get_comm_rank_root()
 
         if rank != root:
-            for tag, buf in enumerate(solnmap.values()):
+            for tag, buf in enumerate(data):
                 comm.Send(buf.copy(), root, tag)
         else:
-            # Recv all of the non-local solution mats
+            # Recv all of the non-local data
             MPI.Prequest.Startall(self._mpi_rreqs)
             MPI.Prequest.Waitall(self._mpi_rreqs)
 
             # Combine local and MPI data
             names = it.chain(self._loc_names, self._mpi_names)
-            solns = it.chain(solnmap.values(), self._mpi_rbufs)
+            dats = it.chain(data, self._mpi_rbufs)
 
             # Convert any metadata to ASCII
             metadata = {k: np.array(v, dtype='S')
                         for k, v in metadata.items()}
 
             # Create the output dictionary
-            outdict = dict(zip(names, solns), **metadata)
+            outdict = dict(zip(names, dats), **metadata)
 
             with h5py.File(path, 'w') as h5file:
                 for k, v in outdict.items():
--- a/pyfr/integrators/base.py
+++ b/pyfr/integrators/base.py
@@ -1,11 +1,9 @@
 # -*- coding: utf-8 -*-
 
 from abc import ABCMeta, abstractmethod, abstractproperty
-from collections import OrderedDict
 
 from pyfr.inifile import Inifile
 from pyfr.mpiutil import get_comm_rank_root, get_mpi
-from pyfr.nputil import range_eval
 from pyfr.util import proxylist
 
 
@@ -20,11 +18,8 @@ class BaseIntegrator(object, metaclass=ABCMeta):
             raise TypeError('Incompatible stepper/controller combination')
 
         # Start time
-        self.tstart = cfg.getfloat('solver-time-integrator', 't0', 0.0)
-
-        # Output times
-        self.tout = sorted(range_eval(cfg.get('soln-output', 'times')))
-        self.tend = self.tout[-1]
+        self.tstart = cfg.getfloat('solver-time-integrator', 'tstart', 0.0)
+        self.tend = cfg.getfloat('solver-time-integrator', 'tend')
 
         # Current time; defaults to tstart unless resuming a simulation
         if initsoln is None or 'stats' not in initsoln:
@@ -33,12 +28,7 @@ class BaseIntegrator(object, metaclass=ABCMeta):
             stats = Inifile(initsoln['stats'])
             self.tcurr = stats.getfloat('solver-time-integrator', 'tcurr')
 
-            # Cull already written output times
-            self.tout = [t for t in self.tout if t > self.tcurr]
-
-        # Ensure no time steps are in the past
-        if self.tout[0] < self.tcurr:
-            raise ValueError('Output times must be in the future')
+        self.tlist = [self.tend]
 
         # Determine the amount of temp storage required by thus method
         nreg = self._stepper_nregs
@@ -47,7 +37,7 @@ class BaseIntegrator(object, metaclass=ABCMeta):
         self.system = systemcls(backend, rallocs, mesh, initsoln, nreg, cfg)
 
         # Extract the UUID of the mesh (to be saved with solutions)
-        self._mesh_uuid = mesh['mesh_uuid']
+        self.mesh_uuid = mesh['mesh_uuid']
 
         # Get a queue for subclasses to use
         self._queue = backend.queue()
@@ -83,10 +73,6 @@ class BaseIntegrator(object, metaclass=ABCMeta):
     def advance_to(self, t):
         pass
 
-    @abstractmethod
-    def output(self, solns, stats):
-        pass
-
     @abstractproperty
     def _controller_needs_errest(self):
         pass
@@ -108,19 +94,9 @@ class BaseIntegrator(object, metaclass=ABCMeta):
         pass
 
     def run(self):
-        for t in self.tout:
+        for t in self.tlist:
             # Advance to time t
-            solns = self.advance_to(t)
-
-            # Map solutions to elements types
-            solnmap = OrderedDict(zip(self.system.ele_types, solns))
-
-            # Collect statistics
-            stats = Inifile()
-            self.collect_stats(stats)
-
-            # Output
-            self.output(solnmap, stats)
+            self.advance_to(t)
 
     def collect_stats(self, stats):
         stats.set('solver-time-integrator', 'tcurr', self.tcurr)
--- a/pyfr/integrators/controllers.py
+++ b/pyfr/integrators/controllers.py
@@ -3,6 +3,8 @@
 import math
 import re
 
+import numpy as np
+
 from pyfr.integrators.base import BaseIntegrator
 from pyfr.mpiutil import get_comm_rank_root, get_mpi
 from pyfr.plugins import get_plugin
@@ -15,7 +17,7 @@ class BaseController(BaseIntegrator):
 
         # Current and minimum time steps
         self._dt = self.cfg.getfloat('solver-time-integrator', 'dt')
-        self._dtmin = 1.0e-14
+        self.dtmin = 1.0e-14
 
         # Solution filtering frequency
         self._fnsteps = self.cfg.getint('soln-filter', 'nsteps', '0')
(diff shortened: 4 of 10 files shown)


examples/couette_flow_2d/couette_flow_2d.ini
```diff
@@ -57,6 +53,12 @@ u = Uw
 type = no-slp-isot-wall
 cpTw = cp*Tw
 
+[soln-plugin-solutionwriter]
+dt = 0.1
+basedir = .
+basename = couette_flow_2d-%(t).1f
+
+
```
FreddieWitherden: Extra newline.

pyfr/integrators/controllers.py
```diff
@@ -39,6 +41,7 @@ def __init__(self, *args, **kwargs):
             m = re.match('soln-plugin-(.+?)(?:-(.+))?$', s)
             if m:
                 cfgsect, name, suffix = m.group(0), m.group(1), m.group(2)
+                print('Loading plugin: {}'.format(name))
```
FreddieWitherden: Print statement.

pyfr/integrators/controllers.py
```diff
@@ -49,6 +52,14 @@ def __init__(self, *args, **kwargs):
         # Delete the memory-intensive elements map from the system
         del self.system.ele_map
 
+    def set_dt(self, dt):
+        tlist = [self.tlist, np.arange(self.tcurr, self.tend, dt)]
+
+        # List of sorted unique times
+        digits = int(np.log10(1.0/self.dtmin))
+        self.tlist = np.sort(np.unique(np.around(np.concatenate(tlist),
```
FreddieWitherden: This will not work.

pyfr/plugins/solutionwriter.py
```diff
@@ -0,0 +1,40 @@
...
+from pyfr.inifile import Inifile
+from pyfr.plugins.base import BasePlugin
+
+
+class SolutionWriterPlugin(BasePlugin, H5Writer):
+    name = 'solutionwriter'
+    systems = ['*']
+
+    def __init__(self, intg, cfgsect, *args, **kwargs):
+        BasePlugin.__init__(self, intg, cfgsect)
```
FreddieWitherden: Old style class initialization.  Use super, like the rest of PyFR.

pyfr/plugins/solutionwriter.py
```diff
@@ -0,0 +1,40 @@
...
+        H5Writer.__init__(self, intg, cfgsect)
+        self.dt = self.cfg.getfloat(self.cfgsect, 'dt', -1)
+        self.tprev = intg.tcurr - self.dt
```
FreddieWitherden: Working in terms of `tnext` as opposed to `tprev` will probably make your life easier.

pyfr/plugins/solutionwriter.py
```diff
@@ -0,0 +1,40 @@
...
+
+from collections import OrderedDict
+
+from pyfr.h5writer import H5Writer
+from pyfr.inifile import Inifile
+from pyfr.plugins.base import BasePlugin
+
+
+class SolutionWriterPlugin(BasePlugin, H5Writer):
+    name = 'solutionwriter'
```
FreddieWitherden: For consistency with the existing config files it is probably better to use the abbreviation `soln`.

pyfr/integrators/controllers.py
```diff
@@ -49,6 +52,14 @@ def __init__(self, *args, **kwargs):
         # Delete the memory-intensive elements map from the system
         del self.system.ele_map
 
+    def set_dt(self, dt):
```
FreddieWitherden: Misleading method name.  It does nothing close to setting `dt`.  Rather, it appends the times when we should aim to invoke plugins.

pyfr/integrators/base.py
```diff
@@ -33,12 +30,7 @@ def __init__(self, backend, systemcls, rallocs, mesh, initsoln, cfg):
             stats = Inifile(initsoln['stats'])
             self.tcurr = stats.getfloat('solver-time-integrator', 'tcurr')
 
-            # Cull already written output times
-            self.tout = [t for t in self.tout if t > self.tcurr]
-
-        # Ensure no time steps are in the past
-        if self.tout[0] < self.tcurr:
-            raise ValueError('Output times must be in the future')
+        self.tlist = np.array([self.tend])
```
FreddieWitherden: This should not be a NumPy array.

pyfr/plugins/solutionwriter.py
```diff
@@ -0,0 +1,40 @@
...
+                     for e in intg.system.ele_types)
+
+        solnmap = OrderedDict(zip(solnnames, intg.soln))
+
+        stats = Inifile()
+        intg.collect_stats(stats)
+
+        metadata = dict(config=self.cfg.tostr(),
+                        stats=stats.tostr(),
+                        mesh_uuid=intg._mesh_uuid)
```
FreddieWitherden: Access of private member variable.

pyfr/h5writer.py
```diff
@@ -0,0 +1,147 @@
...
+        # Deciding if parallel
+        parallel = (h5py.get_config().mpi and
+                    h5py.version.version_tuple >= (2, 5) and
+                    not cfg.getbool(cfgsect, 'serial-h5', False))
+
+        self.fpdtype = intg.backend.fpdtype
+
+        if parallel:
+            self._write = self._write_parallel
+
```
FreddieWitherden: Extra newline.

pyfr/integrators/controllers.py
iyer-arvind: Would you please elaborate on this?

pyfr/integrators/controllers.py
iyer-arvind: agreed, ensure_tstep is better?

pyfr/integrators/base.py
iyer-arvind: Kindly let me know the reason.

pyfr/plugins/solutionwriter.py
iyer-arvind: There was a problem like this:
```python
class A(object): 
    def __init__(self,a):
        print('A')
        super().__init__(a) # Error on instantiation of e
        # super().__init__() # Error on instantiation of d

class B(object):
    def __init__(self, a):
        print('B')
        super().__init__(a)        

class D(A,B):
    def __init__(self, a):
        print('D')
        super().__init__(5)

class E(A):
    def __init__(self, a):
        print('E')
        super().__init__(5)

d = D(5)
e = E(5)

```
```text
A  B          A
| /       OR  |
D             E
```

I could not locate this situation in rest of PyFR.  
(Diamond inheritance is not an issue though!).

FreddieWitherden: In your above example why are you calling `super().__init(...)` for classes that inherit from `object`?  This will attempt to call the constructor for `object` which takes no parameters with a bunch of parameters, hence the error.

pyfr/integrators/controllers.py
FreddieWitherden: If you want to see if two things are within a tolerance of each other rounding is not the correct solution.  A good example of this is in the fuzzy sorting code.  A better solution is along the lines of:

```
for t in np.arange(self.tcurr, self.tend, dt):
    if all(abs(s - t) >= self.dtmin for s in self.tlist):
        self.tlist.append(s)
        
self.tlist.sort()
```
It is also quite easy to extend the method to accept a list of times as opposed to a step.

pyfr/plugins/solutionwriter.py
iyer-arvind: Even that will give an error because D.__init__ calls A.__init__ which calls B.__init__ . 
If A.__init__ does not call super with args, B cannot get the args, and fails. 
If it calls with args, when B is not herited it tries to call object with args and fails.

pyfr/integrators/base.py
FreddieWitherden: If you need a list of values and you intend to extend the list then the most sensible type is a `list`.  Moreover, NumPy arrays can not be extended, lists can.

iyer-arvind: Ok... agreed.

pyfr/plugins/solutionwriter.py
FreddieWitherden: See `EulerElements` where we have this exact situation.  Super works fine.

Although here it looks like inheritance is the wrong solution.  What you want is composition.

iyer-arvind: Problem of `EulerElements` is not the exact same situation, i think, because `__init__` is implemented only by the `BaseElements` class. 

I believe you are referring to the `EulerElements::set_backend` . 
Here the issue does not arise because 
1. The `BaseFluidElements` call does not implement `set_backend`, so is passed up via `BaseAdvectionElements` and finally `BaseElements` so the order is perfectly clear,
2. `object` does not implement set_backend, so no problem differing function signatures.

Composition, however,  will work without any problem. Will switch to that.

FreddieWitherden: Just because there is not an explicit `__init__` method does not change the principle.  Translating how we handle `EulerElements` into your example:

```
class A(object):
    def __init__(self,a):
        print('A')

class B(object):
    def __init__(self, a):
        print('B')
        super().__init__(a)

class D(B, A):
    def __init__(self, a):
        print('D')
        super().__init__(a)

class E(A):
    def __init__(self, a):
        print('E')
        super().__init__(a)

print('Constructing D')
d = D(5)

print('Constructing E')
e = E(5)
```
Which works as expected.  The key is in the ordering of the classes when you perform the inheritance.  You can demonstrate this for yourself by switching `BaseFluidElements` with `BaseAdvectionElements` and seeing what happens.

iyer-arvind: Its not just the right order, its kind of hacky. 
1. Even if `A` and `B` both are similar objects inheriting only from `object`, B calls `super.__init__(a)` whereas `A` does not.
2. There is no way of creating an object of class `B`.

Any way, I have switched to composition. soln-writer will have a _writer memeber object.

pyfr/plugins/solnwriter.py
```diff
@@ -0,0 +1,40 @@
...
+from pyfr.inifile import Inifile
+from pyfr.plugins.base import BasePlugin
+
+
+class SolnWriterPlugin(BasePlugin):
+    name = 'solnwriter'
+    systems = ['*']
+
+    def __init__(self, intg, cfgsect, *args, **kwargs):
+        BasePlugin.__init__(self, intg, cfgsect)
```
FreddieWitherden: Still not using super...

pyfr/plugins/solnwriter.py
```diff
@@ -0,0 +1,40 @@
...
+from pyfr.h5writer import H5Writer
+from pyfr.inifile import Inifile
+from pyfr.plugins.base import BasePlugin
+
+
+class SolnWriterPlugin(BasePlugin):
+    name = 'solnwriter'
+    systems = ['*']
+
+    def __init__(self, intg, cfgsect, *args, **kwargs):
```
FreddieWitherden: Either specify all of the parameters (including suffix) or just `*args` and `**kwargs` and let the parent constructor handle them for you.

pyfr/plugins/solnwriter.py
```diff
@@ -0,0 +1,40 @@
...
+        stats = Inifile()
+        intg.collect_stats(stats)
+
+        metadata = dict(config=self.cfg.tostr(),
+                        stats=stats.tostr(),
+                        mesh_uuid=intg.mesh_uuid)
+
+        self._writer.write(solnmap, metadata, intg.tcurr)
+
+        self.tnext = intg.tcurr + self.tsteps
```
FreddieWitherden: Missing newline at end of file.

pyfr/integrators/controllers.py
```diff
@@ -49,6 +51,12 @@ def __init__(self, *args, **kwargs):
         # Delete the memory-intensive elements map from the system
         del self.system.ele_map
 
+    def ensure_tstep(self, tstep):
+        for t in np.arange(self.tcurr, self.tend, tstep):
+            if all(abs(s - t) > self.dtmin for s in self.tlist):
+                self.tlist.append(t)
+        self.tlist.sort()
```
FreddieWitherden: Newline after loop.

pyfr/h5writer.py
```diff
@@ -0,0 +1,146 @@
...
+        # Append the '.pyfrs' extension
+        if not fname.endswith('.pyfrs'):
+            fname += '.pyfrs'
+
+        return os.path.join(self._basedir, fname)
+
+    def _write_parallel(self, path, datamap, metadata):
+        comm, rank, root = get_comm_rank_root()
+        # Collect shapes of current rank
```
FreddieWitherden: Newline before comment.

pyfr/h5writer.py
```diff
@@ -0,0 +1,146 @@
...
+import os
+
+import h5py
+import numpy as np
+
+from pyfr.mpiutil import get_comm_rank_root
+
+
+class H5Writer(object):
+    def __init__(self, intg, cfgsect, *args, **kwargs):
```
FreddieWitherden: Random `*args` and `**kwargs`.

pyfr/h5writer.py
```diff
@@ -0,0 +1,146 @@
...
+                    h5py.version.version_tuple >= (2, 5) and
+                    not cfg.getbool(cfgsect, 'serial-h5', False))
+
+        self.fpdtype = intg.backend.fpdtype
+
+        if parallel:
+            self._write = self._write_parallel
+        else:
+            self._write = self._write_serial
+            if rank == root:
```
FreddieWitherden: Missing newline before `if`.

pyfr/h5writer.py
```diff
@@ -0,0 +1,146 @@
...
+                comm.Send(buf.copy(), root, tag)
+        else:
+            names = [i for p in names for i in p]
+
+            # Recv all of the non-local solution mats
+            MPI.Prequest.Startall(self._mpi_rreqs)
+            MPI.Prequest.Waitall(self._mpi_rreqs)
+
+            # Combine local and MPI data
+
```
FreddieWitherden: Random newline.

pyfr/h5writer.py
```diff
@@ -0,0 +1,146 @@
...
+            solns = it.chain(datamap.values(), self._mpi_rbufs)
+
+            # Convert any metadata to ASCII
+            metadata = {k: np.array(v, dtype='S')
+                        for k, v in metadata.items()}
+
+            # Create the output dictionary
+            outdict.update(dict(zip(names, solns), **metadata))
+
+        with h5py.File(path, 'w') as h5file:
```
FreddieWitherden: This indentation looks badly wrong.  Here it appears as if **all** ranks will open the file and try to do output even though this is the serial code path.  This is also not what the current writer does.

pyfr/h5writer.py
```diff
@@ -0,0 +1,146 @@
...
+        comm, rank, root = get_comm_rank_root()
+        outdict = {}
+
+        names = comm.gather(tuple(datamap.keys()), root=root)
+
+        if rank != root:
+            for tag, buf in enumerate(datamap.values()):
+                comm.Send(buf.copy(), root, tag)
+        else:
+            names = [i for p in names for i in p]
```
FreddieWitherden: Why are we doing this here rather than `it.chain` as before?

pyfr/h5writer.py
```diff
@@ -0,0 +1,146 @@
...
+        shapes = [(k, v.shape) for k, v in datamap.items()]
+        all_shapes = comm.gather(shapes, root=root)
+
+        if rank == root:
+            shape_dict = dict(s for p in all_shapes for s in p)
+        else:
+            shape_dict = None
+
+        # Dictionary of shapes from all ranks
+        shape_dict = comm.bcast(shape_dict)
```
FreddieWitherden: `comm.gather` followed by `comm.bcast` suggest we should be using `comm.allgather`.  Also, there is no guarantee that `shape_dict` will have the same iteration order across ranks.  `comm.allgather` a list and then construct an `OrderedDict`.

FreddieWitherden:
Why exactly in the write parallel method do we now exchange data about the array shapes?  We compute this in the constructor (`eleinfo`) and it does not change between iterations.

pyfr/h5writer.py
iyer-arvind: To simplify this and to do the next comment, it is best if the h5writer knows at the init phase what the names of the data sets. We have a standard for the names of the datasets : <prefix>_<elem_name>_p<prank>. I am make this as a standard and moving it to the H5Writer itself, so that most of this can be done just once at its init.

iyer-arvind: I guess its best for me to recode h5writer.py . 
It has evolved so much and I think I have completed a full circle.
I will pick the v1.0.0 file and make minimal changes.

pyfr/h5writer.py
```diff
@@ -10,46 +10,48 @@
...
-        self._basename = self.cfg.get('soln-output', 'basename', raw=True)
+        self._basedir = cfg.getpath(cfgsect, 'basedir', '.')
+        self._basename = cfg.get(cfgsect, 'basename', raw=True)
 
         # Output counter (incremented each time output() is called)
         self.nout = 0
 
+        # The float type and the mesh uuid
+        self.fpdtype = intg.backend.fpdtype
+        self.mesh_uuid = intg.mesh_uuid
```
FreddieWitherden: Why do we save this?

pyfr/h5writer.py
```diff
@@ -10,46 +10,48 @@
...
-        super().__init__(*args, **kwargs)
+class H5Writer(object):
+    def __init__(self, intg, cfgsect, prefix):
+        cfg = intg.cfg
+        self.prefix = prefix
 
         # Base output directory and file name
-        self._basedir = self.cfg.getpath('soln-output', 'basedir', '.')
-        self._basename = self.cfg.get('soln-output', 'basename', raw=True)
+        self._basedir = cfg.getpath(cfgsect, 'basedir', '.')
```
FreddieWitherden: Decouple from the specifics of our config file format.  These should be parameters.

pyfr/plugins/solnwriter.py
```diff
@@ -0,0 +1,37 @@
...
+from pyfr.plugins.base import BasePlugin
+
+
+class SolnWriterPlugin(BasePlugin):
+    name = 'solnwriter'
+    systems = ['*']
+
+    def __init__(self, intg, cfgsect, suffix=None):
+        super().__init__(intg, cfgsect, suffix)
+        basedir = self.cfg.getpath(cfgsect, 'basedir', '.')
```
FreddieWitherden: Add newline.

pyfr/plugins/solnwriter.py
```diff
@@ -0,0 +1,37 @@
...
+        basename = self.cfg.get(cfgsect, 'basename', raw=True)
+        serial_h5 = self.cfg.getbool(cfgsect, 'serial-h5', False)
+
+        self._writer = H5Writer(intg, basedir, basename, 'soln', serial_h5)
+        self.tsteps = self.cfg.getfloat(self.cfgsect, 'tsteps')
+        self.tnext = intg.tcurr
+        intg.ensure_tstep(self.tsteps)
```
FreddieWitherden: Add newline before.

pyfr/h5writer.py
```diff
@@ -10,46 +10,48 @@
...
+    def __init__(self, intg, basedir, basename, prefix, serial_h5=False):
+        # Base output directory and file name and prefix for the data-set.
+        self._basedir = basedir
+        self._basename = basename
+        self._prefix = prefix
 
         # Output counter (incremented each time output() is called)
         self.nout = 0
 
+        # The float type and the mesh uuid
```
FreddieWitherden: Misleading comment.

pyfr/h5writer.py
```diff
@@ -10,46 +10,48 @@
...
-        self._basedir = self.cfg.getpath('soln-output', 'basedir', '.')
-        self._basename = self.cfg.get('soln-output', 'basename', raw=True)
+class H5Writer(object):
+    def __init__(self, intg, basedir, basename, prefix, serial_h5=False):
+        # Base output directory and file name and prefix for the data-set.
+        self._basedir = basedir
+        self._basename = basename
+        self._prefix = prefix
 
         # Output counter (incremented each time output() is called)
```
FreddieWitherden: Misleading comment.

pyfr/h5writer.py
```diff
@@ -10,46 +10,48 @@
...
         # Deciding if parallel
         parallel = (h5py.get_config().mpi and
                     h5py.version.version_tuple >= (2, 5) and
-                    not self.cfg.getbool('soln-output', 'serial-h5', False))
+                    not serial_h5)
 
         if parallel:
             self._write = self._write_parallel
+            self._loc_names = loc_names = []
+            self.all_datalist = []
```
FreddieWitherden: Not so sure about this name.  Also, the member is public whereas most of the other variables appear to be 'private'.

pyfr/integrators/controllers.py
```diff
@@ -49,6 +51,13 @@ def __init__(self, *args, **kwargs):
         # Delete the memory-intensive elements map from the system
         del self.system.ele_map
 
+    def ensure_tstep(self, tstep):
```
pv101: I understand tstep to be the separation between invocations of the plugin, and ensure_tstep to construct a list of the times that the solver should hit, so the plugin can be invoked at the appropriate times. Therefore, could tstep be renamed dtinvoke and could ensure_tstep be renamed hit_dtinvokes?

iyer-arvind: call_plugin_dt ?

pv101: OK. And just replace tstep with dt?

Peter

Dr Peter Vincent MSci ARCS DIC PhD
Senior Lecturer and EPSRC Early Career Fellow
Department of Aeronautics
Imperial College London
South Kensington
London
SW7 2AZ
UK

web: www.imperial.ac.uk/aeronautics/research/vincentlab<http://www.imperial.ac.uk/aeronautics/research/vincentlab>
twitter: @Vincent_Lab<https://twitter.com/Vincent_Lab>





On 1 Oct 2015, at 16:27, iyer-arvind <notifications@github.com<mailto:notifications@github.com>> wrote:


In pyfr/integrators/controllers.py<https://github.com/vincentlab/PyFR/pull/71#discussion_r40926736>:

> @@ -49,6 +51,13 @@ def __init__(self, *args, **kwargs):
>          # Delete the memory-intensive elements map from the system
>          del self.system.ele_map
>
> +    def ensure_tstep(self, tstep):


call_plugin_dt ?

iyer-arvind: OK!

pyfr/plugins/solnwriter.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def __init__(self, intg, cfgsect, suffix=None):
+        super().__init__(intg, cfgsect, suffix)
+
+        basedir = self.cfg.getpath(cfgsect, 'basedir', '.')
+        basename = self.cfg.get(cfgsect, 'basename', raw=True)
+        serial_h5 = self.cfg.getbool(cfgsect, 'serial-h5', False)
+
+        self._writer = H5Writer(intg, basedir, basename, 'soln', serial_h5)
+        self.tsteps = self.cfg.getfloat(self.cfgsect, 'tsteps')
```
FreddieWitherden: `tstep` might be better or `dt`.

pyfr/plugins/solnwriter.py
```diff
@@ -0,0 +1,39 @@
...
+        self.tnext = intg.tcurr
+
+        intg.call_plugin_dt(self.tsteps)
```
FreddieWitherden: Comment above.

FreddieWitherden:
I think we're almost ready to go.

pyfr/plugins/solnwriter.py
iyer-arvind: Like you, I am not in favour of 'dt' variable by itself because name clashes with integration dt.
Along with call_plugin_ I guessed the name is explicitly clear. 

Besides nsteps rhymes with tsteps, make it more obvious (-- poetic logic?)

If we prefer, we could name call_plugin_dt to 
call_plugins_at_tstep /
call_plugin_tstep /even 
plugin_tstep ?

FreddieWitherden:
I am more wondering what we put in the config file.  `tsteps` just does not read correctly to me (I'm not saying `nsteps` is any better, either).  They are plural yet you assign them to a scalar value in the file.

pv101:
Let's have a quick chat tomorrow about naming conventions before merging.

pyfr/plugins/solnwriter.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def __init__(self, intg, cfgsect, suffix=None):
+        super().__init__(intg, cfgsect, suffix)
+
+        basedir = self.cfg.getpath(cfgsect, 'basedir', '.')
+        basename = self.cfg.get(cfgsect, 'basename', raw=True)
+        serial_h5 = self.cfg.getbool(cfgsect, 'serial-h5', False)
+
+        self._writer = H5Writer(intg, basedir, basename, 'soln', serial_h5)
+        self.dt_out = self.cfg.getfloat(self.cfgsect, 'dt-out')
```
FreddieWitherden: We seem to use `self.cfgsect` here but `cfgsect` above.

pyfr/plugins/solnwriter.py
```diff
@@ -0,0 +1,39 @@
...
+class SolnWriterPlugin(BasePlugin):
+    name = 'solnwriter'
+    systems = ['*']
+
+    def __init__(self, intg, cfgsect, suffix=None):
+        super().__init__(intg, cfgsect, suffix)
+
+        basedir = self.cfg.getpath(cfgsect, 'basedir', '.')
+        basename = self.cfg.get(cfgsect, 'basename', raw=True)
+        serial_h5 = self.cfg.getbool(cfgsect, 'serial-h5', False)
```
FreddieWitherden: I suggest we move this to an environmental variable such as `PYFR_FORCE_SERIAL_HDF5` and move all of the logic into the writer class.  Otherwise every single plugin that wants to write something will need to copy this piece of boilerplate code and if you want to switch you need to set `serial-h5 = false` in your inifile for all plugins that write stuff.

As it is more of a debugging/performance thing it is probably best left to an environmental variable.

pyfr/plugins/solnwriter.py
```diff
@@ -0,0 +1,40 @@
...
+        stats = Inifile()
+        intg.collect_stats(stats)
+
+        metadata = dict(config=self.cfg.tostr(),
+                        stats=stats.tostr(),
+                        mesh_uuid=intg.mesh_uuid)
+
+        self._writer.write(intg.soln, metadata, intg.tcurr)
+
+        self.tout_next = intg.tcurr + self.dt_out
```
FreddieWitherden: `self.tout_next += self.dt_out`

FreddieWitherden:
Looks good.  Please test on a few boxes and meshes (partitioned mixed grids tend to work nicely for finding bugs) and if there are no bugs feel free to merge.

FreddieWitherden:
Have all of the tests passed?

Aerojspark:
I have  tested box meshes with hex, tet, pyr, pri (not mixed) and these meshes works well.

iyer-arvind:
Yes, the stock cases pass the test. There is no (except close to precision) change in the final output files.

PR #72, opened 2015-09-08
 Fix a bug in residual plugin.

Aerojspark:
It fixes a bug in residual plugin. The previous solutions are deep-copied.


pyfr/plugins/residual.py
```diff
@@ -43,7 +44,7 @@ def __init__(self, intg, cfgsect, suffix):
     def __call__(self, intg):
         # If an output is due next step
         if (intg.nacptsteps + 1) % self.nsteps == 0:
-            self._prev = intg.soln
+            self._prev = copy.deepcopy(intg.soln)
```
FreddieWitherden: `[s.copy() for s in intg.soln]`

FreddieWitherden:
Assuming this has been tested I am happy to merge.

PR #73, opened 2015-09-21
Fluid force plugin to compute viscous force.

Aerojspark:
It improves the fluid force plugin to compute viscous force, as well as pressure force. There is an option to compute pressure force only or pressure and viscous forces.

type = viscous | pressure

### DIFF
--- a/pyfr/plugins/fluidforce.py
+++ b/pyfr/plugins/fluidforce.py
@@ -21,6 +21,15 @@ class FluidForcePlugin(BasePlugin):
         # Output frequency
         self.nsteps = self.cfg.getint(cfgsect, 'nsteps')
 
+        # Check if we need to compute viscous force
+        self._viscous = intg.system.name == 'navier-stokes'
+
+        # Viscous correction
+        self._viscorr = self.cfg.get('solver', 'viscosity-correction', 'none')
+
+        # Constant variables
+        self._constants = self.cfg.items_as('constants', float)
+
         # Underlying elements class
         self.elementscls = intg.system.elementscls
 
@@ -52,11 +61,20 @@ class FluidForcePlugin(BasePlugin):
             # Output a header if required
             if (os.path.getsize(fname) == 0 and
                 self.cfg.getbool(cfgsect, 'header', True)):
-                print(','.join('txyz'[:self.ndims + 1]), file=self.outf)
+                header = ['t', 'px', 'py', 'pz'][:self.ndims + 1]
+
+                if self._viscous:
+                    header += ['vx', 'vy', 'vz'][:self.ndims]
+
+                print(','.join(header), file=self.outf)
 
         # Interpolation matrices and quadrature weights
         self._m0 = m0 = {}
         self._qwts = qwts = defaultdict(list)
+
+        if self._viscous:
+            self._m4 = m4 = {}
+            rcpjact = {}
 
         # If we have the boundary then process the interface
         if bc in mesh:
@@ -73,6 +91,21 @@ class FluidForcePlugin(BasePlugin):
                     m0[etype, fidx] = eles.basis.m0[facefpts]
                     qwts[etype, fidx] = eles.basis.fpts_wts[facefpts]
 
+                if self._viscous and etype not in m4:
+                    m4[etype] = eles.basis.m4
+
+                    # Obtain SMat at the face
+                    smat = eles.smat_at_np('upts')
+
+                    # Transpose SMat (= Smat^T)
+                    smat = smat.transpose(2, 0, 1, 3)
+
+                    # Obtain |J|^-1
+                    rcpdjac = eles.rcpdjac_at_np('upts')
+
+                    # Compute |J|^-1 SMat^T
+                    rcpjact[etype] = smat*rcpdjac
+
                 area = eles.basis.faces[fidx][3]
 
                 # Unit physical normals and their magnitudes (including |J|)
@@ -85,6 +118,10 @@ class FluidForcePlugin(BasePlugin):
             self._eidxs = {k: np.array(v) for k, v in eidxs.items()}
             self._norms = {k: np.array(v) for k, v in norms.items()}
 
+            if self._viscous:
+                self._rcpjact = {k: rcpjact[k[0]][..., v]
+                                 for k, v in self._eidxs.items()}
+
     def __call__(self, intg):
         # Return if no output is due
         if intg.nacptsteps % self.nsteps:
@@ -95,9 +132,10 @@ class FluidForcePlugin(BasePlugin):
 
         # Solution matrices indexed by element type
         solns = dict(zip(intg.system.ele_types, intg.soln))
+        ndims, nvars = self.ndims, self.nvars
 
         # Force vector
-        f = np.zeros(self.ndims)
+        f = np.zeros(2*ndims if self._viscous else ndims)
 
         for etype, fidx in self._m0:
             # Get the interpolation operator
@@ -109,7 +147,7 @@ class FluidForcePlugin(BasePlugin):
 
             # Interpolate to the face
             ufpts = np.dot(m0, uupts.reshape(nupts, -1))
-            ufpts = ufpts.reshape(nfpts, self.nvars, -1)
+            ufpts = ufpts.reshape(nfpts, nvars, -1)
             ufpts = ufpts.swapaxes(0, 1)
 
             # Compute the pressure
@@ -120,7 +158,30 @@ class FluidForcePlugin(BasePlugin):
             norms = self._norms[etype, fidx]
 
             # Do the quadrature
-            f += np.einsum('i...,ij,jik', qwts, p, norms)
+            f[:ndims] += np.einsum('i...,ij,jik', qwts, p, norms)
+
+            if self._viscous:
+                # Get operator and J^-T matrix
+                m4 = self._m4[etype]
+                rcpjact = self._rcpjact[etype, fidx]
+
+                # Transformed gradient at solution points
+                tduupts = np.dot(m4, uupts.reshape(nupts, -1))
+                tduupts = tduupts.reshape(ndims, nupts, nvars, -1)
+
+                # Physical gradient at solution points
+                duupts = np.einsum('ijkl,jkml->ikml', rcpjact, tduupts)
+                duupts = duupts.reshape(ndims, nupts, -1)
+
+                # Interpolate gradient to flux points
+                dufpts = np.array([np.dot(m0, duupts[i]) for i in range(ndims)])
+                dufpts = dufpts.reshape(ndims, nfpts, nvars, -1).swapaxes(1, 2)
+
+                # Viscous stress
+                vis = self.stress_tensor(ufpts, dufpts)
+
+                # Do the quadrature
+                f[ndims:] += np.einsum('i...,klij,jil', qwts, vis, norms)
 
         # Reduce and output if we're the root rank
         if rank != root:
@@ -136,3 +197,27 @@ class FluidForcePlugin(BasePlugin):
 
             # Flush to disk
             self.outf.flush()
+
+    def stress_tensor(self, u, du):
+        c = self._constants
+
+        # Density, Energy
+        rho, E = u[0], u[-1]
+
+        # Gradient of density and momentum
+        gradrho, gradrhou = du[:, 0], du[:, 1:-1]
+
+        # Gradient of velocity
+        gradu = (gradrhou - gradrho[:, None]*u[None, 1:-1]/rho) / rho
+
+        # Bulk tensor
+        bulk = np.eye(self.ndims)[:, :, None, None]*np.trace(gradu)
+
+        mu = c['mu']
+
+        if self._viscorr == 'sutherland':
+            cpT = c['gamma']*(E/rho - 0.5*np.sum(u[1:-1]**2, axis=0))
+            Trat = cpT/c['cpTref']
+            mu *= (c['cpTref'] + c['cpTs'])*Trat**1.5 / (cpT + c['cpTs'])
+
+        return -mu*(gradu + gradu.swapaxes(0, 1) - 2 / 3*bulk)
--- a/pyfr/solvers/base/elements.py
+++ b/pyfr/solvers/base/elements.py
@@ -158,18 +158,25 @@ class BaseElements(object, metaclass=ABCMeta):
                                      tags={expr, 'align'})
 
     @memoize
-    def smat_at(self, name):
-        smat = self._get_smats(getattr(self.basis, name))
-        return self._be.const_matrix(smat, tags={'align'})
+    def smat_at_np(self, name):
+        return self._get_smats(getattr(self.basis, name))
 
     @memoize
-    def rcpdjac_at(self, name):
+    def smat_at(self, name):
+        return self._be.const_matrix(self.smat_at_np(name), tags={'align'})
+
+    @memoize
+    def rcpdjac_at_np(self, name):
         _, djac = self._get_smats(getattr(self.basis, name), retdets=True)
 
         if np.any(djac < -1e-5):
             raise RuntimeError('Negative mesh Jacobians detected')
 
-        return self._be.const_matrix(1.0 / djac, tags={'align'})
+        return 1.0 / djac
+
+    @memoize
+    def rcpdjac_at(self, name):
+        return self._be.const_matrix(self.rcpdjac_at_np(name), tags={'align'})
 
     @memoize
     def ploc_at_np(self, name):
--- a/doc/src/user_guide.rst
+++ b/doc/src/user_guide.rst
@@ -746,8 +746,8 @@ Example::
 [soln-plugin-fluidforce-name]
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Periodically integrates the pressure on the boundary labelled ``name``
-and writes out the resulting force vectors to a CSV file.
+Periodically integrates the pressure and viscous stress on the boundary
+labelled ``name`` and writes out the resulting force vectors to a CSV file.
 
 1. ``nsteps`` --- integrate every ``nsteps``:
 


pyfr/plugins/fluidforce.py
```diff
@@ -21,6 +21,17 @@ def __init__(self, intg, cfgsect, suffix):
         # Output frequency
         self.nsteps = self.cfg.getint(cfgsect, 'nsteps')
 
+        # Pressure only or viscous force
+        self.forcetype = self.cfg.get(cfgsect, 'type', 'viscous')
```
FreddieWitherden: Missing validation.

pyfr/plugins/fluidforce.py
```diff
@@ -73,6 +88,19 @@ def __init__(self, intg, cfgsect, suffix):
                     m0[etype, fidx] = eles.basis.m0[facefpts]
                     qwts[etype, fidx] = eles.basis.fpts_wts[facefpts]
 
+                    if self.forcetype == 'viscous':
+                        m4e = eles.basis.m4.reshape(self.ndims, eles.nupts, -1)
+                        m4[etype, fidx] = np.einsum('ij,kjl->kil',
+                                                    m0[etype, fidx], m4e)
+
+                        # |J|^-1 Smat^T
+                        smat = eles.smat_at('upts').get().swapaxes(1, 2)
```
FreddieWitherden: Why do we not compute the smats at the flux points directly and then pick out the ones we want say by adding a `smat_at_np` function like we have for `ploc`?

pyfr/plugins/fluidforce.py
```diff
@@ -85,6 +113,10 @@ def __init__(self, intg, cfgsect, suffix):
             self._eidxs = {k: np.array(v) for k, v in eidxs.items()}
             self._norms = {k: np.array(v) for k, v in norms.items()}
 
+            if self.forcetype == 'viscous':
+                self._rcpjact = {k: v[:, :, :, self._eidxs[k]]
```
FreddieWitherden: `...` rather than `:, :, :`

pyfr/plugins/fluidforce.py
```diff
@@ -108,9 +140,7 @@ def __call__(self, intg):
             uupts = solns[etype][..., self._eidxs[etype, fidx]]
 
             # Interpolate to the face
-            ufpts = np.dot(m0, uupts.reshape(nupts, -1))
-            ufpts = ufpts.reshape(nfpts, self.nvars, -1)
-            ufpts = ufpts.swapaxes(0, 1)
+            ufpts = np.dot(m0, uupts.swapaxes(0, 1)).swapaxes(0, 1)
```
FreddieWitherden: Not entirely sure I understand this change.

pyfr/plugins/fluidforce.py
```diff
@@ -136,3 +183,32 @@ def __call__(self, intg):
 
             # Flush to disk
             self.outf.flush()
+
+    def stress_tensor(self, u, du):
+        c = self._constants
+
+        # Density and gradient of density
+        rho = u[0]
+        rcprho = 1/u[0]
```
FreddieWitherden: NumPy is slow enough that you can feel free to just divide as and when you need to.

pyfr/plugins/fluidforce.py
```diff
@@ -136,3 +183,32 @@ def __call__(self, intg):
...
+        gradrho = du[:, 0]
+
+        # Gradient of momentum
+        gradrhou = du[:, 1:-1]
+
+        # Gradient of velocity
+        gradu = (gradrhou - gradrho[:, None, :, :]*u[None, 1:-1, :, :]*rcprho)
```
FreddieWitherden: No need for trailing `:, :`.

pyfr/plugins/fluidforce.py
```diff
@@ -136,3 +183,32 @@ def __call__(self, intg):
...
+
+        # Bulk tensor
+        ndims = du.shape[0]
+        bulk = np.eye(ndims)[:, :, None, None]*np.trace(gradu)
+
+        mu = c['mu']
+
+        if self._viscorr == 'sutherland':
+            E = u[-1]
+            cpT = c['gamma']*(rcprho*E - 0.5*sum(u[1:-1]*u[1:-1]))
```
FreddieWitherden: Can probably simplify `sum(x*x)`.

pyfr/plugins/fluidforce.py
```diff
@@ -122,6 +152,23 @@ def __call__(self, intg):
...
+            if self.forcetype == 'viscous':
+                # Get operator and J^-T matrix
+                m4 = self._m4[etype, fidx]
+                rcpjact = self._rcpjact[etype, fidx]
+
+                # Interpolate gradient to the face
+                dufpts = np.dot(m4, uupts.swapaxes(0, 1)).swapaxes(1, 2)
+
+                # Derivative in physical domain
+                dufpts = np.einsum('ijkl, jmkl->imkl', rcpjact, dufpts)
```
FreddieWitherden: Do not usually have whitespace inside of first einsum argument.

pyfr/plugins/fluidforce.py
Aerojspark: I just simplified them. I added the interpolation of gradient to the face in line 161 and implemented it in the same manner.

The dimension of m0, uupts are (nfpts, nupts) and (nupts, nvars, neles).
The dimension of np.dot(...) is (nfpts, nvars, neles). Then I swap axis 0 and 1.

In line 161,the dimension of m4 is (ndim, nfpts, nupts).
The dimension of np.dot(...) is (ndims, nfpts, nvars, neles). Then I swap axis 1 and 2 and obtain (ndims, nvars, nfpts, neles).

FreddieWitherden:
I think it is better if we compute the physical gradients the same way we do in the solver: first by going from the transformed gradient of the physical solution at the solution points to the physical gradient of the physical solution at the solution points and then interpolate this quantity to the flux points.

pyfr/plugins/fluidforce.py
```diff
@@ -21,6 +21,20 @@ def __init__(self, intg, cfgsect, suffix):
         # Output frequency
         self.nsteps = self.cfg.getint(cfgsect, 'nsteps')
 
+        # Pressure only or viscous force
+        self.forcetype = self.cfg.get(cfgsect, 'type', 'viscous')
+        if self.forcetype not in ['pressure', 'viscous']:
```
FreddieWitherden: Use a set `{'a', 'b'}` here.

pyfr/plugins/fluidforce.py
```diff
@@ -21,6 +21,20 @@ def __init__(self, intg, cfgsect, suffix):
...
+            raise ValueError('Invalid fluid force type')
+
+        if intg.system.name == 'euler':
+            self.forcetype = 'pressure'
+
+        # Viscous Correction
```
FreddieWitherden: Capital 'C'.

pyfr/plugins/fluidforce.py
```diff
@@ -136,3 +189,31 @@ def __call__(self, intg):
...
+        ndims = du.shape[0]
+        bulk = np.eye(ndims)[:, :, None, None]*np.trace(gradu)
+
+        mu = c['mu']
+
+        if self._viscorr == 'sutherland':
+            E = u[-1]
+            cpT = c['gamma']*(E/rho - 0.5*sum(c*c for c in u[1:-1]))
+            Trat = 1/c['cpTref']*cpT
+            mu = mu*(c['cpTref'] + c['cpTs'])*pow(Trat, 1.5) / (cpT + c['cpTs'])
```
FreddieWitherden: `*=` for `mu`.  Use `**` instead of `pow`.

pyfr/plugins/fluidforce.py
```diff
@@ -21,6 +21,19 @@ def __init__(self, intg, cfgsect, suffix):
         # Output frequency
         self.nsteps = self.cfg.getint(cfgsect, 'nsteps')
 
+        # Pressure only or viscous force
+        self.forcetype = self.cfg.get(cfgsect, 'type', 'viscous')
+        if self.forcetype not in {'pressure', 'viscous'}:
+            raise ValueError('Invalid fluid force type')
+        elif self.forcetype == 'pressure' and intg.system.name == 'euler':
```
FreddieWitherden: This block does not appear to do anything.

pyfr/plugins/fluidforce.py
```diff
@@ -122,6 +158,26 @@ def __call__(self, intg):
...
+                rcpjact = self._rcpjact[etype, fidx]
+
+                # Transformed gradient at solution points
+                tduupts = np.dot(m4, uupts.swapaxes(0, 1))
+
+                # Physical gradient at solution points
+                duupts = np.einsum('ijkl,jkml->ikml', rcpjact, tduupts)
+
+                # Interpolate gradient to flux points
+                dufpts = np.einsum('ij,kjlm->kilm', m0, duupts).swapaxes(1, 2)
```
FreddieWitherden: Why is `einsum` being used for the interpolation?

pyfr/plugins/fluidforce.py
```diff
@@ -136,3 +192,31 @@ def __call__(self, intg):
...
+
+        # Bulk tensor
+        ndims = du.shape[0]
+        bulk = np.eye(ndims)[:, :, None, None]*np.trace(gradu)
+
+        mu = c['mu']
+
+        if self._viscorr == 'sutherland':
+            E = u[-1]
+            cpT = c['gamma']*(E/rho - 0.5*sum(c*c for c in u[1:-1]))
```
FreddieWitherden: See if the `sum` can be written in terms of `np.dot` or a relative; might clean things up.

pyfr/plugins/fluidforce.py
```diff
@@ -136,3 +192,31 @@ def __call__(self, intg):
...
+            Trat = 1/c['cpTref']*cpT
```
FreddieWitherden: `cpT/cp['cpTref']`?

pyfr/plugins/fluidforce.py
```diff
@@ -136,3 +192,31 @@ def __call__(self, intg):
 
             # Flush to disk
             self.outf.flush()
+
+    def stress_tensor(self, u, du):
+        c = self._constants
+
+        # Density and gradient of density
+        rho = u[0]
```
FreddieWitherden: Group some assignments, so `rho, E = u[0], u[-1]`.

pyfr/plugins/fluidforce.py
```diff
@@ -136,3 +192,31 @@ def __call__(self, intg):
...
+        gradrho = du[:, 0]
```
FreddieWitherden: See above about grouping.

pyfr/plugins/fluidforce.py
```diff
@@ -136,3 +192,31 @@ def __call__(self, intg):
...
+
+        # Gradient of momentum
+        gradrhou = du[:, 1:-1]
+
+        # Gradient of velocity
+        gradu = (gradrhou - gradrho[:, None]*u[None, 1:-1]/rho)
+        gradu /= rho
+
+        # Bulk tensor
+        ndims = du.shape[0]
```
FreddieWitherden: Is this not available as `self.ndims`?

pyfr/plugins/fluidforce.py
```diff
@@ -73,6 +90,24 @@ def __init__(self, intg, cfgsect, suffix):
                     m0[etype, fidx] = eles.basis.m0[facefpts]
                     qwts[etype, fidx] = eles.basis.fpts_wts[facefpts]
 
+                if self.forcetype == 'viscous':
+                    if etype not in m4:
```
FreddieWitherden: Not sure about this `if` block.

pyfr/plugins/fluidforce.py
FreddieWitherden: Okay.  Check that the `np.dot` call still goes through to BLAS.

pyfr/plugins/fluidforce.py
```diff
@@ -122,6 +164,29 @@ def __call__(self, intg):
...
+                # Transformed gradient at solution points
+                tduupts = np.dot(m4, uupts.reshape(nupts, -1))
+                tduupts = tduupts.reshape(ndims, nupts, nvars, -1)
+
+                # Physical gradient at solution points
+                duupts = np.einsum('ijkl,jkml->ikml', rcpjact, tduupts)
+
+                # Interpolate gradient to flux points
+                dufpts = np.dot(m5, duupts.reshape(ndims*nupts, -1))
+                dufpts = dufpts.reshape(ndims, nfpts, nvars, -1).swapaxes(1, 2)
```
Aerojspark: Previously, I implemented this using einsum to mutiply block diagonal matrix of m0. Now, I stored m5 matrix (block diagonal matrix of [m0]*ndim) and computed this with np.dot(). 
It might be possible to reduce the lines 178 ~ 182 with one np.einsum() but it seems too complicated and we might lose the procedure to compute these parts.

pyfr/plugins/fluidforce.py
```diff
@@ -109,7 +151,7 @@ def __call__(self, intg):
 
             # Interpolate to the face
             ufpts = np.dot(m0, uupts.reshape(nupts, -1))
-            ufpts = ufpts.reshape(nfpts, self.nvars, -1)
+            ufpts = ufpts.reshape(nfpts, nvars, -1)
             ufpts = ufpts.swapaxes(0, 1)
```
Aerojspark: I tried to confirm the previous implementation (one line of np.dot()) uses BLAS, but it was not easy for me. As an alternative, I checked timeit of this and the previous implementation and found that the current one is much faster. Thus, I rolled back these lines and fixed line 174~175 to multiply m4 matrix to uupts.

pyfr/plugins/fluidforce.py
```diff
@@ -136,3 +201,29 @@ def __call__(self, intg):
...
+
+        # Density, Energy and gradient of density
+        rho, E = u[0], u[-1]
+        gradrho = du[:, 0]
+
+        # Gradient of momentum
+        gradrhou = du[:, 1:-1]
+
+        # Gradient of velocity
+        gradu = (gradrhou - gradrho[:, None]*u[None, 1:-1]/rho)
```
FreddieWitherden: Brackets around expression.

pyfr/plugins/fluidforce.py
```diff
@@ -122,6 +164,29 @@ def __call__(self, intg):
...
+
+                # Transformed gradient at solution points
+                tduupts = np.dot(m4, uupts.reshape(nupts, -1))
+                tduupts = tduupts.reshape(ndims, nupts, nvars, -1)
+
+                # Physical gradient at solution points
+                duupts = np.einsum('ijkl,jkml->ikml', rcpjact, tduupts)
+
+                # Interpolate gradient to flux points
+                dufpts = np.dot(m5, duupts.reshape(ndims*nupts, -1))
```
FreddieWitherden: I think we should be doing multiple applications of `m0` here.  It avoids the need for `m5` and the inefficiencies associated with it.

pyfr/plugins/fluidforce.py
```diff
@@ -122,6 +160,29 @@ def __call__(self, intg):
...
+                tduupts = np.dot(m4, uupts.reshape(nupts, -1))
+                tduupts = tduupts.reshape(ndims, nupts, nvars, -1)
+
+                # Physical gradient at solution points
+                duupts = np.einsum('ijkl,jkml->ikml', rcpjact, tduupts)
+
+                # Interpolate gradient to flux points
+                m5 = block_diag([m0]*ndims)
+                dufpts = np.dot(m5, duupts.reshape(ndims*nupts, -1))
+                dufpts = dufpts.reshape(ndims, nfpts, nvars, -1).swapaxes(1, 2)
```
Aerojspark: I have compared these three lines (176~ 178) and multiple application of m0, which is followings.

```python
duupts = duupts.reshape(ndims, nupts, -1)
dufpts = np.zeros((ndims, nfpts) + duupts.shape[2:])
for i in range(ndims):
    dufpts[i] = np.dot(m0, duupts[i])
    
dufpts = dufpts.reshape(ndims, nfpts, nvars, -1).swapaxes(1, 2)
```

For (nfpts, nupts, ndims, nvars, neles) = (16, 64, 3, 5, 1024), current implementation and multiple application of m0 tooks 16.6ns and 20.6ns, respectively. The current one is slightly faster. Do we need to change these lines?

FreddieWitherden: Anything below around ~1ms is noise and not really suitable for benchmarking purposes.

pyfr/plugins/fluidforce.py
```diff
@@ -122,6 +160,32 @@ def __call__(self, intg):
...
+                tduupts = tduupts.reshape(ndims, nupts, nvars, -1)
+
+                # Physical gradient at solution points
+                duupts = np.einsum('ijkl,jkml->ikml', rcpjact, tduupts)
+
+                # Interpolate gradient to flux points
+                duupts = duupts.reshape(ndims, nupts, -1)
+                dufpts = np.zeros((ndims, nfpts) + duupts.shape[2:])
+                for i in range(ndims):
+                    dufpts[i] = np.dot(m0, duupts[i])
```
FreddieWitherden: `dufpts = np.something([np.dot(m0, u) for u in duupts])` or something similar to this should work.

pyfr/plugins/fluidforce.py
```diff
@@ -21,6 +22,19 @@ def __init__(self, intg, cfgsect, suffix):
         # Output frequency
         self.nsteps = self.cfg.getint(cfgsect, 'nsteps')
 
+        # Pressure only or viscous force
+        self.forcetype = self.cfg.get(cfgsect, 'type', 'viscous')
+        if self.forcetype not in {'pressure', 'viscous'}:
+            raise ValueError('Invalid fluid force type')
+        elif self.forcetype == 'viscous' and intg.system.name == 'euler':
+            self.forcetype = 'pressure'
```
FreddieWitherden: I think we should raise an error here.

pyfr/plugins/fluidforce.py
```diff
@@ -122,6 +160,29 @@ def __call__(self, intg):
...
+
+                # Transformed gradient at solution points
+                tduupts = np.dot(m4, uupts.reshape(nupts, -1))
+                tduupts = tduupts.reshape(ndims, nupts, nvars, -1)
+
+                # Physical gradient at solution points
+                duupts = np.einsum('ijkl,jkml->ikml', rcpjact, tduupts)
+
+                # Interpolate gradient to flux points
+                duupts = duupts.reshape(ndims, nupts, -1)
```
FreddieWitherden: Move reshape up a couple of lines.

FreddieWitherden:
Update the documentation and we're basically ready to go.

pyfr/plugins/fluidforce.py
```diff
@@ -52,12 +61,19 @@ def __init__(self, intg, cfgsect, suffix):
             # Output a header if required
             if (os.path.getsize(fname) == 0 and
                 self.cfg.getbool(cfgsect, 'header', True)):
-                print(','.join('txyz'[:self.ndims + 1]), file=self.outf)
+                header = ['t', 'px', 'py', 'pz'][:self.ndims + 1]
+                if self._viscous:
+                    header += ['vx', 'vy', 'vz'][:self.ndims]
+                print(','.join(header), file=self.outf)
```
FreddieWitherden: Add some spacing around the if.

pyfr/plugins/fluidforce.py
```diff
@@ -95,9 +130,13 @@ def __call__(self, intg):
 
         # Solution matrices indexed by element type
         solns = dict(zip(intg.system.ele_types, intg.soln))
+        ndims, nvars = self.ndims, self.nvars
 
         # Force vector
-        f = np.zeros(self.ndims)
```
FreddieWitherden: Inline if; `f = np.zeros(2*self.ndims if self._viscous else self.ndims)`

pyfr/plugins/fluidforce.py
```diff
@@ -120,7 +158,30 @@ def __call__(self, intg):
...
+
+                # Interpolate gradient to flux points
+                dufpts = np.array([np.dot(m0, duupts[i]) for i in range(ndims)])
+                dufpts = dufpts.reshape(ndims, nfpts, nvars, -1).swapaxes(1, 2)
+
+                # Viscous stress
+                vis = self.stress_tensor(ufpts, dufpts)
+
+                # Do the quadrature
+                f[ndims:] += np.einsum('i,klij,jil', qwts, vis, norms)
```
FreddieWitherden: One final comment before we merge.  We should probably be consistent about if we do `i,` or `i...,` for `einsum` when we do quadrature.  Either is fine.

pyfr/plugins/fluidforce.py
```diff
@@ -136,3 +197,29 @@ def __call__(self, intg):
...
+
+    def stress_tensor(self, u, du):
+        c = self._constants
+
+        # Density, Energy and gradient of density
+        rho, E = u[0], u[-1]
+        gradrho = du[:, 0]
+
+        # Gradient of momentum
+        gradrhou = du[:, 1:-1]
```
FreddieWitherden: Move up to line 206?

pyfr/solvers/base/elements.py
```diff
@@ -158,18 +158,26 @@ def opmat(self, expr):
                                      tags={expr, 'align'})
 
     @memoize
-    def smat_at(self, name):
+    def smat_at_np(self, name):
         smat = self._get_smats(getattr(self.basis, name))
-        return self._be.const_matrix(smat, tags={'align'})
+        return smat
```
FreddieWitherden: Can we return directly without first assigning to `smat`?

FreddieWitherden:
A couple of changes appear to have reverted (see comments).

Aerojspark:
I made a mistake to rebase. I also fixed f = np.zeros(...). I am sorry for this.

PR #74, opened 2015-10-04
Add support for solution-dependent source terms

FreddieWitherden:
This commit allows for source term expressions to be a function of the solution variables in conservative form.

It is, without question, one of the more difficult things I've had to implement into PyFR due to the huge number of edge cases.  More testing is therefore certainly required before merging.


Aerojspark:
This plugin seems working fine for 2d quad and 3d hex. Below images are density contours of Rayleigh-Taylor instabilities, which requires a gravity force. I imposed source term as $$(0,0,\rho, \rhov)$$.

rti2d  rti3d

FreddieWitherden:
Looks very nice, Jin Seok.  I have a couple of minor changes that I would like to make to the source term processing code but when these are in we can merge.

FreddieWitherden:
I have updated this request to improve how the expressions are fetched/processed.  Feedback welcome.  Ideally I would like to get this merged ASAP.

Aerojspark:
I have a problem when I impose my source term as 
[solver-source-terms]
rhov = rho
E = rhov

getexpr() function, called at line 120 in pyfr.solvers.base.elements.py, replaces rho as u[0.] not u[0].

FreddieWitherden:
Thanks for this.  I'll go back and play with my regex to avoid this.

PR #75, opened 2015-10-07
Fixes export for numpy 1.10 version.

iyer-arvind:
Numpy 1.10.0 version changes the default casting for inplace operations to
‘same_kind’. This breaks the paraview writer. This commit fixes this.

### DIFF
--- a/pyfr/writers/paraview.py
+++ b/pyfr/writers/paraview.py
@@ -120,7 +120,6 @@ class ParaviewWriter(BaseWriter):
 
                 write_s_to_fh('</PUnstructuredGrid>\n</VTKFile>\n')
 
-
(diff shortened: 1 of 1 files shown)


pyfr/writers/paraview.py
```diff
@@ -277,7 +276,7 @@ def subnodes(cls, n):
             conbase = np.hstack((conbase, conbase + (1 + n)**2))
 
         # Calculate offset of each subdivided element's nodes
-        nodeoff = np.zeros((n,)*cls.ndim)
+        nodeoff = np.zeros((n,)*cls.ndim, dtype='int')
```
FreddieWitherden: `dtype=np.int32` or `dtype=np.int` or `dtype=np.int64` (whatever is most appropriate.

PR #76, opened 2015-10-19
Prevents rewriting restart file.

iyer-arvind:
This commit modifies the solution writer plugin to make sure
it does not rewrite the initial file if the solution is restarted.

### DIFF
--- a/pyfr/integrators/base.py
+++ b/pyfr/integrators/base.py
@@ -15,6 +15,7 @@ class BaseIntegrator(object, metaclass=ABCMeta):
         self.backend = backend
         self.rallocs = rallocs
         self.cfg = cfg
+        self.isrestart = initsoln is not None
 
         # Sanity checks
         if self._controller_needs_errest and not self._stepper_has_errest:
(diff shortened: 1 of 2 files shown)


FreddieWitherden:
As an alternative can you see if using `rm` from `pyfr.util` on the file is sufficient before we attempt to open it for writing? See https://github.com/vincentlab/PyFR/blob/a6fe5f6b2562076af8a4f1ae43048a383565ee55/pyfr/integrators/writers.py#L131

iyer-arvind:
​In my opinion, we should either close the file if not needed, or not write
an open file.
I feel a run, should not delete/rewrite or modify an existing file.

For instance, the user may have added many more information in the hdf5
file, which would be lost.

In any case if we are deleting the file, we need to close it first.​


On Mon, Oct 19, 2015 at 10:56 AM, Freddie Witherden <
notifications@github.com> wrote:

> As an alternative can you see if using rm from pyfr.util on the file is
> sufficient before we attempt to open it for writing? See
> https://github.com/vincentlab/PyFR/blob/a6fe5f6b2562076af8a4f1ae43048a383565ee55/pyfr/integrators/writers.py#L131
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/vincentlab/PyFR/pull/76#issuecomment-149171080>.
>



-- 
================================
Dr. Iyer Arvind Sundaram

The reasonable man adapts himself to the world; the unreasonable one
persists in trying to adapt the world to himself.

Therefore all progress depends on the unreasonable man

- George Bernard Shaw.

PR #78, opened 2015-10-20
Updated User Guide and Developer Guide.

pv101:


FreddieWitherden:
Would you be okay to document the residual plugin?

pv101:
Is this with reference to definition of ICs et al. in the User Guide, or available functions within Mako templates defined in the Developer Guide?

pv101:
@FreddieWitherden I have now documented the residual plugin.

iyer-arvind:
I think the MAKO can have whatever functionality the backend(s) can provide, I am not sure why we mention them under MAKO documentation specifically. 

I guessed we mention the specific few which are explicitly available for ICs as well as BCs and same as MAKO and guaranteed to work identically on all backends. 

This list is available in nputil file, which is the list I mentioned.

FreddieWitherden:
Are we okay to merge?

PR #79, opened 2015-10-23
Fix for first arg for pow in char-riemann-inv.

iyer-arvind:
This makes sure that the first argument to `pow` is always a double in the
characteristic boundary condition. In its absence, setting rho to an
integer causes cuda kernel compilation errors.

### DIFF
--- a/pyfr/solvers/euler/kernels/bcs/char-riem-inv.mako
+++ b/pyfr/solvers/euler/kernels/bcs/char-riem-inv.mako
@@ -6,7 +6,7 @@
 
 <%pyfr:macro name='bc_rsolve_state' params='ul, nl, ur, ploc, t'>
     fpdtype_t cs = sqrt(${gamma}*${c['p']}/${c['rho']});
-    fpdtype_t s = ${c['p']}*pow(${c['rho']}, -${gamma});
+    fpdtype_t s = ${c['p']}*pow((fpdtype_t)${c['rho']}, -${gamma});
     fpdtype_t ratio = cs*${2.0/gmo};
 
     fpdtype_t inv = 1.0/ul[0];


FreddieWitherden:
Why are we casting to a double?

iyer-arvind:
Not casting gives a CUDA kernel where if rho is an int, it maps pow expression to pow(int, double). Apparrently this them maps to math.h pow (somehow) and we get the CUDA kernel error

[command: nvcc --cubin -arch sm_35 -I/home/arvind/Codes/PyFR/PyFR-v1.1.0/env/lib/python3.5/site-packages/pycuda/cuda kernel.cu] 
[stderr: 
kernel.cu(62) (col. 15): error: calling a __host__ function("std::pow<int, double> ") from a __global__ function("bool") is not allowed 

kernel.cu(177) (col. 15): error: calling a __host__ function("std::pow<int, double> ") from a __global__ function("bool") is not allowed

FreddieWitherden:
This is a known issue (and indeed one that is currently holding up source dependent source terms — although there is a partial fix included as part of the request see `_src_exprs`).

My problem is not that you're casting but that you're casting to a double.  Everywhere else in PyFR we use `fpdtype_t` which can be typed to either float or double.

FreddieWitherden:
This is still a poor solution.  We are simply patching over the problem in one specific instance.  Even then we are not doing a particularly good job.  Case in point try setting `rho` to be `2/3`.  There are a variety of places where integer expressions inside of substituted terms result in trouble.

It is better to more explicitly document this as part of v1.1.0 with a proper solution coming as part of v1.2.0.

iyer-arvind:
I am OK with reg-ex replacement, but not pushing it to 1.2.0. Many use integers while using non-dimensional version. Especially noting that this particular error comes in the ini files which have worked perfectly till v1.0.0.

PR #80, opened 2015-10-29
Implementation of free-stream metric.

Aerojspark:
I have implemented the free-stream metric, which satisfies the discrete metric identity.

### DIFF
--- a/pyfr/shapes.py
+++ b/pyfr/shapes.py
@@ -53,6 +53,22 @@ class BaseShape(object):
             self.nsptsord = nsptord = self.order_from_nspts(nspts)
             self.sbasis = get_polybasis(self.name, nsptord, self.spts)
 
+            # Basis for free-stream metric
+            # We need p-th order pseudo grid points, which includes
+            # p-th order points on faces.
+            # It guarantees th q-th order collocation projection on the face
+            # on the both adjacent cells.
+            # Ref. 1 JCP 281, 28-54, Sec 4.2
+            # Ref. 2 JSC 26(3), 301-327, Definition 1
+            if nsptord >= self.order + 1:
+                # Construct metric basis when q > p
+                self.mbasis = get_polybasis(
+                    self.name, self.order + 1, self.mpts
+                )
+            else:
+                # Use sbasis, spts when q <= p
+                self.mbasis = self.sbasis
+
     @classmethod
     def nspts_from_order(cls, sptord):
         return np.polyval(cls.npts_coeffs, sptord) // cls.npts_cdenom
@@ -301,6 +317,17 @@ class BaseShape(object):
     @property
     def nfpts(self):
         return sum(self.nfacefpts)
+
+    @lazyprop
+    def mpts(self):
+        if self.nsptsord >= self.order + 1:
+            return self.std_ele(self.order)
+        else:
+            return self.spts
+
+    @lazyprop
+    def nmpts(self):
+        return len(self.mpts)
 
 
 class TensorProdShape(object):
--- a/pyfr/solvers/base/elements.py
+++ b/pyfr/solvers/base/elements.py
@@ -46,6 +46,7 @@ class BaseElements(object, metaclass=ABCMeta):
         self.nqpts = basis.nqpts if haveqpts else None
         self.nfpts = basis.nfpts
         self.nfacefpts = basis.nfacefpts
+        self.nmpts = basis.nmpts
 
         # Physical normals at the flux points
         self._gen_pnorm_fpts()
@@ -211,47 +212,78 @@ class BaseElements(object, metaclass=ABCMeta):
         self._norm_pnorm_fpts = pnorm_fpts / mag_pnorm_fpts[..., None]
         self._mag_pnorm_fpts = mag_pnorm_fpts
 
-    def _get_jac_eles_at(self, pts):
-        nspts, neles, ndims = self.nspts, self.neles, self.ndims
+    def _get_smats(self, pts, retdets=False):
         npts = len(pts)
+        smats_mpts, djacs_mpts = self._smats_djacs_mpts
 
-        # Form the Jacobian operator
-        jacop = np.rollaxis(self.basis.sbasis.jac_nodal_basis_at(pts), 2)
+        # Interpolation matrix to pts
+        M0 = self.basis.mbasis.nodal_basis_at(pts)
+
+        # Interpolate smats
+        smats = np.array([np.dot(M0, smat) for smat in smats_mpts])
+        smats = smats.reshape(self.ndims, npts, self.ndims, -1)
+
+        # Interpolate djacs
+        if retdets:
+            return smats, np.dot(M0, djacs_mpts)
+        else:
+            return smats
+
+    @lazyprop
+    def _smats_djacs_mpts(self):
+        # Metric basis with grid point (q<=p) or pseudo grid points (q>p)
+        mpts = self.basis.mpts
+        mbasis = self.basis.mbasis
+
+        # Dimensions, number of elements and number of mpts
+        ndims, neles, nmpts = self.ndims, self.neles, self.nmpts
+
+        # Coordinate at pts
+        x = self.ploc_at_np('mpts')
+
+        # Jacobian at pts
+        jacop = np.rollaxis(mbasis.jac_nodal_basis_at(mpts), 2)
+        jacop = jacop.reshape(-1, nmpts)
 
         # Cast as a matrix multiply and apply to eles
-        jac = np.dot(jacop.reshape(-1, nspts), self.eles.reshape(nspts, -1))
+        jac = np.dot(jacop, x.reshape(nmpts, -1))
 
         # Reshape (npts*ndims, neles*ndims) => (npts, ndims, neles, ndims)
-        jac = jac.reshape(npts, ndims, neles, ndims)
+        jac = jac.reshape(nmpts, ndims, ndims, neles)
 
         # Transpose to get (ndims, npts, ndims, neles)
-        return jac.transpose(3, 0, 1, 2)
+        jac = jac.transpose(2, 0, 1, 3)
 
-    def _get_smats(self, pts, retdets=False):
-        jac = self._get_jac_eles_at(pts)
         smats = np.empty_like(jac)
 
-        if self.ndims == 2:
+        if ndims == 2:
+            # Just compute smats whose order is q-1
             a, b, c, d = jac[0,:,0], jac[0,:,1], jac[1,:,0], jac[1,:,1]
 
             smats[0,:,0], smats[0,:,1] = d, -b
             smats[1,:,0], smats[1,:,1] = -c, a
 
-            if retdets:
-                djacs = a*d - b*c
+            djacs = a*d - b*c
         else:
-            # We note that J = [x0, x1, x2]
-            x0, x1, x2 = jac[:,:,0], jac[:,:,1], jac[:,:,2]
+            # Compute x cross x_(chi)
+            jac = np.rollaxis(jac, 2)
+            tt = [np.cross(x, dx, axisa=1, axisb=0, axisc=1) for dx in jac]
+            tt = np.array(tt).reshape(ndims, nmpts, -1)
 
-            smats[0] = np.cross(x1, x2, axisa=0, axisb=0, axisc=1)
-            smats[1] = np.cross(x2, x0, axisa=0, axisb=0, axisc=1)
-            smats[2] = np.cross(x0, x1, axisa=0, axisb=0, axisc=1)
+            # Derivative of x cross x_(chi) at (pseudo) grid points
+            dtt = np.array([np.dot(jacop, tn) for tn in tt])
+            dtt = dtt.reshape(ndims, nmpts, ndims, ndims, -1).swapaxes(1, 2)
 
-            if retdets:
-                # Exploit the fact that det(J) = x0 . (x1 ^ x2)
-                djacs = np.einsum('ij...,ji...->j...', x0, smats[0])
+            # Kopriva's invariant form of smats
+            # Ref. JSC 26(3), 301-327, Eq. (37)
+            smats[0] = 0.5*(dtt[2][1] - dtt[1][2])
+            smats[1] = 0.5*(dtt[0][2] - dtt[2][0])
+            smats[2] = 0.5*(dtt[1][0] - dtt[0][1])
 
-        return (smats, djacs) if retdets else smats
+            # Exploit the fact that det(J) = x0 . (x1 ^ x2)
+            djacs = np.einsum('ij...,ji...->j...', jac[0], smats[0])
+
+        return smats.reshape(ndims, self.nmpts, -1), djacs
 
     def get_mag_pnorms(self, eidx, fidx):
         fpts_idx = self.basis.facefpts[fidx]


Aerojspark:
Any comments for this?

I have tested free-stream flow over the serrated nozzle with P2 scheme. Free-stream preserving metric successfully remove the errors around curved elements.

One caution is that even free-stream preserving metric may not completely suppress these errors for single precision computation.

[ PyFR 1.0]
non_freestream

[Free-stream preserving metric]
free-stream

FreddieWitherden:
If I recall we had a few action points from our meeting.  The first was to remove the configuration file option for free stream preservation.  It should be on all the time.  When this is done (it should allow for quite a few things to be simplified) I'll go through the request again and make suggestions.

pyfr/shapes.py
```diff
@@ -53,6 +53,24 @@ def __init__(self, nspts, cfg):
...
+                # collocation projection on the face on the both adjacent
+                # cells.
+                # Ref. 1 JCP 281, 28-54, Sec 4.2
+                # Ref. 2 JSC 26(3), 301-327, Definition 1
+                self.mpts = self.std_ele(self.order)
+                self.mbasis = get_polybasis(
+                    self.name, self.order + 1, self.mpts
+                )
+            else:
+                # Use sbasis, spts when q <= p
```
FreddieWitherden: Move comment up.

pyfr/solvers/base/elements.py
```diff
@@ -211,47 +214,98 @@ def _gen_pnorm_fpts(self):
...
+
+        # Interpolate djacs
+        if retdets:
+            djacs = np.dot(M0, self._djacs)
 
-        # Form the Jacobian operator
-        jacop = np.rollaxis(self.basis.sbasis.jac_nodal_basis_at(pts), 2)
+        return (smats, djacs) if retdets else smats
+
+    def _plot_at_pts_np(self, pts):
```
FreddieWitherden: Plot?

pyfr/solvers/base/elements.py
```diff
@@ -211,47 +214,98 @@ def _gen_pnorm_fpts(self):
...
-                # Exploit the fact that det(J) = x0 . (x1 ^ x2)
-                djacs = np.einsum('ij...,ji...->j...', x0, smats[0])
-
-        return (smats, djacs) if retdets else smats
+            # Cpmpute x cross x_(chi)
+            tt = np.zeros((ndims,) + x.shape)
+            tt[0] = np.cross(x, x0, axisa=1, axisb=0, axisc=1)
+            tt[1] = np.cross(x, x1, axisa=1, axisb=0, axisc=1)
+            tt[2] = np.cross(x, x2, axisa=1, axisb=0, axisc=1)
+            tt = tt.reshape(ndims, nmpts, -1)
```
FreddieWitherden: Rework.  If we assign every element we should use `np.empty`.  If we are assigning from a bunch of similar function calls an np.array([np.cross(x, xn, axisa=1, axisb=0, axisc=1) for xn in [x0, x1, x2]) is probably cleaner.  It might be possible to do it with a single `np.cross` call, too.

pyfr/solvers/base/elements.py
```diff
@@ -211,47 +214,98 @@ def _gen_pnorm_fpts(self):
...
+
+            # Derivative of x cross x_(chi) at (pseudo) grid points
+            dtt = np.zeros((ndims, nmpts*ndims, ndims*neles))
+            dtt[0] = np.dot(jacop, tt[0])
+            dtt[1] = np.dot(jacop, tt[1])
+            dtt[2] = np.dot(jacop, tt[2])
+            dtt = dtt.reshape(ndims, nmpts, ndims, ndims, -1).swapaxes(1, 2)
```
FreddieWitherden: See above comment.

pyfr/solvers/base/elements.py
```diff
@@ -211,47 +214,98 @@ def _gen_pnorm_fpts(self):
...
+
+            # Kopriva's invariant form of smats
+            # Ref. JSC 26(3), 301-327, Eq. (37)
+            smats[0] = -0.5*(dtt[1][2] - dtt[2][1])
```
FreddieWitherden: Any reason for this over `0.5*(dtt[2][1] - dtt[1][2]))`?

pyfr/solvers/base/elements.py
```diff
@@ -211,47 +214,93 @@ def _gen_pnorm_fpts(self):
...
+    def _get_smats(self, pts, retdets=False):
         npts = len(pts)
+        smats = np.empty((self.ndims, npts, self.ndims*self.neles))
+
+        # Interpolation matrix to pts
+        M0 = self.basis.mbasis.nodal_basis_at(pts)
+
+        # Interpolate smats
+        for i in range(self.ndims):
+            smats[i] = np.dot(M0, self._smats[i])
```
FreddieWitherden: Create list comprehension.

pyfr/solvers/base/elements.py
```diff
@@ -211,47 +214,93 @@ def _gen_pnorm_fpts(self):
...
 
             smats[0,:,0], smats[0,:,1] = d, -b
             smats[1,:,0], smats[1,:,1] = -c, a
 
-            if retdets:
-                djacs = a*d - b*c
+            self._djacs = a*d - b*c
         else:
             # We note that J = [x0, x1, x2]
             x0, x1, x2 = jac[:,:,0], jac[:,:,1], jac[:,:,2]
```
FreddieWitherden: Can we avoid this definition?  Might clean up the cross product as well.

pyfr/solvers/base/elements.py
```diff
@@ -211,47 +214,93 @@ def _gen_pnorm_fpts(self):
...
 
-            smats[0] = np.cross(x1, x2, axisa=0, axisb=0, axisc=1)
-            smats[1] = np.cross(x2, x0, axisa=0, axisb=0, axisc=1)
-            smats[2] = np.cross(x0, x1, axisa=0, axisb=0, axisc=1)
+            # Cpmpute x cross x_(chi)
+            tt = np.array([np.cross(x, xn,  axisa=1, axisb=0, axisc=1)
```
FreddieWitherden: Double space.

pyfr/solvers/base/elements.py
```diff
@@ -211,47 +214,88 @@ def _gen_pnorm_fpts(self):
...
-            smats[2] = np.cross(x0, x1, axisa=0, axisb=0, axisc=1)
-
-            if retdets:
-                # Exploit the fact that det(J) = x0 . (x1 ^ x2)
-                djacs = np.einsum('ij...,ji...->j...', x0, smats[0])
-
-        return (smats, djacs) if retdets else smats
+            # Cpmpute x cross x_(chi)
+            jac = np.rollaxis(jac, 2)
+            tt = np.array([np.cross(x, dx, axisa=1, axisb=0, axisc=1)
```
FreddieWitherden: Make this a list then call `np.array(tt).reshape(...)` on the next line.  Should save us a line

pyfr/solvers/base/elements.py
```diff
@@ -47,6 +47,9 @@ def __init__(self, basiscls, eles, cfg):
         self.nfpts = basis.nfpts
         self.nfacefpts = basis.nfacefpts
 
+        # Compute metric (Smat, djac on pseudo grid points)
+        self._smat_base()
```
FreddieWitherden: Try to avoid generating in the constructor.  We are trying to transition away from this so that constructing Elements instances is a relatively cheap operation.

Aerojspark: I am wondering that even _smat_base() is removed in the constructor, the next line, _gen_pnorm_fpts(), requires smats and eventually calls _smat_base(). The constructor cost will be the same.

Do you think it may be better to remove _get_pnorm_fpts() in the constructor, as well ?

FreddieWitherden: I have a commit to remove _gen_pnorm_fpts from the constructor along with the ploc stuff.  I will push later tonight.

pyfr/solvers/base/elements.py
```diff
@@ -211,47 +211,87 @@ def _gen_pnorm_fpts(self):
...
+    def _ploc_at_pts_np(self, pts):
+        op = self.basis.sbasis.nodal_basis_at(pts)
+
+        ploc = np.dot(op, self.eles.reshape(self.nspts, -1))
+        ploc = ploc.reshape(-1, self.neles, self.ndims).swapaxes(1, 2)
+
+        return ploc
+
+    @lazyprop
+    def _smats(self):
```
FreddieWitherden: Having `_smats` and `get_smats` is somewhat confusing.  Suggestions?

Aerojspark: Sorry for the confusing naming.
Now, get_smats() computes Smats by interpolation of Smats at pseudo grid points (_smats). 
How about renaming _smats as _smats_mpts? mpts denotes pseudo grid points.
Please let me know if you have a good naming for them.

FreddieWitherden: I'm good with `_smat_mpts`.

pyfr/solvers/base/elements.py
```diff
@@ -211,47 +211,87 @@ def _gen_pnorm_fpts(self):
...
             smats[0,:,0], smats[0,:,1] = d, -b
             smats[1,:,0], smats[1,:,1] = -c, a
 
-            if retdets:
-                djacs = a*d - b*c
+            self._djacs_mpts = a*d - b*c
         else:
-            # We note that J = [x0, x1, x2]
-            x0, x1, x2 = jac[:,:,0], jac[:,:,1], jac[:,:,2]
+            # Cpmpute x cross x_(chi)
```
FreddieWitherden: Typo.

pyfr/solvers/base/elements.py
```diff
@@ -211,47 +211,87 @@ def _gen_pnorm_fpts(self):
...
-                djacs = np.einsum('ij...,ji...->j...', x0, smats[0])
+            # Kopriva's invariant form of smats
+            # Ref. JSC 26(3), 301-327, Eq. (37)
+            smats[0] = 0.5*(dtt[2][1] - dtt[1][2])
+            smats[1] = 0.5*(dtt[0][2] - dtt[2][0])
+            smats[2] = 0.5*(dtt[1][0] - dtt[0][1])
 
-        return (smats, djacs) if retdets else smats
+            # Exploit the fact that det(J) = x0 . (x1 ^ x2)
+            self._djacs_mpts = np.einsum('ij...,ji...->j...', jac[0], smats[0])
```
FreddieWitherden: Do not set; return a tuple.

pyfr/solvers/base/elements.py
```diff
@@ -211,47 +211,87 @@ def _gen_pnorm_fpts(self):
...
+        smats = np.array([np.dot(M0, smat) for smat in self._smats_mpts])
+        smats = smats.reshape(self.ndims, npts, self.ndims, -1)
+
+        # Interpolate djacs
+        if retdets:
+            djacs = np.dot(M0, self._djacs_mpts)
+
+        return (smats, djacs) if retdets else smats
+
+    def _ploc_at_pts_np(self, pts):
```
FreddieWitherden: We should not need this function.

pyfr/solvers/base/elements.py
Aerojspark: If I remove ()
```python
return smats, djacs if retdets else smats
```
it also returns smats as tuple (smats,) even when det is not required. So, I need to modify other lines, which call self._get_smats(self, pts). 
One way to avoid this is to expand it as following four lines.
```python
if retdets:
    return smats, djacs
else:
    return smats
```
Can you give me a suggestion for it?

FreddieWitherden: Adding some extra lines is fine.  The key is to remove the side effect (which is an assignment inside of the function).

pyfr/solvers/base/elements.py
```diff
@@ -211,47 +211,81 @@ def _gen_pnorm_fpts(self):
...
+        if retdets:
+            djacs = np.dot(M0, djacs_mpts)
+            return smats, djacs
+        else:
+            return smats
 
-        # Form the Jacobian operator
-        jacop = np.rollaxis(self.basis.sbasis.jac_nodal_basis_at(pts), 2)
+    @lazyprop
+    def _smats_mpts(self):
```
FreddieWitherden: Rename to note that we also get the djacs.  Maybe `_smats_djacs_mpts`, but I am not 100%.

pyfr/shapes.py
FreddieWitherden: See above.

Aerojspark: Although I have already moved up the comment, this message does not disappear.

FreddieWitherden: When I look at https://github.com/Aerojspark/PyFR/blob/develop/pyfr/shapes.py it still seems to be below.  Try pushing again.

pyfr/solvers/base/elements.py
```diff
@@ -211,47 +211,81 @@ def _gen_pnorm_fpts(self):
...
+        # Interpolation matrix to pts
+        M0 = self.basis.mbasis.nodal_basis_at(pts)
+
+        # Interpolate smats
+        smats = np.array([np.dot(M0, smat) for smat in smats_mpts])
+        smats = smats.reshape(self.ndims, npts, self.ndims, -1)
+
+        # Interpolate djacs
+        if retdets:
+            djacs = np.dot(M0, djacs_mpts)
```
FreddieWitherden: No need to assign.

pyfr/solvers/base/elements.py
```diff
@@ -211,47 +211,81 @@ def _gen_pnorm_fpts(self):
...
+            return smats
 
-        # Form the Jacobian operator
-        jacop = np.rollaxis(self.basis.sbasis.jac_nodal_basis_at(pts), 2)
+    @lazyprop
+    def _smats_djacs_mpts(self):
+        # Metric basis with grid point (q<=p) or pseudo grid points (q>p)
+        mpts = self.basis.mpts
+        mbasis = self.basis.mbasis
+        self.nmpts = nmpts = len(mpts)
```
FreddieWitherden: Perform this assignment in __init__ along with the other `nfoo` assignments.

pyfr/solvers/base/elements.py
```diff
@@ -211,47 +211,81 @@ def _gen_pnorm_fpts(self):
...
+
+        # Dimensions and number of elements
+        ndims = self.ndims
```
FreddieWitherden: Can compact; `ndims, neles = self.ndims, self.neles`.

pyfr/shapes.py
```diff
@@ -53,6 +53,26 @@ def __init__(self, nspts, cfg):
...
+            # Basis for free-stream metric
+            # We need p-th order pseudo grid points, which includes
+            # p-th order points on faces.
+            # It guarantees th q-th order collocation projection on the face
+            # on the both adjacent cells.
+            # Ref. 1 JCP 281, 28-54, Sec 4.2
+            # Ref. 2 JSC 26(3), 301-327, Definition 1
+            if nsptord >= self.order + 1:
+                # Construct metric basis when q > p
+                self.mpts = self.std_ele(self.order)
```
FreddieWitherden: Make these attributes lazy (mpts, nmpts) so it fits with spts, fpts, upts.

FreddieWitherden:
Can you rebase to avoid the conflicts?

PR #81, opened 2015-11-09
Added a time-averaging plugin.

pv101:
Note that 'stats' are different to those added by the soon to be depreciated time-averaging utility. We can discuss what is most appropriate going forwards (in the context of the proposed improvements to this time-averaging plugin).

I have tested with the 2D Euler Vortex. Should probably test further before accepting.

### DIFF
--- a/pyfr/plugins/tavg.py
+++ b/pyfr/plugins/tavg.py
@@ -1,0 +1,56 @@
+# -*- coding: utf-8 -*-
+
+import numpy as np
+
+from pyfr.h5writer import H5Writer
+from pyfr.inifile import Inifile
+from pyfr.plugins.base import BasePlugin
+
+
+class TavgPlugin(BasePlugin):
+    name = 'tavg'
+    systems = ['*']
+
+    def __init__(self, intg, cfgsect, suffix=None):
+        super().__init__(intg, cfgsect, suffix)
+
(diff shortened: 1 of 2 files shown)


pyfr/plugins/tavg.py
```diff
@@ -0,0 +1,56 @@
...
+                a += 0.5*(intg.tcurr - self.pt)*(p + c)
+
+            self.psoln = [s.copy() for s in intg.soln]
+            self.pt = intg.tcurr
+
+            if dowrite:
+                self.asoln = [a / self.dtout for a in self.asoln]
+
+                stats = Inifile()
+                stats.set('tavg', 'start', intg.tcurr - self.dtout)
```
FreddieWitherden: Do we want to include the regular stats here, too?

FreddieWitherden:
It is a bit light on comments/documentation although given I will be extending it as soon as we merge that is probably fine.

pyfr/plugins/tavg.py
pv101: I don't think they (tcurr, nsteps, nfevals) make much sense for an averaged solution, so I elected to omit for the moment. We can discuss exactly what to include going forwards when your extension to this is merged?

pv101:
Yes, I think comment situation is fine given it will be extended soon after merging.

PR #83, opened 2015-11-23
Possible bug fix for RoeM

Aerojspark:
It fixes a possible bug to compute function f in RoeM flux.

### DIFF
--- a/pyfr/solvers/euler/kernels/rsolvers/roem.mako
+++ b/pyfr/solvers/euler/kernels/rsolvers/roem.mako
@@ -43,7 +43,9 @@
     fpdtype_t contraa = ${pyfr.dot('n[{i}]', 'va[{i}]', i=ndims)};
     fpdtype_t aa      = sqrt(${c['gamma'] - 1}*(ha - 0.5*qq));
     fpdtype_t rcp_aa  = 1.0/aa;
-    fpdtype_t abs_ma  = fabs(contraa*rcp_aa);
+
+    // Compute |M|, add a small number to avoid a possible singularity of f
+    fpdtype_t abs_ma  = fabs(contraa*rcp_aa) + 1e-15;
 
     // Eigen structure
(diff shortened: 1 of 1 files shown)


FreddieWitherden:
My preference is generally against small fudge factors.  Would something along the lines of

```
    f = (f == f) ? f : 1.0
```

resolve the issue?  So `f == f` is always true except for when `f` is NaN when we instead return 1.

Aerojspark:
Unfortunately, that does not work. Because I cannot use printf in OpenCL with NVIDIA, I didn't find why it does not work.

I think another way to fix is just enforcing positive h.
```
fpdtype_t h = max(1.0 - SDST, 0.0);
```
Could you let me know your thought about it?

FreddieWitherden:
I did consider this however is it ever possible for `abs_ma` to be zero?  As this would leave us with 0^0 which would still give us NaN.

Aerojspark:
abs_ma becomes zero at the stagnation point. It also occurs when flow is tangential to mesh.

When base is zero, especially with negative exponent (see http://en.cppreference.com/w/c/numeric/math/pow), it may yield error but pow(0.0, 0.0) = 1. 
I confirmed that adding small number to base or enforcing positive exponent both work well.

FreddieWitherden:
According to the C standard (http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf) section 7.12.7.4 "domain error may occur if x is zero and y is zero" which basically means it is a crapshoot.  We can therefore not rely on this to produce the correct behaviour across platforms.

Aerojspark:
Could we go for the initial commit (adding a small number) ?

FreddieWitherden:
Although I am not a fan of adding small fudge factors I also can not think of a more sensible solution at this time.  Although please resubmit the request to merge into release/v1.2.0.

PR #85, opened 2015-12-11
Fixes the plugin dt_out accumulated error.

iyer-arvind:
This remedies the accumulated error in dt_out which is compared with
intg.tcurr( which does not accumulate error).


pyfr/plugins/writer.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, intg, cfgsect, suffix=None):
         if not intg.isrestart:
             self(intg)
         else:
-            self.tout_next += self.dt_out
+            self.tout_next = intg.tcurr + self.dt_out
```
FreddieWitherden: No possibility for error here.  We assign just above.

FreddieWitherden:
Please squish into a single commit.

pyfr/plugins/writer.py
iyer-arvind: I understand. Done for consistency.

PR #87, opened 2015-12-16
Compute the common normal vector

Aerojspark:
This commit change the way to save normal vector.
Previous: |J|J^-T n_ref
Now: |J|J^-T n_ref dS_ref
The proposed one can be interpreted as the physical infinitesimal volume.
dV_phy = n_phy dS_phy = |J| dV_ref = |J|J^-T n_ref dS_ref.
Thus, magnl = magnr

### DIFF
--- a/pyfr/plugins/fluidforce.py
+++ b/pyfr/plugins/fluidforce.py
@@ -106,14 +106,12 @@ class FluidForcePlugin(BasePlugin):
                     # Compute |J|^-1 SMat^T
                     rcpjact[etype] = smat*rcpdjac
 
-                area = eles.basis.faces[fidx][3]
-
                 # Unit physical normals and their magnitudes (including |J|)
                 npn = eles.get_norm_pnorms(eidx, fidx)
                 mpn = eles.get_mag_pnorms(eidx, fidx)
 
                 eidxs[etype, fidx].append(eidx)
-                norms[etype, fidx].append(mpn[:, None]*npn*area)
+                norms[etype, fidx].append(mpn[:, None]*npn)
 
             self._eidxs = {k: np.array(v) for k, v in eidxs.items()}
             self._norms = {k: np.array(v) for k, v in norms.items()}
--- a/pyfr/shapes.py
+++ b/pyfr/shapes.py
@@ -103,7 +103,7 @@ class BaseShape(object):
         if 'surf-flux' in self.antialias:
             fp = [_proj_l2(self._iqrules[kind],
                            self.facebases[kind])
-                  for kind, proj, norm, area in self.faces]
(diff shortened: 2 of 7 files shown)


FreddieWitherden:
Code is fine.

However, I would like to see just a few sanity checks confirming that (for all element types at, say order 3) that M6 is unchanged and the product M3*M2 is unchanged.

FreddieWitherden:
If we no longer need it we should absolutely get rid of it.

FreddieWitherden:
Merge into a single commit and revise the commit message slightly and I am happy to commit.

pv101:
Does this now effectively shift 'responsibility' for knowing the relative side lengths (in 2D), or face areas (in 3D) of the standard elements, from magnl(r) to the correction functions? Specifically, is it the case that g_i.n(x_i) (normal component of correction function i at flux point i) no longer has unit magnitude, but some magnitude based on relative side lengths (in 2D), or face areas (in 3D) of the relevant standard element? I understand the memory saving benefits of the approach, but if what I just wrote is correct, then we do need to be careful. I guess M3 and M6 now change?

pv101:
Correction - M6 will be unchanged, since it has an n in its definition?

FreddieWitherden:
M3 changes, but since M6 is a product of M3 and the normals (whose definition has now changed) the differences should cancel.  M2 has also changed as it also involves the normals.  However, in the code M2 and M3 occur together as the product M3*M2.

Of course, changes only occur in specific element types where faces/edges have different relative areas.

pv101:
Is my textual interpretation correct: i.e. "Does this now effectively shift 'responsibility' for knowing the relative side lengths (in 2D), or face areas (in 3D) of the standard elements, from magnl(r) to the correction functions? Specifically, is it the case that g_i.n(x_i) (normal component of correction function i at flux point i) no longer has unit magnitude, but some magnitude based on relative side lengths (in 2D), or face areas (in 3D) of the relevant standard element?"

FreddieWitherden:
That is my understanding.  The correction function knows which flux point it is correcting and so can slot in the relevant relative area/length factor.

pv101:
OK. I am happy with the change, provided the sanity checks discussed in the thread above are satisfied, and we still pass all our other tests, and we remember what we have done when we come to define new correction functions!!

Aerojspark:
Sorry for late reply. Correction function is the order of area/length and M2 has normal vector. 
Previously
M2 -  |J|J^-T n_ref
M3 - dS_ref
Now
M2 - |J|J^-T n_ref*dS_ref
M3 - 1
Thus, M3*M2 is order of volume (dV_phy) and it is same for both.

PR #89, opened 2015-12-21
Eliminate the dependency on mpmath

FreddieWitherden:
This eliminates the moribund mpmath library as a dependency.  Start up time is reduced significantly.  There is a marginal loss of accuracy in the solver (but a potential improvement in the post-processing code/anyone that imports fragments of PyFR).


pv101:
Just to confirm - this is possible because of our switch over to analytical derivatives, and hence we no longer need to perform numerical differentiation at high precision? So long as all tests in @iyer-arvind framework pass, then I am happy with this.

FreddieWitherden:
Yes.

PR #92, opened 2015-12-22
Implementation of free-stream metric.

Aerojspark:
I have implemented the free-stream metric, which satisfies the discrete metric identity.

### DIFF
--- a/pyfr/shapes.py
+++ b/pyfr/shapes.py
@@ -53,6 +53,22 @@ class BaseShape(object):
             self.nsptsord = nsptord = self.order_from_nspts(nspts)
             self.sbasis = get_polybasis(self.name, nsptord, self.spts)
 
+            # Basis for free-stream metric
+            # We need p-th order pseudo grid points, which includes
+            # p-th order points on faces.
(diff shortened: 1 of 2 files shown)


pyfr/shapes.py
```diff
@@ -53,6 +53,22 @@ def __init__(self, nspts, cfg):
...
+            # on the both adjacent cells.
+            # Ref. 1 JCP 281, 28-54, Sec 4.2
+            # Ref. 2 JSC 26(3), 301-327, Definition 1
+            if nsptord >= self.order + 1:
+                # Construct metric basis when q > p
+                self.mbasis = get_polybasis(
+                    self.name, self.order + 1, self.mpts
+                )
+            else:
+                # Use sbasis, spts when q <= p
```
FreddieWitherden: Move above.

Aerojspark: Sorry for my confusion.
Do you mean to move comments (lines 56~62) to above if clause (line 52) or move comment (line 65 or 69) to line above 63?

FreddieWitherden: Shift the comment up one line so it is before the else.

PR #96, opened 2016-03-09
Implement CGNS reader.

Aerojspark:
I have implemented high-order CGNS mesh reader. It can deal with single zone, unstructured mesh.


pyfr/readers/__init__.py
```diff
@@ -2,6 +2,7 @@
 
 from pyfr.readers.base import BaseReader
 from pyfr.readers.gmsh import GmshReader
+from pyfr.readers.cgns import CGNSReader
```
FreddieWitherden: Move up one line.

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,167 @@
+# -*- coding: utf-8 -*-
+
+import numpy as np
+import re
```
FreddieWitherden: from collections import defaultdict
import re

import numpy as np

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,167 @@
...
+
+from collections import defaultdict
+
+from pyfr.readers.gmsh import GmshReader
+from pyfr.readers.nodemaps import CGNSNodeMaps
+from pyfr.readers.cgns_wrapper import CGNSWrapper
+
+
+class PYFR_CGNS_ZONE:
```
FreddieWitherden: Rename class to fit with convention.

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,167 @@
...
+
+            elenode = {k: np.array(v) for k, v in mele.items()}
+        else:
+            spts = self._cgns_map[cgns_type][1]
+            elenode[cgns_type] = conn.reshape(-1, spts)
+
+        return elerng, elenode
+
+
+class CGNSReader(GmshReader):
```
FreddieWitherden: This should not inherit from, or touch, the GmshReader.  If both the GmshReader and a CGNSReader can share some functionality it should be moved into a common base class.

pyfr/readers/cgns_wrapper.py
```diff
@@ -0,0 +1,235 @@
...
+                raise OSError
+
+    def _errcheck(self, status, fn, arg):
+        if status != 0:
+            errmsg = self.lib.cg_get_error().decode('utf-8')
+            raise RuntimeError('''CGNS Error detected. Error code {}
+            Error message: {}'''.format(status, errmsg))
+
+    def open(self, name, mode):
+        # Read cgns file
```
FreddieWitherden: Lone coment.

pyfr/readers/cgns_wrapper.py
```diff
@@ -0,0 +1,235 @@
...
+
+        # Zone type check
+        self.lib.cg_zone_type(base['file'], base['idx'], idx + 1,
+                              byref(zonetype))
+        if zonetype.value != self.cgnsUnstructured:
+            raise RuntimeError('ReadCGNS_read: Incorrect zone type for file')
+
+        return {'base': base, 'idx': idx + 1,
+                'name': name.value.decode('utf-8'),
+                'size': tuple(i for i in size)}
```
FreddieWitherden: list(size)

pyfr/readers/nodemaps.py
```diff
@@ -187,3 +187,73 @@ class GmshNodeMaps(object):
     }
 
     from_pyfr = {k: np.argsort(v) for k, v in to_pyfr.items()}
+
+
+class CGNSNodeMaps:
```
FreddieWitherden: Inherit from object.

pyfr/readers/cgns_wrapper.py
```diff
@@ -0,0 +1,235 @@
...
+        lib.cg_coord_read.argtypes = [c_int, c_int, c_int, c_char_p, c_int,
+                                      c_void_p, c_void_p, c_void_p]
+        lib.cg_coord_read.errcheck = self._errcheck
+
+        # cg_nbocos
+        lib.cg_nbocos.argtypes = [c_int, c_int, c_int, c_void_p]
+        lib.cg_nbocos.errcheck = self._errcheck
+
+        # cg_boco_info
+        lib.cg_boco_info.argtypes = [c_int, c_int, c_int, c_int, c_char_p,
```
FreddieWitherden: Use POINTER(c_int) where appropriate.  Avoids having to use byref.  See the METIS partitioner interface for further details.

pyfr/readers/cgns_wrapper.py
```diff
@@ -0,0 +1,235 @@
...
+
+import os
+
+from ctypes import (RTLD_GLOBAL, c_size_t, c_int, c_void_p, c_char_p,
+                    byref, create_string_buffer)
+
+from pyfr.ctypesutil import load_library
+
+
+class CGNSWrapper:
```
FreddieWitherden: Inherit from object.  Move into cgns.py.

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,167 @@
...
+    # Nodemap to PyFR node order
+    _nodemaps_from_pyfr = CGNSNodeMaps.from_pyfr
+
+    def __init__(self, msh):
+        cgns = CGNSWrapper()
+
+        file = cgns.open(msh.name, 0)
+        base = cgns.base_read(file, 0)
+
+        nz = cgns.nzones(base)
```
FreddieWitherden: No need to assign if we only use once.

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,423 @@
...
+        # cg_elements_read
+        lib.cg_elements_read.argtypes = [c_int, c_int, c_int, c_int,
+                                         c_void_p, c_void_p]
+        lib.cg_elements_read.errcheck = self._errcheck
+
+    def _hdf5_load(self):
+        try:
+            load_library('hdf5', mode=RTLD_GLOBAL)
+        except OSError:
+            import pkgconfig
```
FreddieWitherden: Wrap this in a try except as otherwise we could give some misleading exceptions.

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,423 @@
...
+
+            if pkgconfig.exists('hdf5'):
+                # Load HDF5 as RTLD_GLOBAL
+                hdf5_dir = pkgconfig.parse('hdf5')['library_dirs']
```
FreddieWitherden: setenv from pyfr.util might help out quite a bit here.

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,423 @@
...
+        14: ('pri', 6), 16: ('pri', 18), 36: ('pri', 40), 53: ('pri', 75),
+        17: ('hex', 8), 19: ('hex', 27), 39: ('hex', 64), 56: ('hex', 125),
+        20: ('mixed',),
+    }
+
+    def __init__(self, cgns, base, idx):
+        self._cgns = cgns
+        zone = cgns.zone_read(base, idx)
+
+        # Read Node
```
FreddieWitherden: # Read nodes

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,423 @@
...
+class CGNSReader(BaseMeshConverter, BaseReader):
+    # Supported file types and extensions
+    name = 'cgns'
+    extn = ['.cgns']
+
+    # CGNS element types to PyFR type (petype) and node counts
+    _etype_map = CGNSZoneReader._cgns_map
+
+    # Number of nodes in the first-order representation an element
+    _petype_focount = {'line': 2, 'tri': 3, 'quad': 4,
```
FreddieWitherden: I think this can be moved into the BaseMeshReader class as both the Gmsh reader and the CGNS reader appear to need it.

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,423 @@
...
+import re
+
+import numpy as np
+
+from pyfr.ctypesutil import load_library
+from pyfr.readers import BaseMeshConverter, BaseReader
+from pyfr.readers.nodemaps import CGNSNodeMaps
+
+
+class CGNSWrapper(object):
```
FreddieWitherden: Rename to `CGNSWrappers` to fit with convention.

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,423 @@
...
+        self.lib.cg_zone_read(base['file'], base['idx'], idx + 1, name, size)
+
+        # Zone type check
+        self.lib.cg_zone_type(base['file'], base['idx'], idx + 1, zonetype)
+        if zonetype.value != self.Unstructured:
+            raise RuntimeError('ReadCGNS_read: Incorrect zone type for file')
+
+        return {'base': base, 'idx': idx + 1,
+                'name': name.value.decode('utf-8'),
+                'size': list(i for i in size)}
```
FreddieWitherden: Just list(size) should do it.

pyfr/readers/base.py
```diff
@@ -22,3 +26,209 @@ def to_pyfrm(self):
...
+        'tri': {'line': [0, 1, 2]},
+        'quad': {'line': [0, 1, 2, 3]},
+        'tet': {'tri': [0, 1, 2, 3]},
+        'hex': {'quad': [0, 1, 2, 3, 4, 5]},
+        'pri': {'quad': [2, 3, 4], 'tri': [0, 1]},
+        'pyr': {'quad': [0], 'tri': [1, 2, 3, 4]}
+    }
+
+    # First-order node numbers associated with each element face
+    _petype_fnmap = {
```
FreddieWitherden: It has been a while: are these our node numbers or Gmsh node numbers?

pyfr/readers/base.py
```diff
@@ -22,3 +26,209 @@ def to_pyfrm(self):
         mesh['mesh_uuid'] = np.array(str(uuid.uuid4()), dtype='S')
 
         return mesh
+
+
+class BaseMeshConverter(object):
```
FreddieWitherden: The design of this class is a bit fragile.  It requires anyone inheriting from it to set a bunch of internal variables.  Could do with a minor rework to abstract this.  (Perhaps the constructor can take element and face data as parameters to avoid sub-classes have to set various properties themselves.

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,423 @@
...
+        for idx in range(ns):
+            elerng, elenode = self._read_element(zone, idx)
+
+            name = 'fluid'
+            for bcname, bcrng in bc.items():
+                if elerng[0] >= bcrng[0] and elerng[1] <= bcrng[1]:
+                    name = bcname
+
+            if name in pents:
+                pent = pents[name]
```
FreddieWitherden: If the pents do not need to be consecutive then the dictionary setdefault method should clean this up.

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,423 @@
...
+
+        # Construct elenodes and physical entity
+        newp = 0
+        for idx in range(ns):
+            elerng, elenode = self._read_element(zone, idx)
+
+            name = 'fluid'
+            for bcname, bcrng in bc.items():
+                if elerng[0] >= bcrng[0] and elerng[1] <= bcrng[1]:
+                    name = bcname
```
FreddieWitherden: Should we break after assigning?

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,423 @@
...
+        s = self._cgns.section_read(zone, idx)
+
+        elerng = s['range']
+        conn = np.zeros(s['dim'], dtype=np.int32)
+        self._cgns.elements_read(s, conn)
+
+        try:
+            cgns_type = s['etype']
+            petype = self._cgns_map[cgns_type][0]
+        except KeyError as err:
```
FreddieWitherden: Just let the error throw, no need to catch and reformat.

pyfr/readers/base.py
Aerojspark: I removed inheritance and modified that BaseMeshConverter is constructed in raw_to_pyfrm of MeshReaders.

pyfr/readers/cgns.py
Aerojspark: I am wondering that there may be multiple sections for a boundary.

pyfr/readers/base.py
Aerojspark: I think petype_fnmap is the node numbering of each mesh format and both are the same counter-clockwise first-order node numbering. I move it to each MeshReader.

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,435 @@
...
+            except ImportError:
+                raise('Cannot import pkgconfig, failed to load hdf5')
+
+            if pkgconfig.exists('hdf5'):
+                # Load HDF5 as RTLD_GLOBAL
+                with setenv(PYFR_LIBRARY_PATH=
+                            ':'.join(pkgconfig.parse('hdf5')['library_dirs'])):
+                    load_library('hdf5', mode=RTLD_GLOBAL)
+            else:
+                raise OSError('Failed to load hdf5')
```
FreddieWitherden: Just raise (this will rethrow the current exception).

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,435 @@
...
+        except OSError:
+            try:
+                import pkgconfig
+            except ImportError:
+                raise('Cannot import pkgconfig, failed to load hdf5')
+
+            if pkgconfig.exists('hdf5'):
+                # Load HDF5 as RTLD_GLOBAL
+                with setenv(PYFR_LIBRARY_PATH=
+                            ':'.join(pkgconfig.parse('hdf5')['library_dirs'])):
```
FreddieWitherden: Maybe assign this to a local variable and then do with setenv(PYFR_LIBRARY_PATH=path).

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,435 @@
...
+        try:
+            load_library('hdf5', mode=RTLD_GLOBAL)
+        except OSError:
+            try:
+                import pkgconfig
+            except ImportError:
+                raise('Cannot import pkgconfig, failed to load hdf5')
+
+            if pkgconfig.exists('hdf5'):
+                # Load HDF5 as RTLD_GLOBAL
```
FreddieWitherden: Move comment down.

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,435 @@
...
+        zone = sect['zone']['idx']
+        idx = sect['idx']
+
+        self.lib.cg_elements_read(file, base, zone, idx,
+                                  conn.ctypes.data, None)
+
+
+class CGNSZoneReader(object):
+    # CGNS element types to PyFR type (petype) and node counts
+    _cgns_map = {
```
FreddieWitherden: Remove underscore.  This variable is accessed by another class.

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,435 @@
...
+        'hex': {'quad': [[0, 1, 2, 3], [0, 1, 4, 5], [1, 2, 5, 6],
+                         [2, 3, 6, 7], [0, 3, 4, 7], [4, 5, 6, 7]]},
+        'pri': {'quad': [[0, 1, 3, 4], [1, 2, 4, 5], [0, 2, 3, 5]],
+                'tri': [[0, 1, 2], [3, 4, 5]]},
+        'pyr': {'quad': [[0, 1, 2, 3]],
+                'tri': [[0, 1, 4], [1, 2, 4], [2, 3, 4], [0, 3, 4]]}
+    }
+
+    # Nodemap to PyFR node order
+    nodemaps_from_pyfr = CGNSNodeMaps.from_pyfr
```
FreddieWitherden: Might be better if we do `nodemaps = CGNSNodeMaps` (and a similar thing in the Gmsh reader).

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,435 @@
...
+        newp = 0
+        for idx in range(cgns.nsections(zone)):
+            elerng, elenode = self._read_element(zone, idx)
+
+            name = 'fluid'
+            for bcname, bcrng in bc.items():
+                if elerng[0] >= bcrng[0] and elerng[1] <= bcrng[1]:
+                    name = bcname
+
+            pent = pents.setdefault(name)
```
FreddieWitherden: pent = pents.setdefault(name, idx).  So long as the pents do not need to be consecutive this should clean up the code as it allows us to get rid of newp.

pyfr/readers/cgns.py
FreddieWitherden: The current code just appears to take the last one it finds?

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,435 @@
...
+
+from pyfr.ctypesutil import load_library
+from pyfr.readers import BaseMeshConverter, BaseReader
+from pyfr.readers.nodemaps import CGNSNodeMaps
+from pyfr.util import setenv
+
+
+class CGNSWrappers(object):
+    def __init__(self):
+        self._hdf5_load()
```
FreddieWitherden: Comment me.  Maybe _load_hdf5 or _reload_hdf5 is a better name?

pyfr/readers/base.py
FreddieWitherden: Great.  Do we need to do the same for _petype_fnums?

Aerojspark: I don't think so. Face numbering (_petype_fnums) is only related with PyFR mesh file format. I think this numbering is associated with faces of Shape class in shape.py.

pyfr/readers/cgns.py
Aerojspark: I am sorry that I might misunderstand your comment. I am considering the following situation.
- section 0 : (Tri elements, elerng=(0, 10))
- section 1: (Quad elements, elerng(11, 20))
- other sections...

The section 1 and 2 are the wall bc, thus bcname and bcrng are 'wall' and (0, 20). I try to keep the bc dictionary, not popping the bcname during the section loop. (for idx rnage(...)).
I might misunderstand what I break in here. Could you please correct me and explain this comment a little bit more?

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,433 @@
...
+from pyfr.ctypesutil import load_library
+from pyfr.readers import BaseMeshConverter, BaseReader
+from pyfr.readers.nodemaps import CGNSNodeMaps
+from pyfr.util import setenv
+
+
+class CGNSWrappers(object):
+    def __init__(self):
+        # Load CGNS
+        try:
```
Aerojspark: I found that CGNS 3.3 fix this issue. Comparing to CGNS 3.3-rc1, they changed cmake option related with HDF5 and now it can automatically load hdf5 as well. I try to call _load_hdf5 (I prefer this name) only when we got OSError.
Regarding CGNS version, CGNS 3.2+ supports cubic (P3) elements and CGNS 3.2.2+ supports quartic (P4) elements.

pyfr/readers/cgns.py
FreddieWitherden: So if we look at the loop we see:
```
            for bcname, bcrng in bc.items():
                if elerng[0] >= bcrng[0] and elerng[1] <= bcrng[1]:
                    name = bcname
```

so `name` is assigned to the `bcname` of the last entry in `bc` that satisfies the condition `elerng[0] >= bcrng[0] and elerng[1] <= bcrng[1]`.  Why is this?  Why do we not stop after finding the first `bc` entry that satisfies the condition?  Instead, we keep going and always iterate over all of `bc`.

Aerojspark: I am really sorry that I totally misunderstood this comment. I will add break to go out side the loop after finding name.

FreddieWitherden: With the break a trick you can play is:

```
            for bcname, bcrng in bc.items():
                if elerng[0] >= bcrng[0] and elerng[1] <= bcrng[1]:
                    name = bcname
                    break
            else:
                name = 'fluid'

File: pyfr/readers/cgns.py [hunk 30]
```diff
@@ -0,0 +1,435 @@
...
+
+        if petype == 'mixed':
+            i = 0
+            mele = defaultdict(list)
+
+            while i < s['dim']:
+                try:
+                    cgns_type = conn[i]
+                    petype, spts = self.cgns_map[cgns_type]
+                except KeyError as err:
```
FreddieWitherden: Just let this raise.

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,435 @@
...
+
+        # Construct elenodes and physical entity
+        for idx in range(cgns.nsections(zone)):
+            elerng, elenode = self._read_element(zone, idx)
+
+            for bcname, bcrng in bc.items():
+                if elerng[0] >= bcrng[0] and elerng[1] <= bcrng[1]:
+                    name = bcname
+                    break
+                else:
```
FreddieWitherden: Unindent else.

pyfr/readers/base.py
FreddieWitherden: Okay.  Lets rename this class from BaseMeshReader to something else.  As if we construct an instance of it, rather than inherit, it should probably not be a base class.  Also, it is less of a reader and more of an assembler for taking a bunch of nodes and using these to build up a mesh.  (If we ever support an input file format which already has face connectivity we would not use this class, for example.)

pyfr/readers/cgns.py
FreddieWitherden: Lets require the new version of CGNS.  Solves a lot of issues.  Although the code for loading HDF5 is clean it still exposes people to things such as RTLD_GLOBAL and will become harder to test as time goes on (as more of us upgrade our CGNS versions).  Drop the HDF5 loading code, let it through an exception, and we can put a note in the documentation stating the required version.

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,435 @@
...
+                self.pfacespents[p.group(1)].append(name)
+            # Other boundary faces
+            else:
+                self.bfacespents[name] = pent
+
+        if any(len(pf) != 2 for pf in self.pfacespents.values()):
+            raise ValueError('Unpaired periodic boundary in mesh')
+
+    def __del__(self):
+        if self._file:
```
FreddieWitherden: Check file exists with hasattr.

pyfr/ctypesutil.py
```diff
@@ -13,18 +13,19 @@ def find_libc():
         return ctypes.util.find_library('c')
 
 
-def load_library(name):
+def load_library(name, **kwargs):
```
FreddieWitherden: No need for these changes.  Not that they're bad, but unused features are untested features.

pyfr/readers/base.py
```diff
@@ -22,3 +26,195 @@ def to_pyfrm(self):
         mesh['mesh_uuid'] = np.array(str(uuid.uuid4()), dtype='S')
 
         return mesh
+
+
+class MeshDBAssembler(object):
```
FreddieWitherden: DB?

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,414 @@
+# -*- coding: utf-8 -*-
+
+from collections import defaultdict
+from ctypes import POINTER, create_string_buffer, c_char_p, c_int, c_void_p
+import re
+
+import numpy as np
+
+from pyfr.ctypesutil import load_library
+from pyfr.readers import MeshDBAssembler, BaseReader
```
FreddieWitherden: Alphabetical order.

pyfr/readers/base.py
```diff
@@ -22,3 +26,195 @@ def to_pyfrm(self):
...
+        'hex': {'quad': [0, 1, 2, 3, 4, 5]},
+        'pri': {'quad': [2, 3, 4], 'tri': [0, 1]},
+        'pyr': {'quad': [0], 'tri': [1, 2, 3, 4]}
+    }
+
+    # Number of nodes in the first-order representation an element
+    _petype_focount = {'line': 2, 'tri': 3, 'quad': 4,
+                       'tet': 4, 'pyr': 5, 'pri': 6, 'hex': 8}
+
+    def __init__(self, mesh):
```
FreddieWitherden: Pass required arrays in directly as opposed to an opaque mesh argument.

pyfr/readers/base.py
Aerojspark: Sorry for the poor name of this class. Could you please suggest a good name for this?

FreddieWitherden: Good question.  NodalMeshAssembler, perhaps, to indicate that it is for meshes specified in terms of a bunch of nodes?

Aerojspark: Thanks, I will change the name and push it soon.

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,416 @@
...
+
+        # cg_elements_read
+        lib.cg_elements_read.argtypes = [c_int, c_int, c_int, c_int,
+                                         c_void_p, c_void_p]
+        lib.cg_elements_read.errcheck = self._errcheck
+
+    def _errcheck(self, status, fn, arg):
+        if status != 0:
+            errmsg = self.lib.cg_get_error().decode('utf-8')
+            raise RuntimeError('''CGNS Error detected. Error code {}
```
FreddieWitherden: See if you can do something more expansive here like we do for cuBLAS and METIS.

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,416 @@
...
+        'pri': {'quad': [[0, 1, 3, 4], [1, 2, 4, 5], [0, 2, 3, 5]],
+                'tri': [[0, 1, 2], [3, 4, 5]]},
+        'pyr': {'quad': [[0, 1, 2, 3]],
+                'tri': [[0, 1, 4], [1, 2, 4], [2, 3, 4], [0, 3, 4]]}
+    }
+
+    # Nodemap to PyFR node order
+    _nodemaps = CGNSNodeMaps
+
+    def __init__(self, msh):
```
FreddieWitherden: A comment or two would be helpful in this method.

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,416 @@
...
+        'tet': {'tri': [[0, 1, 2], [0, 1, 3], [0, 2, 3], [1, 2, 3]]},
+        'hex': {'quad': [[0, 1, 2, 3], [0, 1, 4, 5], [1, 2, 5, 6],
+                         [2, 3, 6, 7], [0, 3, 4, 7], [4, 5, 6, 7]]},
+        'pri': {'quad': [[0, 1, 3, 4], [1, 2, 4, 5], [0, 2, 3, 5]],
+                'tri': [[0, 1, 2], [3, 4, 5]]},
+        'pyr': {'quad': [[0, 1, 2, 3]],
+                'tri': [[0, 1, 4], [1, 2, 4], [2, 3, 4], [0, 3, 4]]}
+    }
+
+    # Nodemap to PyFR node order
```
FreddieWitherden: Misleading comment (it is a Nodemap that can go either way to/from PyFR).

pyfr/readers/cgns.py
```diff
@@ -0,0 +1,416 @@
...
+        self._file = file = cgns.open(msh.name)
+        base = cgns.base_read(file, 0)
+
+        if cgns.nzones(base) > 1:
+            raise ValueError('Only single zone is supported')
+
+        zone = CGNSZoneReader(cgns, base, 0)
+
+        self._nodepts = {i + 1: v
+                        for i, v in enumerate(zone.nodepts.swapaxes(0, 1))}
```
FreddieWitherden: Add a space before the `for`.

pyfr/readers/gmsh.py
```diff
@@ -76,6 +57,9 @@ class GmshReader(BaseReader):
                 'tri': [[0, 1, 4], [1, 2, 4], [2, 3, 4], [0, 3, 4]]}
     }
 
+    # Nodemap to PyFR node order
```
FreddieWitherden: Misleading comment.

pyfr/readers/gmsh.py
```diff
@@ -206,179 +190,13 @@ def _read_eles(self, msh):
...
-            arr = np.array([[nodepts[i] for i in nn] for nn in peles])
-            arr = arr.swapaxes(0, 1)
-            arr = arr[..., :ndim]
-
-            spts['spt_{0}_p0'.format(petype)] = arr
-
-        return spts
-
     def _to_raw_pyfrm(self):
+        # Converter to pyfrm connectivity and shape points
```
FreddieWitherden: Rework comment.

FreddieWitherden:
If you can address these open issues and fold it all down into a single commit I think we're ready to go.

PR #102, opened 2016-04-19
Implement support for dual time stepping

FreddieWitherden:
This request adds support for dual time stepping, developed by myself and Niki.  Various implicit schemes are implemented within this formulation and the existing systems have been extended to support dual time stepping formalism.

Support for the use of an uniform error norm inside of the adaptive RK schemes is also included.  This may yield better results for large scale systems however it is likely that the `atol` and `rtol` values will need to be relaxed slightly when this norm is employed.

### DIFF
--- a/doc/src/user_guide.rst
+++ b/doc/src/user_guide.rst
@@ -379,6 +379,10 @@ Parameterises the time-integration scheme used by the solver with
 
            *float*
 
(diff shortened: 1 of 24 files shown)


FreddieWitherden:
Everyone okay if we go ahead and merge this?  Documentation for the dual time stepping is still pending but everything else is there.

PR #103, opened 2016-04-27
Cleanups and CUDA-aware-MPI Support

FreddieWitherden:
This request cleans up various parts of PyFR, including adding support for mpi4py v2.0.  The main new feature is the ability for PyFR to exploit the fact that many MPI libraries are now CUDA aware.  This permits the CUDA backend to directly pass device pointers to various MPI routines and have the MPI library determine the most efficient means of transferring the data.  This may include the use of GPUDirect and GPUDirect over RDMA where available.

Support is enabled by setting
```
[backend-cuda]
mpi-type = cuda-aware
```
which must of course be used with a CUDA aware MPI library (OpenMPI, MVAPICH, MVAPICH-GDR).

Performance depends heavily on the problem type, the topology of the system, and the MPI library.  As such it may not always yield an improvement.


FreddieWitherden:
Support is now included for post-actions which permits commands, either blocking or non-blocking, to be invoked by the root rank after the writer plugin has finished dumping a solution to disk.  Details can be found in the documentation.

FreddieWitherden:
As a means of improving the start up time of PyFR on large clusters we now directly implement the ``__contains__`` method inside of ``NativeReader``.  This method is called when one executes a statement along the lines of ``name in file``.

The default implementation of this method provided by the base class works by attempting to get the item in question.  If this is successful then ``True`` is returned otherwise, if an ``AttributeError`` is raised, it is caught and ``False`` is returned.  From the standpoint of PyFR this means that statements along the lines of ``name in file`` end up reading the relevant array/record from disk should it exist.

This is a problem as the rank allocation code executes a large number of statements along these lines in order to ensure that if a mesh contains a data set ``con_pxpy`` that it also contains ``con_pypx``.  With the default implementation of ``__contains__`` this means that the root rank (which is the one performing the rank allocation) ends up reading all of the MPI interfaces.  This can result in a lot of disk I/O for a sufficiently large grid.

The new implementation avoids this reading and just asks ``h5py`` directly if the record in question exists in the file.

FreddieWitherden:
What is the status of this request?  Have all of the relevant tests passed?

PR #104, opened 2016-08-19
DTS bugfix and cleanup. Convergence criterion for dual controller.

nloppi:


pyfr/integrators/dual/controllers.py
```diff
@@ -41,22 +45,58 @@ class DualNoneController(BaseDualController):
...
-            for i in range(self._niters):
+            for i in range(self._maxniters):
                 dt = max(min(t - self.tcurr, self._dt), self.dtmin)
                 dtau = max(min(t - self.tcurr, self._dtau), self.dtaumin)
 
                 # Take the step
-                self._idxcurr = self.step(self.tcurr, dt, dtau)
+                self._idxcurr, self._idxprev = self.step(self.tcurr, dt, dtau)
+                self._innersteps += 1
+                if self._innersteps % self._chkevery == 0:
```
FreddieWitherden: Add newline.

pyfr/integrators/dual/controllers.py
```diff
@@ -41,22 +45,58 @@ class DualNoneController(BaseDualController):
...
+                    # Subtract the current solution from the previous solution
+                    self._add(1.0, self._idxprev, -1.0, self._idxcurr)
+                    # Compute the normalised residual
```
FreddieWitherden: Add newline.

pyfr/integrators/dual/controllers.py
```diff
@@ -41,22 +45,58 @@ class DualNoneController(BaseDualController):
...
+                    resid = self._resid(self._idxprev, self._idxcurr)
```
FreddieWitherden: No need to assign, just move the expression into the if.

pyfr/integrators/dual/controllers.py
```diff
@@ -41,22 +45,58 @@ class DualNoneController(BaseDualController):
...
 
-        self._niters = self.cfg.getint('solver-time-integrator', 'niters')
+        sect = 'solver-time-integrator'
+
+        self._innersteps = 0
+        self._idxprev = 0
+        self._maxniters = self.cfg.getint(sect, 'niters')
+        self._atol = self.cfg.getfloat(sect, 'atol', default=1e-6)
+        self._rtol = self.cfg.getfloat(sect, 'rtol', default=1e-6)
+        self._chkevery = self.cfg.getfloat(sect, 'chk', default=5)
```
FreddieWitherden: Break up some of these blocks.  Check every should be an int rather than a float.

FreddieWitherden:
First line of the commit message needs shortening.

pyfr/integrators/dual/controllers.py
```diff
@@ -41,22 +45,58 @@ class DualNoneController(BaseDualController):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        self._niters = self.cfg.getint('solver-time-integrator', 'niters')
+        sect = 'solver-time-integrator'
+
+        self._innersteps = 0
+        self._idxprev = 0
+        self._maxniters = self.cfg.getint(sect, 'niters')
+        self._atol = self.cfg.getfloat(sect, 'atol', default=1e-6)
```
FreddieWitherden: Probably want to rename the variables to something like `_iatol` and the config parameter as `inner-atol` so that we can be sure it will not clash with the regular atol.  Also not sure if we should have a default or not.

pyfr/integrators/dual/controllers.py
```diff
@@ -41,22 +45,58 @@ class DualNoneController(BaseDualController):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        self._niters = self.cfg.getint('solver-time-integrator', 'niters')
+        sect = 'solver-time-integrator'
+
+        self._innersteps = 0
+        self._idxprev = 0
```
FreddieWitherden: Do we need to store this?

pyfr/integrators/dual/controllers.py
```diff
@@ -41,22 +45,60 @@ class DualNoneController(BaseDualController):
...
-        self._niters = self.cfg.getint('solver-time-integrator', 'niters')
+        sect = 'solver-time-integrator'
+
+        self._innersteps = 0
+        self._maxniters = self.cfg.getint(sect, 'niters')
+
+        self._iatol = self.cfg.getfloat(sect, 'inner-atol')
+        self._irtol = self.cfg.getfloat(sect, 'inner-rtol')
+
+        self._chkevery = self.cfg.getint(sect, 'chk', default=5)
```
FreddieWitherden: Rename this variable/config option.  Also, do we want to force the user to specify a value rather than having a default?

pyfr/integrators/dual/controllers.py
```diff
@@ -41,22 +45,60 @@ class DualNoneController(BaseDualController):
...
         while self.tcurr < t:
-            for i in range(self._niters):
+            for i in range(self._maxniters):
                 dt = max(min(t - self.tcurr, self._dt), self.dtmin)
                 dtau = max(min(t - self.tcurr, self._dtau), self.dtaumin)
 
                 # Take the step
-                self._idxcurr = self.step(self.tcurr, dt, dtau)
+                self._idxcurr, self._idxprev = self.step(self.tcurr, dt, dtau)
+                self._innersteps += 1
```
FreddieWitherden: Why do we need this variable?  Why not just use ``i``?

nloppi:
I changed the controller according to your latest notes. I think we should force specifying the tolerances as they are very case dependent especially in AC. Users might not even notice the option if there is a default and it is set very low. What do you think? Thanks Freddie!

pyfr/integrators/dual/controllers.py
```diff
@@ -41,22 +45,58 @@ class DualNoneController(BaseDualController):
...
-            for i in range(self._niters):
+            for i in range(self._maxniters):
                 dt = max(min(t - self.tcurr, self._dt), self.dtmin)
                 dtau = max(min(t - self.tcurr, self._dtau), self.dtaumin)
 
                 # Take the step
-                self._idxcurr = self.step(self.tcurr, dt, dtau)
+                self._idxcurr, self._idxprev = self.step(self.tcurr, dt, dtau)
+
+                if (i+1) % self._itol_chk == 0:
```
FreddieWitherden: space around ``+``.

FreddieWitherden:
I agree.  Lets remove the default for the residual checking.

nloppi:
Added the space around +. I need to be more precise with my syntax. Thanks!

pyfr/integrators/dual/controllers.py
```diff
@@ -41,22 +45,58 @@ class DualNoneController(BaseDualController):
...
 
-        self._niters = self.cfg.getint('solver-time-integrator', 'niters')
+        sect = 'solver-time-integrator'
+
+        self._maxniters = self.cfg.getint(sect, 'niters')
+
+        self._iatol = self.cfg.getfloat(sect, 'inner-atol')
+        self._irtol = self.cfg.getfloat(sect, 'inner-rtol')
+
+        self._itol_chk = self.cfg.getint(sect, 'tol-nsteps', default=5)
```
FreddieWitherden: Remove this default.

pyfr/integrators/dual/controllers.py
```diff
@@ -41,22 +45,58 @@ class DualNoneController(BaseDualController):
...
-            for i in range(self._niters):
+            for i in range(self._maxniters):
                 dt = max(min(t - self.tcurr, self._dt), self.dtmin)
                 dtau = max(min(t - self.tcurr, self._dtau), self.dtaumin)
 
                 # Take the step
-                self._idxcurr = self.step(self.tcurr, dt, dtau)
+                self._idxcurr, self._idxprev = self.step(self.tcurr, dt, dtau)
+
+                if (i + 1) % self._itol_chk == 0:
```
FreddieWitherden: Comment me.

FreddieWitherden:
I think we're getting close.  The only things left to finalise are the variable names/configuration file keys.  The current ones for the maximum number of inner iterations and the frequency with which to check for convergence are certainly not ideal.

One thing it might be worth investigating is the performance impact from checking for convergence every iteration.  If it is only a percent or two it is probably worth it (as that will be more than offset by the fact that we'll be saving iterations and simplifying the code).

When we have sorted this out we can update the documentation and merge the request.

CC: @pv101

nloppi:
Great!
Here's my suggestions:
Because we already use scheme and pseudo-scheme. Keyword 'pseudo' should be repeating.
(max)niters -> pseudo-niters,
inner-atol -> pseudo-atol,
inner-rtol  -> pseudo-rtol,

The keyword  'nsteps' always refers to sample frequency. We could use for instance,
tol-nsteps -> pseudo-tol-nsteps.

I investigated the impact of convergence monitoring frequency. I set pseudo-niters = 40 and pseudo-atol = pseudo-rtol - 1e-9 in the compressible Euler vortex case. The convergence is achieved in iter 14. Changing the frequency  from 5 to 1 resulted in ~10% slowdown. In AC when we typically hit almost the maximum number of iterations we want to have a large tol-nsteps or even disable it completely. I think 5 is a reasonable default.

nloppi:
I did some changes that I think are good. If you are not happy with them let's go back to the original plan. 

Previously we had error tolerance for the flow variables as
max(abs((Q^{n+1} - Q^{n})/(atol+rtolQ^{n+1}))) < 1

Now I changed it to residual tolerance:
max(abs((1/dtauQ^{n+1} - 1/dtauQ^{n})/(atol+rtolQ^{n+1}))) < 1 .

I think pseudo-aresid and pseudo-rresid are good keywords, as cfd users are so used to the term residual.

nloppi:
The .ini file looks like this:

[solver-time-integrator]
formulation = dual
controller = none

scheme = bdf2
tstart = 0.0
tend = 20.0
dt = 0.01

pseudo-scheme = rk4
pseudo-dt = 0.005
pseudo-aresid = 1e-9
pseudo-rresid = 0   
pseudo-niters-max = 20
pseudo-niters-min = 10

pyfr/integrators/dual/controllers.py
```diff
@@ -41,22 +45,62 @@ class DualNoneController(BaseDualController):
...
             self._accept_step(dt, self._idxcurr)
+
+    @memoize
+    def _get_errest_kerns(self):
+        return self._get_kernels('errest', nargs=3, norm='uniform')
+
+    def _resid(self, x, y):
+        comm, rank, root = get_comm_rank_root()
+
+        # Get an errest kern to compute the maximum residual
```
FreddieWitherden: compute the square of the maximum residual.

pyfr/integrators/dual/controllers.py
```diff
@@ -41,22 +45,62 @@ class DualNoneController(BaseDualController):
...
                 # Take the step
-                self._idxcurr = self.step(self.tcurr, dt, dtau)
+                self._idxcurr, self._idxprev = self.step(self.tcurr, dt, dtau)
+
+                # Activate convergence monitoring after pseudo-niters-min
+                if i >= self._minniters - 1:
+                    # Subtract the current solution from the previous solution
+                    self._add(-1.0/dtau, self._idxprev, 1.0/dtau, self._idxcurr)
+
+                    # Compute the normalised residual and check convergence
```
FreddieWitherden: check for convergence.

pyfr/integrators/dual/controllers.py
```diff
@@ -41,22 +45,62 @@ class DualNoneController(BaseDualController):
...
+
+        self._maxniters = self.cfg.getint(sect, 'pseudo-niters-max')
+        self._minniters = self.cfg.getint(sect, 'pseudo-niters-min')
+
+        self._iatol = self.cfg.getfloat(sect, 'pseudo-aresid')
+        self._irtol = self.cfg.getfloat(sect, 'pseudo-rresid')
+
+        if self._maxniters < self._minniters:
+            raise ValueError('The maximum number of pseudo-iterations must'
+                             ' be greater than or equal to the minimum')
```
FreddieWitherden: We normally put the space on the line above.

pyfr/integrators/dual/controllers.py
```diff
@@ -41,22 +45,62 @@ class DualNoneController(BaseDualController):
...
-        self._niters = self.cfg.getint('solver-time-integrator', 'niters')
+        sect = 'solver-time-integrator'
+
+        self._maxniters = self.cfg.getint(sect, 'pseudo-niters-max')
+        self._minniters = self.cfg.getint(sect, 'pseudo-niters-min')
+
+        self._iatol = self.cfg.getfloat(sect, 'pseudo-aresid')
+        self._irtol = self.cfg.getfloat(sect, 'pseudo-rresid')
+
+        if self._maxniters < self._minniters:
```
FreddieWitherden: Move this block up a couple of lines.

pyfr/integrators/dual/controllers.py
```diff
@@ -41,22 +45,62 @@ class DualNoneController(BaseDualController):
...
+        sect = 'solver-time-integrator'
+
+        self._maxniters = self.cfg.getint(sect, 'pseudo-niters-max')
+        self._minniters = self.cfg.getint(sect, 'pseudo-niters-min')
+
+        if self._maxniters < self._minniters:
+            raise ValueError('The maximum number of pseudo-iterations must '
+                             'be greater than or equal to the minimum')
+
+        self._iatol = self.cfg.getfloat(sect, 'pseudo-aresid')
```
FreddieWitherden: Lets rename these.

FreddieWitherden:
Are you okay with making these final changes.  Otherwise, we are basically ready to merge.

nloppi:
I changed the variable names. Did you mean that you also want to rename the config-keyword? My other proposal is pseudo-resid-atol, pseudo-resid-rtol. Which ones do you prefer or do you have suggestions?  Thanks

PR #105, opened 2016-09-02
Cleanups

FreddieWitherden:
This is quite a large request which refactors a substantial portion of the code.  Shock capturing has been rewritten and several of the underlying data structures have been rewritten.  GiMMiK is now a hard dependency and is used automatically.  As such performance improvements should be observed across a range of scenarios.


pyfr/backends/mic/blasext.py
```diff
@@ -12,8 +12,8 @@ def axnpby(self, *arr, subdims=None):
             raise ValueError('Incompatible matrix types')
 
         nv = len(arr)
-        ncola, ncolb = arr[0].datashape[1:]
-        nrow, ldim, lsdim, dtype = arr[0].traits
+        nrow, ldim, dtype = arr[0].traits
+        ncola, ncolb = arr[0].ioshape[1:]
 
         # Render the kernel template
         src = self.backend.lookup.get_template('axnpby').render(
```
Aerojspark: ncola=ncola is needed in arguments of render().

pyfr/backends/mic/kernels/axnpby.mako
```diff
@@ -18,25 +18,47 @@ axnpby(long *nrowp, long *ncolbp, long *ldimp, long *lsdimp,
     #pragma omp parallel
     {
         int align = PYFR_ALIGN_BYTES / sizeof(fpdtype_t);
-        int rb, re, cb, ce;
-        loop_sched_2d(*nrowp, *ncolbp, align, &rb, &re, &cb, &ce);
+        int rb, re, cb, ce, idx;
+        fpdtype_t axn;
+        loop_sched_2d(nrow, ncolb, align, &rb, &re, &cb, &ce);
```
Aerojspark: *nrowp, *ncolbp is needed instead of nrow, ncolb

pyfr/backends/mic/blasext.py
FreddieWitherden: Thanks.  Will fix shortly; the code itself for the MIC was written blind.

Aerojspark:
I have tested this pull request on the testing framework for all backends and confirmed that it works fine. One minor thing is that we need gcc >=4.9 for openmp backend.

PR #107, opened 2016-09-20
Update User Guide.

pv101:


pv101:
This documents the DTS setup. I'm still a bit uncomfortable using  'controller' to describe what to me are two things. The double use is reflected in the docs by the fact that the description of 'controller' is contextual on whether formulation is 'std' or 'dual', and the parameterisation of the 'none' controller is is contextual on whether formulation is 'std' or 'dual'. However, if there are no major objections, then I think it is OK.

PR #109, opened 2016-10-02
Integers to floats to support mic.

nloppi:


FreddieWitherden:
This issue is best fixed in the add kernel.  Or, better still, the run() method of the axnpby kernel for the MIC backend.  Rather than doing `list(consts)` we should instead do: `[float(c) for c in consts]`.

nloppi:
I'll run few of the text cases with mic over night just in case. Everything seems to be working though.

nloppi:
All cases started running on mic and I exported some mid run results. They seem good, so it is pretty safe to say that everything is working. Thanks.

nloppi:
Peter requested these changes in AUTHORS and README.

FreddieWitherden:
Rename mic => MIC and submit against the release/v1.5.0 branch so that it makes it into the next release.

PR #116, opened 2017-01-24
Support for errest/residual of individual field variables.

nloppi:


nloppi:
Hi Freddie,

I opened the pull request early just see if you are happy with the structure. I am still polishing the code, checking formatting and renaming inconsistencies. Phi backend still needs to be tested as well.
Thanks!

pyfr/backends/cuda/blasext.py
```diff
@@ -56,29 +56,40 @@ def errest(self, x, y, z, *, norm):
...
-
-        # Build the reduction kernel
-        rkern = ReductionKernel(
-            x.dtype, neutral='0', reduce_expr=reduce_expr,
-            map_expr='pow(x[i]/(atol + rtol*max(fabs(y[i]), fabs(z[i]))), 2)',
-            arguments='{0}* x, {0}* y, {0}* z, {0} atol, {0} rtol'
-                      .format(npdtype_to_ctype(x.dtype))
+        # Determine the blocksize
+        block = (128, 1, 1)
+        # Determine the gridsize
```
FreddieWitherden: Add newline.

pyfr/backends/cuda/blasext.py
```diff
@@ -56,29 +56,40 @@ def errest(self, x, y, z, *, norm):
...
+
+        # Get the kernel template
+        src = self.backend.lookup.get_template('errest').render(
+            norm=norm, ncola=ncola, sharesz=block[0]
         )
-
+        
+        # Build the reduction kernel
+        rkern = self._build_kernel(
+            'errest', src, [np.int32]*3 + [np.intp]*4 + [dtype]*2)
```
FreddieWitherden: Move `)` onto newline.

pyfr/backends/cuda/blasext.py
```diff
@@ -56,29 +56,40 @@ def errest(self, x, y, z, *, norm):
...
-        rkern = ReductionKernel(
-            x.dtype, neutral='0', reduce_expr=reduce_expr,
-            map_expr='pow(x[i]/(atol + rtol*max(fabs(y[i]), fabs(z[i]))), 2)',
-            arguments='{0}* x, {0}* y, {0}* z, {0} atol, {0} rtol'
-                      .format(npdtype_to_ctype(x.dtype))
+        # Determine the blocksize
+        block = (128, 1, 1)
+        # Determine the gridsize
+        grid = (int((ncolb + block[0] - 1)/block[0]), 1, 1)
+        # Empty result buffer on host, shape = (nvars, nblocks)
```
FreddieWitherden: Add newline.

pyfr/backends/cuda/blasext.py
```diff
@@ -56,29 +56,40 @@ def errest(self, x, y, z, *, norm):
...
+        err_host = cuda.pagelocked_empty((ncola, grid[0]), dtype, 'C')
+        # Device memory allocation
```
FreddieWitherden: Add newline.

pyfr/backends/cuda/kernels/errest.mako
```diff
@@ -0,0 +1,93 @@
...
+            sdata${k}[tid] = max(sdata${k}[tid], sdata${k}[tid + 64]);
+        % else:
+            sdata${k}[tid] += sdata${k}[tid + 64];
+        % endif
+        % endfor
+        __syncthreads();
+        }
+
+        // Last warp
+        if (tid < 32){
```
FreddieWitherden: Move { onto newline.

pyfr/backends/cuda/kernels/errest.mako
```diff
@@ -0,0 +1,93 @@
...
+        % else:
+            sdata${k}[tid] += r;
+        % endif
+        % endfor
+        }
+        __syncthreads();
+    }
+
+    // Unrolled reduction within blocks except the last block
+    if (blockIdx.x != gridDim.x - 1){
```
FreddieWitherden: Move { onto linewline.

pyfr/backends/cuda/kernels/errest.mako
```diff
@@ -0,0 +1,93 @@
...
+        if (tid < 64){
```
FreddieWitherden: Move { onto newline.

pyfr/backends/cuda/kernels/errest.mako
```diff
@@ -0,0 +1,93 @@
...
+                sdata${k}[tid] += sdata${k}[tid + s];
+            % endif
+            % endfor
+            }
+            __syncthreads();
+        }
+    }
+
+    // Copy to global memory
+    if (tid == 0){
```
FreddieWitherden: Move { onto newline.

pyfr/backends/cuda/kernels/errest.mako
```diff
@@ -0,0 +1,93 @@
...
+            sdata${k}[tid] += sdata${k}[tid + ${n}];
+        % endif
+        % endfor
+            __syncthreads();
+        % endfor
+        }
+    }
+
+    // Last block reduced with a variable sized loop
+    else{
```
FreddieWitherden: Move { onto newline.

pyfr/backends/cuda/kernels/errest.mako
```diff
@@ -0,0 +1,93 @@
...
+            sdata${k}[tid] = max(sdata${k}[tid], sdata${k}[tid + ${n}]);
+        % else:
+            sdata${k}[tid] += sdata${k}[tid + ${n}];
+        % endif
+        % endfor
+            __syncthreads();
+        % endfor
+        }
+    }
+
```
FreddieWitherden: Remove newline.

pyfr/backends/cuda/kernels/errest.mako
```diff
@@ -0,0 +1,93 @@
...
+    // Last block reduced with a variable sized loop
+    else{
+        for(int s = 1; s < lastblksize; s *= 2){
```
FreddieWitherden: Move { onto newline.

pyfr/backends/cuda/kernels/errest.mako
```diff
@@ -0,0 +1,93 @@
...
+            if (tid % (2*s) == 0) {
```
FreddieWitherden: Move { onto newline.

pyfr/backends/opencl/blasext.py
```diff
@@ -50,33 +50,43 @@ def errest(self, x, y, z, *, norm):
...
 
-        cnt = x.leaddim*x.nrow
-        dtype = x.dtype
-
-        # Norm type
-        reduce_expr = 'a + b' if norm == 'l2' else 'max(a, b)'
+        nrow, ldim, dtype = x.traits
+        ncola, ncolb = x.ioshape[1:]
+
+        # Determine the "blocksize"
```
FreddieWitherden: Use OpenCL terminology.

pyfr/integrators/dual/controllers.py
```diff
@@ -68,35 +69,50 @@ def advance_to(self, t):
...
         # Prepare and run the kernel
-        self._prepare_reg_banks(x, y, y)
-        self._queue % errest(self._pseudo_aresid, self._pseudo_rresid)
+        self._prepare_reg_banks(x, x, x)
+        self._queue % errest(dtau, 0.0)
+
+        # L2 norm
+        if self._norm == 'l2':
+            # Reduce locally (element types) and globally (MPI ranks)
+
```
FreddieWitherden: Remove newline.

pyfr/integrators/dual/controllers.py
```diff
@@ -68,35 +69,50 @@ def advance_to(self, t):
...
+            rl = [np.sum(e) for e in zip(*errest.retval)]
+            rg = [comm.allreduce(r, op=get_mpi('sum')) for r in rl]
+
+            # Normalise
+            res = np.sqrt(rg/self._gndofs)
+
+        # Linfinity norm
```
FreddieWitherden: L^<infinity symbol here>

pyfr/integrators/dual/controllers.py
```diff
@@ -68,35 +69,50 @@ def advance_to(self, t):
...
+        # L2 norm
+        if self._norm == 'l2':
+            # Reduce locally (element types) and globally (MPI ranks)
+
+            rl = [np.sum(e) for e in zip(*errest.retval)]
+            rg = [comm.allreduce(r, op=get_mpi('sum')) for r in rl]
+
+            # Normalise
+            res = np.sqrt(rg/self._gndofs)
+
```
FreddieWitherden: Remove newline.

pyfr/integrators/dual/controllers.py
```diff
@@ -68,35 +69,50 @@ def advance_to(self, t):
...
+        # L2 norm
+        if self._norm == 'l2':
+            # Reduce locally (element types) and globally (MPI ranks)
+
+            rl = [np.sum(e) for e in zip(*errest.retval)]
+            rg = [comm.allreduce(r, op=get_mpi('sum')) for r in rl]
+
+            # Normalise
+            res = np.sqrt(rg/self._gndofs)
```
FreddieWitherden: Space around /

pyfr/integrators/dual/controllers.py
```diff
@@ -68,35 +69,50 @@ def advance_to(self, t):
...
-        self._queue % errest(self._pseudo_aresid, self._pseudo_rresid)
+        self._prepare_reg_banks(x, x, x)
+        self._queue % errest(dtau, 0.0)
+
+        # L2 norm
+        if self._norm == 'l2':
+            # Reduce locally (element types) and globally (MPI ranks)
+
+            rl = [np.sum(e) for e in zip(*errest.retval)]
+            rg = [comm.allreduce(r, op=get_mpi('sum')) for r in rl]
```
FreddieWitherden: MPI can do vector -> vector reductions (see force calculation plugin for an example).

pyfr/integrators/std/controllers.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import math
+import numpy as np
```
FreddieWitherden: Add newline.  (NumPy is not a standard Python package and so goes into its own stanza.)

pyfr/backends/opencl/blasext.py
```diff
@@ -50,33 +50,43 @@ def errest(self, x, y, z, *, norm):
...
+                else:
+                    return np.sum(err_host, axis=1)
 
-                self._retarr = rkern(xarr, yarr, zarr, atol, rtol,
-                                     queue=qcomp)
+            def run(self, queue, atol, rtol):
+                self.queue = queue
+                rkern(queue.cl_queue_comp, (ncolb, 1, 1), block,
+                      nrow, ncolb, ldim, err_dev, x.data, y.data,
+                      z.data, atol, rtol)
```
FreddieWitherden: Move ) onto newline.

pyfr/backends/opencl/blasext.py
```diff
@@ -50,33 +50,43 @@ def errest(self, x, y, z, *, norm):
...
 
-            def run(self, queue, atol, rtol):
-                qcomp = queue.cl_queue_comp
-
-                xarr = Array(qcomp, cnt, dtype, data=x.data)
-                yarr = Array(qcomp, cnt, dtype, data=y.data)
-                zarr = Array(qcomp, cnt, dtype, data=z.data)
+                # Reduction across blocks outsourced to numpy
+                if norm == 'uniform':
+                    return np.max(err_host, axis=1)
```
FreddieWitherden: Outside the class define something line

  reducer = np.max if norm == 'uniform' else np.sum

then just do return reducer(err_host, axis=1).

FreddieWitherden:
Biggest things: Blocksize is currently always 128.  This could be a problem in relation to running out of shared memory (think MHD where we have 9 field variables).  Adapt the code to support any power of two blocksize.  The result should be one for loop with an if statement to decide if we need to sync threads or not.

Secondly, the OpenCL code is probably wrong in the sense that it, unlike NVIDIA, does not have warps and does not guarantee that groups of 32 threads execute synchronously.  As such you always need the barrier.

nloppi:
Thanks for the comments, Freddie! 

Yes, actually I always found a difference in 5th or 6th digit between OpenCL (my desktop) and CUDA (nighthawk)  and  was planning to double check it on the test machine that has both backends. It must be related to this synchronisation feature.

nloppi:
Now the OpenCL and CUDA outputs match as well. I still need to test Phi

nloppi:
Sorry, I forgot to change OpenCL terminology in the comments. I'm on it.

pyfr/backends/cuda/kernels/errest.mako
```diff
@@ -0,0 +1,98 @@
...
+        % endif
+        % endfor
+        % if sharesz/(2**(n + 1)) > 32:
+            __syncthreads();
+	}
+        % endif
+    % endfor
+    	}
+    }
+
```
FreddieWitherden: Remove newline.

pyfr/backends/opencl/kernels/errest.mako
```diff
@@ -0,0 +1,93 @@
...
+            sdata${k}[tid] = max(sdata${k}[tid], sdata${k}[tid + tstride]);
+        % else:
+            sdata${k}[tid] += sdata${k}[tid + tstride];
+        % endif
+        % endfor
+            barrier(CLK_LOCAL_MEM_FENCE);
+        }
+    % endfor
+    }
+
```
FreddieWitherden: Remove newline.

pyfr/integrators/dual/controllers.py
```diff
@@ -68,35 +69,61 @@ def advance_to(self, t):
...
+
+            # Reduce across MPI ranks
+            if rank != root:
+                comm.Reduce(rl, None, op=get_mpi('sum'), root=root)
+            else:
+                comm.Reduce(get_mpi('in_place'), rl, op=get_mpi('sum'),
+                            root=root)
+            # Normalise
+            res = np.sqrt(rl / self._gndofs)
+
```
FreddieWitherden: Remove newline.

pyfr/integrators/std/controllers.py
```diff
@@ -120,21 +122,30 @@ def _errest(self, x, y, z):
         # L2 norm
         if self._norm == 'l2':
             # Reduce locally (element types) and globally (MPI ranks)
-            rl = sum(errest.retval)
-            rg = comm.allreduce(rl, op=get_mpi('sum'))
-
+            rl = [np.sum(e) for e in zip(*errest.retval)]
+            # Reduce across MPI ranks
+            if rank != root:
+                comm.Reduce(rl, None, op=get_mpi('sum'), root=root)
```
FreddieWitherden: You want to reduce a NumPy array.

pyfr/integrators/dual/controllers.py
```diff
@@ -68,35 +69,61 @@ def advance_to(self, t):
...
+        # L^infinity norm
+        else:
+            # Reduce locally (element types) and globally (MPI ranks)
+            rl = np.maximum(np.zeros(self.system.nvars), *errest.retval)
```
FreddieWitherden: np.max

nloppi:
errest.retval returns a list of np.arrays, such as 
[array([  2.,   3,   4]), array([  2.,   6.,   2.])]. That's why I chose to use np.maximum which takes in multiple NumPy arrays as an input, rather than converting the proxylist as that the np.max(*, axis=0) requires. I assumed it would be quicker.

nloppi:
Actually, it is a proxylist isn't it? np.max does not like and even converting it to ndarray yields invalid __array_struct__.

pyfr/backends/base/makoutil.py
```diff
@@ -13,6 +14,8 @@
 def ndrange(context, *args):
     return util.ndrange(*args)
 
+def logtwo(context, x):
```
FreddieWitherden: Rename ilog2 (as it gives an integer).

pyfr/backends/base/makoutil.py
```diff
@@ -13,6 +14,8 @@
 def ndrange(context, *args):
     return util.ndrange(*args)
```
FreddieWitherden: Add newline.

pyfr/backends/base/makoutil.py
```diff
@@ -13,6 +14,8 @@
...
 
+def logtwo(context, x):
+    return int(math.log(x, 2))
```
FreddieWitherden: Add newline.

pyfr/backends/mic/kernels/errest.mako
```diff
@@ -3,24 +3,85 @@
...
+            % for i in range(ncola):
+                idx = r*ldim + X_IDX_AOSOA(${i}, ${ncola});
+                res${i} = max(res${i}, pow(x[idx]/(atol + rtol*max(fabs(y[idx]), fabs(z[idx]))), 2));
+            % endfor
+            }
+        }
+    }
 
-    *out = err;
+// Copy
```
FreddieWitherden: Indent.

pyfr/backends/openmp/kernels/errest.mako
```diff
@@ -2,21 +2,83 @@
...
+                idx = r*ldim + X_IDX_AOSOA(${i}, ${ncola});
+                res${i} = max(res${i}, pow(x[idx]/(atol + rtol*max(fabs(y[idx]), fabs(z[idx]))), 2));
+            % endfor
+            }
+        }
+    }
 % endif
 
-    return out;
+// Copy
```
FreddieWitherden: Indent.

pyfr/integrators/dual/controllers.py
```diff
@@ -68,35 +69,58 @@ def advance_to(self, t):
...
+            rl = np.array([np.sum(e) for e in zip(*errest.retval)])
+            # Reduce across MPI ranks
+            if rank != root:
+                comm.Reduce(rl, None, op=get_mpi('sum'), root=root)
+            else:
+                comm.Reduce(get_mpi('in_place'), rl, op=get_mpi('sum'),
+                            root=root)
+            # Normalise
+            res = np.sqrt(rl / self._gndofs)
+        # L^infinity norm
```
FreddieWitherden: Use actual infinity character.

pyfr/integrators/dual/controllers.py
```diff
@@ -68,35 +69,58 @@ def advance_to(self, t):
...
+        else:
+            # Reduce locally (element types) and globally (MPI ranks)
+            rl = np.maximum(np.zeros(self.system.nvars), *errest.retval)
+            # Reduce across MPI ranks
+            if rank != root:
+                comm.Reduce(rl, None, op=get_mpi('max'), root=root)
+            else:
+                comm.Reduce(get_mpi('in_place'), rl, op=get_mpi('max'),
```
FreddieWitherden: We need all reduce rather than reduce.

pyfr/integrators/std/controllers.py
```diff
@@ -120,21 +122,29 @@ def _errest(self, x, y, z):
         # L2 norm
         if self._norm == 'l2':
             # Reduce locally (element types) and globally (MPI ranks)
-            rl = sum(errest.retval)
-            rg = comm.allreduce(rl, op=get_mpi('sum'))
-
+            rl = np.array([np.sum(e) for e in zip(*errest.retval)])
+            # Reduce across MPI ranks
+            if rank != root:
+                comm.Reduce(rl, None, op=get_mpi('sum'), root=root)
```
FreddieWitherden: We need all reduce rather than reduce.

FreddieWitherden:
np.maximum does not accept multiple arrays; it accepts a, b, [out], where out is optional.  Thus this is probably not doing what you expect.

nloppi:
Hi Freddie,
I could not find a way to do the elementwise max reduction for NumPy arrays across mpi ranks. Now it is done via
rg = [comm.allreduce(e, op=get_mpi('max')) for e in rl]

In addition, I have been struggling to get the pymic working on the test machine. Problem was just the intel compiler... Anyway, Now I will test everything with Arvind's framework.

FreddieWitherden:
You need to use comm.Allreduce.  First, get everything into  NumPy array, then call comm.Allreduce using the in-place technique.

nloppi:
The last push was just fixes in the mic backend. I think I also found my bug with the MPI reduction and I'm working on it.

nloppi:
This now works, but I'm  not sure if it's ok. After Allreduce we still have two ranks all containg the same reduced array, right? Do I need to make an if statement that only root rank normalises and returns the value? What does the the other ranks do? return pass?

Thanks

pyfr/integrators/dual/controllers.py
```diff
@@ -68,35 +69,50 @@ def advance_to(self, t):
...
+            # Reduce locally (element types) and globally (MPI ranks)
+            rl = np.array([np.sum(e) for e in zip(*errest.retval)])
+            comm.Allreduce(get_mpi('in_place'), rl, op=get_mpi('sum'))
+
+            # Normalise
+            res = np.sqrt(rl / self._gndofs)
+        # L^∞ norm
+        else:
+            # Reduce locally (element types) and globally (MPI ranks)
+            rl = np.max(list(errest.retval), axis=0)
```
FreddieWitherden: This and line 104 is still ugly.  Start with `np.array(list(errest.retval))` and go from there.

nloppi:
Hi Freddie,

I cleaned up the kernels in OpenMP and MIC to fit the maximum line length. Particularly

                res${i} = max(
                          res${i},
                          pow(x[idx]/(atol + rtol*max(fabs(y[idx]),
                                                      fabs(z[idx]))), 2)
                          );
looks pretty awful. Do you have a suggestion for these? Shall I just store the 
pow(x[idx]/(atol + rtol*max(fabs(y[idx]), fabs(z[idx]))), 2) to another variable?

In addition, the previous pragma omp simd did not work on MIC. I changed it to #pragma omp parallel for. Is this right?

Sorry for these several iterations... and thanks again

pyfr/backends/cuda/blasext.py
```diff
@@ -56,29 +56,43 @@ def errest(self, x, y, z, *, norm):
...
+        nrow, ldim, dtype = x.traits
+        ncola, ncolb = x.ioshape[1:]
 
-        # Norm type
-        reduce_expr = 'a + b' if norm == 'l2' else 'max(a, b)'
+        # Determine the block size
+        block = (128, 1, 1)
+
+        # Determine the grid size
+        grid = (int((ncolb + block[0] - 1)/block[0]), 1, 1)
```
FreddieWitherden: What is wrong with the get grid for block function?  Is a block size of 128 always going to work?

pyfr/backends/cuda/blasext.py
```diff
@@ -56,29 +56,43 @@ def errest(self, x, y, z, *, norm):
...
         )
 
+        # Reduction across blocks outsourced to NumPy
+        reducer = np.max if norm == 'uniform' else np.sum
+
         class ErrestKernel(ComputeKernel):
             @property
             def retval(self):
-                return self._retarr.get()
+                cuda.memcpy_dtoh(err_host, err_dev)
```
FreddieWitherden: Do an async copy in run.

pyfr/backends/cuda/kernels/errest.mako
```diff
@@ -0,0 +1,97 @@
...
+                                    for i in range(ncola))};
+    fpdtype_t r;
+    int idx, tstride;
+    int tid = threadIdx.x;
+    int i = blockIdx.x*blockDim.x + tid;
+    int lastblksize = ncolb % blockDim.x;
+
+    if (i < ncolb)
+    {
+        // All threads load the residuals of the first row to shared memory
```
FreddieWitherden: Are these residuals?

pyfr/backends/cuda/kernels/errest.mako
```diff
@@ -0,0 +1,97 @@
...
+        % endfor
+            __syncthreads();
+        }
+    }
+
+    // Unrolled reduction within blocks
+    if (blockIdx.x != gridDim.x - 1)
+    {
+    % for n in range(pyfr.ilog2(sharesz)):
+        % if sharesz/(2**(n + 1)) > 32:
```
FreddieWitherden: No indent.

pyfr/backends/cuda/kernels/errest.mako
```diff
@@ -0,0 +1,97 @@
...
+        % endif
+        % endfor
+            __syncthreads();
+        }
+    }
+
+    // Unrolled reduction within blocks
+    if (blockIdx.x != gridDim.x - 1)
+    {
+    % for n in range(pyfr.ilog2(sharesz)):
```
FreddieWitherden: This block is messy.

pyfr/backends/opencl/kernels/errest.mako
```diff
@@ -0,0 +1,92 @@
...
+        % endif
+        % endfor
+            barrier(CLK_LOCAL_MEM_FENCE);
+        }
+    }
+
+    // Unrolled reduction within blocks
+    if (bid != gdim - 1)
+    {
+    % for n in range(pyfr.ilog2(sharesz)):
```
FreddieWitherden: This block is messy.  Should be a nicer way to iterate over the powers of two up to a number sharsz.

FreddieWitherden:
Line length are not such a big deal in the kernels themselves, so don't worry.

#pragma omp simd should work for the MIC (we use it all over the place).  This is something you will need to debug.

nloppi:
Now the kernels should be cleaner. I did all the changes except the simd pragma, which I'm still working on. Thanks

nloppi:
Okay, this is weird... 
The pragma omp simd works perfectly in l2 norm (there was a bug i->idx), but with uniform norm it results just zeros (or any value that res${i} are initialised with). In the uniform norm, If I change the pragma omp simd to pragma openmp parallel for, I get the right answer. So the error seems to be related to the combination of  max statement and simd. In addition, if I place a barrier in
                #pragma omp simd
                for (int cj = 0; cj < SOA_SZ; cj++)
                {
                % for i in range(ncola):
                    idx = r*ldim + X_IDX_AOSOA(${i}, ${ncola});
                    res${i} = max(pow(x[idx]/(atol + rtol*max(fabs(y[idx]), fabs(z[idx]))), 2), pow(x[idx]/(atol + rtol*max(fabs(y[idx]), fabs(z[idx]))), 2));
                    #pragma omp barrier
                % endfor
                }
the code cannot go past the barrier. Doesn't this mean that some threads never finish? race condition?

Do you have any ideas? It will be so embarrassing if it's a trivial bug e.g. syntax error.

pyfr/integrators/std/controllers.py
```diff
@@ -117,24 +119,26 @@ def _errest(self, x, y, z):
         self._prepare_reg_banks(x, y, z)
         self._queue % errest(self._atol, self._rtol)
 
+        error = np.array(list(errest.retval))
```
FreddieWitherden: Is the cast necessary?  A proxylist is an instance of a list.

pyfr/integrators/std/controllers.py
```diff
@@ -117,24 +119,26 @@ def _errest(self, x, y, z):
...
-            rg = comm.allreduce(rl, op=get_mpi('max'))
+            rl = np.max(error, axis=0)
+            comm.Allreduce(get_mpi('in_place'), rl, op=get_mpi('sum'))
 
             # Normalise
-            err = math.sqrt(rg)
+            err = np.sqrt(rl)
 
-        return err if not math.isnan(err) else 100
+        return err if not np.isnan(err).any() else 100
```
FreddieWitherden: This is wrong.

FreddieWitherden:
You need to repeat the reduction clause in the #pragma omp simd block so that it knows that it must do the reduction over the vector lanes.

nloppi:
Hi Freddie,
without the explicitly using the list(), the conversion to numpy array yields a alueError: invalid __array_struct__.

I still have not encountered a case where the 128 block/workgroup size did not work, but I am yet to try a 3D NS on my s10000 which is probably the most restrictive case. When MHD system comes we may want to tune it down at least for that system.

FreddieWitherden:
It would be interesting to know what the type of this variable is; can you pull it out?

We can not commit until we know exactly what the failure modes are in terms of required shared/local memory.  You will need to look these up.  Chances are the code will end up doing some computation to determine what the maximum size actually is based off of the number of field variables and the hardware in question.

pyfr/backends/cuda/kernels/errest.mako
```diff
@@ -0,0 +1,96 @@
...
+            __syncthreads();
+        }
+    }
+
+    // Unrolled reduction within blocks
+    if (blockIdx.x != gridDim.x - 1)
+    {
+    % for n in [int(sharesz/(2**(m + 1))) for m in range(pyfr.ilog2(sharesz))]:
+    % if n > 32:
+        if (tid < ${n})
```
FreddieWitherden: Remove extra space.

pyfr/backends/cuda/kernels/errest.mako
```diff
@@ -0,0 +1,96 @@
...
+	{
+        % elif n == 32:
+        if (tid < 32)
```
FreddieWitherden: Remove extra space.

pyfr/backends/cuda/kernels/errest.mako
```diff
@@ -0,0 +1,96 @@
...
+        % endif
+        % endfor
+            __syncthreads();
+        }
+    }
+
+    // Unrolled reduction within blocks
+    if (blockIdx.x != gridDim.x - 1)
+    {
+    % for n in [int(sharesz/(2**(m + 1))) for m in range(pyfr.ilog2(sharesz))]:
```
FreddieWitherden: Messy.  Write a generator function to do this.

pyfr/integrators/std/controllers.py
```diff
@@ -117,24 +119,26 @@ def _errest(self, x, y, z):
...
-            rg = comm.allreduce(rl, op=get_mpi('max'))
+            rl = np.max(error, axis=0)
+            comm.Allreduce(get_mpi('in_place'), rl, op=get_mpi('sum'))
 
             # Normalise
-            err = math.sqrt(rg)
+            err = np.sqrt(rl)
 
-        return err if not math.isnan(err) else 100
+        return err if not np.isnan(err).any() else [100]*self.system.nvars
```
FreddieWitherden: This is still wrong.  The result should always be a scalar.

nloppi:
Hi Freddie,
I checked that 128 work unit (block/workgroup) size is fine. In the kerneles, each work unit can only load simultaneously
8bytes*128*nvars,
so even with 8 field variables in MHD it is only 8192bytes.

Old Nvidia cards have 16kBytes of shared memory per work unit (opencl/cuda). In opencl AMD cards have at least 32kBytes of local memory per workgroup

references:
http://amd-dev.wpengine.netdna-cdn.com/wordpress/media/2013/12/AMD_OpenCL_Programming_User_Guide2.pdf
http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#axzz4Y0jmRCPP

nloppi:
I unified the variable names and fixed the standard integrator part. Now it works just like before, returning a reduced scalar that is the global maximum of all variables or l2 norm of all variables. The naming convention here is error/errest. 

In dual integrator we use the same errest kernel, but in a way that it return a vector of residuals (l2 or linfinity).  This entire vector is printed into dt_stats and the maximum of the of the value vector works as the convergence criterion. In the dual integrator the naming is residual.

FreddieWitherden:
We're getting close.  Please confirm the shared memory size for the various CPU OpenCL implementations.

The last thing to work through is to see if the trick whereby we avoid calls to __syncthreads() on NVIDIA GPUs when everything is happening inside of a warp is actually beneficial or not on modern hardware.

nloppi:
Fixed a minor typo bug in the cuda kernel. 

I did few tests with compressible couette flow using pi controller. The effect of avoiding __syncthreads() was negligible. I tested the new kernel against the old one in pycuda and found a tiny speedup ~6% (4:31/4:16)

FreddieWitherden:
The couette flow test case is somewhat pathological.  It might be better to test with a larger system.  However, if avoiding the __syncthreads() really does have very little impact (and I suspect it does), I suggest sticking them back in.  This will simplify the code and make it more similar to the OpenCL code.

nloppi:
About local memory on CPU OpenCL, I found:
" CPUs do not have a special physical memory designed as local memory,
which  means  that  all  memory  objects  in  local  memory  are
mapped into sections of global memory."   " CPU local memory is region of physical memory(DDR)"

https://pdfs.semanticscholar.org/8ce8/3dd85535034f7c482a0ff5e794802c20ddaf.pdf

So I guess the reduction in "local memory" for CPUs  is fine, but it does not benefit from it like GPUs do.

pyfr/backends/cuda/kernels/errest.mako
```diff
@@ -0,0 +1,96 @@
...
+        % for k in range(ncola):
+            idx = j*ldim + SOA_IX(i, ${k}, ${ncola});
+            r = pow(x[idx]/(atol + rtol*max(fabs(y[idx]), fabs(z[idx]))), 2);
+        % if norm == 'uniform':
+            sdata${k}[tid] = max(sdata${k}[tid], r);
+        % else:
+            sdata${k}[tid] += r;
+        % endif
+        % endfor
+            __syncthreads();
```
FreddieWitherden: Is this needed?  Everyone is just accessing their own bit of share memory and so the threads at this point are independent.  (Same for the above?)

pyfr/integrators/dual/controllers.py
```diff
@@ -68,35 +69,53 @@ def advance_to(self, t):
...
+        resid = np.array(list(errest.retval))
+
+        # L2 norm
+        if self._norm == 'l2':
+            # Reduce locally (element types) and globally (MPI ranks)
+            rl = np.sum(resid, axis=0)
+            comm.Allreduce(get_mpi('in_place'), rl, op=get_mpi('sum'))
+
+            # Normalise
+            res = np.sqrt(rl / self._gndofs)
```
FreddieWitherden: Put return here.

pyfr/integrators/dual/controllers.py
```diff
@@ -68,35 +69,53 @@ def advance_to(self, t):
...
+        # L^∞ norm
+        else:
+            # Reduce locally (element types) and globally (MPI ranks)
+            rl = np.max(resid, axis=0)
+            comm.Allreduce(get_mpi('in_place'), rl, op=get_mpi('max'))
+
+            # Normalise
+            res = np.sqrt(rl)
```
FreddieWitherden: Put return here.

pyfr/integrators/std/controllers.py
```diff
@@ -156,7 +160,7 @@ def advance_to(self, t):
             # Take the step
             idxcurr, idxprev, idxerr = self.step(self.tcurr, dt)
 
-            # Estimate the error
+            # Estimate the maximum error
```
FreddieWitherden: Fix comment.

nloppi:
Indeed, it did not need quite that many synchronisation points, but it seems that one is necessary just before starting the blockwise reduction.

I did a larger scale test with a 3D cylinder and there is a negligible difference in the speed (<2%) still favouring the new kernel.

pyfr/backends/opencl/blasext.py
```diff
@@ -50,33 +50,45 @@ def errest(self, x, y, z, *, norm):
...
+        block = (128, 1, 1)
 
-        # Build the reduction kernel
-        rkern = ReductionKernel(
-            self.backend.ctx, dtype, neutral='0', reduce_expr=reduce_expr,
-            map_expr='pow(x[i]/(atol + rtol*max(fabs(y[i]), fabs(z[i]))), 2)',
-            arguments='__global {0}* x, __global {0}* y, __global {0}* z, '
-                      '{0} atol, {0} rtol'.format(npdtype_to_ctype(dtype))
+        # Determine the number of groups
+        grid = (int((ncolb + block[0] - 1)/block[0]), 1, 1)
```
FreddieWitherden: Use // division.

pyfr/backends/opencl/blasext.py
```diff
@@ -50,33 +50,45 @@ def errest(self, x, y, z, *, norm):
...
+                            err_host.nbytes)
+
+        # Get the kernel template
+        src = self.backend.lookup.get_template('errest').render(
+            norm=norm, ncola=ncola, sharesz=block[0]
         )
 
+        # Build the reduction kernel
+        rkern = self._build_kernel(
+            'errest', src, [np.int32]*3 + [np.intp]*4 + [dtype]*2)
```
FreddieWitherden: Move ) down.

pyfr/backends/opencl/blasext.py
```diff
@@ -50,33 +50,45 @@ def errest(self, x, y, z, *, norm):
...
+
+        # Reduction across workgroups outsourced to NumPy
+        reducer = np.max if norm == 'uniform' else np.sum
+
         class ErrestKernel(ComputeKernel):
             @property
             def retval(self):
-                return self._retarr.get()
+                cl.enqueue_copy(self.queue.cl_queue_comp, err_host, err_dev)
```
FreddieWitherden: Make this copy async.

pyfr/backends/opencl/blasext.py
```diff
@@ -50,33 +50,45 @@ def errest(self, x, y, z, *, norm):
...
-                yarr = Array(qcomp, cnt, dtype, data=y.data)
-                zarr = Array(qcomp, cnt, dtype, data=z.data)
-
-                self._retarr = rkern(xarr, yarr, zarr, atol, rtol,
-                                     queue=qcomp)
+                self.queue = queue
+                rkern(queue.cl_queue_comp, (ncolb, 1, 1), block,
+                      nrow, ncolb, ldim, err_dev, x.data, y.data,
+                      z.data, atol, rtol
+                      )
```
FreddieWitherden: Identation.

pyfr/backends/opencl/kernels/errest.mako
```diff
@@ -0,0 +1,89 @@
...
+            r = pow(x[idx]/(atol + rtol*max(fabs(y[idx]), fabs(z[idx]))), 2);
+        % if norm == 'uniform':
+            sdata${k}[tid] = max(sdata${k}[tid], r);
+        % else:
+            sdata${k}[tid] += r;
+        % endif
+        % endfor
+        }
+    }
+    barrier(CLK_LOCAL_MEM_FENCE);
```
FreddieWitherden: Add newline.

pyfr/backends/mic/kernels/errest.mako
```diff
@@ -3,24 +3,87 @@
...
     fpdtype_t *x = *xp, *y = *yp, *z = *zp;
     fpdtype_t atol = *atolp, rtol = *rtolp;
 
-    fpdtype_t err = 0.0;
-
 % if norm == 'l2':
-    #pragma omp parallel for reduction(+:err)
-    for (int i = 0; i < *n; i++)
-        err += pow(x[i]/(atol + rtol*max(fabs(y[i]), fabs(z[i]))), 2);
+    #pragma omp parallel reduction(+ : ${','.join('err{0}'.format(i) for i in range(ncola))})
```
FreddieWitherden: Too much code duplication between L2 and uniform norms.  Refactor.

pyfr/integrators/std/controllers.py
```diff
@@ -117,22 +119,24 @@ def _errest(self, x, y, z):
         self._prepare_reg_banks(x, y, z)
         self._queue % errest(self._atol, self._rtol)
 
+        error = np.array(list(errest.retval))
+
         # L2 norm
         if self._norm == 'l2':
             # Reduce locally (element types) and globally (MPI ranks)
-            rl = sum(errest.retval)
+            rl = np.sum(error)
```
FreddieWitherden: Move onto the next line.

nloppi:
You wanted to know about the errest.retval type. It is

<class 'pyfr.util.proxylist'>
['__add__', '__call__', '__class__', '__contains__', '__delattr__', '__delitem__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattr__', '__getattribute__', '__getitem__', '__gt__', '__hash__', '__iadd__', '__imul__', '__init__', '__iter__', '__le__', '__len__', '__lt__', '__module__', '__mul__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__reversed__', '__rmul__', '__setattr__', '__setitem__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', 'append', 'clear', 'copy', 'count', 'extend', 'index', 'insert', 'pop', 'remove', 'reverse', 'sort']

No idea why we need to convert it to a normal list.

pyfr/integrators/std/controllers.py
```diff
@@ -117,22 +119,22 @@ def _errest(self, x, y, z):
...
             # Normalise
-            err = math.sqrt(rg / self._gndofs)
-        # Uniform norm
+            err = np.sqrt(rg / self._gndofs)
+        # L^∞ norm
         else:
             # Reduce locally (element types) and globally (MPI ranks)
-            rl = max(errest.retval)
-            rg = comm.allreduce(rl, op=get_mpi('max'))
+            rg = comm.allreduce(np.max(error), op=get_mpi('max'))
```
FreddieWitherden: Use Allreduce here for performance and symmetry.

nloppi:
I chose to use the lower case allreduce because then it is not necessary to pass a numpy array to the function.  Passing a float to Allreduce gives an error 

Traceback (most recent call last):
  File "/home/niki/python3venv/venv/bin/pyfr", line 9, in <module>
    load_entry_point('pyfr==1.5.0', 'console_scripts', 'pyfr')()
  File "/home/niki/python3venv/venv/lib/python3.5/site-packages/pyfr-1.5.0-py3.5.egg/pyfr/__main__.py", line 110, in main
  File "/home/niki/python3venv/venv/lib/python3.5/site-packages/pyfr-1.5.0-py3.5.egg/pyfr/__main__.py", line 235, in process_run
  File "/home/niki/python3venv/venv/lib/python3.5/site-packages/pyfr-1.5.0-py3.5.egg/pyfr/__main__.py", line 227, in _process_common
  File "/home/niki/python3venv/venv/lib/python3.5/site-packages/pyfr-1.5.0-py3.5.egg/pyfr/integrators/base.py", line 199, in run
  File "/home/niki/python3venv/venv/lib/python3.5/site-packages/pyfr-1.5.0-py3.5.egg/pyfr/integrators/std/controllers.py", line 165, in advance_to
  File "/home/niki/python3venv/venv/lib/python3.5/site-packages/pyfr-1.5.0-py3.5.egg/pyfr/integrators/std/controllers.py", line 129, in _errest
  File "MPI/Comm.pyx", line 714, in mpi4py.MPI.Comm.Allreduce (src/mpi4py.MPI.c:99618)
  File "MPI/msgbuffer.pxi", line 699, in mpi4py.MPI._p_msg_cco.for_allreduce (src/mpi4py.MPI.c:36296)
  File "MPI/msgbuffer.pxi", line 652, in mpi4py.MPI._p_msg_cco.for_cro_recv (src/mpi4py.MPI.c:35845)
  File "MPI/msgbuffer.pxi", line 148, in mpi4py.MPI.message_simple (src/mpi4py.MPI.c:30349)
  File "MPI/msgbuffer.pxi", line 86, in mpi4py.MPI.message_basic (src/mpi4py.MPI.c:29359)
  File "MPI/asbuffer.pxi", line 200, in mpi4py.MPI.getbuffer (src/mpi4py.MPI.c:9310)
  File "MPI/asbuffer.pxi", line 113, in mpi4py.MPI.PyObject_GetBufferEx (src/mpi4py.MPI.c:8070)
BufferError: Object is not writable.

Which one do you prefer? If we want to use Allreduce then we need to do another numpy conversion,  since numpy.sum over all axes returns a float.
 
rl = np.sum(error)
comm.Allreduce(get_mpi('in_place'), np.array([rl]), op=get_mpi('sum')) 

Or did you mean doing the reduction over MPI ranks before reducing across elements/variables?

Thanks.

FreddieWitherden:
Try allocating a NumPy float and passing this to Allreduce.  Failing this, allocate a single entry array and pass this.

nloppi:
Allreduce seems to always require a NumPy array. The array needs to be converted back to float immediately after the reduction as with the std integrator, the error estimate is used to compute dt. This causes errors for instance in the progress_bar module.

FreddieWitherden:
Okay, create a single element NumPy array, apply Allreduce, and then when you return cast as float(array[0]).

nloppi:
I just found out that reduction on the OpenCL backend goes haywire with single precision. I'm working on it.

nloppi:
Hi Freddie,

I heard that you had a talk with Peter about the residual output. 

Here is my proposal:
New plugin called pseudostats, that works just as dtstats. The controller calculates the errest/residual and appends stepinfo/pseudostep info depending on the formulation. Then dtstats and pseudostats plugins just handle the output. When the plugins are initialised they also check that the formulation is supported.

pyfr/backends/base/makoutil.py
```diff
@@ -14,6 +15,11 @@ def ndrange(context, *args):
     return util.ndrange(*args)
 
 
+def ilog2range(context, x):
+    n = int(math.log(x, 2))
+    return [x//(2**(i + 1)) for i in range(n)]
```
FreddieWitherden: I still do not like this.  Try using ``x.bit_length()`` as an alternative to log.

pyfr/backends/cuda/blasext.py
```diff
@@ -56,29 +56,44 @@ def errest(self, x, y, z, *, norm):
...
-        # Wrap
-        xarr = GPUArray(x.leaddim*x.nrow, x.dtype, gpudata=x)
-        yarr = GPUArray(y.leaddim*y.nrow, y.dtype, gpudata=y)
-        zarr = GPUArray(z.leaddim*z.nrow, z.dtype, gpudata=z)
+        nrow, ldim, dtype = x.traits
+        ncola, ncolb = x.ioshape[1:]
 
-        # Norm type
-        reduce_expr = 'a + b' if norm == 'l2' else 'max(a, b)'
+        # Determine the block size
```
FreddieWitherden: Misleading comment.

pyfr/backends/mic/kernels/errest.mako
```diff
@@ -3,24 +3,69 @@
...
+        int align = PYFR_ALIGN_BYTES / sizeof(fpdtype_t);
+        int rb, re, cb, ce, idx;
+        loop_sched_2d(*nrowp, *ncolbp, align, &rb, &re, &cb, &ce);
+        int nci = ((ce - cb) / SOA_SZ)*SOA_SZ;
+
+        for (int r = rb; r < re; r++)
+        {
+            for (int ci = cb; ci < cb + nci; ci += SOA_SZ)
+            {
+                #pragma omp simd
```
FreddieWitherden: Ditch the simd.

pyfr/backends/mic/kernels/errest.mako
```diff
@@ -3,24 +3,69 @@
...
+                for (int cj = 0; cj < SOA_SZ; cj++)
+                {
+                % if norm == 'uniform':
+                    #pragma omp parallel reduction(max : ${','.join('err{0}'.format(i) for i in range(ncola))})
```
FreddieWitherden: Should be no need for this.

pyfr/backends/openmp/blasext.py
```diff
@@ -55,23 +55,31 @@ def errest(self, x, y, z, *, norm):
...
+        nrow, ldim, dtype = x.traits
+        ncola, ncolb = x.ioshape[1:]
 
         # Render the reduction kernel template
-        src = self.backend.lookup.get_template('errest').render(norm=norm)
+        src = self.backend.lookup.get_template('errest').render(norm=norm,
+                                                                ncola=ncola)
+
+        # Array for the error estimate
+        error = np.zeros((1, ncola))
```
FreddieWitherden: np.zeros(ncola)

pyfr/backends/openmp/blasext.py
```diff
@@ -55,23 +55,31 @@ def errest(self, x, y, z, *, norm):
...
-            'errest', src, [np.int32] + [np.intp]*3 + [dtype]*2, restype=dtype
+            'errest', src, [np.int32]*3 + [np.intp]*4 + [dtype]*2,
+            restype=dtype
         )
 
         class ErrestKernel(ComputeKernel):
             @property
             def retval(self):
-                return self._retval
+                return error.get()[0]
```
FreddieWitherden: See above and then remove [0].

pyfr/backends/openmp/kernels/errest.mako
```diff
@@ -2,21 +2,61 @@
...
+        int align = PYFR_ALIGN_BYTES / sizeof(fpdtype_t);
+        int rb, re, cb, ce, idx;
+        loop_sched_2d(nrow, ncolb, align, &rb, &re, &cb, &ce);
+        int nci = ((ce - cb) / SOA_SZ)*SOA_SZ;
+
+        for (int r = rb; r < re; r++)
+        {
+            for (int ci = cb; ci < cb + nci; ci += SOA_SZ)
+            {
+                #pragma omp simd
```
FreddieWitherden: Remove.

pyfr/integrators/dual/controllers.py
```diff
@@ -15,13 +15,12 @@ def __init__(self, *args, **kwargs):
...
 
         # Stats on the most recent step
-        self.stepinfo = []
+        self.pseudostepinfo = []
 
     def _accept_step(self, dt, idxcurr):
         self.tcurr += dt
         self.nacptsteps += 1
         self.nacptchain += 1
-
```
FreddieWitherden: ?

pyfr/integrators/dual/controllers.py
```diff
@@ -50,8 +52,8 @@ def __init__(self, *args, **kwargs):
             raise ValueError('The maximum number of pseudo-iterations must '
                              'be greater than or equal to the minimum')
 
-        self._pseudo_aresid = self.cfg.getfloat(sect, 'pseudo-aresid')
-        self._pseudo_rresid = self.cfg.getfloat(sect, 'pseudo-rresid')
+        self._residtol= self.cfg.getfloat(sect, 'pseudo-residtol')
```
FreddieWitherden: Prefix with pseudo.

pyfr/integrators/dual/controllers.py
```diff
@@ -50,8 +52,8 @@ def __init__(self, *args, **kwargs):
...
+        self._norm = self.cfg.get(sect, 'resid-norm', 'uniform')
```
FreddieWitherden: Check user is not a moron.

pyfr/integrators/dual/controllers.py
```diff
@@ -68,10 +70,13 @@ def advance_to(self, t):
                 # Activate convergence monitoring after pseudo-niters-min
                 if i >= self._minniters - 1:
                     # Subtract the current solution from the previous solution
-                    self._add(-1.0/dtau, self._idxprev, 1.0/dtau, self._idxcurr)
```
FreddieWitherden: Is this a bug fix?

pyfr/integrators/dual/controllers.py
```diff
@@ -82,21 +87,33 @@ def advance_to(self, t):
...
-        return math.sqrt(rg)
+        self._prepare_reg_banks(x, x, x)
+        self._queue % errest(dtau, 0.0)
+
+        resid = np.array(list(errest.retval))
+
+        # L2 norm
+        if self._norm == 'l2':
+            # Reduce locally (element types) and globally (MPI ranks)
+            rl = np.sum(resid, axis=0)
```
FreddieWitherden: Move into Allreduce.

pyfr/integrators/dual/controllers.py
```diff
@@ -82,21 +87,33 @@ def advance_to(self, t):
...
+            comm.Allreduce(get_mpi('in_place'), rl, op=get_mpi('sum'))
+
+            # Normalise and return
+            return np.sqrt(rl / self._gndofs)
+        # L^∞ norm
+        else:
+            # Reduce locally (element types) and globally (MPI ranks)
+            rl = np.max(resid, axis=0)
```
FreddieWitherden: As above.

pyfr/integrators/dual/controllers.py
```diff
@@ -82,21 +87,33 @@ def advance_to(self, t):
...
-        # Reduce locally (element types) and globally (MPI ranks)
-        rl = max(errest.retval)
-        rg = comm.allreduce(rl, op=get_mpi('max'))
-
-        # Normalise
-        return math.sqrt(rg)
+        self._prepare_reg_banks(x, x, x)
+        self._queue % errest(dtau, 0.0)
+
+        resid = np.array(list(errest.retval))
```
FreddieWitherden: Move down.

pyfr/integrators/std/controllers.py
```diff
@@ -117,24 +119,26 @@ def _errest(self, x, y, z):
...
         self._queue % errest(self._atol, self._rtol)
 
+        error = np.array(list(errest.retval))
+
         # L2 norm
         if self._norm == 'l2':
             # Reduce locally (element types) and globally (MPI ranks)
-            rl = sum(errest.retval)
-            rg = comm.allreduce(rl, op=get_mpi('sum'))
+            rl = np.array([np.sum(error)])
```
FreddieWitherden: Still messy.

pyfr/plugins/pseudostats.py
```diff
@@ -0,0 +1,57 @@
...
+        # The root rank needs to open the output file
+        if rank == root:
+            self.outf = init_csv(self.cfg, cfgsect,
+                                 'total_niter,t,pseudo-iter,' + fvars)
+        else:
+            self.outf = None
+
+    def __call__(self, intg):
+        # Process the sequence of pseudo residuals
+        for i, (iternr, pseudo_resid) in enumerate(intg.pseudostepinfo,
```
FreddieWitherden: Why are we using enumerate here?

pyfr/plugins/pseudostats.py
```diff
@@ -0,0 +1,57 @@
...
+                                                   start=self.count):
+            self.stats.append((iternr + self.count + 1, self.tprev, iternr + 1)
+                              + tuple(pseudo_resid))
+
+        # Update the total step count and save the current time
+        self.count += intg.pseudostepinfo[-1][0] + 1
```
FreddieWitherden: len(self.stats)?

pyfr/plugins/pseudostats.py
```diff
@@ -0,0 +1,57 @@
...
+            raise RuntimeError('Formulation {0} not supported by plugin {1}'
+                               .format(intg.formulation, self.name))
+
+        self.flushsteps = self.cfg.getint(self.cfgsect, 'flushsteps', 500)
+
+        self.count = 0
+        self.stats = []
+        self.tprev = intg.tcurr
+
+        fvars = ','.join(intg.system.elementscls.privarmap[self.ndims])
```
FreddieWitherden: I think this should be the conservative variable map as that is what the reduction is based on.

pyfr/plugins/pseudostats.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class PseudoStatsPlugin(BasePlugin):
+    name = 'pseudostats'
+    systems = ['*']
+
+    def __init__(self, intg, cfgsect, prefix):
+        super().__init__(intg, cfgsect, prefix)
+
+        if intg.formulation != 'dual':
```
FreddieWitherden: Add generic support for limiting plugins to formulation (as we do at the moment with systems).

pyfr/backends/openmp/blasext.py
nloppi: The backend.matrix that we use for the output array buffer wants the second dimension.

pyfr/integrators/dual/controllers.py
nloppi: Do you mean by removing the default norm? Yes, defaults make user think/experiment less...

pyfr/integrators/dual/controllers.py
nloppi: No, the break condition is the same and just reformulated, residual=resid tol. Before we computed the residual with this add kernel and then used the errest kernel to relate the residual to the tolerances and the break criterion was normalised so that were comparing to unity.  This was unecessarily confusing.  In addition, we want to write to file the absolute residuals, not the ones that are normalised with the tolerances.

pyfr/plugins/pseudostats.py
nloppi: Residuals are computed and stats accumulated only after pseudo-niters-min. I want one of columns of the file to be the total number of iterations.

FreddieWitherden: For controlling the process I agree that we should always be sure to do a minimum number of iterations.  Although, now that the accumulation is done on the device it is somewhat confusing for the stats plugin to miss out on a bunch of data at the start of an iteration.

pyfr/backends/openmp/blasext.py
FreddieWitherden: We should not be using a backend matrix type for this.  A simple numpy array of the correct type will do the job.  Will need to pass it in to the kernel explicitly as as err.ctypes.data, however.

pyfr/plugins/base.py
```diff
@@ -63,6 +64,12 @@ def __init__(self, intg, cfgsect, suffix=None):
             raise RuntimeError('System {0} not supported by plugin {1}'
                                .format(intg.system.name, self.name))
 
+        # Check that we support this particular integrator formulation
+        if not ('*' in self.formulations
```
FreddieWitherden: Overkill to support *.  Just check for intg.formulation in self.formulations and require plugins to specify both std and dual.

pyfr/plugins/pseudostats.py
nloppi: I think this is still the most flexible approach. Sometimes it is useful not to output all the residual just to prevent flooding the file, for instance when performing long simulations. If user wants to monitor the entire process with fixed number of iterations, it can be done by setting the niters-min=1 and tolerance to zero. If only the last residual or a few just before hitting the creterion is wanted it is easy to set niters-min = niters-max or close to each other.

FreddieWitherden: So the plugin should not cause any more issues than the current dtstats plugin which outputs after every time step (even rejected steps) as each inner iteration is of comparable cost to each explicit iteration.  Performance issues can be addressed by simply reducing the flush frequency.

pyfr/backends/base/makoutil.py
```diff
@@ -14,6 +15,11 @@ def ndrange(context, *args):
     return util.ndrange(*args)
 
 
+def ilog2range(context, x):
+    n = x.bit_length() - 1
+    return [int(x/(2**(i + 1))) for i in range(n)]
```
FreddieWitherden: In [8]: x = 256

In [9]: [2**i for i in range(x.bit_length())]
Out[9]: [1, 2, 4, 8, 16, 32, 64, 128, 256]

pyfr/backends/cuda/blasext.py
```diff
@@ -56,29 +56,44 @@ def errest(self, x, y, z, *, norm):
...
-        # Wrap
-        xarr = GPUArray(x.leaddim*x.nrow, x.dtype, gpudata=x)
-        yarr = GPUArray(y.leaddim*y.nrow, y.dtype, gpudata=y)
-        zarr = GPUArray(z.leaddim*z.nrow, z.dtype, gpudata=z)
+        nrow, ldim, dtype = x.traits
+        ncola, ncolb = x.ioshape[1:]
 
-        # Norm type
-        reduce_expr = 'a + b' if norm == 'l2' else 'max(a, b)'
+        # Determine the reduction block dimensions
```
FreddieWitherden: Misleading comment.

pyfr/backends/mic/kernels/errest.mako
```diff
@@ -3,24 +3,63 @@
...
+                % endif
+                % endfor
+                }
+            }
+
+            for (int ci = cb + nci, cj = 0; cj < ce - ci; cj++)
+            {
+            % for i in range(ncola):
+                idx = r*ldim + X_IDX_AOSOA(${i}, ${ncola});
+                % if norm == 'uniform':
```
FreddieWitherden: Identation.

pyfr/backends/opencl/blasext.py
```diff
@@ -50,33 +50,46 @@ def errest(self, x, y, z, *, norm):
...
+        nrow, ldim, dtype = x.traits
+        ncola, ncolb = x.ioshape[1:]
 
-        # Norm type
-        reduce_expr = 'a + b' if norm == 'l2' else 'max(a, b)'
+        # Determine the reduction workgroup dimensions
+        block = (128, 1, 1)
+
+        # Determine the number of groups
+        grid = ((ncolb + block[0] - 1)//block[0], 1, 1)
```
FreddieWitherden: Space around //

pyfr/backends/opencl/blasext.py
```diff
@@ -50,33 +50,46 @@ def errest(self, x, y, z, *, norm):
...
             def run(self, queue, atol, rtol):
-                qcomp = queue.cl_queue_comp
-
-                xarr = Array(qcomp, cnt, dtype, data=x.data)
-                yarr = Array(qcomp, cnt, dtype, data=y.data)
-                zarr = Array(qcomp, cnt, dtype, data=z.data)
-
-                self._retarr = rkern(xarr, yarr, zarr, atol, rtol,
-                                     queue=qcomp)
+                self.queue = queue
```
FreddieWitherden: ?

pyfr/backends/openmp/kernels/errest.mako
```diff
@@ -2,21 +2,60 @@
...
+                % endif
+                % endfor
+                }
+            }
+
+            for (int ci = cb + nci, cj = 0; cj < ce - ci; cj++)
+            {
+            % for i in range(ncola):
+                idx = r*ldim + X_IDX_AOSOA(${i}, ${ncola});
+                % if norm == 'uniform':
```
FreddieWitherden: Indentation.

pyfr/plugins/base.py
```diff
@@ -63,6 +64,11 @@ def __init__(self, intg, cfgsect, suffix=None):
             raise RuntimeError('System {0} not supported by plugin {1}'
                                .format(intg.system.name, self.name))
 
+        # Check that we support this particular integrator formulation
+        if not (intg.formulation in self.formulations):
```
FreddieWitherden: not in.

pyfr/plugins/pseudostats.py
nloppi: The current approach allows controlling the flush frequency for the iterations themselves and also skipping real time-steps, while still monitoring the total number of pseudo iterations. I do not see the disadvantage.

If we want the feature that dualstats always outputs all the residuals when it is called, we need to do it with an if clause in the controller because the residual kernel is called there. 

 if i >= self._minniters - 1 or  if intg.nacptsteps % flushsteps == 0:
   calculate residual
   append pseudostats

This differs from the overall plugin methodology.

FreddieWitherden: Very well.  However, the use of a list here is confusing.  In the integrator should either use a dictionary or, alternatively, append None to pseudostats if we did not compute the residual that iteration.  

Either way, the current design is not flexible enough to handle dual time stepping with an error estimating outer controller.

FreddieWitherden:
In terms of the stats plugin I would change the code to firstly keep track of the total (global) number of pseudo iterations.  Say, self.npseudoiter.  Then I would do:

<take step>


```
if monitoring:
    <stuff>
    self.pseudostats.append((self.npseudoiter, i, ...))
else:
    self.pseudostats.append((self.npseudoiter, i, None)

self.npseudoiter += 1
```

this way any plugin always knows the number of iterations taken and there is no issue with rejected steps (i resets, but that's fine, as npseudoiter is always increasing no matter what);

nloppi:
I did as you suggested. Now it prints (None,)*nvars if the residuals are not computed.

pyfr/backends/base/makoutil.py
```diff
@@ -2,6 +2,7 @@
 
 from collections import Iterable
 import itertools as it
+import math
```
FreddieWitherden: Remove import.

pyfr/backends/base/makoutil.py
```diff
@@ -14,6 +15,10 @@ def ndrange(context, *args):
     return util.ndrange(*args)
 
 
+def ilog2range(context, x):
+    return [2**i for i in range(x.bit_length())][-2::-1]
```
FreddieWitherden: Rework range: ``range(x.bit_length() - 2, -1, -1)``.

pyfr/backends/opencl/blasext.py
```diff
@@ -50,33 +50,45 @@ def errest(self, x, y, z, *, norm):
...
+        nrow, ldim, dtype = x.traits
+        ncola, ncolb = x.ioshape[1:]
 
-        # Norm type
-        reduce_expr = 'a + b' if norm == 'l2' else 'max(a, b)'
+        # Reduction workgroup dimensions
+        block = (128, 1, 1)
+
+        # Determine the number of groups
+        grid = ((ncolb + block[0] - 1) // block[0], 1, 1)
```
FreddieWitherden: Expression is different to that used by the CUDA backend, check equivalence.

pyfr/backends/opencl/blasext.py
```diff
@@ -50,33 +50,45 @@ def errest(self, x, y, z, *, norm):
...
 
-        cnt = x.leaddim*x.nrow
-        dtype = x.dtype
+        nrow, ldim, dtype = x.traits
+        ncola, ncolb = x.ioshape[1:]
 
-        # Norm type
-        reduce_expr = 'a + b' if norm == 'l2' else 'max(a, b)'
+        # Reduction workgroup dimensions
+        block = (128, 1, 1)
```
FreddieWitherden: I do not believe that PyOpenCL requires the additional dimensions if they're 1.

pyfr/integrators/dual/controllers.py
```diff
@@ -15,7 +15,9 @@ def __init__(self, *args, **kwargs):
         self._fnsteps = self.cfg.getint('soln-filter', 'nsteps', '0')
 
         # Stats on the most recent step
-        self.stepinfo = []
+        self.pseudostepinfo = []
+        # Pseudo-step counter
```
FreddieWitherden: Add newline.

pyfr/integrators/dual/controllers.py
```diff
@@ -50,8 +55,8 @@ def __init__(self, *args, **kwargs):
             raise ValueError('The maximum number of pseudo-iterations must '
                              'be greater than or equal to the minimum')
 
-        self._pseudo_aresid = self.cfg.getfloat(sect, 'pseudo-aresid')
-        self._pseudo_rresid = self.cfg.getfloat(sect, 'pseudo-rresid')
+        self._pseudo_residtol= self.cfg.getfloat(sect, 'pseudo-resid-tol')
+        self._norm = self.cfg.get(sect, 'pseudo-resid-norm', 'uniform')
```
FreddieWitherden: Prefix with pseudo.

pyfr/plugins/pseudostats.py
```diff
@@ -0,0 +1,53 @@
...
+
+        fvars = ','.join(intg.system.elementscls.convarmap[self.ndims])
+
+        # MPI info
+        comm, rank, root = get_comm_rank_root()
+
+        # The root rank needs to open the output file
+        if rank == root:
+            self.outf = init_csv(self.cfg, cfgsect,
+                                 'total_niter,t,pseudo-iter,' + fvars)
```
FreddieWitherden: Inconsistent column names.

pyfr/integrators/dual/controllers.py
FreddieWitherden: Also dump into stats file (see collect_stats method).

pyfr/backends/opencl/blasext.py
nloppi: File "/home/niki/python3venv/venv/lib/python3.5/site-packages/pyopencl-2016.1-py3.5-linux-x86_64.egg/pyopencl/cffi_cl.py", line 1192, in enqueue_nd_range_kernel
    work_dim = len(global_work_size)
TypeError: object of type 'int' has no len()

pyfr/backends/opencl/blasext.py
nloppi: Changed so that it is passed to the kernel

pyfr/integrators/dual/controllers.py
nloppi: Do you mean dumping the global iteration number self.npseudosteps into stats?

nloppi:
Now the npseudosteps is dumped into the stats file.

Because the writer plugin is called for t=0.0 in the __init__ of BaseIntegrator, I need to initialise npseudosteps = 0 there. I can also do it in DualController/BaseDualIntegrator by switching the order
 
        self.npseudosteps = 0
        super().__init__(*args, **kwargs)

Which one do you prefer?

pyfr/backends/opencl/blasext.py
FreddieWitherden: Try (128,).

pyfr/backends/opencl/blasext.py
```diff
@@ -50,33 +50,45 @@ def errest(self, x, y, z, *, norm):
...
             def run(self, queue, atol, rtol):
-                qcomp = queue.cl_queue_comp
-
-                xarr = Array(qcomp, cnt, dtype, data=x.data)
-                yarr = Array(qcomp, cnt, dtype, data=y.data)
-                zarr = Array(qcomp, cnt, dtype, data=z.data)
-
-                self._retarr = rkern(xarr, yarr, zarr, atol, rtol,
-                                     queue=qcomp)
+                rkern(queue.cl_queue_comp, (ncolb, 1, 1), block,
```
FreddieWitherden: Might be able to get away with (ncolb,)

FreddieWitherden:
Putting it in the init function before we call super will be fine.

PR #118, opened 2017-02-28
Update documentation.

nloppi:


doc/src/user_guide.rst
```diff
@@ -1091,13 +1165,38 @@ with
 
 1. ``type`` --- type of boundary condition:
 
-    ``char-riem-inv`` | ``no-slp-adia-wall`` | ``no-slp-isot-wall`` |
-    ``slp-adia-wall`` | ``sub-in-frv`` | ``sub-in-ftpttang`` |
-    ``sub-out-fp`` | ``sup-in-fa`` | ``sup-out-fn``
+    ``ac-in-fv`` | ``ac-out-fp`` | ``char-riem-inv`` |
```
pv101: Are we happy with solver specific 'ac' prefix on ac-in-fv and ac-out-fp?

FreddieWitherden: I believe we had a discussion over this and the conclusion was that it is the best of a bad bunch of options for some of the AC BCs will have different params to regular BCs.

PR #119, opened 2017-03-10
Example case for the AC-Navier-Stokes system.

nloppi:


FreddieWitherden:
For the mesh can you compress the .msh with gzip (-9) to give .msh.gz.  Then replace the import command with

gzcat mesh.gz | pyfr import -tgmsh - mesh.pyfrm

which is a useful trick to know, anyway.

pv101:
How much space do we save by zipping? Extra commands might confuse new users?

FreddieWitherden:
As big as the rest of PyFR.  Either way, knowing the extra commands is useful and most users are just going to copy and paste anyway.

nloppi:
Seems reduce by quite a lot, from 456K to 186K

FreddieWitherden:
It also marks the file binary (although we can do that explicitly, and absolutely should if we decide not to go with the zipped option) which prevents goo from showing up in git diffs.

doc/src/user_guide.rst
```diff
@@ -1477,3 +1477,50 @@ simulation on a structured mesh:
...
+3. Copy the `Gmsh <http:http://geuz.org/gmsh/>`_ mesh file
+   ``PyFR/examples/inc_cylinder_2d/inc_cylinder_2d.msh`` into
+   ``inc_cylinder_2d/``
+
+4. Run pyfr to covert the `Gmsh <http:http://geuz.org/gmsh/>`_
+   mesh file into a PyFR mesh file called ``inc_cylinder_2d.pyfrm``::
+
+        pyfr import inc_cylinder_2d.msh inc_cylinder_2d.pyfrm
+
+5. Run pyfr to solve the Navier-Stokes equations on the mesh,
```
pv101: incompressible Navier-Stokes

pv101:
OK - fine with me.

pv101:
Should we now change 'Example --- 2D Couette Flow' -> 'Example --- 2D Compressible Couette Flow'?

pv101:
I have just tested, and the website seems to build fine.

nloppi:
Which one of these commands should we use in the user guide: gunzip -c, zcat or gzcat? At least on my Ubuntu 16.04 gzcat is installed as just zcat.

FreddieWitherden:
Any of them are fine, all widely available and do the same thing; gzcat is arguably the most direct in terms of the command saying what it does.

nloppi:
Peter, I don't think we should change it to Compressible Couette Flow. I remember that we once had a discussion about the system names. Navier-Stokes by itself refers to the compressible system and Incompressible Navier-Stokes is just its special case. Same analogy should be with case naming, Couette flow and Incompressible Couette flow.

pv101:
Yes, I agree about the system naming. However, I think colloquially at least, most people would take Couette Flow on its own to be the incompressible case! But happy to leave as it is at the moment if you think it is better.

examples/inc_cylinder_2d/inc_cylinder_2d.ini
```diff
@@ -0,0 +1,80 @@
...
+flushsteps = 20
+file = residual.csv
+header = true
+
+[soln-plugin-writer]
+dt-out = 5.0
+basedir = .
+basename = cylinder_2d_{t:.2f}
+
+[solver-source-terms]
```
FreddieWitherden: Source terms?

pv101: I think the case has a time dependent sponge layer, which switches off once the initial transient has passed.

nloppi: Yes, it is a time dependent sponge layer. The sponge is on just for the first 5 time units near the edges of the domain. When you start to develop the flow from scratch, a very strong pressure wave is created. Since we don't have a non-reflective boundary condition (tricky in AC), it is convenient to kill this initial wave to prevent it bouncing back. 

Here is a snapshot at t=5 when the sponge switches off. 
sponge

nloppi: I'm sure there will be questions about damping the strong initial wave in the transient phase. Then we can just refer to this case. One could pseudo march it to death, but being high-order and using rk4 as a pseudo-stepper, it would take a long time to dissipate it!

FreddieWitherden: Lets add a comment noting why we have the source terms.

pv101: OK, @nloppi if we add a comment to the .ini file I think we are good to merge the PR.

PR #120, opened 2017-03-30
Renaming the field names for tavg.

iyer-arvind:
This renames the fields while being written by tavg. It replaces the '-'
with '_' so that paraview does not get confused.


FreddieWitherden:
Push a more general fix into the VTK writer; gradients may have a similar problem.  This will also solve the problem of backwards compatibility between new and old time averages.

PR #121, opened 2017-04-10
Fixes (p)vtu writer field names.

iyer-arvind:
Replaces all non alphanumeric characters in the name of a field to '_'
while writing to .(p)vtu files.

### DIFF
--- a/pyfr/writers/vtk.py
+++ b/pyfr/writers/vtk.py
@@ -244,7 +244,7 @@ class VTKWriter(BaseWriter):
         array.tofile(vtuf)
 
     def _process_name(self, name):
-        return re.sub('[^0-9a-zA-Z]+', '_', name)
+        return re.sub('\W+', '_', name)
 
     def _write_serial_header(self, vtuf, mk, off):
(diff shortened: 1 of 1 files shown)


pyfr/writers/vtk.py
```diff
@@ -242,6 +243,9 @@ def _write_darray(self, array, vtuf, dtype):
         np.uint32(array.nbytes).tofile(vtuf)
         array.tofile(vtuf)
 
+    def _process_name(self, name):
+        return re.sub('[^0-9a-zA-Z]+', '_', name)
```
FreddieWitherden: \W

pyfr/writers/vtk.py
```diff
@@ -242,6 +243,9 @@ def _write_darray(self, array, vtuf, dtype):
         np.uint32(array.nbytes).tofile(vtuf)
         array.tofile(vtuf)
 
+    def _process_name(self, name):
+        return re.sub('\W+', '_', name)
```
FreddieWitherden: Use a raw string here.

FreddieWitherden:
Squish into one commit and we're good to go.

PR #122, opened 2017-04-13
Polynomial multigrid for dual time stepping.

nloppi:
I changed the structure so that only the DualIntegrator is affected. Now, DualIntegrator uses variable length lists for systems and registers. Even without multigrid they are lists with length one, just to make the necessary methods suitable for both cases without too many if statements.

Another bigger change is in the pseudostepper implementation. Previously we added the real time derivative source term at each stage addition using the method _add_with_dts. Now, the real-time derivative is added directly to the register containing -div(f) with _rhs_with_dts.  This makes pseudosteppers cleaner and also facilitates local time stepping, so that it can be handled with a single pointwise multiplication.

Thanks!

### DIFF
--- a/pyfr/integrators/base.py
+++ b/pyfr/integrators/base.py
@@ -58,9 +58,6 @@ class BaseIntegrator(object, metaclass=ABCMeta):
         # Construct the relevant mesh partition
         self.system = systemcls(backend, rallocs, mesh, initsoln, nreg, cfg)
 
-        # Storage register banks
-        self._regs, self._regidx = self._get_reg_banks(nreg)
-
         # Extract the UUID of the mesh (to be saved with solutions)
         self.mesh_uuid = mesh['mesh_uuid']
 
@@ -70,9 +67,6 @@ class BaseIntegrator(object, metaclass=ABCMeta):
         # Global degree of freedom count
         self._gndofs = self._get_gndofs()
 
-        # Bank index of solution
-        self._idxcurr = 0
-
         # Solution cache
         self._curr_soln = None
 
@@ -81,12 +75,6 @@ class BaseIntegrator(object, metaclass=ABCMeta):
 
         # Record the starting wall clock time
         self._wstart = time.time()
-
-        # Event handlers for advance_to
-        self.completed_step_handlers = proxylist(self._get_plugins())
-
-        # Delete the memory-intensive elements map from the system
-        del self.system.ele_map
 
     def _get_reg_banks(self, nreg):
         regs, regidx = [], list(range(nreg))
@@ -122,7 +110,7 @@ class BaseIntegrator(object, metaclass=ABCMeta):
 
         return plugins
 
-    def _get_kernels(self, name, nargs, **kwargs):
+    def _get_kernels(self, name, nargs, level=None, **kwargs):
         # Transpose from [nregs][neletypes] to [neletypes][nregs]
         transregs = zip(*self._regs)
 
@@ -133,20 +121,21 @@ class BaseIntegrator(object, metaclass=ABCMeta):
 
         return kerns
 
-    def _prepare_reg_banks(self, *bidxes):
+    def _prepare_reg_banks(self, *bidxes, level=None):
         for reg, ix in zip(self._regs, bidxes):
             reg.active = ix
 
     @memoize
-    def _get_axnpby_kerns(self, n, subdims=None):
-        return self._get_kernels('axnpby', nargs=n, subdims=subdims)
+    def _get_axnpby_kerns(self, n, level=None, subdims=None):
+        return self._get_kernels('axnpby', nargs=n, level=level,
+                                 subdims=subdims)
 
-    def _add(self, *args):
+    def _add(self, *args, level=None):
         # Get a suitable set of axnpby kernels
-        axnpby = self._get_axnpby_kerns(len(args) // 2)
+        axnpby = self._get_axnpby_kerns(len(args) // 2, level=level)
 
         # Bank indices are in odd-numbered arguments
-        self._prepare_reg_banks(*args[1::2])
+        self._prepare_reg_banks(*args[1::2], level=level)
 
         # Bind and run the axnpby kernels
         self._queue % axnpby(*args[::2])
--- a/pyfr/integrators/dual/base.py
+++ b/pyfr/integrators/dual/base.py
@@ -3,24 +3,99 @@
 from abc import abstractmethod
 
 from pyfr.integrators.base import BaseIntegrator
+from pyfr.util import memoize, proxylist
 
 
 class BaseDualIntegrator(BaseIntegrator):
     formulation = 'dual'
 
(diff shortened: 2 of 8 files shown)


pyfr/inifile.py
```diff
@@ -98,6 +98,10 @@ def getfloat(self, section, option, default=_sentinel):
     def getint(self, section, option, default=_sentinel):
         return int(self.get(section, option, default))
 
+    def getintlist(self, section, option, default=_sentinel):
```
FreddieWitherden: What is wrong with getliteral?

pyfr/integrators/dual/base.py
```diff
@@ -3,24 +3,116 @@
...
+            self._leveliters = self.cfg.getintlist(sect, 'mgcycle')
+            # Number of levels
+            self.levels = len(self._leveliters)
+
+        for level in range(self.levels):
+            self.dualsystems.append(systemcls(backend, rallocs, mesh,
+                                              initsoln, nreg, cfg, level))
+            self._dualregs.append(self._get_reg_banks(nreg, level)[0])
+            self._dualregidx.append(self._get_reg_banks(nreg, level)[1])
+            self._dualidxcurr.append(0)
```
FreddieWitherden: Outside of the loop you can just do:

self._dualidxcurr = [0]*self.levels
self._dualidxprev = [0]*self.levels

pyfr/integrators/dual/base.py
```diff
@@ -3,24 +3,116 @@
...
+        axnpby = self._get_axnpby_kerns(len(args) // 2, level=level)
+
+        # Bank indices are in odd-numbered arguments
+        self._prepare_reg_banks(*args[1::2], level=level)
+
+        # Bind and run the axnpby kernels
+        self._queue % axnpby(*args[::2])
+
+    @property
+    def soln(self):
```
FreddieWitherden: How is this different to what we usually do?

pyfr/integrators/dual/base.py
```diff
@@ -29,3 +121,97 @@ def _dual_time_source(self):
     @abstractmethod
     def finalise_step(self, currsoln):
         pass
+
+    @memoize
+    def prolrest(self, l1, l2):
+        prolrestkerns = proxylist([])
+
+        for etype in self.system.ele_types:
+            l1sys = self.dualsystems[l1]
```
FreddieWitherden: Should be inside the loop.

pyfr/integrators/dual/base.py
```diff
@@ -29,3 +121,97 @@ def _dual_time_source(self):
     @abstractmethod
     def finalise_step(self, currsoln):
         pass
+
+    @memoize
+    def prolrest(self, l1, l2):
+        prolrestkerns = proxylist([])
+
+        for etype in self.system.ele_types:
```
FreddieWitherden: Probably looping over the wrong thing, want to loop over l1eles, l2eles directly.

pyfr/integrators/dual/base.py
```diff
@@ -29,3 +121,97 @@ def _dual_time_source(self):
...
+                                    out=l2eles.scal_upts_inb)
+            )
+
+        return prolrestkerns
+
+    def _get_prolrestmatrix(self, l1eles, l2eles):
+        ub1 = l1eles.basis.ubasis
+        l2b = l2eles.basis
+
+        mat = ub1.nodal_basis_at(l2b.upts)
```
FreddieWitherden: Wont this introduce aliasing when restricting?

pyfr/integrators/dual/base.py
```diff
@@ -29,3 +121,97 @@ def _dual_time_source(self):
...
+        add, rhs = self._add, self._rhs_with_dts
+
+        # Registers [level][(stepper_regs)(source_regs)(multigrid_regs)]
+        # R = [l(1)][mg(1)] = -∇·f - dQ/dt
+        rhs(self.tcurr, idxcurr[l1], regidx[l1][mgregidx[1]],
+            c=1/dt, level=l1, passmg=True)
+
+        l = 0 if l1 == 0 else 1
+        # d = [l(1)][mg(2)] = r - R
+        add(-1, regidx[l1][mgregidx[1]], l, regidx[l1][mgregidx[0]], level=l1)
```
FreddieWitherden: This indexing is sufficiently messy that it is likely we're abstracting the problem incorrectly.

pyfr/integrators/base.py
```diff
@@ -82,12 +76,6 @@ def __init__(self, backend, systemcls, rallocs, mesh, initsoln, cfg):
         # Record the starting wall clock time
         self._wstart = time.time()
 
-        # Event handlers for advance_to
-        self.completed_step_handlers = proxylist(self._get_plugins())
-
-        # Delete the memory-intensive elements map from the system
-        del self.system.ele_map
```
FreddieWitherden: This does not get deleted for the dual integrator.  Non-starter.

pyfr/integrators/dual/base.py
```diff
@@ -3,24 +3,116 @@
...
     @property
     def _stepper_regidx(self):
-        return self._regidx[:self._pseudo_stepper_nregs]
+        return self._dualregidx[0][:self._pseudo_stepper_nregs]
 
     @property
     def _source_regidx(self):
-        return self._regidx[self._pseudo_stepper_nregs:]
+        pnreg, dsrc = self._pseudo_stepper_nregs, self._dual_time_source
+        return self._dualregidx[0][pnreg:pnreg + len(dsrc) - 1]
```
FreddieWitherden: These 0's suggest that we've got the wrong abstraction.

FreddieWitherden:
As it stands there is too much code duplication and too much fancy indexing for this to go mainline.

pyfr/integrators/dual/base.py
```diff
@@ -3,24 +3,123 @@
...
+        sect = 'solver-time-integrator'
 
-        self._dtau = self.cfg.getfloat('solver-time-integrator', 'pseudo-dt')
+        self._dtau = self.cfg.getfloat(sect, 'pseudo-dt')
         self.dtaumin = 1.0e-12
 
+        # Dual Integrator treats following variables as lists to accommodate MG
+        self.dualsystems = []
+        self._dualregs = []
+        self._dualidxcurr = []
```
FreddieWitherden: Reinitalized below.

pyfr/solvers/base/elements.py
```diff
@@ -144,8 +145,8 @@ def set_backend(self, backend, nscal_upts, nonce):
 
         # Convenience functions for scalar/vector allocation
         alloc = lambda ex, n: abufs.append(
-            backend.matrix(n, extent=nonce + ex, tags={'align'})
-        ) or abufs[-1]
+            backend.matrix(n, extent='{0}_{1}'.format(nonce + ex, self.level),
```
FreddieWitherden: Why is the nonce not sufficient?

PR #124, opened 2017-06-06
Remove the abstract base classes.

nloppi:


FreddieWitherden:
For the moment can we just remove the abstract base classes from the integrators?  It may well be that going forwards we also decide to remove them from other classes, but for the moment lets just handle the integrators.

nloppi:
Ok!

FreddieWitherden:
When removing @abstractproperty you need to replace it with @property.

PR #127, opened 2017-07-12
Fix an issue with multigrid specific element sections.

nloppi:


FreddieWitherden:
What is the issue here that this fixes?

PR #128, opened 2017-07-13
Improvements for Windows Platforms

NauticalMile64:
- Further modified the way the C runtime library is loaded to be friendly to Python 3.3 and 3.4 users on Windows
- Updated progress bar display
- Instruct to use mpiexec for parallel runs, as mpirun is not available from some distributions (i.e. MS-MPI)
- using the recommended "Development Mode" from distutils works well, but creates an egg-info folder in the root dir. gitignore'd it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 MANIFEST
 dist/
 build/
+pyfr.egg-info/
 
 # Documentation
 doc/build/
--- a/doc/src/user_guide.rst
+++ b/doc/src/user_guide.rst
@@ -151,7 +151,7 @@ Running in Parallel
 ^^^^^^^^^^^^^^^^^^^
 
 ``pyfr`` can be run in parallel. To do so prefix ``pyfr`` with
-``mpirun -n <cores/devices>``. Note that the mesh must be
+``mpiexec -n <cores/devices>``. Note that the mesh must be
 pre-partitioned, and the number of cores or devices must be equal to
 the number of partitions.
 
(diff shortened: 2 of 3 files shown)


FreddieWitherden:
Regarding mpiexec vs mpirun: are there distributions which ship with one but not the other?

Also, on Windows what is the specific issue with having the progress bar go the end of the console?  Is PyFR overshooting or is there a bug in the console itself?

NauticalMile64:
At very least Microsoft's MPI distribution does not include an mpirun executable. I don't know of any other mpi distributions that have mpirun but not mpiexec. While the name mpiexec is not required by the MPI standard, it comes more highly reccomended than mpirun (See section 8.8 of the 2.2 and 3.1 standards).

OpenMPI, MPICH, and MVAPICH2 all have both mpirun and mpiexec. In OpenMPI they are synonomous, and in MPICH mpiexec is reccomended in all cases over mpirun.

As for the console line I believe there's an extra space printed on every newline in pyfr on Windows 10, So when the progress bar + text is allowed to fill the width of command window it spills over onto the next line, so using '\r' to return the cursor moves it back to the beginning of the newline it just created (+ the space it likes to insert) and it prints a new progress bar, which overfills again, and so on. I looked for a little bit, but could not find where the extra space is being inserted. Maybe you can spot the root cause, but for me the simple fix of shortening the progress bar width seems good enough (although I would never cut corners like this for the simulation code).

FreddieWitherden:
Thank you for this.  I am happy to cherry pick everything except the progress bar tweaks (with a view to getting this issue resolved completely for the next point release).

NauticalMile64:
Sounds good.

NauticalMile64:
Do I need to do anything else, or can you just pull in the commits of your choosing?

FreddieWitherden:
If you could update the pull request (ideally rebasing from the current head of release/v1.7.0 to avoid an unnecessary merge), that would be great.

NauticalMile64:
Re-opening request after new commits added.

FreddieWitherden:
Thanks for this.  One final thing is by convention we usually treat the first line of the commit message as a proper sentence.  Would you be able to add a full stop to the end of the three commit messages?

NauticalMile64:
Updated commit messages to include full stops.

PR #129, opened 2017-07-13
Updated the User Guide.

pv101:
Here's a shot at the updated User Guide. Niki/Freddie, please have a look.

### DIFF
--- a/doc/src/user_guide.rst
+++ b/doc/src/user_guide.rst
@@ -537,6 +537,31 @@ Example::
     min-fact = 0.3
     max-fact = 2.5
 
+[solver-dual-time-integrator-multip]
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Parameterises multi-p for dual time-stepping with
+
+1. ``pseudo-dt-fact`` --- factor by which the pseudo time-step size
+   changes between multi-p levels:
+
+    *float*
+
+2. ``cycle`` --- nature of a single multi-p cycle:
+
+    ``[(`` *int* ``,`` *int* ``), (`` *int* ``,`` *int* ``), ... (``
+    *int* ``,`` *int* ``)]``
+
+    where the first *int* in each bracketed pair is the multi-p level,
+    and the second *int* in each bracketed pair is the number of pseudo
+    time-steps taken at that multi-p level
+
+Example::
+
+    [solver-dual-time-integrator-multip]
+    pseudo-dt-fact = 2.3
+    cycle = [(3, 1), (2, 1), (1, 2), (2, 1), (3, 3)]
+
 [solver-interfaces]
 ^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


doc/src/user_guide.rst
```diff
@@ -537,6 +537,31 @@ Example::
...
+    *float*
+
+2. ``cycle`` --- nature of a single multi-p cycle:
+
+    ``[(`` *int* ``,`` *int* ``), (`` *int* ``,`` *int* ``), ... (``
+    *int* ``,`` *int* ``)]``
+
+    where the first *int* in each bracketed pair is the multi-p level,
+    and the second *int* in each bracketed pair is the number of pseudo
+    time-steps taken at that multi-p level
```
FreddieWitherden: Remark that must start and end on the polynomial order of the simulation.  Also can only change ±1 between levels.

doc/src/user_guide.rst
```diff
@@ -592,10 +617,11 @@ Example::
     quad-deg = 10
     quad-pts = gauss-legendre
 
-[solver-interfaces-tri]
-^^^^^^^^^^^^^^^^^^^^^^^
+[solver-interfaces-tri-{mg-*p*}]
```
FreddieWitherden: Confusing. As we have -mg-p<x> whereas this looks like *p* is substituted.

doc/src/user_guide.rst
```diff
@@ -1097,7 +1130,13 @@ where available.  Parameterised with
 
 2. ``samp-pts`` --- list of points to sample:
 
-    ``[(x, y), (x, y), ...]`` | ``[(x, y, z), (x, y, z), ...]``
+    ``[(`` *float* ``,`` *float* ``), (`` *float* ``,`` *float* ``),
```
FreddieWitherden: Ugly.  Harder to read than before.

doc/src/user_guide.rst
```diff
@@ -537,6 +537,31 @@ Example::
...
+Parameterises multi-p for dual time-stepping with
+
+1. ``pseudo-dt-fact`` --- factor by which the pseudo time-step size
+   changes between multi-p levels:
+
+    *float*
+
+2. ``cycle`` --- nature of a single multi-p cycle:
+
+    ``[(`` *int* ``,`` *int* ``), (`` *int* ``,`` *int* ``), ... (``
```
FreddieWitherden: [(order, nsteps), ..., (order, nsteps)]

doc/src/user_guide.rst
pv101: Is solver-interfaces-tri{-mg-*p*} better? The idea was to show that the whole -mg-*p* is optional, and *p* is replaced by the level you want to set params for?

doc/src/user_guide.rst
pv101: (level,nsteps) ?

pv101: Since order could be confusing?

doc/src/user_guide.rst
pv101: OK, will revert

doc/src/user_guide.rst
FreddieWitherden: Level is meaningless, it is a term we have never used before.  Order, as in polynomial order, is unambiguous.

doc/src/user_guide.rst
FreddieWitherden: I did not do a particularly good job explaining what I meant.  An example of a config line is [solver-interfaces-tri-mg-p3].  Hence why it is important to indicate that the p is required, rather than something which is substituted.

pv101: My fault, I got the syntax wrong, I did not realise we actually have a 'p'. So we could do:
solver-interfaces-tri{-mg-p*p*}
or following your other comment
solver-interfaces-tri{-mg-p*order*}

doc/src/user_guide.rst
pv101: I think the fact that order is already used and defined as the top level order could be confusing ... but no strong feelings, as long as explained in subsquent text.

PR #142, opened 2018-10-05
Documentation.

pv101:


pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -117,7 +117,7 @@ def __init__(self, *args, **kwargs):
         tplargs['maxf'] = self.cfg.getfloat(sect, 'max-fact', 1.01)
         tplargs['minf'] = self.cfg.getfloat(sect, 'min-fact', 0.98)
         tplargs['saff'] = self.cfg.getfloat(sect, 'safety-fact', 0.8)
-        tplargs['dtau_maxf'] = self.cfg.getfloat(sect, 'pseudo-dt-max-fact',
+        tplargs['dtau_maxf'] = self.cfg.getfloat(sect, 'pseudo-dt-max',
                                                  3.0)
```
FreddieWitherden: Can we now bring this up a line?

doc/src/user_guide.rst
```diff
@@ -515,25 +514,24 @@ Parameterises the time-integration scheme used by the solver with
...
                *float*
 
-            - ``max-fact`` --- maximum factor that the local
+            - ``max-fact`` --- maximum factor by which the local
               pseudo time-step can change between iterations
               (suitable range 1.001-1.01)
 
                *float*
 
-            - ``pseudo-dt-max-fact`` --- maximum permissible
```
FreddieWitherden: Should this still be suffixed?

pv101: Removal is intentional, and thus we rely on the explanation to tell user that this is a multiplier as opposed to an absolute value (as is the case for the physical analogue). Reason for removal was due to associations of max-fact above with incremental time-step changes. But debatable - there is not a completely elegant solution ...

FreddieWitherden: Any reason we do not just make it an absolute value?

pv101: I defer to @nloppi

FreddieWitherden: So my view here is that we either come up with a sensible name for this constant or move to absolute values with the current name (which is consistent with the standard time stepper).

pv101: I'm personally not averse to moving to an absolute value, and leaving the user to do the maths. Also has advantage of being consistent with the current stepper.

nloppi: Sorry, I prefer a relative value. I think having an absolute value with a relative default value is more confusing than a slightly akward name.

FreddieWitherden: What do you mean by a relative default value?  If we went with an absolute value I would see the key becoming mandatory with no default.

nloppi: self.dtaumax = self.cfg.getfloat(sect, 'pseudo-dt-max', 3.0*self._dtau)

A factor of 3 works great in most cases, and I think it should be the default. We can't really specify an absolute default as in dtmax

self.dtmax = self.cfg.getfloat(sect, 'dt-max', 1e2)

FreddieWitherden: So due to the way default values are handled in PyFR this may be somewhat confusing especially if one tries to restart or reuse the resulting config file later on (since an absolute value will now be hardcoded).  If the value is absolute then we can not reasonably have a variable default.

To this end we should look at getting a better name for this factor that is consistent with the rest of PyFR.

nloppi: Maybe pseudo-dt-limit (or rlimit) and explanation in the user guide?

FreddieWitherden: What about -mult?

pv101: I've updated the PR with -mult.

FreddieWitherden:
Can we get this squished into one or two commits?

pv101:
@nloppi I've updated the PR with -mult. Can you review quickly. Do you want the commits squashed into one before merging to develop?

nloppi:
@pv101, looks good to me. Thanks.

pv101:
I have squashed into a single commit. @FreddieWitherden if all looks OK please merge into develop of the main repo, and mint an RC. @nloppi please then run tests on the RC and we are hopefully good to go.

PR #143, opened 2018-10-11
Fix a minor issue with the dualcontroller.

nloppi:


FreddieWitherden:
Can you amend so we have "DualController" in the commit message.

PR #144, opened 2018-10-12
Fixed the time step sanity check with DualController.

nloppi:


pyfr/integrators/dual/phys/base.py
```diff
@@ -44,3 +46,11 @@ def soln(self):
             )
 
         return self._curr_soln
+
+    def call_plugin_dt(self, dt):
+        rem = math.fmod(dt, self._dt)
+        tol = 5.0*self.dtmin
+        if rem > tol and (self._dt - rem) > tol:
+            raise ValueError('Output times must be multiplies of dt')
```
FreddieWitherden: Plugin times.

pv101: multiplies -> multiples

nloppi: Plugin output times?

pyfr/integrators/dual/phys/base.py
```diff
@@ -44,3 +46,11 @@ def soln(self):
             )
 
         return self._curr_soln
+
+    def call_plugin_dt(self, dt):
+        rem = math.fmod(dt, self._dt)
+        tol = 5.0*self.dtmin
+        if rem > tol and (self._dt - rem) > tol:
+            raise ValueError('Plugin output times must be multiples of dt')
```
FreddieWitherden: Maybe call is better than output?

FreddieWitherden:
Okay!  Final sanity check is to convince ourselves that self._dt is always set when call_plugin_dt is invoked.

nloppi:
> Okay! Final sanity check is to convince ourselves that self._dt is always set when call_plugin_dt is invoked.

Yes.

self._dt is set in the BaseIntegrator constructor. The super().__init__(*) in BaseDualIntegrator is called before self.completed_step_handlers = proxylist(self._get_plugins()). You can see it in

class BaseDualIntegrator(BaseIntegrator):
    formulation = 'dual'

    def __init__(self, backend, systemcls, rallocs, mesh, initsoln, cfg):
        super().__init__(backend, rallocs, mesh, initsoln, cfg)

        # Get the pseudo-integrator
        self.pseudointegrator = get_pseudo_integrator(
            backend, systemcls, rallocs, mesh,
            initsoln, cfg, self._stepper_coeffs, self._dt
        )

        # Event handlers for advance_to
        self.completed_step_handlers = proxylist(self._get_plugins())

        # Delete the memory-intensive elements map from the system
        del self.system.ele_map

PR #147, opened 2019-02-12
Fixed another issue with gradient averaging

GiorgioGiangaspero:
same fix done for the file nputil.py


FreddieWitherden:
If you could amend the commit message to end with a . this should be good to go.

PR #148, opened 2019-03-18
Added optimal RK schemes (Vermeire et al. 2019)

nloppi:


pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +337,309 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
     @property
     def _pseudo_stepper_order(self):
         return 4
+
+
+class DualOptRKPseudoStepper(BaseDualPseudoStepper):
+    def __init__(self, *args, **kwargs):
```
FreddieWitherden: No need to override; this is the default.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +337,309 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+        r0, r1, r2, r3, r4, r5, r6, r7 = self._pseudo_stepper_regidx
+
+        # Ensure r0 references the bank containing u(n+1,m)
+        if r0 != self._idxcurr:
+            r0, r1 = r1, r0
+
+        # First stage
+        rhs(t, r0, r1)
+
+        # Second stage
```
FreddieWitherden: A loop is probably best here — especially given the regularity in the second to eighth stage.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +337,309 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+            self.b[6]*dtau, r7)
+
+        # Return the index of the bank containing u(n+1,m+1)
+        return r1, r0
+
+
+class DualOptRKp1(DualOptRKPseudoStepper):
+    pseudo_stepper_name = 'optrk_p1'
+
+    a = np.array([
```
FreddieWitherden: Just use a list of lists here.

pyfr/integrators/dual/pseudo/multip.py
```diff
@@ -61,6 +61,10 @@ def __init__(self, backend, systemcls, rallocs, mesh, initsoln, cfg,
             if l == self._order:
                 bases = [cc, pc]
             else:
+                if pn.startswith('optrk'):
```
FreddieWitherden: This smells funny.  Add a property to the pseudo steppers which describes the attribute of the scheme you are looking for rather than relying on a naming convention.

nloppi:
Sorry. Should not have amended. 

I managed to do implement the RK schemes using loops. I left the expanded version as a comment code block for now.

However, I am still trying to figure out what is the best way to do the special cases for rkopt since subclass_where uses the names. Do you have any suggestions? Thanks.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +337,316 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+
+
+class DualOptRKPseudoStepper(BaseDualPseudoStepper):
+    @property
+    def _pseudo_stepper_has_lerrest(self):
+        return False
+
+    @property
+    def _stepper_nfevals(self):
+        return 7*self.nsteps
```
FreddieWitherden: Lets do a generic dense explicit RK scheme class and then have the magic schemes be a subclass thereof.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +337,316 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+        # Ensure r0 references the bank containing u(n+1,m)
+        if r[0] != self._idxcurr:
+            r[0], r[1] = r[1], r[0]
+
+        # First stage
+        rhs(t, r[0], r[1])
+
+        # Stages two to nregs
+        for i in range(1, self._pseudo_stepper_nregs - 1):
+            rc = [ar for ars in zip([a*dtau for a in self.a[i][:i]],
```
FreddieWitherden: This looks like it can be simplified.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +337,316 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+
+        # Return the index of the bank containing u(n+1,m+1)
+        return r[1], r[0]
+
+
+class DualOptRKp1(DualOptRKPseudoStepper):
+    pseudo_stepper_name = 'optrk1'
+
+    a = [
+        [0, 0, 0, 0, 0, 0, 0],
```
FreddieWitherden: Drop 0's.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +337,316 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+        [0.2161768331428258, 0, 0, 0, 0, 0, 0],
+        [0.2161763099027749, 0.2161763099027749, 0, 0, 0, 0, 0],
+        [0.216173556224235, 0.216173556224235, 0.2161740794576209, 0, 0, 0, 0],
+        [0.2161735562337322, 0.2161735562337322, 0.2161734529385136,
+         0.2161762066157379, 0, 0, 0],
+        [0.2161732559391219, 0.2161732559391219, 0.2161728401742457,
+         0.2161755938436643, 0.2161762203689764, 0, 0],
+        [0.2161732559425067, 0.2161732559425067, 0.2161728401882148,
+         0.2161755938576336, 0.2161737836501301, 0.2161743964170724, 0]
```
FreddieWitherden: These numbers look dirty.  Are you sure they are not (in reality) all supposed to be the same value?

FreddieWitherden:
> Sorry. Should not have amended.
> 
> I managed to do implement the RK schemes using loops. I left the expanded version as a comment code block for now.
> 
> However, I am still trying to figure out what is the best way to do the special cases for rkopt since subclass_where uses the names. Do you have any suggestions? Thanks.

We may want to think about trying design a mechanism which groups a collection of schemes together as a family.  Then, in the config file, something should be done to make it clear that a family is being employed as opposed to just a single scheme.

pyfr/integrators/dual/pseudo/pseudosteppers.py
nloppi: The Butcher tableaus are generated from the optimised stability polynomial coefficients. For a given stability polynomial, the Butcher tableau is under defined. I have plotted the stability regions of these tableaus and they should be fine.

FreddieWitherden: Okay.  What is typically done in the literature here is for rational coefficients to be employed (see Carpenter and Kennedy for instance).

nloppi:
Hi Freddie. I think this is the cleanest approach.

First you find the pseudo_stepper or family as:
         pc = subclass_where(BaseDualPseudoStepper, pseudo_stepper_name=pn)

Then with dualintegrator you always try to find whether there is p-optimised version of that scheme as
        try:
            pc = subclass_where(pc, pseudo_stepper_porder=porder)
        except KeyError:
            pass

I still need to do some clean up, but what do you think about the approach?

FreddieWitherden:
That seems like a reasonable approach.  Lets proceed with that and see how it looks.

nloppi:
Hi Freddie,

That should do it. I have tested it for a few different p individually and for a pmg cycle between p0-p4 (all levels picking the correct scheme). ~2x speed-up in all cases.

nloppi:
Switching to rational coefficients would make the tableaus quite ugly and long. Moreover, now they are consistent with those listed in the supplementary material of the original paper.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +337,234 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+    def _pseudo_stepper_order(self):
+        return 1
+
+    @property
+    def _pseudo_stepper_nregs(self):
+        return 8
+
+
+class DualRKp1OptPseudoStepper(DualRKOptPseudoStepper):
+    pseudo_stepper_porder = '1'
```
FreddieWitherden: This should be an integer.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +337,234 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+    ]
+
+    b = [
+        0.1888425763820983, 0.1267393084745009, 0.1061061571001407,
+        0.1525740388611983, 0.1650271578073516, 0.1178619741653911,
+        0.1428487872093191
+    ]
+
+
+class DualRKp0OptPseudoStepper(DualRKp1OptPseudoStepper):
```
FreddieWitherden: Might be cleaner if we put the p thing as a suffix so DualRKOptPseudoStepper_p0.  We'll never use the class directly by name so it should not be an issue.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +337,234 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+        # Ensure r0 references the bank containing u(n+1,m)
+        if r[0] != self._idxcurr:
+            r[0], r[1] = r[1], r[0]
+
+        # First stage
+        rhs(t, r[0], r[1])
+
+        # Other stages
+        for i in range(self._pseudo_stepper_nregs - 2):
+            ar = [ar for ars in zip([a*dtau for a in self.a[i]],
```
FreddieWitherden: This could still be simplifiedl

FreddieWitherden:
> Switching to rational coefficients would make the tableaus quite ugly and long. Moreover, now they are consistent with those listed in the supplementary material of the original paper.

Okay, if we want to stick with rational coefficients I suggest adopting the same approach we've taken with quadrature rules.  Whack the coefficients themselves into text files whose names describe the properties of the scheme.  The init code can then get a list of these files and proceed accordingly.  This also avoids the need to define explicit classes for each scheme.

As a first step it is probably worth going over the quadrature rule code to familiarise yourself with the approach.  Then lets pick a naming convention along with a set of suffixes.

pyfr/integrators/dual/pseudo/pseudosteppers.py
nloppi: Do you have any suggestions? I can't think of a nice alternative one liner without using numpy. Thanks.

FreddieWitherden: chain from itertools is what you need.  Specifically, `ar = chain(*zip(..., ...))`.  May want to factor out the multiplications by dtau to the top of the function.  So `a = [self._dtau*ai for ai in self.a]`.  Also saves some needless multiplications in the inner for loop.

pyfr/denserkschemes/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+import re
+
+import numpy as np
+
+
+class BaseTabulatedDensePseudoStepper(object):
+    def __init__(self, rule):
+        a = []
+
+        rule = re.sub(r'(?<=\))\s*,?\s*(?!$)', r'\n', rule)
```
FreddieWitherden: Unsure how much of this is needed.  (We have it for quadrature rules as you can specify them in the ini file.)

pyfr/denserkschemes/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+        rule = re.sub(r'\(|\)|,', '', rule).strip()
+        rule = rule[1:-1] if rule.startswith('[') else rule
+
+        for l in rule.splitlines()[:-1]:
```
FreddieWitherden: Read all at once in a coeffs variable array and then split as a, b = coeffs[:-1], coeffs[-1]

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +340,67 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+
+        # Final accumulation
+        br = chain(*zip(b[1:], r[2:]))
+        add(b[0], r[1], 1, r[0], *br)
+
+        # Return the index of the bank containing u(n+1,m+1)
+        return r[1], r[0]
+
+
+class DualRKOptPseudoStepper(DualDenseRKPseudoStepper):
```
FreddieWitherden: Not sure we need to make this 'opt' specific.

pyfr/denserkschemes/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+            m = re.match(r'([a-zA-Z0-9\-~+]+)-s(\d+)'
+                         r'(?:-p(\d+))?\.txt$', path)
+            if m:
+                yield (path, int(m.group(2)), int(m.group(3) or -1))
+
+    def __init__(self, porderin):
+        best = None
+        for rpath, rnstages, rporder in self._iter_scheme():
+            # See if this rule fulfils the required criterion
+            if rporder == porderin:
```
FreddieWitherden: Might want to think about this logic.

pyfr/denserkschemes/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+            raise ValueError('No pseudostepper found for order = '
+                             + str(porderin))
+
+        # Load the pseudostepper
+        scheme = resource_string(__name__, '{}/{}'.format(self.pstype, best))
+        super().__init__(scheme.decode('utf-8'))
+
+
+def get_densepseudostepper(pstypein, porderin):
+    class DensePseudoStepper(BaseDensePseudoStepper):
```
FreddieWitherden: Use type() here as it enables us to give the class a better name.

pyfr/denserkschemes/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+            if not l:
+                continue
+
+            # Store the Butcher A matrix
+            a.append([float(f) for f in l.split()])
+
+        # The last line is the b vector
+        b = [float(f) for f in rule.splitlines()[-1].split()]
+
+        # Cast
```
FreddieWitherden: Do we want/need these as NumPy arrays?

pyfr/denserkschemes/rkopt/vermeire-loppi-vincent-s7-p0.txt
```diff
@@ -0,0 +1,7 @@
+0.2328811281838825 0.0 0.0 0.0 0.0 0.0 0.0
```
FreddieWitherden: Strip the 0's (don't want to encourage people to try and write out an implicit scheme!)

pyfr/denserkschemes/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+        self.a = np.array(a, dtype=np.float)
+        self.b = np.array(b, dtype=np.float)
+
+
+class BaseDensePseudoStepper(BaseTabulatedDensePseudoStepper):
```
FreddieWitherden: Should be general enough to handle a random dense scheme (no p-optimal variants).

pyfr/denserkschemes/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+    @classmethod
+    def _iter_scheme(cls):
+        rpaths = getattr(cls, '_rpaths', None)
+        if rpaths is None:
+            cls._rpaths = rpaths = resource_listdir(__name__, cls.pstype)
+
+        for path in rpaths:
+            m = re.match(r'([a-zA-Z0-9\-~+]+)-s(\d+)'
+                         r'(?:-p(\d+))?\.txt$', path)
```
FreddieWitherden: p might be confusing as it could also refer to the order of accuracy of the scheme.  Should probably settle on a suffix for denoting the optimal spatial polynomial order.

nloppi:
Hi Freddie,

Now the implementation supports arbitrary dense RK-schemes. Dense schemes can be activated by setting pseudo-stepper=dense. In addition, the user needs to specify psname which is matched to the .txt file name e.g. 

pseudo-stepper=dense
psname=rkopt

The .txt file names must specify the name of the scheme (psname) and stagecount  e.g. jameson-s4.txt. The implementation also looks for p-optimised variants which can be given as e.g. rkopt-s7-p4.txt

FreddieWitherden:
How about o or po as a suffix?

pyfr/denserkschemes/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+            cls._rpaths = rpaths = resource_listdir(__name__, 'schemes')
+        for path in rpaths:
+            m = re.match(r'([a-zA-Z0-9\-~+]+)-s(\d+)'
+                         r'(?:-p(\d+))?\.txt$', path)
+            if m:
+                yield (path, m.group(1), int(m.group(2)), int(m.group(3) or -1))
+
+    def __init__(self, namein, porderin):
+        best = None
+        for rpath, rname, rnstages, rporder in self._iter_scheme():
```
FreddieWitherden: This logic needs looking over.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +340,66 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
     @property
     def _pseudo_stepper_order(self):
         return 4
+
+
+class DualDenseRKPseudoStepper(BaseDualPseudoStepper):
+    pseudo_stepper_name = 'dense'
```
FreddieWitherden: Not sure about this.  We might want to simply synthesise classes on the fly.  So as part of the init process we can list all of the dense schemes in PyFR and then generate (using the type function) suitable subclasses.  Everything else then works as expected with no changes needed to inifiles or scheme-selection logic.

nloppi:
Hi Freddie,
This approach allows us to use the current inifile format e.g. pseudo-stepper=rk-opt. However, it does make the multip class modifications slightly messy.

I also tried combining the class as
        try:
            pc = subclass_where(BaseDualPseudoStepper, pseudo_stepper_name=pn)
        except KeyError:
            pc = type(pn, (DensePseudoScheme, DualDenseRKPseudoStepper),
                      dict())
...
        pseudointegrator = type(name, (cc, pc), dict(name=name))

        # Construct and return an instance of this new integrator class
        return pseudointegrator(backend, systemcls, rallocs, mesh,
                                initsoln, cfg, tcoeffs, dt)

where DensePseudoScheme would take polynomial order from cfg at construction. However, that multi-inheritance caused several other issues.

FreddieWitherden:
So I think the best approach is to detach the code which enables us to have families of schemes from the code which enables us to have tabulated RK schemes.  I suggest starting with the former and putting together a clean commit which just adds this functionality.  We can then iterate on what we feel is the cleanest design.  Then, when we are happy, we can iterate on the best way of getting the coefficients from disk.

nloppi:
Hi Freddie, 

What do you think about this approach? I changed subclass_where such that it can find the stepper names from an array of names. The idea is to include all the names of all schemes that fit into the category as

class DualDenseRKPseudoStepper(BaseDualPseudoStepper):
    pseudo_stepper_name = ['rkopt', 'other_densefamily', 'myscheme']

Then the constructor sets self.a, self.b and self.nstages for a given scheme by calling the parser method (to be written). Since multip has already updated the p-order in the cfgs before construction, multip.py does not have to be modified

nloppi:
I will rebase the commits after we have agreed on the approach. Thanks.

FreddieWitherden:
So finding all of the dense schemes is easy and very much a solved problem.  The function `pkg_resources.resource_listdir` can easily let us find all of the coefficient files and from this we can obtain a list of available dense schemes.

What we need to decide on is how we handle the fact that there can be multiple schemes with the same name (one scheme per polynomial order).  This is our primary issue.

nloppi:
The commit above should have everything. It finds the correct scheme based on the polynomial order and the name given in the config file and multip is handled automatically.

If one wants to add more schemes, it can be done by adding the name of the family to the name list under DualDenseRKPseudoStepper class, and writing the coefficients in a text file with naming convention:

familyname-s{nstages}-po{polyorder}.txt

If the schemes are not polynomial order specific, you can write 

familyname-s{nstages}-po*.txt

pyfr/integrators/dual/pseudo/densepseudoschemes/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+from pkg_resources import resource_listdir, resource_string
+import re
+
+
+def _get_coefficients(scheme):
+        a = []
+
+        scheme = re.sub(r'(?<=\))\s*,?\s*(?!$)', r'\n', scheme)
+        scheme = re.sub(r'\(|\)|,', '', scheme).strip()
+        scheme = scheme[1:-1] if scheme.startswith('[') else scheme
```
FreddieWitherden: Why do we have this line?

pyfr/integrators/dual/pseudo/densepseudoschemes/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+
+        # The last row is the b vector
+        return a[:-1], a[-1]
+
+def get_densepseudoscheme(namein, porderin):
+    rpaths = resource_listdir(__name__, 'schemes')
+    scheme = None
+
+    for path in rpaths:
+        m = re.match(r'([a-zA-Z0-9\-~+]+)-s(\d+)'r'(?:-po(\d+))?\.txt$', path)
```
FreddieWitherden: Two strings?

pyfr/integrators/dual/pseudo/densepseudoschemes/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+        name, nstages, porder = m.group(1), m.group(2), m.group(3)
+        if name == namein and (int(porder) == porderin or porder == '*'):
```
FreddieWitherden: Don't use * for any porder; some systems such as Windows do not permit * in file names.  Just drop the -po suffix for a scheme which can work at any order.

pyfr/util.py
```diff
@@ -152,7 +152,8 @@ def subclass_where(cls, **kwargs):
     k, v = next(iter(kwargs.items()))
 
     for s in subclasses(cls):
-        if hasattr(s, k) and getattr(s, k) == v:
+        if hasattr(s, k) and (getattr(s, k) == v or
+                              any(i == v for i in getattr(s, k))):
```
FreddieWitherden: This is probably not a good design; if only because strings are iterable.

nloppi: I'm still working on this. Everything else should be ok.

FreddieWitherden: So I think we should consider extending `subclass_where` to process multiple arguments so we could say `subclass_where(Base, name=name, po=po)`.  This should not be too much more work and would provide us with a simple means of differentiating schemes within a family.

nloppi: I am not sure. I don't think differentiating schemes within a family is the issue here. 

The main challenge is finding the DensePseudoStepper subclass for all dense families. Lets say we have 'rkopt-s7-px.txt', 'otherdensescheme-s5.txt',  'magic-s4-px.txt' which are all dense schemes. The subclass where need to match all these names 'rkopt', 'otherdensescheme' and 'magic' to the DensePseudoStepper class. This now can be done by adding them into

class DualDenseRKPseudoStepper(BaseDualPseudoStepper):
    pseudo_stepper_name = ['rkopt', 'otherdensescheme', 'magic-s4' ]

And having the modified sub_class where that can handle lists:

def subclass_where(cls, **kwargs):
    k, v = next(iter(kwargs.items()))

    for s in subclasses(cls):
        if hasattr(s, k):
            if getattr(s, k) == v:
                return s
            elif type(getattr(s, k)) == list and any(i == v for i
                                                     in getattr(s, k)):
                return s

    raise KeyError("No subclasses of {0} with cls.{1} == '{2}'"
                   .format(cls.__name__, k, v))

The constructor queries the polynomial order from the cfg and the right scheme within gets picked since multip passes p-order modifified cfgs to the pseudo-integrators.

FreddieWitherden: This is not a huge challenge.  All we need to do in an __init__ block somewhere is list all of the available dense schemes and then create (sub) classes for them using the type method.  This will register them with PyFR (indeed, it will not be able to tell the difference between a class whose coefficients are hard coded and one which reads them from disk).

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,15 +1,26 @@
...
-from pyfr.integrators.dual.pseudo.pseudosteppers import BaseDualPseudoStepper
+from pyfr.integrators.dual.pseudo.pseudosteppers import BaseDualPseudoStepper, DualDenseRKPseudoStepper
 from pyfr.integrators.dual.pseudo.multip import DualMultiPIntegrator
 from pyfr.util import subclass_where
 
 
 def get_pseudo_integrator(backend, systemcls, rallocs, mesh,
                           initsoln, cfg, tcoeffs, dt):
+    # Create subclasses for all dense pseudoscheme families
+    schemefams = []
```
FreddieWitherden: Do this once outside the function in the global __init__ block.

pyfr/integrators/dual/pseudo/densepseudoschemes/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+
+from pkg_resources import resource_listdir, resource_string
+import re
+
+
+def _get_coefficients(scheme):
+        a = []
+
+        scheme = re.sub(r'(?<=\))\s*,?\s*(?!$)', r'\n', scheme)
+        scheme = re.sub(r'\(|\)|,', '', scheme).strip()
```
FreddieWitherden: Why is there here?

pyfr/integrators/dual/pseudo/densepseudoschemes/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+            if not l:
+                continue
+
+            a.append([float(f) for f in l.split()])
+
+        # The last row is the b vector
+        return a[:-1], a[-1]
+
+def get_densepseudoscheme(namein, porderin):
+    rpaths = resource_listdir(__name__, 'schemes')
```
FreddieWitherden: This should not be here.  Once the classes are registered (created with type) the rest of the code should not need to make a distinction between different types of pseudo schemes.

FreddieWitherden:
Just going through the PR I note we still have not settled on a way of handling families of schemes.  Can we please get this resolved entirely independently of **any** work around stored or tabulated schemes.  Until we get this right there is no point going further as we risk getting ahead of ourselves and having to needlessly rewrite/review code.

nloppi:
> Just going through the PR I note we still have not settled on a way of handling families of schemes. Can we please get this resolved entirely independently of **any** work around stored or tabulated schemes. Until we get this right there is no point going further as we risk getting ahead of ourselves and having to needlessly rewrite/review code.

I not quite following. Do you mean handling schemes within a family?

The families of schemes will be handled by creating a DenseRKStepper subclass with an pseudo_stepper_name='familyname', which the subclass_where then finds.

For handling schemes within a family we can do two approaches:
1. What we currently have. Constructor of DenseRKStepper calls

        porder = cfg.getint('solver', 'order')
        pn = cfg.get('solver-time-integrator', 'pseudo-scheme')

        self.a, self.b, self.nstages = get_densepseudoscheme(pn, porder)

   which reads the correct scheme and returns correct Butcher coefficients.

2. The DenseRKStepper subclass will be created not just for different families but all schemes with attributes pseudo_stepper_name = familyname and pseudo_stepper_porder = polyorder
   Then the subclass_where looks subclass_where(pseudo_stepper_name = cfg.name, pseudo_stepper_porder= cfg.polyorder)
   In this approach parsing the butcher tableaus would be moved to the __init__ of DenseRKStepper.

Have not tried the second approach but it should be ok.

FreddieWitherden:
So what I do not quite understand is why families have anything to do with dense RK schemes.  They are separate.  You could have a family of schemes which are not dense, or not RK schemes, for instance.  Thus we should first decide on a fully generic mechanism by which a scheme can say what (solution) polynomial order it is designed for.  Then we can handle the dense stuff later.

nloppi:
Ah, Ok. Now we are on the same page.

In this release, my idea was to include only DenseRKPseudoStepper class which is a baseclass for all schemes and families of schemes which have dense Butcher Tableau. 

For any other type of families we'd write a new class. For example,
TypeXRKPseudoStepper class would be a base class for all schemes or families of schemes which have a have a Butcher tableau and step method of type X. 
To include the schemes you'd just need to add a TypeXpseudoSchemes dir with the schemes in it and generate them as subclasses. The  __init__.py in the folder would implement an appropriate parser/logic for the selection. 

Moreover, being dependent on the polynomial order is specific to the p-optimised family of schemes so I don't think we need a generic mechanism to check this for all different types of schemes.  For example, TypeX schemes might want to use completely different logic for selecting the schemes e.g. smaller stage counts in the intermediate multip stages.

FreddieWitherden:
> Moreover, being dependent on the polynomial order is specific to the p-optimised family of schemes so I don't think we need a generic mechanism to check this for all different types of schemes. 

So there are p-optimised RK schemes for SD and (if I recall) DG which are time accurate.  Thus I see no reason why our approach should not be generic; especially as it has the potential to reduce the amount of boiler plate code.  To this end I suggest extending the `subclass_with` function to support multiple properties.  This is nice as it will also work in case we want to query for more attributes in the future (such as the system type or the number of stages for those looking to minimise memory use).

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,12 +1,45 @@
...
 
 from pyfr.integrators.dual.pseudo.pseudocontrollers import BaseDualPseudoController
-from pyfr.integrators.dual.pseudo.pseudosteppers import BaseDualPseudoStepper
+from pyfr.integrators.dual.pseudo.pseudosteppers import BaseDualPseudoStepper, DualDenseRKPseudoStepper
 from pyfr.integrators.dual.pseudo.multip import DualMultiPIntegrator
 from pyfr.util import subclass_where
 
+# Create subclasses for all tabulated dense schemes
+tabschemes = []
+rpaths = resource_listdir(__name__, 'densepseudoschemes/schemes')
```
FreddieWitherden: Have the schemes a level up.  They should just be schemes (with appropriate flags to mark them as being not time accurate).

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,12 +1,45 @@
...
+for path in rpaths:
+    m = re.match(r'([a-zA-Z0-9\-~+]+)-s(\d+)(?:-po(\d+))?\.txt$', path)
+    name, nstages, porder = m.group(1), m.group(2), m.group(3)
+
+    attrs = dict(pseudo_stepper_name=name, nstages=int(nstages),
+                 path=path)
+
+    if porder:
+        attrs['pseudo_stepper_porder'] = int(porder)
+        name = name + porder
```
FreddieWitherden: name += '_po{}'.format(porder)

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,12 +1,45 @@
...
+
+    tabschemes.append(type(name, (DualDenseRKPseudoStepper,), attrs))
+
```
FreddieWitherden: Missing newline

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,12 +1,45 @@
...
+from pkg_resources import resource_listdir
 import re
 
 from pyfr.integrators.dual.pseudo.pseudocontrollers import BaseDualPseudoController
-from pyfr.integrators.dual.pseudo.pseudosteppers import BaseDualPseudoStepper
+from pyfr.integrators.dual.pseudo.pseudosteppers import BaseDualPseudoStepper, DualDenseRKPseudoStepper
 from pyfr.integrators.dual.pseudo.multip import DualMultiPIntegrator
 from pyfr.util import subclass_where
 
+# Create subclasses for all tabulated dense schemes
```
FreddieWitherden: Move into a function and call this function; avoids polluting the global namespace.  Store the schemes as a member of the function (yes, that works).

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,12 +1,45 @@
...
+
+    for p in reversed(range(porder + 1)):
+        try:
+            pc = subclass_where(BaseDualPseudoStepper,
+                                pseudo_stepper_name=name,
+                                pseudo_stepper_porder=p)
+            break
+        except KeyError:
+            pass
+    if not pc:
```
FreddieWitherden: else

pyfr/integrators/dual/pseudo/multip.py
```diff
@@ -49,15 +48,16 @@ def __init__(self, backend, systemcls, rallocs, mesh, initsoln, cfg,
...
         cn = cfg.get(sect, 'pseudo-controller')
 
-        pc = subclass_where(BaseDualPseudoStepper, pseudo_stepper_name=pn)
         cc = subclass_where(BaseDualPseudoController,
                             pseudo_controller_name=cn)
         cc_none = subclass_where(BaseDualPseudoController,
                                  pseudo_controller_name='none')
 
+        from pyfr.integrators.dual.pseudo import get_pseudo_steppercls
         # Construct a pseudo-integrator for each level
```
FreddieWitherden: Add a newline or move the comment up one.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +341,73 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+
+def _get_coefficients(scheme):
+        a = []
+
+        scheme = re.sub(r'(?<=\))\s*,?\s*(?!$)', r'\n', scheme)
+        for l in scheme.splitlines():
+            if not l:
+                continue
+
+            a.append([float(f) for f in l.split()])
```
FreddieWitherden: a = [[float(f) for f in l.split()] for l in scheme.splitlines() if l]

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +341,73 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
     @property
     def _pseudo_stepper_order(self):
         return 4
+
+def _get_coefficients(scheme):
+        a = []
```
FreddieWitherden: Wacky indentation.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +341,73 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
     @property
     def _pseudo_stepper_order(self):
         return 4
+
```
FreddieWitherden: Add newline.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +341,73 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+def _get_coefficients(scheme):
```
FreddieWitherden: Move up (usually these utility functions go up before the classes).

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +341,73 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+
+class DualDenseRKPseudoStepper(BaseDualPseudoStepper):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        scheme = resource_string(
+            __name__, '{}/{}'.format('densepseudoschemes/schemes', self.path)
+        )
+
+        self.a, self.b = _get_coefficients(scheme.decode('utf-8'))
```
FreddieWitherden: .decode() (utf-8 is the default).

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +341,73 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+        # The last row is the b vector
+        return a[:-1], a[-1]
+
+
+class DualDenseRKPseudoStepper(BaseDualPseudoStepper):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        scheme = resource_string(
+            __name__, '{}/{}'.format('densepseudoschemes/schemes', self.path)
```
FreddieWitherden: In new code we try to write out indices explicitly ('{0}/{1}').

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +341,73 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+        # Ensure r0 references the bank containing u(n+1,m)
+        if r[0] != self._idxcurr:
+            r[0], r[1] = r[1], r[0]
+
+        # First stage
+        rhs(t, r[0], r[1])
+
+        # Pre-multiply a matrix and b vector with dtau
+        a = [[self._dtau*j for j in i] for i in self.a]
+        b = [self._dtau*i for i in self.b]
```
FreddieWitherden: These are floats so maybe ar is better than i and bi is better than i.

FreddieWitherden:
I think we are getting close.

One other thing is the name of the schemes.  I do not think we can call these schemes 'rkopt'.  At best they are optimal in the sense of some number of stages for some particular flux function and some particular correction function.  Pick a different name for them.

pyfr/integrators/dual/pseudo/__init__.py
nloppi: I am just about to push. This is the only point that I would like to keep as it is. I have come up with better names though. My logic is to have schemes in folders

tabulatedpseudosteppers/type/ 

under intergrators/dual/pseudo.

If we want to add any type of pseudosteppers we may just add a new folder and make register_tabulated_pseudo_steppers() to include them.

Similarly for the time-accurate schemes. The stepper containers would be

tabulateschemes/type/

under integrators/std/.

pyfr/integrators/dual/pseudo/pseudosteppers.py
nloppi: I missed this. Will push again soon.

pyfr/integrators/dual/pseudo/__init__.py
FreddieWitherden: I'm not sure why.  A stepper is a stepper.  Sure, some steppers have different properties to others, but that does not change the fact that they are still a stepper.  I am also unsure about putting schemes in folders; any scheme which does the same thing should be together.  (To play devil's advocate: where would the classical RK4 scheme go in this arrangement?  What 'type' is it?  What would go in the folder itself?)  Folders should really only be used where name disambiguation is required.  For instance, in the quadrature rules we have several "gauss-legendre" rules (line, quad, hex).  Thus we need to group first by folder.  But beyond that all schemes which integrate a hex to together, etc, etc.

nloppi: Ok. All schemes are now in one folder. 
How about pseudo-vermeire-s7-po0.txt as a flag?

FreddieWitherden: So the name is `pseudo-vermeire`?  That works with me if you guys are happy.  However, be sure that we do not find ourself in an awkward place if you guys develop some more schemes in the future!

For flags I think `s` is good for stages, `p` for (temporal order of accuracy), `sp` for solution order of accuracy (optional).  We may also want to define an `e` flag now for error estimation support.

nloppi: I meant pseudo as a prefix for all schemes which should not be used as a time accurate stepper. The name would be just vermeire.

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,25 +1,64 @@
...
 from pyfr.integrators.dual.pseudo.pseudocontrollers import BaseDualPseudoController
-from pyfr.integrators.dual.pseudo.pseudosteppers import BaseDualPseudoStepper
+from pyfr.integrators.dual.pseudo.pseudosteppers import BaseDualPseudoStepper, DualDenseRKPseudoStepper
 from pyfr.integrators.dual.pseudo.multip import DualMultiPIntegrator
 from pyfr.util import subclass_where
 
 
+def register_tabulated_pseudo_steppers():
+    # Create subclasses for all tabulated dense schemes
+    rpaths = resource_listdir('pyfr.integrators', 'tabulatedschemes')
```
FreddieWitherden: Just call it schemes?

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,25 +1,64 @@
...
+
+    for path in rpaths:
+        m = re.match(r'([a-zA-Z0-9\-~+]+)-s(\d+)(?:-po(\d+))?\.txt$', path)
+        name, nstages, porder = m.group(1), m.group(2), m.group(3)
+
+        attrs = dict(pseudo_stepper_name=name, nstages=int(nstages),
```
FreddieWitherden: Regular {} dict construction is probably cleaner here.

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,25 +1,64 @@
...
+                     path=path)
+
+        if porder:
+            attrs['pseudo_stepper_porder'] = int(porder)
+            name += '_po{}'.format(porder)
+
+        register_tabulated_pseudo_steppers.__setattr__(
```
FreddieWitherden: Probably worth explaining what you're doing here.  Any reason we can't use setattr here (the ordinary function)?

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,25 +1,64 @@
...
+            name, type(name, (DualDenseRKPseudoStepper,), attrs)
+        )
+
+
+def get_pseudo_steppercls(name, porder):
+    for p in reversed(range(porder + 1)):
```
FreddieWitherden: range can count backwards.

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,25 +1,64 @@
...
+            pass
+    else:
+        pc = subclass_where(BaseDualPseudoStepper, pseudo_stepper_name=name)
+
+    return pc
+
+
 def get_pseudo_integrator(backend, systemcls, rallocs, mesh,
                           initsoln, cfg, tcoeffs, dt):
+    register_tabulated_pseudo_steppers()
```
FreddieWitherden: Have the function set a flag so it does not run again if called more than once.

pyfr/integrators/dual/pseudo/__init__.py
FreddieWitherden: Also, why not set register_tabulated_pseudo_steppers._schemes = [] and then .append to the list in this loop?

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -1,11 +1,23 @@
...
+
 import numpy as np
 
 from pyfr.integrators.dual.pseudo.base import BaseDualPseudoIntegrator
 from pyfr.util import proxylist
 
 
+def _get_coefficients(scheme):
+    scheme = re.sub(r'(?<=\))\s*,?\s*(?!$)', r'\n', scheme)
+    a = [[float(f) for f in l.split()] for l in scheme.splitlines() if l]
```
FreddieWitherden: I'd group this with the return block.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +349,61 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
     @property
     def _pseudo_stepper_order(self):
         return 4
+
+
+class DualDenseRKPseudoStepper(BaseDualPseudoStepper):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
FreddieWitherden: Does anything higher up the chain expect .a and .b to exist?  If so we may want to load the scheme first.

pyfr/integrators/dual/pseudo/pseudosteppers.py
nloppi: It becomes very messy?

pyfr/integrators/dual/pseudo/pseudosteppers.py
nloppi: self.a and self.b are not used higher up. Should be fine.

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,25 +1,74 @@
...
+    for path in rpaths:
+        m = re.match(r'([a-zA-Z0-9\-~+]+)-s(\d+)-p(\d+)'
+                     r'(?:-sp(\d+))(?:-([e]+))?\.txt', path)
+        name = m.group(1)
+
+        attrs = {'pseudo_stepper_name': name, 'path': path}
+        attrs['_stepper_nfevals'] = int(m.group(2)) + 1
+        attrs['_pseudo_stepper_nregs'] = int(m.group(2))
+        attrs['_pseudo_stepper_order'] = int(m.group(3))
+        attrs['_pseudo_stepper_has_lerrest'] = True if m.group(5) else False
```
FreddieWitherden: bool(m.group(5))

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,25 +1,74 @@
...
+
+        if m.group(4):
+            attrs['pseudo_stepper_porder'] = int(m.group(4))
+            name += '_ps{}'.format(m.group(4))
+
+        register_tabulated_pseudo_steppers._schemes.append(
+            type(name, (DualDenseRKPseudoStepper,), attrs)
+        )
+    register_tabulated_pseudo_steppers._has_been_called = True
```
FreddieWitherden: No need for this.

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,25 +1,74 @@
...
 
 
+def register_tabulated_pseudo_steppers():
+    if register_tabulated_pseudo_steppers._has_been_called:
+        return
+
+    # Create subclasses for all tabulated dense schemes
+    rpaths = resource_listdir('pyfr.integrators', 'schemes')
+
+    register_tabulated_pseudo_steppers._schemes = []
```
FreddieWitherden: = schemes = [] then use schemes down below.

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,25 +1,74 @@
...
 
 from pyfr.integrators.dual.pseudo.pseudocontrollers import BaseDualPseudoController
-from pyfr.integrators.dual.pseudo.pseudosteppers import BaseDualPseudoStepper
+from pyfr.integrators.dual.pseudo.pseudosteppers import BaseDualPseudoStepper, DualDenseRKPseudoStepper
 from pyfr.integrators.dual.pseudo.multip import DualMultiPIntegrator
 from pyfr.util import subclass_where
 
 
+def register_tabulated_pseudo_steppers():
+    if register_tabulated_pseudo_steppers._has_been_called:
```
FreddieWitherden: hasattr(register_tabulated_pseudo_steppers, '_schemes'):

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +349,44 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+
+        scheme = resource_string(
+           'pyfr.integrators', '{0}/{1}'.format('schemes', self.path)
```
FreddieWitherden: format with hardcode string.

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,25 +1,72 @@
...
+
+        schemes.append(type(name, (DualDenseRKPseudoStepper,), attrs))
+
+    register_tabulated_pseudo_steppers._has_been_called = True
+
+
+def get_pseudo_steppercls(name, porder):
+    for p in range(porder, -1, -1):
+        try:
+            pc = subclass_where(BaseDualPseudoStepper,
```
FreddieWitherden: return

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,25 +1,72 @@
...
+
+        attrs = {'pseudo_stepper_name': name, 'path': path}
+        attrs['_stepper_nfevals'] = int(m.group(2)) + 1
+        attrs['_pseudo_stepper_nregs'] = int(m.group(2))
+        attrs['_pseudo_stepper_order'] = int(m.group(3))
+        attrs['_pseudo_stepper_has_lerrest'] = bool(m.group(5))
+
+        if m.group(4):
+            attrs['pseudo_stepper_porder'] = int(m.group(4))
+            name += '_ps{}'.format(m.group(4))
```
FreddieWitherden: '{0}'

pyfr/integrators/dual/pseudo/pseudosteppers.py
nloppi: What do you mean? 
resource_string('pyfr.integrators.schemes', self.path)
doesn't work unless we place __init__.py within /schemes, right? This can mess up the regex.

FreddieWitherden: '/schemes/' + self.path.  The first argument to format is hardcoded.

pyfr/integrators/dual/pseudo/__init__.py
FreddieWitherden: ^

pyfr/integrators/dual/pseudo/__init__.py
FreddieWitherden: ^

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -1,11 +1,23 @@
...
+from pkg_resources import resource_string
+import re
+
 import numpy as np
 
 from pyfr.integrators.dual.pseudo.base import BaseDualPseudoIntegrator
 from pyfr.util import proxylist
 
 
+def _get_coefficients(scheme):
```
FreddieWitherden: _get_coefficients_from_txt may be a better name?  get_coefficients seems a tad generic.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +349,42 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+        for i in range(self._pseudo_stepper_nregs - 2):
+            ar = chain(*zip(a[i], r[1:]))
+            add(0, r[i + 2], 1, r[0], *ar)
+            rhs(t, r[i + 2], r[i + 2])
+
+        # Final accumulation
+        br = chain(*zip(b[1:], r[2:]))
+        add(b[0], r[1], 1, r[0], *br)
+
+        # Return the index of the bank containing u(n+1,m+1)
```
FreddieWitherden: Update comment.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +349,42 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+
+    def step(self, t):
+        add = self._add
+        rhs = self._rhs_with_dts
+
+        r = self._pseudo_stepper_regidx
+
+        # Ensure r0 references the bank containing u(n+1,m)
+        if r[0] != self._idxcurr:
+            r[0], r[1] = r[1], r[0]
```
FreddieWitherden: Do we want to be messing with the entries of _pseudo_stepper_regidx?  In other code we just swap variables, but here I think we are actually operating on _pseudo_stepper_regidx which makes me uncomfortable.  Perhaps wrap L366 with list(...)?

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -337,3 +349,42 @@ class DualRK45PseudoStepper(DualRKVdH2RPseudoStepper):
...
+        # First stage
+        rhs(t, r[0], r[1])
+
+        # Pre-multiply a matrix and b vector with dtau
+        a = [[self._dtau*aj for aj in ar] for ar in self.a]
+        b = [self._dtau*bi for bi in self.b]
+
+        # Other stages
+        for i in range(self._pseudo_stepper_nregs - 2):
+            ar = chain(*zip(a[i], r[1:]))
```
FreddieWitherden: Will the chain fit in the next line?

pyfr/integrators/dual/pseudo/pseudosteppers.py
FreddieWitherden: I think that's fine for the moment but later we might need to move them up when we start consider error-estimating schemes.

pyfr/integrators/dual/pseudo/pseudosteppers.py
FreddieWitherden: Thinking about it, please add some validation code like we have in the BaseTabulatedQuadRule class.  Specifically, have the function take the number of stages as an argument and check the number of lines matches up with what we expect, as does the number of elements in each line.

pyfr/integrators/dual/pseudo/pseudosteppers.py
nloppi: I don't think we are modifying anything, since

    @property
    def _pseudo_stepper_regidx(self):
        return self._regidx[:self._pseudo_stepper_nregs]

r = self._pseudo_stepper_regidx
Just copies all pseudo stepper register indices from the self._regidx list that contains all register indices.

pyfr/integrators/dual/pseudo/__init__.py
nloppi: Ah, now I got it. Pushing one more time.

pyfr/integrators/dual/pseudo/pseudosteppers.py
FreddieWitherden: So that's an implementation detail in the sense that we are implicitly relying on a copy being made.  However, should _pseudo_stepper_regidx become a list in its own right or if self._regidx should be changed from a list to a NumPy array of ints this would no longer be the case.  Neither of these are things which one would expect to break code, but in this case it would.

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,25 +1,67 @@
...
 
 
+def register_tabulated_pseudo_steppers():
+    if hasattr(register_tabulated_pseudo_steppers, '_schemes'):
+        return
+
+    # Create subclasses for all tabulated dense schemes
+    rpaths = resource_listdir('pyfr.integrators', 'schemes')
+
+    register_tabulated_pseudo_steppers._schemes = schemes = []
```
FreddieWitherden: Move this up so it is grouped with the check for schemes.

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,25 +1,67 @@
...
+
+def get_pseudo_steppercls(name, porder):
+    for p in range(porder, -1, -1):
+        try:
+            return subclass_where(BaseDualPseudoStepper,
+                                  pseudo_stepper_name=name,
+                                  pseudo_stepper_porder=p)
+        except KeyError:
+            pass
+    else:
```
FreddieWitherden: Can remove else and un-indent the next line.

pyfr/integrators/dual/pseudo/multip.py
```diff
@@ -49,15 +48,17 @@ def __init__(self, backend, systemcls, rallocs, mesh, initsoln, cfg,
...
         cc_none = subclass_where(BaseDualPseudoController,
                                  pseudo_controller_name='none')
 
         # Construct a pseudo-integrator for each level
+        from pyfr.integrators.dual.pseudo import get_pseudo_steppercls
+
         self.pintgs = {}
         for l in self.levels:
+            pc = get_pseudo_steppercls(pn, l)
             if l == self._order:
```
FreddieWitherden: Add newline.

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,25 +1,67 @@
...
+        attrs['_pseudo_stepper_has_lerrest'] = bool(m.group(5))
+
+        if m.group(4):
+            attrs['pseudo_stepper_porder'] = int(m.group(4))
+            name += '_ps{0}'.format(m.group(4))
+
+        schemes.append(type(name, (DualDenseRKPseudoStepper,), attrs))
+
+
+def get_pseudo_steppercls(name, porder):
```
FreddieWitherden: _cls?

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -1,11 +1,31 @@
...
 
 from pyfr.integrators.dual.pseudo.base import BaseDualPseudoIntegrator
 from pyfr.util import proxylist
 
 
+def _get_coefficients_from_txt(scheme, nstages):
+    scheme = re.sub(r'(?<=\))\s*,?\s*(?!$)', r'\n', scheme)
+    a_mat = []
+
+    for i, l in zip(list(range(1, nstages + 1)), scheme.splitlines()):
```
FreddieWitherden: enumerate with start arg.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -1,11 +1,31 @@
...
+import re
+
 import numpy as np
 
 from pyfr.integrators.dual.pseudo.base import BaseDualPseudoIntegrator
 from pyfr.util import proxylist
 
 
+def _get_coefficients_from_txt(scheme, nstages):
+    scheme = re.sub(r'(?<=\))\s*,?\s*(?!$)', r'\n', scheme)
```
FreddieWitherden: Do we need this re.sub?

pyfr/integrators/dual/pseudo/pseudosteppers.py
FreddieWitherden: Can also create `a` the old way and then do an `if any(len(ai) != i for i, ai in enumerate(a, start=1))` or something to that effect.

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,25 +1,67 @@
...
 from pyfr.util import subclass_where
 
 
+def register_tabulated_pseudo_steppers():
+    if hasattr(register_tabulated_pseudo_steppers, '_schemes'):
+        return
+    register_tabulated_pseudo_steppers._schemes = schemes = []
+
+    # Create subclasses for all tabulated dense schemes
+    rpaths = resource_listdir('pyfr.integrators', 'schemes')
```
FreddieWitherden: Will this fit on L20?

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -1,11 +1,30 @@
 # -*- coding: utf-8 -*-
 
+from itertools import chain
+from pkg_resources import resource_string
+import re
```
FreddieWitherden: Remove.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -1,11 +1,30 @@
...
+
 import numpy as np
 
 from pyfr.integrators.dual.pseudo.base import BaseDualPseudoIntegrator
 from pyfr.util import proxylist
 
 
+def _get_coefficients_from_txt(scheme):
+    a_mat = []
```
FreddieWitherden: coeffs_mat may be better as it also ends up storing b.

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,25 +1,67 @@
...
-from pyfr.integrators.dual.pseudo.pseudosteppers import BaseDualPseudoStepper
+from pyfr.integrators.dual.pseudo.pseudosteppers import BaseDualPseudoStepper, DualDenseRKPseudoStepper
 from pyfr.integrators.dual.pseudo.multip import DualMultiPIntegrator
 from pyfr.util import subclass_where
 
 
+def register_tabulated_pseudo_steppers():
+    if hasattr(register_tabulated_pseudo_steppers, '_schemes'):
+        return
+    register_tabulated_pseudo_steppers._schemes = schemes = []
```
FreddieWitherden: Add newline.

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,25 +1,66 @@
...
+def register_tabulated_pseudo_steppers():
+    if hasattr(register_tabulated_pseudo_steppers, '_schemes'):
+        return
+
+    register_tabulated_pseudo_steppers._schemes = schemes = []
+
+    # Create subclasses for all tabulated dense schemes
+    for path in resource_listdir('pyfr.integrators', 'schemes'):
+        m = re.match(r'([a-zA-Z0-9\-~+]+)-s(\d+)-p(\d+)'
+                     r'(?:-sp(\d+))(?:-([e]+))?\.txt', path)
```
FreddieWitherden: `(?:-sp(\d+))` => `(?:-sp(\d+))?`

pyfr/integrators/dual/pseudo/__init__.py
```diff
@@ -1,25 +1,66 @@
...
+
+        attrs = {'pseudo_stepper_name': name, 'path': path}
+        attrs['_stepper_nfevals'] = int(m.group(2)) + 1
+        attrs['_pseudo_stepper_nregs'] = int(m.group(2))
+        attrs['_pseudo_stepper_order'] = int(m.group(3))
+        attrs['_pseudo_stepper_has_lerrest'] = bool(m.group(5))
+
+        if m.group(4):
+            attrs['pseudo_stepper_porder'] = int(m.group(4))
+            name += '_ps{0}'.format(m.group(4))
```
FreddieWitherden: sp for consistency?  May also be able to get away with + m.group(4) as it is a string.

PR #149, opened 2019-03-27
Fix other NumPy warnings in the VTK writer.

GiorgioGiangaspero:


FreddieWitherden:
Capitalise VTK and we should be good to go.

FreddieWitherden:
Probably also worth checking the other hstack instances too (of which there are 3/4 more which do not pass lists but generators).

FreddieWitherden:
> tested on a mesh with pyramids and it runs fine. Since with the square brackets we are doing a list comprehension, aren't we sure the results is going to be a list?
> Anyhow, do you mean like the following?
> np.hstack(list(range() for k in range()))

So the result is a list (due to the square brackets) of generator objects (the output from range) as opposed to a list of lists.  I am wondering if this causes NumPy to complain.

GiorgioGiangaspero:
> > tested on a mesh with pyramids and it runs fine. Since with the square brackets we are doing a list comprehension, aren't we sure the results is going to be a list?
> > Anyhow, do you mean like the following?
> > np.hstack(list(range() for k in range()))
> 
> So the result is a list (due to the square brackets) of generator objects (the output from range) as opposed to a list of lists. I am wondering if this causes NumPy to complain.

np.hstack(list(range(i) for i in range(3)))
np.hstack([list(range(i)) for i in range(3)])
np.hstack([range(i) for i in range(3)])

these three all produce the same result without warning. The following makes NumPy throw a warning (as the original implementation):
np.hstack(list(range(i)) for i in range(3))

FreddieWitherden:
Thanks for this.

Can we squish into a single commit?

PR #150, opened 2019-07-24
Writer fixes.

nloppi:


pyfr/integrators/base.py
```diff
@@ -4,6 +4,7 @@
 import re
 import time
 
+import itertools as it
```
FreddieWitherden: Itertools is part of Python.

pyfr/plugins/tavg.py
```diff
@@ -150,5 +151,6 @@ def __call__(self, intg):
 
                 self._writer.write(accmex, metadata, intg.tcurr)
 
-                self.tout = intg.tcurr + self.dtout
```
FreddieWitherden: Why are we moving this line?

pyfr/plugins/writer.py
```diff
@@ -19,9 +19,9 @@ def __init__(self, intg, cfgsect, suffix=None):
         self._writer = NativeWriter(intg, self.nvars, basedir, basename,
                                     prefix='soln')
 
-        # Output time step and next output time
+        # Output time step and last output time
         self.dt_out = self.cfg.getfloat(cfgsect, 'dt-out')
-        self.tout_next = intg.tcurr
+        self.tout = intg.tcurr
```
FreddieWitherden: Rename to tout_last or last_tout? (Same for tavg.)

pyfr/plugins/writer.py
```diff
@@ -32,11 +32,9 @@ def __init__(self, intg, cfgsect, suffix=None):
         # If we're not restarting then write out the initial solution
         if not intg.isrestart:
```
FreddieWitherden: Double check that behaviour here has not changed (normal run and restart for when start/restart time is/is not a multiple).

nloppi: The sanity checks seem to work and the tavg fields look right. I tried
1. Fresh start with dt-out that is multiple of dt works
2. Fresh start with dt-out that is not multiple of dt fails
3. Restart with dt-out that is multiple of dt works. (if dt is changed you need to also change dt-out)
4. Restart with dt-out that is not multiple of dt fails

Standard stepper works with any dt-out.

PR #151, opened 2019-08-05
Update documentation.

nloppi:


pv101:
Just to confirm that I have tested, and this builds the website OK.

FreddieWitherden:
Please include some notes regarding the fact that schemes can now refer to families.

doc/src/developer_guide.rst
```diff
@@ -162,8 +162,28 @@ inheritance diagram:
                          pyfr.integrators.dual.phys.steppers
     :parts: 1     
                          
-There are five types of pseudo-time `Stepper`_ available in PyFR |release|:                      
-                                           
+
+
+PseudoStepper
+-------
```
FreddieWitherden: Do we need more _?

nloppi: True

FreddieWitherden:
Fix line length and we should be good to go.

doc/src/developer_guide.rst
```diff
@@ -224,6 +219,11 @@ The families of schemes follow a naming convention
     :exclude-members: _abc_cache, _abc_negative_cache,
                       _abc_negative_cache_version, _abc_registry
 
+Note that :code:`DualDenseRKPseudoStepper` includes families of
```
pv101: I don't think we should have the :code: here. We seem to reserve that just for function names.

doc/src/developer_guide.rst
```diff
@@ -224,6 +219,11 @@ The families of schemes follow a naming convention
...
+`PseudoStepper`_ whose coefficients are read from .txt files named thus:
+
+`{scheme name}-s{stage count}-p{temporal order}-sp{optimal spatial polynomial order}.txt`
```
pv101: We could perhaps ass :code: here to format the file name spec?

nloppi: Not sure - especially if we use it only with functions? It's now in italics so it pops up from the main text.

doc/src/developer_guide.rst
nloppi: Removed.

doc/src/developer_guide.rst
pv101: OK - although not sure it is currently in italics? But fine without anyway.

pv101:
If @FreddieWitherden is happy with this then I think good to merge. I think we can then mint RC?

pv101:
And you should probably squash into a single commit.

doc/src/developer_guide.rst
nloppi: Ah ok. Not sure how it looks on the website. If I build an html using sphinx it appears in italics. Let's just remove the ` `?

nloppi:
Yes, will do if Freddie approves. Thanks.

PR #152, opened 2019-08-11
Fix writer output at t=0.

nloppi:


pyfr/plugins/writer.py
```diff
@@ -31,6 +31,7 @@ def __init__(self, intg, cfgsect, suffix=None):
 
         # If we're not restarting then write out the initial solution
         if not intg.isrestart:
+            self.tout_last = intg.tcurr - self.dt_out
```
FreddieWitherden: -= self.dt_out

PR #154, opened 2019-09-11
Migrate away from the Python abc module.

FreddieWitherden:
This never really caught any issues but came with a few annoying
quirks.


pv101:
I have tested, and the docs build fine.

PR #155, opened 2019-09-23
Improve strong scaling by enabling additional overlap

FreddieWitherden:
This PR aims to improve the strong scaling of PyFR.  Specifically, it should improve performance in instances where network bandwidth is a limiting factor.  It accomplishes this by splitting elements into two groups: those whose who have faces which are on the exterior of a partition and everything else.  The elements in the first group are then subject to priority processing which enables us to start MPI send/recv operations sooner.  This allows more time for them to complete.

### DIFF
--- a/pyfr/backends/base/types.py
+++ b/pyfr/backends/base/types.py
@@ -99,6 +99,12 @@ class MatrixBase(object):
 
         return ary
 
+    def slice(self, ra=None, rb=None, ca=None, cb=None):
+        ra, rb = ra or 0, rb or self.nrow
+        ca, cb = ca or 0, cb or self.ncol
+
+        return self.backend.matrix_slice(self, ra, rb, ca, cb)
+
 
 class Matrix(MatrixBase):
     _base_tags = {'dense'}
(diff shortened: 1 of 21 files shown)


pyfr/backends/base/types.py
```diff
@@ -192,8 +207,11 @@ def __getitem__(self, idx):
     def __getattr__(self, attr):
         return getattr(self._curr_mat, attr)
 
-    def rslice(self, p, q):
-        raise RuntimeError('Matrix banks can not be sliced')
+    def slice(self, ra=None, rb=None, ca=None, cb=None):
```
nloppi: Could do 

class MatrixBank(Sequence):
    slice = MatrixBase.slice

above to avoid  duplicating slice method

pyfr/solvers/base/system.py
```diff
@@ -68,6 +68,19 @@ def __init__(self, backend, rallocs, mesh, initsoln, nregs, cfg):
         self._bc_inters = bc_inters
         del bc_inters.elemap
 
+    def _compute_int_offets(self, rallocs, mesh):
```
nloppi: Probably mean offset?

pyfr/solvers/base/system.py
```diff
@@ -105,8 +118,12 @@ def _load_eles(self, rallocs, mesh, initsoln, nregs, nonce):
         else:
             eles.set_ics_from_cfg()
 
+        # Compute the index of first strictly interior element
+        intoffs = self._compute_int_offets(rallocs, mesh)
```
nloppi: Probably mean offset?

pyfr/backends/openmp/types.py
```diff
@@ -37,12 +37,11 @@ def hdata(self):
         return self.data
 
 
-class OpenMPMatrixRSlice(base.MatrixRSlice):
-    @lazyprop
-    def data(self):
-        return self.parent.data[self.p:self.q]
+class OpenMPMatrixSlice(base.MatrixSlice):
+    def _init_data(self, mat):
+        return mat.data[self.ra:self.rb, self.ca:self.cb]
```
nloppi: Flux AA seems to break the implementation here with attribute error
AttributeError: 'OpenMPMatrix' object has no attribute 'data'.

Traces back to base/elements.py line 183
self._scal_fpts = self._scal_fqpts.slice(0, nfpts)

Looks like self.data of the OpenMP matrix to which we are applying the slice is not initialised.

FreddieWitherden:
Thanks for the review.  I've fixed the issues you've identified.  Please do go over my revisions to the MatrixSlice class carefully as I've changed how some things work in order to fix anti-aliasing.

pyfr/backends/opencl/types.py
```diff
@@ -38,24 +44,17 @@ def _set(self, ary):
...
 
-class OpenCLMatrixRSlice(base.MatrixRSlice):
-    @lazyprop
-    def data(self):
-        return self.parent.basedata.get_sub_region(self.offset,
-                                                   self.nrow*self.pitch)
+class OpenCLMatrixSlice(_OpenCLMatrixCommon, base.MatrixSlice):
+    def _init_data(self, mat):
+        start = self.ra*self.pitch + self.ca*self.itemsize
+        end = (self.nrow - 1)*self.pitch + self.cb*self.itemsize
```
nloppi: This still needs fixing. Can lead to invalid buffer size.

nloppi:
Also spotted a "bug" that we missed in the previous PR. Apparently 
"In-place matrix multiplication is not (yet) supported. Use 'a = a @ b' instead of 'a @= b'"
We use this in shapes.py m3 so currently surf-flux AA is fails with this error.

FreddieWitherden:
@nloppi What are you thoughts regarding merging this?

nloppi:
Yes, please merge. We can open another pull request for any possible tweaks after quantifying the speed-ups.

pyfr/backends/opencl/types.py
```diff
@@ -7,7 +7,13 @@
 from pyfr.util import lazyprop
 
 
-class OpenCLMatrixBase(base.MatrixBase):
+class _OpenCLMatrixCommon(object):
+    @property
+    def _as_parameter_(self):
+        return self.data
```
FreddieWitherden: We don't seem to have .int_ptr here.  @nloppi Thoughts?  Does this cause any problems?

nloppi: Yes, it is required by CLBlast.

PR #157, opened 2020-01-13
p0 pyramid quadrature support

tarikdzanic:


FreddieWitherden:
I think for this only Legendre makes sense.  Also title commit something like "Add p0 pyramid quadrature support." for consistency.

PR #158, opened 2020-01-14
Add p0 pyramid quadrature support.

tarikdzanic:


AUTHORS
```diff
@@ -1,3 +1,4 @@
+Tarik Dzanic <tdzanic@tamu.edu>
```
FreddieWitherden: Move down a line to maintain alphabetical order. (PS You can use git amend and force-push to update your PR without having to open a new one.)

FreddieWitherden:
Is it possible to make this rule -d1 by adding a weight?

tarikdzanic:
Just for p0 or also for all GL pyramid quadratures? Since the rest of them don't have weights.

FreddieWitherden:
Ah yes, it appears as if I never got around to calculating them.  Lets leave as is for the moment.

AUTHORS
FreddieWitherden: Are you okay to tweak this?

tarikdzanic: I think I did it properly, can you confirm?

FreddieWitherden:
Looks good.  You'll need to resubmit against develop before we can merge (master is only used for brining forward releases).

PR #159, opened 2020-01-25
Add support for writing out the partitioner remapping table.

FreddieWitherden:
This PR adds an optional argument to the partition command to have it write out a .csv file containing the mapping between old element partition indices to the new element partition indices.


tarikdzanic:
Everything looks OK and works.

PR #160, opened 2020-02-08
Added flags for tagging periodic boundary faces.

tarikdzanic:


pyfr/readers/base.py
```diff
@@ -138,6 +138,13 @@ def _pair_periodic_fluid_faces(self, bpart, resid):
         pfaces = defaultdict(list)
         nodepts = self._nodepts
 
+        # Setup flags for periodic faces 
+        pfaceflags = {}
```
FreddieWitherden: Why do we need two loops here?

pyfr/readers/base.py
```diff
@@ -150,13 +157,21 @@ def _pair_periodic_fluid_faces(self, bpart, resid):
...
                 for lfn, rfn in zip(lfnodes[lfidx], rfnodes[rfidx]):
-                    lf = resid.pop(tuple(sorted(lfn)))
-                    rf = resid.pop(tuple(sorted(rfn)))
+                    lf = list(resid.pop(tuple(sorted(lfn))))
+                    rf = list(resid.pop(tuple(sorted(rfn))))
 
-                    pfaces[pftype].append([lf, rf])
+                    # Add periodic face flags
+                    if lpent in pfaceflags:
+                        lf[-1] = pfaceflags[lpent]
```
FreddieWitherden: When is this condition false?

pyfr/readers/base.py
tarikdzanic: Updated

pyfr/readers/base.py
```diff
@@ -182,8 +188,8 @@ def get_connectivity(self):
         # Pair the fluid-fluid faces
         fpairs, resid = self._pair_fluid_faces(ffaces)
 
-        # Identify periodic boundary face pairs
-        pfpairs = self._pair_periodic_fluid_faces(bpart, resid)
+        # Tag and pair periodic boundary faces
+        pfpairs = self._pair_periodic_fluid_faces(bpart, resid)
```
FreddieWitherden: Extra whitespace.

pyfr/readers/base.py
```diff
@@ -150,13 +150,19 @@ def _pair_periodic_fluid_faces(self, bpart, resid):
                 rfidx = fuzzysort(rfpts.mean(axis=1).T, range(len(rfnodes)))
 
                 for lfn, rfn in zip(lfnodes[lfidx], rfnodes[rfidx]):
-                    lf = resid.pop(tuple(sorted(lfn)))
-                    rf = resid.pop(tuple(sorted(rfn)))
+                    lf = list(resid.pop(tuple(sorted(lfn))))
```
FreddieWitherden: `resid.pop(tuple(sorted(lfn)))[:-1] + (int(k) + 1,)` allows you to work with tuples directly.

pyfr/readers/base.py
```diff
@@ -138,7 +138,7 @@ def _pair_periodic_fluid_faces(self, bpart, resid):
         pfaces = defaultdict(list)
         nodepts = self._nodepts
 
-        for lpent, rpent in self._pfacespents.values():
+        for k, [lpent, rpent] in self._pfacespents.items():
```
FreddieWitherden: Customary to use a tuple here.

FreddieWitherden:
One thing we also may want to check is if the pents regular exceed 127.

tarikdzanic:
> One thing we also may want to check is if the pents regular exceed 127.

Do you want to raise a runtime error or just print a warning and resume without flagging the boundaries exceeding 127?

FreddieWitherden:
> > One thing we also may want to check is if the pents regular exceed 127.
> 
> Do you want to raise a runtime error or just print a warning and resume without flagging the boundaries exceeding 127?

Another option would be to simply persist these as 16- or 32-bit integers.  It would only require changes in 3/4 places.

tarikdzanic:
Would that affect backwards compatibility with meshes imported beforehand? Because the i1 type seems to pop up in the partitioner and the base system.

FreddieWitherden:
> Would that affect backwards compatibility with meshes imported beforehand? Because the i1 type seems to pop up in the partitioner and the base system.

So we'd replace all of the i1's with i2's or i4's.  Everything would be upward compatible since it is valid to cast an i1 to an i2 or i4 and so old meshes would continue to function.

pyfr/readers/base.py
```diff
@@ -150,13 +150,18 @@ def _pair_periodic_fluid_faces(self, bpart, resid):
                 rfidx = fuzzysort(rfpts.mean(axis=1).T, range(len(rfnodes)))
 
                 for lfn, rfn in zip(lfnodes[lfidx], rfnodes[rfidx]):
-                    lf = resid.pop(tuple(sorted(lfn)))
-                    rf = resid.pop(tuple(sorted(rfn)))
+                    # Add periodic face flags
+                    # Left = +, right = -, periodic BC number offset plus 1 to avoid +/- 0
```
FreddieWitherden: Move the left/right bit down to just above lf and rf (but keep the +1 bit).

pyfr/readers/base.py
```diff
@@ -150,13 +150,18 @@ def _pair_periodic_fluid_faces(self, bpart, resid):
...
+                    flg = int(k) + 1
+
+                    lf = resid.pop(tuple(sorted(lfn)))[:-1] + ( flg,)
+                    rf = resid.pop(tuple(sorted(rfn)))[:-1] + (-flg,)
 
                     pfaces[pftype].append([lf, rf])
 
         return pfaces
-
+
```
FreddieWitherden: ?

FreddieWitherden:
@GiorgioGiangaspero would be good to get your feedback here.

FreddieWitherden:
It's also worth checking if the pent can ever be negative.

tarikdzanic:
> It's also worth checking if the pent can ever be negative.

Current implementation in the Gmsh and CGNS readers does not allow it:

`p = re.match(r'periodic[ _-]([a-z0-9]+)[ _-](l|r)$', name)`

PR #163, opened 2020-05-18
Update soln-plugin-writer documentation.

pv101:


doc/src/user_guide.rst
```diff
@@ -1076,6 +1076,14 @@ Parameterised with
 
     ``blocking`` | ``non-blocking``
 
+4. ``region`` --- optional specification of a region to be written,
+   defined via either diametrically opposite vertices of a bounding
+   cuboid, or the name of a domain boundary (in the latter case the
+   solution in all elements that have faces on the domain boundary
+   is written):
+
+    [(x, y, z),(x, y, z)] | *string*
```
FreddieWitherden, line 1085: ``* | ..`` where * corresponds to everything.

doc/src/user_guide.rst
```diff
@@ -1076,6 +1076,14 @@ Parameterised with
 
     ``blocking`` | ``non-blocking``
 
+4. ``region`` --- optional specification of a region to be written,
+   defined via either diametrically opposite vertices of a bounding
+   cuboid, or the name of a domain boundary (in the latter case the
+   solution in all elements that have faces on the domain boundary
+   is written):
```
FreddieWitherden, line 1083: are written?

doc/src/conf.py
```diff
@@ -78,7 +78,7 @@
 
 # General information about the project.
 project = u'PyFR'
-copyright = u'2013–2019, Imperial College London'
+copyright = u'2013–2020, Imperial College London'
```
FreddieWitherden, line 81: Check if any other copyrights need bumping (LICENSE and README).

doc/src/user_guide.rst
pv101: So just to check, you can actually put * and it will dump everything? I thought that to dump everything you just don't specify a region? Or do both work?

doc/src/user_guide.rst
pv101: 'is' refers to solution (singular) taken from a bunch of elements, so I think correct?

doc/src/user_guide.rst
FreddieWitherden: The default is ``*`` which corresponds to dump everything.

pv101:
@FreddieWitherden I have made updates and force pushed the revised commit.

PR #164, opened 2020-06-07
Add region support to the time averaging plugin.

FreddieWitherden:


pyfr/plugins/base.py
```diff
@@ -91,3 +94,111 @@ def _invoke_postaction(self, **kwargs):
...
+
+        # Region of interest
+        region = self.cfg.get(self.cfgsect, 'region', '*')
+
+        # All elements
+        if region == '*':
+            mdata = self._prepare_mdata_all(intg, nout, prefix)
+            self._add_region_data = self._add_region_data_all
+        # All elements inside a box
+        elif ',' in region:
```
tarikdzanic, line 117: Boundary names are allowed to have ',' in them so in some rare cases it could interfere with the box region tag. Possibly elif all(x in region for x in ['(', ',', ')']) ?

FreddieWitherden: I think this should be resolved now, thanks.

PR #170, opened 2020-09-23
Documentation for examples

wigging:
I changed the main README for the repository from `rst` to the more standard `md` filetype. I also added sections for Examples and Contributing.

For the examples, I added a README explaining how to run the simulation using the OpenMP backend. I also made some minor corrections to some of the `ini` files. I know the website has instructions for each example, but I find it more helpful to have the instructions within the repository along with the example files.


pv101:
Thanks for the PR! I think changing the README to .md is fine. Regarding the additional .md files and images for the examples, these replicate content that is already in the /doc directory, hence we would have two places to maintain the same content, so I'm not so keen on that. The principle of keeping instructions for running the examples in their respective directories is, however, possibly not a bad one. Might there be a way of having it there, and then including it in the sphinx doc build process? I'm open to suggestions on that one. Finally, I'm not so keen on having the instructions for a specific test case special cased to CPU only, I think it sends the wrong message in terms of PyFR being platform independent.

examples/inc_cylinder_2d/inc_cylinder_2d.ini
```diff
@@ -62,7 +62,7 @@ header = true
 [soln-plugin-writer]
 dt-out = 5.0
 basedir = .
-basename = cylinder_2d_{t:.2f}
+basename = inc_cylinder_2d_{t:.2f}
```
pv101, line 65: Well spotted! Currently the .ini is not consistent with the docs.

wigging:
@pv101 Yes, I saw the `doc` folder but I wasn't sure how it was being used. Usually the folder is named `docs` which can be used with GitHub Pages to host a website. Or with Sphinx it can be used with Read the Docs to automatically build the website from the GitHub repo. Can you clarify how you're using the `doc` directory?

Regarding the instructions I submitted for each example, I focused on the CPU because that's just what I've done so far. I'm trying to learn how to run PyFR simulations so this Pull Request is my notes from running the examples on the CPU. The next step is to run the examples on the other supported platforms and submit another Pull Request with instructions for those platforms too.

As a new PyFR user, reading as many details as possible about each example is helpful. So I'm trying to help other newcomers by adding to your example documentation as I learn the material myself. Right now the examples on your website are buried at the bottom of the User Guide page which is not very noticeable. More prominent example documentation would be beneficial to new users.

pv101:
Content of doc can be built with Sphinx. On each release we build updated docs and host on PyFR website.

Regarding the examples being buried, I agree, good point. We should probably split out into their own .rst file in doc/ and have them under their own tab on the website.

wigging:
Now that I know how you're using the `doc` folder, I will update this Pull Request to put my example notes into `rst` format for Sphinx.

wigging:
I'm closing this Pull Request based on our conversations in the forum. I'll submit a new Pull Request related to the documentation at a later date.

PR #174, opened 2020-10-04
added instaneous global integrator plugin

WillTrojak:
This allows for the integration of functions over the domain to be calculated and output to a file. Some slightly strange behaviour is observed in some cases, for example, energy in the TGV case, although enstrophy is fine. It is unclear if this is consistent or inconsistent with the solution.

### DIFF
--- a/pyfr/plugins/__init__.py
+++ b/pyfr/plugins/__init__.py
@@ -9,6 +9,7 @@ from pyfr.plugins.residual import ResidualPlugin
 from pyfr.plugins.sampler import SamplerPlugin
 from pyfr.plugins.tavg import TavgPlugin
 from pyfr.plugins.writer import WriterPlugin
+from pyfr.plugins.integrate import IntegratePlugin
 from pyfr.util import subclass_where
 
 
--- a/pyfr/plugins/__init__.py
+++ b/pyfr/plugins/__init__.py
@@ -3,13 +3,13 @@
 from pyfr.plugins.base import BasePlugin
 from pyfr.plugins.dtstats import DtStatsPlugin
 from pyfr.plugins.fluidforce import FluidForcePlugin
+from pyfr.plugins.integrate import IntegratePlugin
 from pyfr.plugins.nancheck import NaNCheckPlugin
 from pyfr.plugins.pseudostats import PseudoStatsPlugin
 from pyfr.plugins.residual import ResidualPlugin
 from pyfr.plugins.sampler import SamplerPlugin
 from pyfr.plugins.tavg import TavgPlugin
 from pyfr.plugins.writer import WriterPlugin
-from pyfr.plugins.integrate import IntegratePlugin
 from pyfr.util import subclass_where
 
 
--- a/pyfr/plugins/integrate.py
+++ b/pyfr/plugins/integrate.py
@@ -40,8 +40,8 @@ class IntegratePlugin(BasePlugin):
         
         # The root rank needs to open the output file
         if rank == root:
-            header = ['t'] + [k for k in self.cfg.items(cfgsect) if \
-                              k.startswith('int-')]
+            header = ['t'] + [k for k in self.cfg.items(cfgsect)
+                              if k.startswith('int-')]
 
             # Open
             self.outf = init_csv(self.cfg, cfgsect, ','.join(header))
@@ -90,9 +90,10 @@ class IntegratePlugin(BasePlugin):
 
         exprs = np.zeros(len(self.exprs))
         # Iterate over each element type in the simulation
-        for i, (soln, (plocs, jacs, wts)) in \
-            enumerate(zip(intg.soln, self.elminfo)):
-    
+        for i, elminfo in enumerate(zip(intg.soln, self.elminfo)):
+            soln  = elminfo[0]
+            plocs = elminfo[1][0]
+            
             # Convert from conservative to primitive variables
             psolns = self.elementscls.con_to_pri(soln.swapaxes(0, 1),
                                                  self.cfg)
@@ -121,6 +122,8 @@ class IntegratePlugin(BasePlugin):
                     for dim, grad in zip('xyz', gradpn):
                         subs[f'grad_{pname}_{dim}'] = grad
 
+            jacs = elminfo[1][1]
+            wts  = elminfo[1][2]
             for j, v in enumerate(self.exprs):
                 # Accumulate integrated evaluated expressions
                 exprs[j] += np.sum(wts[:, None]*jacs*npeval(v, subs))
@@ -140,7 +143,7 @@ class IntegratePlugin(BasePlugin):
             if rank != root:
                 comm.Reduce(iintex, None, op=get_mpi('sum'), root=root)
             else:
-                comm.Reduce(get_mpi('in_place'), iintex, op=get_mpi('sum'), \
+                comm.Reduce(get_mpi('in_place'), iintex, op=get_mpi('sum'),
                             root=root)
 
                 # Build the row
--- a/pyfr/plugins/integrate.py
+++ b/pyfr/plugins/integrate.py
@@ -1,0 +1,156 @@
+# -*- coding: utf-8 -*-
+
+import re
+
+import numpy as np
+
+from pyfr.inifile import Inifile
+from pyfr.mpiutil import get_comm_rank_root, get_mpi
+from pyfr.plugins.base import BasePlugin, init_csv
+from pyfr.quadrules import get_quadrule
+from pyfr.nputil import npeval
+
+class IntegratePlugin(BasePlugin):
+    name = 'integrate'
+    systems = ['*']
+    formulations = ['dual', 'std']
+
+    def __init__(self, intg, cfgsect, suffix=None):
+        super().__init__(intg, cfgsect, suffix)
+
+        comm, rank, root = get_comm_rank_root()
+        
+        # Underlying elements class
+        self.elementscls = intg.system.elementscls
+
+        # Expressions to integrate
+        c = self.cfg.items_as('constants', float)
+        self.exprs = [(k, self.cfg.getexpr(cfgsect, k, subs=c))
+                      for k in self.cfg.items(cfgsect)
+                      if k.startswith('int-')]
+
+        # Gradient pre-processing
+        self._init_gradients(intg)
+
+        # Save a reference to the physical solution point locations
+        self.plocs = intg.system.ele_ploc_upts
+
+        # Integration parameters
+        self.nsteps = self.cfg.getint(cfgsect, 'nsteps')
+        
+        # The root rank needs to open the output file
+        if rank == root:
+            header = ["t"] + [(k) for k in self.cfg.items(cfgsect) if k.startswith('int-')]
+
+            # Open
+            self.outf = init_csv(self.cfg, cfgsect, ','.join(header))
+            
+        self.elminfo = []
+        # Iterate over the element types
+        for ename, eles in intg.system.ele_map.items():
+            # Locations of each solution point
+            plocupts = eles.ploc_at_np('upts')
+            
+            # Jacobians
+            rjacs = 1/eles.rcpdjac_at_np('upts')
+            
+            # Weights
+            rname = self.cfg.get('solver-elements-' + ename, 'soln-pts')
+            wts = get_quadrule(ename, rname, eles.nupts).wts
+            
+            # Save
+            self.elminfo.append((plocupts, rjacs, wts))
+        
+
+    def _init_gradients(self, intg):
+        # Determine what gradients, if any, are required
+        self._gradpnames = gradpnames = set()
+        for k, ex in self.exprs:
+            gradpnames.update(re.findall(r'\bgrad_(.+?)_[xyz]\b', ex))
+
+        # If gradients are required then form the relevant operators
+        if gradpnames:
+            self._gradop, self._rcpjact = [], []
+
+            for eles in intg.system.ele_map.values():
+                self._gradop.append(eles.basis.m4)
+
+                # Get the smats at the solution points
+                smat = eles.smat_at_np('upts').transpose(2, 0, 1, 3)
+
+                # Get |J|^-1 at the solution points
+                rcpdjac = eles.rcpdjac_at_np('upts')
+
+                # Product to give J^-T at the solution points
+                self._rcpjact.append(smat*rcpdjac)
+
+    def _eval_exprs(self, intg):
+
+        # Get the primitive variable names
+        pnames = self.elementscls.privarmap[self.ndims]
+
+        exprs = np.zeros(len(self.exprs))
+        # Iterate over each element type in the simulation
+        for i, (soln, (plocs, rjacs, wts)) in enumerate(zip(intg.soln, self.elminfo)):
+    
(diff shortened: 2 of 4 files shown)


pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,156 @@
...
+
+        # Save a reference to the physical solution point locations
+        self.plocs = intg.system.ele_ploc_upts
+
+        # Integration parameters
+        self.nsteps = self.cfg.getint(cfgsect, 'nsteps')
+        
+        # The root rank needs to open the output file
+        if rank == root:
+            header = ["t"] + [(k) for k in self.cfg.items(cfgsect) if k.startswith('int-')]
```
FreddieWitherden, line 43: Single quotes for "t", no need for () around k.

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,156 @@
...
+        super().__init__(intg, cfgsect, suffix)
+
+        comm, rank, root = get_comm_rank_root()
+        
+        # Underlying elements class
+        self.elementscls = intg.system.elementscls
+
+        # Expressions to integrate
+        c = self.cfg.items_as('constants', float)
+        self.exprs = [(k, self.cfg.getexpr(cfgsect, k, subs=c))
```
FreddieWitherden, line 28: Is k needed here?

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,156 @@
...
+            self.outf = init_csv(self.cfg, cfgsect, ','.join(header))
+            
+        self.elminfo = []
+        # Iterate over the element types
+        for ename, eles in intg.system.ele_map.items():
+            # Locations of each solution point
+            plocupts = eles.ploc_at_np('upts')
+            
+            # Jacobians
+            rjacs = 1/eles.rcpdjac_at_np('upts')
```
FreddieWitherden, line 55: r may imply reciprocal, but I think these are regular Jacobian determinants?

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,156 @@
...
+
+                    for dim, grad in zip('xyz', gradpn):
+                        subs['grad_{0}_{1}'.format(pname, dim)] = grad
+
+            for j,(k,v) in enumerate(self.exprs):
+                # Accumulate integrated evaluated expressions
+                exprs[j] += np.sum(wts[:, None]*rjacs*npeval(v, subs))
+            
+            
+        # Stack up the expressions for each element type and return
```
FreddieWitherden, line 129: Are we stacking anything here?

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,156 @@
...
+        return exprs
+
+    def __call__(self, intg):
+        doiint = intg.nacptsteps % self.nsteps == 0
```
FreddieWitherden, line 133: Think we can move this into the if below?

FreddieWitherden:
For the energy issue, it might be worth seeing if the issue persists if you work with conservative variables instead of primitive variables.

pyfr/plugins/__init__.py
```diff
@@ -9,6 +9,7 @@
 from pyfr.plugins.sampler import SamplerPlugin
 from pyfr.plugins.tavg import TavgPlugin
 from pyfr.plugins.writer import WriterPlugin
+from pyfr.plugins.integrate import IntegratePlugin
```
FreddieWitherden, line 12: Alphabetical ordering.

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,156 @@
...
+import re
+
+import numpy as np
+
+from pyfr.inifile import Inifile
+from pyfr.mpiutil import get_comm_rank_root, get_mpi
+from pyfr.plugins.base import BasePlugin, init_csv
+from pyfr.quadrules import get_quadrule
+from pyfr.nputil import npeval
+
```
FreddieWitherden, line 12: Add newline.

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,156 @@
...
+import re
+
+import numpy as np
+
+from pyfr.inifile import Inifile
+from pyfr.mpiutil import get_comm_rank_root, get_mpi
+from pyfr.plugins.base import BasePlugin, init_csv
+from pyfr.quadrules import get_quadrule
+from pyfr.nputil import npeval
```
FreddieWitherden, line 11: Alphabetical ordering.

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,156 @@
...
+            # Jacobians
+            rjacs = 1/eles.rcpdjac_at_np('upts')
+            
+            # Weights
+            rname = self.cfg.get('solver-elements-' + ename, 'soln-pts')
+            wts = get_quadrule(ename, rname, eles.nupts).wts
+            
+            # Save
+            self.elminfo.append((plocupts, rjacs, wts))
+
```
FreddieWitherden, line 63: Remove newline.

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,156 @@
...
+                smat = eles.smat_at_np('upts').transpose(2, 0, 1, 3)
+
+                # Get |J|^-1 at the solution points
+                rcpdjac = eles.rcpdjac_at_np('upts')
+
+                # Product to give J^-T at the solution points
+                self._rcpjact.append(smat*rcpdjac)
+
+    def _eval_exprs(self, intg):
+
```
FreddieWitherden, line 88: Remove newline.

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,156 @@
...
+                    gradpn = np.einsum('ijkl,jkl->ikl', rcpjact, tgradpn)
+                    gradpn = gradpn.reshape(self.ndims, nupts, -1)
+
+                    for dim, grad in zip('xyz', gradpn):
+                        subs['grad_{0}_{1}'.format(pname, dim)] = grad
+
+            for j,(k,v) in enumerate(self.exprs):
+                # Accumulate integrated evaluated expressions
+                exprs[j] += np.sum(wts[:, None]*rjacs*npeval(v, subs))
+
```
FreddieWitherden, line 127: Remove newline.

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,156 @@
...
+                    # Compute the transformed gradient
+                    tgradpn = gradop @ psoln
+                    tgradpn = tgradpn.reshape(self.ndims, nupts, -1)
+
+                    # Untransform this to get the physical gradient
+                    gradpn = np.einsum('ijkl,jkl->ikl', rcpjact, tgradpn)
+                    gradpn = gradpn.reshape(self.ndims, nupts, -1)
+
+                    for dim, grad in zip('xyz', gradpn):
+                        subs['grad_{0}_{1}'.format(pname, dim)] = grad
```
FreddieWitherden, line 122: Can use new f-string syntax here.

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,156 @@
...
+            if rank != root:
+                comm.Reduce(iintex, None, op=get_mpi('sum'), root=root)
+            else:
+                comm.Reduce(get_mpi('in_place'), iintex, op=get_mpi('sum'), root=root)
+
+                # Build the row
+                row = [intg.tcurr] + iintex.tolist()
+
+                # Write
+                print(','.join(str(r) for r in row), file=self.outf)
```
FreddieWitherden, line 152: Maybe `print(intg.tcurr, *iintex.tolist(), sep=',', file=self.outf)` will do the job here?

pyfr/plugins/integrate.py
WillTrojak: This is taken directly from `tavg.py`

pyfr/plugins/integrate.py
WillTrojak: again taken directly from `tavg.py`, so it might be best to keep the syntax the same

pyfr/plugins/integrate.py
WillTrojak: agian direct copy from `tavg.py` for evaluating expressions from the ini file. Maybe a potential feature development point would be some plugin utility functions, to abstract this boiler plate stuff.

FreddieWitherden: The comment is confusing as nothing is being stacked.  In `tavg.py` the line reads:

```
# Stack up the expressions for each element type and return
return [np.dstack(exs).swapaxes(1, 2) for exs in exprs]
```
which makes sense as the expressions for each element type are kept together in an array, and thus we have a list of arrays (one for each element type).  Here we just have
```
# Stack up the expressions for each element type and return
return exprs
```
which is very confusing, as the comment mentions stacking and element types, but the code that follows is just a flat return.

pyfr/plugins/integrate.py
FreddieWitherden: `tavg.py` stores `k` as it needs it to generate the `fields` key for the `stats` file (see https://github.com/PyFR/PyFR/blob/develop/pyfr/plugins/tavg.py#L145).  However, this is not something which the integration plugin needs to bother itself with as it's not writing a `.pyfrs` type file.

pyfr/plugins/integrate.py
FreddieWitherden: Use the cleaner syntax in new code (we're gradually switching over as code is updated/added).

FreddieWitherden:
I am still working to get to the bottom of what is causing the oscillations.  In the interim are you okay to action the changes?

pyfr/plugins/integrate.py
FreddieWitherden: I think we can also pre-compute `wts/eles.rcpdjac_at_np('upts')` and just append this quotient to elminfo.

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,152 @@
+# -*- coding: utf-8 -*-
+
+import re
+
+import numpy as np
+
+from pyfr.inifile import Inifile
+from pyfr.mpiutil import get_comm_rank_root, get_mpi
```
FreddieWitherden, line 8: Can remove get_mpi?

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,152 @@
+# -*- coding: utf-8 -*-
+
+import re
+
+import numpy as np
+
+from pyfr.inifile import Inifile
```
FreddieWitherden, line 7: Can remove import?

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,152 @@
...
+
+        # Save a reference to the physical solution point locations
+        self.plocs = intg.system.ele_ploc_upts
+
+        # Integration parameters
+        self.nsteps = self.cfg.getint(cfgsect, 'nsteps')
+        
+        # The root rank needs to open the output file
+        if rank == root:
+            header = ['t'] + [k for k in self.cfg.items(cfgsect) if k.startswith('int-')]
```
FreddieWitherden, line 44: Check line length.

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,152 @@
...
+                # Product to give J^-T at the solution points
+                self._rcpjact.append(smat*rcpdjac)
+
+    def _eval_exprs(self, intg):
+        # Get the primitive variable names
+        pnames = self.elementscls.privarmap[self.ndims]
+
+        exprs = np.zeros(len(self.exprs))
+        # Iterate over each element type in the simulation
+        for i, (soln, (plocs, jacs, wts)) in enumerate(zip(intg.soln, self.elminfo)):
```
FreddieWitherden, line 93: Check line length.

pyfr/plugins/integrate.py
FreddieWitherden: Prod comment.

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,152 @@
...
+            comm, rank, root = get_comm_rank_root()
+            
+            # Evaluate the integation expressions
+            iintex = self._eval_exprs(intg)
+
+            # Reduce and output if we're the root rank
+            if rank != root:
+                comm.Reduce(iintex, None, op=get_mpi('sum'), root=root)
+            else:
+                comm.Reduce(get_mpi('in_place'), iintex, op=get_mpi('sum'), root=root)
```
FreddieWitherden, line 142: Check line length.

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,152 @@
...
+                    tgradpn = tgradpn.reshape(self.ndims, nupts, -1)
+
+                    # Untransform this to get the physical gradient
+                    gradpn = np.einsum('ijkl,jkl->ikl', rcpjact, tgradpn)
+                    gradpn = gradpn.reshape(self.ndims, nupts, -1)
+                    
+                    for dim, grad in zip('xyz', gradpn):
+                        subs[f'grad_{pname}_{dim}'] = grad
+
+            for j,v in enumerate(self.exprs):
```
FreddieWitherden, line 123: Space around j,v

FreddieWitherden:
Thanks for following up.  Few minor comments, but I think we're getting close to being able to merge.

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,154 @@
...
+
+        # Save a reference to the physical solution point locations
+        self.plocs = intg.system.ele_ploc_upts
+
+        # Integration parameters
+        self.nsteps = self.cfg.getint(cfgsect, 'nsteps')
+        
+        # The root rank needs to open the output file
+        if rank == root:
+            header = ['t'] + [k for k in self.cfg.items(cfgsect) if \
```
FreddieWitherden, line 43: No need for \ in Python if we're inside brackets.  Normally if statement is brought down too for clarity.

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,154 @@
...
+            comm, rank, root = get_comm_rank_root()
+            
+            # Evaluate the integation expressions
+            iintex = self._eval_exprs(intg)
+
+            # Reduce and output if we're the root rank
+            if rank != root:
+                comm.Reduce(iintex, None, op=get_mpi('sum'), root=root)
+            else:
+                comm.Reduce(get_mpi('in_place'), iintex, op=get_mpi('sum'), \
```
FreddieWitherden, line 143: No need for \

pyfr/plugins/integrate.py
```diff
@@ -0,0 +1,154 @@
...
+                # Product to give J^-T at the solution points
+                self._rcpjact.append(smat*rcpdjac)
+
+    def _eval_exprs(self, intg):
+        # Get the primitive variable names
+        pnames = self.elementscls.privarmap[self.ndims]
+
+        exprs = np.zeros(len(self.exprs))
+        # Iterate over each element type in the simulation
+        for i, (soln, (plocs, jacs, wts)) in \
```
FreddieWitherden, line 93: Try to avoid \ as they can be a source of subtle issues.  Here I might assign the output of enumerate to a variable first.

PR #177, opened 2020-11-29
Add attribute support to mesh and solution files.

FreddieWitherden:


pyfr/partitioners/base.py
```diff
@@ -153,23 +173,49 @@ def _renumber_verts(self, mesh, vetimap, vparts):
...
-        for eti, part in zip(vetimap, vparts):
-            if eti not in bndeti:
-                nvetimap.append(eti)
-                nvparts.append(part)
+        # Next, process the internal vertices
+        for k, v in mesh.items():
+            if k.startswith('spt'):
+                etype = k.split('_')[1]
+
+                # Start with curved elements, followed by linear elements
```
semi-h, line 190: Are curved elements in the exterior taken into account somewhere?

FreddieWitherden: The plan is to treat all exterior elements as if they were curved.  (Although as discussed offline we're probably not going to make use of internal/external in the solver, but will keep it in the file.)

PR #179, opened 2020-12-10
 Separate out linear elements from curved elements. 

FreddieWitherden:
This enables us to reduce the amount of memory bandwidth required
for a step by computing Jacobian entries on the fly from the vertex
data.

### DIFF
--- a/pyfr/backends/cuda/generator.py
+++ b/pyfr/backends/cuda/generator.py
@@ -12,9 +12,8 @@ class CUDAKernelGenerator(BaseKernelGenerator):
             self._ix = 'int _x = blockIdx.x*blockDim.x + threadIdx.x;'
             self._limits = 'if (_x < _nx)'
         else:
-            self._ix = ('int _x = blockIdx.x*blockDim.x + threadIdx.x;'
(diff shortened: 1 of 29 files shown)


pyfr/backends/cuda/generator.py
```diff
@@ -12,9 +12,8 @@ def __init__(self, *args, **kwargs):
             self._ix = 'int _x = blockIdx.x*blockDim.x + threadIdx.x;'
             self._limits = 'if (_x < _nx)'
         else:
-            self._ix = ('int _x = blockIdx.x*blockDim.x + threadIdx.x;'
-                        'int _y = blockIdx.y*blockDim.y + threadIdx.y;')
-            self._limits = 'if (_x < _nx && _y < _ny)'
+            self._ix = 'int _x = blockIdx.x*blockDim.x + threadIdx.x;'
```
semi-h, line 15: _ix can be placed outside of the if statement. Same applies for opencl and hip backends

FreddieWitherden: Fixed.

PR #181, opened 2020-12-18
Improve the OpenMP implementation of axnpby.

FreddieWitherden:


pyfr/backends/openmp/kernels/axnpby.mako
```diff
@@ -25,14 +36,8 @@ axnpby(int nrow, int ncolb, int ldim,
...
                     idx = r*ldim + X_IDX_AOSOA(${k}, ${ncola});
-                    axn = ${pyfr.dot('a{l}', 'x{l}[idx]', l=(1, nv))};
 
-                    if (a0 == 0.0)
-                        x0[idx] = axn;
-                    else if (a0 == 1.0)
-                        x0[idx] += axn;
-                    else
-                        x0[idx] = a0*x0[idx] + axn;
+                    x0[idx] = ${pyfr.dot('a{l}', 'x{l}[idx]', l=nv)};
```
WillTrojak, line 40: Will this optimise out the cases when `a{l}==0.` and `a{l}=1.`? Same question for the two other occurrences. Maybe pyfr.dot could be modified to handle this rather than relying on the compiler.

FreddieWitherden: No, but they don't do anything on CPU's due to the compiler not issuing streaming stores.  So we might sometimes do some extra FLOPs, but they're free.

PR #188, opened 2021-01-26
integrate: use corrected velocity grads with Navier-Stokes

GonzaloSaez:
Hi everyone,

It seems that when analyzing the pressure-dilatation terms  of the default TGV vortex [1] using the integrate plugin as

`int-dissipationPressure = p*(grad_u_x + grad_v_y + grad_w_z)`

, the results are very far from the expected ones (two-order of magnitudes higher), although other quantities such as kinetic energy dissipation are very close to those shown in the literature. I also observed these issues with my in-house code (which uses a similar approach to compute the gradients in the postprocessing stages). Therefore, I tried to fix the issue by relying on the corrected gradients of the velocity at solution points, instead of the gradients computed using the nodal solution basis. When using these corrected gradients, the results are much closer to those obtained in the literature. 

I suppose that this is not the cleanest implementation (only compatible with the `navier-stokes` solver) and it will harm the performance (extra copy operations) and also increase the spatial complexity of the implementation (due to the need of storing a copy of `_vect_upts`). Anyway, I'll leave this implementation in the PR in case anyone needs to postprocess this pressure dilation term with PyFR.

[1] DeBonis, J. (2013), “Solutions of the Taylor-Green vortex problem using high-resolution explicit finite
difference methods”, In 51st AIAA Aerospace Sciences Meeting including the New Horizons Forum and Aerospace Exposition, p. 382.

### DIFF
--- a/pyfr/integrators/base.py
+++ b/pyfr/integrators/base.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from collections import deque
+from collections import Iterable
 import itertools as it
 import re
 import time
@@ -133,6 +134,38 @@ class BaseIntegrator(object):
         else:
             return {'config': cfg, 'config-0': cfg}
 
+    def grad_pvars(self, pnames):
+        pnames = pnames if isinstance(pnames, Iterable) else [pnames]
+        for pname in pnames:
+            if pname not in self.system.elementscls.privarmap[self.system.ndims]:
+                raise ValueError(f'Gradient of non-primitive variable {pname} can not be computed')
+
+        try:
+            # Fetch the corrected gradients only if needed
+            grad_soln = self.grad_soln
+        except NotImplementedError:
+            # If the considered solver does not explictly compute gradients
+            # of primitive variables, then raise error
+            raise RuntimeError(f'Solver {self.system.name} not compatible with corrected gradients computation')
+
+        grads_eles = []
+        for soln, grad_soln in zip(self.soln, self.grad_soln):
+            grads_ele = []
+            # Subset and transpose the solution
+            soln = soln.swapaxes(0, 1)
+
+            # Subset and transpose the solution gradients
+            grad_soln = grad_soln.swapaxes(0, 2).swapaxes(1, 2)
+
+            # Transform from conservative to primitive gradients
+            pgrads = self.system.elementscls.grad_con_to_pri(soln, grad_soln, self.cfg)
+
+            # Store the gradients
+            for pname in pnames:
+                idx = self.system.elementscls.privarmap[self.system.ndims].index(pname)
+                grads_ele.append(pgrads[idx])
+            grads_eles.append(grads_ele)
+        return grads_eles
 
 class BaseCommon(object):
     def _init_reg_banks(self):
--- a/pyfr/integrators/base.py
+++ b/pyfr/integrators/base.py
@@ -134,12 +134,7 @@ class BaseIntegrator(object):
         else:
             return {'config': cfg, 'config-0': cfg}
 
-    def grad_pvars(self, pnames):
-        pnames = pnames if isinstance(pnames, Iterable) else [pnames]
-        for pname in pnames:
-            if pname not in self.system.elementscls.privarmap[self.system.ndims]:
-                raise ValueError(f'Gradient of non-primitive variable {pname} can not be computed')
-
+    def grad_pvars(self):
         try:
             # Fetch the corrected gradients only if needed
             grad_soln = self.grad_soln
@@ -161,9 +156,8 @@ class BaseIntegrator(object):
             pgrads = self.system.elementscls.grad_con_to_pri(soln, grad_soln, self.cfg)
 
             # Store the gradients
-            for pname in pnames:
-                idx = self.system.elementscls.privarmap[self.system.ndims].index(pname)
-                grads_ele.append(pgrads[idx])
+            for grad in pgrads:
+                grads_ele.append(grad)
             grads_eles.append(grads_ele)
         return grads_eles
 
--- a/pyfr/integrators/base.py
+++ b/pyfr/integrators/base.py
@@ -135,13 +135,8 @@ class BaseIntegrator(object):
             return {'config': cfg, 'config-0': cfg}
 
     def grad_pvars(self):
-        try:
-            # Fetch the corrected gradients only if needed
-            grad_soln = self.grad_soln
-        except NotImplementedError:
-            # If the considered solver does not explictly compute gradients
-            # of primitive variables, then raise error
-            raise RuntimeError(f'Solver {self.system.name} not compatible with corrected gradients computation')
+        # Fetch the corrected gradients
+        grad_soln = self.grad_soln
 
         grads_eles = []
         for soln, grad_soln in zip(self.soln, self.grad_soln):
@@ -149,16 +144,15 @@ class BaseIntegrator(object):
             # Subset and transpose the solution
             soln = soln.swapaxes(0, 1)
 
-            # Subset and transpose the solution gradients
-            grad_soln = grad_soln.swapaxes(0, 2).swapaxes(1, 2)
+            # Rearrange gradient data
+            grad_soln = np.rollaxis(grad_soln, 2)
 
             # Transform from conservative to primitive gradients
             pgrads = self.system.elementscls.grad_con_to_pri(soln, grad_soln, self.cfg)
 
             # Store the gradients
-            for grad in pgrads:
-                grads_ele.append(grad)
-            grads_eles.append(grads_ele)
+            grads_eles.append(pgrads)
+
         return grads_eles
 
 class BaseCommon(object):
--- a/pyfr/integrators/base.py
+++ b/pyfr/integrators/base.py
@@ -54,6 +54,9 @@ class BaseIntegrator(object):
         # Solution cache
         self._curr_soln = None
 
+        # Solution gradients cache
+        self._curr_grad_soln = None
+
         # Record the starting wall clock time
         self._wstart = time.time()
 
--- a/pyfr/integrators/base.py
+++ b/pyfr/integrators/base.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 
 from collections import deque
-from collections import Iterable
 import itertools as it
 import re
 import time
@@ -134,26 +133,6 @@ class BaseIntegrator(object):
         else:
             return {'config': cfg, 'config-0': cfg}
 
-    def grad_pvars(self):
-        # Fetch the corrected gradients
-        grad_soln = self.grad_soln
-
-        grads_eles = []
-        for soln, grad_soln in zip(self.soln, self.grad_soln):
-            grads_ele = []
-            # Subset and transpose the solution
-            soln = soln.swapaxes(0, 1)
-
-            # Rearrange gradient data
-            grad_soln = np.rollaxis(grad_soln, 2)
-
-            # Transform from conservative to primitive gradients
-            pgrads = self.system.elementscls.grad_con_to_pri(soln, grad_soln, self.cfg)
-
-            # Store the gradients
-            grads_eles.append(pgrads)
-
-        return grads_eles
 
 class BaseCommon(object):
     def _init_reg_banks(self):
--- a/pyfr/integrators/std/base.py
+++ b/pyfr/integrators/std/base.py
@@ -52,7 +52,7 @@ class BaseStdIntegrator(BaseCommon, BaseIntegrator):
         # If we do not have the solution gradients cached then compute and fetch them
         if not self._curr_grad_soln:
             self.system.compute_grads(self.tcurr, self._idxcurr)
-            self._curr_grad_soln = self.system.eles_vect_upts_inb.get()
+            self._curr_grad_soln = self.system.eles_vect_upts.get()
 
         return self._curr_grad_soln
 
--- a/pyfr/integrators/std/base.py
+++ b/pyfr/integrators/std/base.py
@@ -52,7 +52,7 @@ class BaseStdIntegrator(BaseCommon, BaseIntegrator):
         # If we do not have the solution gradients cached then compute and fetch them
         if not self._curr_grad_soln:
             self.system.compute_grads(self.tcurr, self._idxcurr)
-            self._curr_grad_soln = self.system.ele_vect_upts()
+            self._curr_grad_soln = self.system.eles_vect_upts_inb.get()
 
         return self._curr_grad_soln
 
--- a/pyfr/integrators/std/base.py
+++ b/pyfr/integrators/std/base.py
@@ -48,6 +48,15 @@ class BaseStdIntegrator(BaseCommon, BaseIntegrator):
         return self._curr_soln
 
     @property
+    def grad_soln(self):
+        # If we do not have the solution gradients cached then compute and fetch them
+        if not self._curr_grad_soln:
+            self.system.compute_grads(self.tcurr, self._idxcurr)
+            self._curr_grad_soln = self.system.ele_vect_upts()
+
+        return self._curr_grad_soln
+
+    @property
     def _controller_needs_errest(self):
         pass
 
--- a/pyfr/plugins/integrate.py
+++ b/pyfr/plugins/integrate.py
@@ -8,7 +8,6 @@ from pyfr.mpiutil import get_comm_rank_root, get_mpi
 from pyfr.nputil import npeval
 from pyfr.plugins.base import BasePlugin, init_csv
 from pyfr.quadrules import get_quadrule
-from pyfr.solvers.baseadvecdiff import BaseAdvectionDiffusionSystem
 
 
 class IntegratePlugin(BasePlugin):
@@ -103,9 +102,13 @@ class IntegratePlugin(BasePlugin):
 
     def _init_gradients(self, intg, rinfo):
         # Determine what gradients, if any, are required
-        self._gradpnames = gradpnames = set()
+        gradpnames = set()
         for ex in self.exprs:
             gradpnames.update(re.findall(r'\bgrad_(.+?)_[xyz]\b', ex))
+
+        privarmap = self.elementscls.privarmap[self.ndims]
+        self._gradpinfo = [(pname, privarmap.index(pname)) 
+                            for pname in gradpnames]
 
         # If gradients are required then form the relevant operators
         if gradpnames:
@@ -131,8 +134,8 @@ class IntegratePlugin(BasePlugin):
         pnames = self.elementscls.privarmap[self.ndims]
 
         # Compute primitive gradients if required
-        if self._gradpnames:
-            grads_eles = intg.grad_pvars()
+        if self._gradpinfo:
+            grads_eles = self._grad_pvars(intg)
 
         # Iterate over each element type in the simulation
         for i, (soln, eleinfo) in enumerate(zip(intg.soln, self.eleinfo)):
@@ -148,14 +151,10 @@ class IntegratePlugin(BasePlugin):
             subs = dict(zip(pnames, psolns))
             subs.update(zip('xyz', plocs))
 
-            # Prepare any required gradient
-            if self._gradpnames:
-                gradps = grads_eles[i]
-                for pname in self._gradpnames:
-                    idx = self.elementscls.privarmap[self.ndims].index(pname)
-                    gradpn = gradps[idx][..., eset]
-                    for dim, grad in zip('xyz', gradpn):
-                        subs[f'grad_{pname}_{dim}'] = grad
+            # Prepare any required gradients
+            for pname, idx in self._gradpinfo:
+                for dim, grad in zip('xyz', grads_eles[i][idx]):
+                    subs[f'grad_{pname}_{dim}'] = grad
 
             for j, v in enumerate(self.exprs):
                 # Evaluate the expression at each point
@@ -165,6 +164,27 @@ class IntegratePlugin(BasePlugin):
                 intvals[j] += np.sum(iex) - np.sum(iex[emask])
 
         return intvals
+
+    def _grad_pvars(self, intg):
+        grads_eles = []
+
+        # Iterate over each element type in the simulation
+        for i, (soln, eleinfo) in enumerate(zip(intg.soln, self.eleinfo)):
+            eset, emask = eleinfo[2:]
+
+            # Subset and transpose the solution
+            soln = soln[..., eset].swapaxes(0, 1)
+
+            # Rearrange and subset gradient data
+            grad_soln = np.rollaxis(intg.grad_soln[i], 2)[..., eset]
+
+            # Transform from conservative to primitive gradients
+            pgrads = self.elementscls.grad_con_to_pri(soln, grad_soln, self.cfg)
+
+            # Store the gradients
+            grads_eles.append(pgrads)
+        
+        return grads_eles
 
     def __call__(self, intg):
         if intg.nacptsteps % self.nsteps == 0:
--- a/pyfr/plugins/integrate.py
+++ b/pyfr/plugins/integrate.py
@@ -130,9 +130,10 @@ class IntegratePlugin(BasePlugin):
         # Get the primitive variable names
         pnames = self.elementscls.privarmap[self.ndims]
 
-        # Compute any required gradients
+        # Compute primitive gradients if required
         if self._gradpnames:
             grads_eles = intg.grad_pvars()
+
         # Iterate over each element type in the simulation
         for i, (soln, eleinfo) in enumerate(zip(intg.soln, self.eleinfo)):
             plocs, wts, eset, emask = eleinfo
@@ -147,7 +148,7 @@ class IntegratePlugin(BasePlugin):
             subs = dict(zip(pnames, psolns))
             subs.update(zip('xyz', plocs))
 
-            # Compute any required gradients
+            # Prepare any required gradient
             if self._gradpnames:
                 gradps = grads_eles[i]
                 for pname in self._gradpnames:
--- a/pyfr/plugins/integrate.py
+++ b/pyfr/plugins/integrate.py
@@ -66,11 +66,8 @@ class IntegratePlugin(BasePlugin):
             rname = self.cfg.get(f'solver-elements-{ename}', 'soln-pts')
             wts = get_quadrule(ename, rname, eles.nupts).wts
 
-            # Corrected gradients at upts 
-            grads = eles._vect_upts
-
             # Save
-            self.eleinfo.append((ploc, wts[:, None] / rcpdjacs, eset, emask, grads))
+            self.eleinfo.append((ploc, wts[:, None] / rcpdjacs, eset, emask))
 
     def _prepare_region_info(self, intg):
         # All elements
@@ -133,8 +130,8 @@ class IntegratePlugin(BasePlugin):
         pnames = self.elementscls.privarmap[self.ndims]
 
         # Iterate over each element type in the simulation
-        for i, (soln, eleinfo) in enumerate(zip(intg.soln, self.eleinfo)):
-            plocs, wts, eset, emask, grads = eleinfo
+        for i, (soln, grad_soln, eleinfo) in enumerate(zip(intg.soln, intg.grad_soln, self.eleinfo)):
+            plocs, wts, eset, emask = eleinfo
 
             # Subset and transpose the solution
             soln = soln[..., eset].swapaxes(0, 1)
@@ -149,10 +146,10 @@ class IntegratePlugin(BasePlugin):
             # Compute any required gradients
             if self._gradpnames:
                 # Subset and transpose the solution gradients
-                grads = grads._get()[..., eset].swapaxes(0, 2).swapaxes(1, 2)
+                grad_soln = grad_soln[..., eset].swapaxes(0, 2).swapaxes(1, 2)
 
                 # Transform from conservative to primitive gradients
-                pgrads = self.elementscls.grad_con_to_pri(soln, grads, self.cfg)
+                pgrads = self.elementscls.grad_con_to_pri(soln, grad_soln, self.cfg)
 
                 # Gradient operator and J^-T matrix
                 gradop, rcpjact = self._gradop[i], self._rcpjact[i]
@@ -196,15 +193,6 @@ class IntegratePlugin(BasePlugin):
         if intg.nacptsteps % self.nsteps == 0:
             # MPI info
             comm, rank, root = get_comm_rank_root()
-
-            # Compute corrected gradients if needed
-            if len(self._gradpnames) > 0:
-                # Choose the appropriate register to store the gradients
-                regidx = 0 if intg._idxcurr != 0 else 1
-
-                # Compute gradients using the solution corresponding to
-                # tcurr
-                self.system.compute_grads(intg.tcurr, intg._idxcurr, regidx)
 
             # Evaluate the integation expressions
             iintex = self._eval_exprs(intg)
--- a/pyfr/plugins/integrate.py
+++ b/pyfr/plugins/integrate.py
@@ -130,6 +130,9 @@ class IntegratePlugin(BasePlugin):
         # Get the primitive variable names
         pnames = self.elementscls.privarmap[self.ndims]
 
+        # Compute any required gradients
+        if self._gradpnames:
+            grads_eles = intg.grad_pvars(self._gradpnames)
         # Iterate over each element type in the simulation
         for i, (soln, eleinfo) in enumerate(zip(intg.soln, self.eleinfo)):
             plocs, wts, eset, emask = eleinfo
@@ -146,40 +149,9 @@ class IntegratePlugin(BasePlugin):
 
             # Compute any required gradients
             if self._gradpnames:
-                # Gradient operator and J^-T matrix
-                gradop, rcpjact = self._gradop[i], self._rcpjact[i]
-                nupts = gradop.shape[1]
-
-                for pname in self._gradpnames:
-                    psoln = subs[pname]
-
-                    # In solvers deriving from BaseAdvecDiff, 
-                    # rely on the LDG corrected gradients to compute
-                    # the primitive gradients, instead of
-                    # evaluating them from the solution nodal basis
-                    if pname in self.elementscls.privarmap[self.ndims] \
-                    and isinstance(intg.system, BaseAdvectionDiffusionSystem):
-                        # Fetch the corrected gradients only if needed
-                        grad_soln = intg.grad_soln[i]
-
-                        # Subset and transpose the solution gradients
-                        grad_soln = grad_soln[..., eset].swapaxes(0, 2).swapaxes(1, 2)
-
-                        # Transform from conservative to primitive gradients
-                        pgrads = self.elementscls.grad_con_to_pri(soln, grad_soln, self.cfg)
-                        idx = self.elementscls.privarmap[self.ndims].index(pname)
-                        gradpn = pgrads[idx]
-                    else:
-                        # If the gradient does not correspond to
-                        # any primitive gradient, then use uncorrected gradient
-                        # Compute the transformed gradient
-                        tgradpn = gradop @ psoln
-                        tgradpn = tgradpn.reshape(self.ndims, nupts, -1)
-
-                        # Untransform this to get the physical gradient
-                        gradpn = np.einsum('ijkl,jkl->ikl', rcpjact, tgradpn)
-                    gradpn = gradpn.reshape(self.ndims, nupts, -1)
-
+                gradps = grads_eles[i]
+                for gradpn, pname in zip(gradps, self._gradpnames):
+                    gradpn = gradpn[..., eset]
                     for dim, grad in zip('xyz', gradpn):
                         subs[f'grad_{pname}_{dim}'] = grad
 
--- a/pyfr/plugins/integrate.py
+++ b/pyfr/plugins/integrate.py
@@ -132,7 +132,7 @@ class IntegratePlugin(BasePlugin):
 
         # Compute any required gradients
         if self._gradpnames:
-            grads_eles = intg.grad_pvars(self._gradpnames)
+            grads_eles = intg.grad_pvars()
         # Iterate over each element type in the simulation
         for i, (soln, eleinfo) in enumerate(zip(intg.soln, self.eleinfo)):
             plocs, wts, eset, emask = eleinfo
@@ -150,8 +150,9 @@ class IntegratePlugin(BasePlugin):
             # Compute any required gradients
             if self._gradpnames:
                 gradps = grads_eles[i]
-                for gradpn, pname in zip(gradps, self._gradpnames):
-                    gradpn = gradpn[..., eset]
+                for pname in self._gradpnames:
+                    idx = self.elementscls.privarmap[self.ndims].index(pname)
+                    gradpn = gradps[idx][..., eset]
                     for dim, grad in zip('xyz', gradpn):
                         subs[f'grad_{pname}_{dim}'] = grad
 
--- a/pyfr/plugins/integrate.py
+++ b/pyfr/plugins/integrate.py
@@ -8,6 +8,7 @@ from pyfr.mpiutil import get_comm_rank_root, get_mpi
 from pyfr.nputil import npeval
 from pyfr.plugins.base import BasePlugin, init_csv
 from pyfr.quadrules import get_quadrule
+from pyfr.solvers.baseadvecdiff import BaseAdvectionDiffusionSystem
 
 
 class IntegratePlugin(BasePlugin):
@@ -21,10 +22,10 @@ class IntegratePlugin(BasePlugin):
         comm, rank, root = get_comm_rank_root()
 
         # Underlying system
-        self.system = intg.system
+        system = intg.system
 
         # Underlying system elements class
-        self.elementscls = self.system.elementscls
+        self.elementscls = system.elementscls
 
         # Expressions to integrate
         c = self.cfg.items_as('constants', float)
@@ -39,7 +40,7 @@ class IntegratePlugin(BasePlugin):
         self._init_gradients(intg, rinfo)
 
         # Save a reference to the physical solution point locations
-        self.plocs = self.system.ele_ploc_upts
+        self.plocs = system.ele_ploc_upts
 
         # Integration parameters
         self.nsteps = self.cfg.getint(cfgsect, 'nsteps')
@@ -54,7 +55,7 @@ class IntegratePlugin(BasePlugin):
 
         # Prepare the per element-type info list
         self.eleinfo = []
-        for (ename, eles), (eset, emask) in zip(self.system.ele_map.items(), rinfo):
+        for (ename, eles), (eset, emask) in zip(system.ele_map.items(), rinfo):
             # Locations of each solution point
             ploc = eles.ploc_at_np('upts')[..., eset]
             ploc = ploc.swapaxes(0, 1)
@@ -130,7 +131,7 @@ class IntegratePlugin(BasePlugin):
         pnames = self.elementscls.privarmap[self.ndims]
 
         # Iterate over each element type in the simulation
-        for i, (soln, grad_soln, eleinfo) in enumerate(zip(intg.soln, intg.grad_soln, self.eleinfo)):
+        for i, (soln, eleinfo) in enumerate(zip(intg.soln, self.eleinfo)):
             plocs, wts, eset, emask = eleinfo
 
             # Subset and transpose the solution
@@ -145,12 +146,6 @@ class IntegratePlugin(BasePlugin):
 
             # Compute any required gradients
             if self._gradpnames:
-                # Subset and transpose the solution gradients
-                grad_soln = grad_soln[..., eset].swapaxes(0, 2).swapaxes(1, 2)
-
-                # Transform from conservative to primitive gradients
-                pgrads = self.elementscls.grad_con_to_pri(soln, grad_soln, self.cfg)
-
                 # Gradient operator and J^-T matrix
                 gradop, rcpjact = self._gradop[i], self._rcpjact[i]
                 nupts = gradop.shape[1]
@@ -158,13 +153,21 @@ class IntegratePlugin(BasePlugin):
                 for pname in self._gradpnames:
                     psoln = subs[pname]
 
-                    # In Navier-Stokes, rely on the
-                    # LDG corrected gradients to compute
+                    # In solvers deriving from BaseAdvecDiff, 
+                    # rely on the LDG corrected gradients to compute
                     # the primitive gradients, instead of
-                    # evaluating them from the solution 
-                    # nodal basis
-                    if f'grad_{pname}' in self.elementscls.gradprivarmap[self.ndims]:
-                        idx = self.elementscls.gradprivarmap[self.ndims].index(f'grad_{pname}')
+                    # evaluating them from the solution nodal basis
+                    if pname in self.elementscls.privarmap[self.ndims] \
+                    and isinstance(intg.system, BaseAdvectionDiffusionSystem):
+                        # Fetch the corrected gradients only if needed
+                        grad_soln = intg.grad_soln[i]
+
+                        # Subset and transpose the solution gradients
+                        grad_soln = grad_soln[..., eset].swapaxes(0, 2).swapaxes(1, 2)
+
+                        # Transform from conservative to primitive gradients
+                        pgrads = self.elementscls.grad_con_to_pri(soln, grad_soln, self.cfg)
+                        idx = self.elementscls.privarmap[self.ndims].index(pname)
                         gradpn = pgrads[idx]
                     else:
                         # If the gradient does not correspond to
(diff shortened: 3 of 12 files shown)


FreddieWitherden:
So I've been thinking about this for a while now.  The final solution I came up with (although never implemented) was to add a new method alongside rhs() which simply does enough work to compute the gradients and then stops.  Then, if a plugin needs gradients, this method can be called.

The advantages here are twofold: there is no performance hit when gradients are not needed by a plugin or when the plugin is only run periodically; and there is no time lag between the gradients and the solution.  (The gradients your code picks up are typically from the final RK stage which is usually working on a very different solution to the final solution generated by the RK scheme.)

Some extra work would also be required in each of the solvers to state how to convert conserved gradients into primitive gradients, although this should not be too difficult.

GonzaloSaez:
Hi Freedie,

Thank you for your answer. I understand your concerns regarding the lag between the gradients and the conservative variables values at the end of the RK step (although for postprocessing purposes and if the time step is small enough the differences might not be very important). To better implement this PR, I suppose that one might want to add the following function

```python
class BaseAdvectionDiffusionSystem(BaseAdvectionSystem):
...
def compute_grads(self, t, uinbank, foutbank):
        runall = self.backend.runall
        q1, q2 = self._queues
        kernels = self._kernels

        self._bc_inters.prepare(t)

        self.eles_scal_upts_inb.active = uinbank
        self.eles_scal_upts_outb.active = foutbank

        if ('eles', 'copy_soln') in kernels:
            q1.enqueue(kernels['eles', 'copy_soln'])
        if ('iint', 'copy_fpts') in kernels:
            q1.enqueue(kernels['iint', 'copy_fpts'])
        q1.enqueue(kernels['iint', 'con_u'])
        q1.enqueue(kernels['bcint', 'con_u'], t=t)
        if ('eles', 'shocksensor') in kernels:
            q1.enqueue(kernels['eles', 'shocksensor'])
            q1.enqueue(kernels['mpiint', 'artvisc_fpts_pack'])
        q1.enqueue(kernels['eles', 'tgradpcoru_upts'])
        q2.enqueue(kernels['mpiint', 'scal_fpts_send'])
        q2.enqueue(kernels['mpiint', 'scal_fpts_recv'])
        q2.enqueue(kernels['mpiint', 'scal_fpts_unpack'])

        runall([q1, q2])

        q1.enqueue(kernels['mpiint', 'con_u'])
        q1.enqueue(kernels['eles', 'tgradcoru_upts'])
        q1.enqueue(kernels['eles', 'gradcoru_upts'])

        runall([q1])
````

which should be called by the considered plugin using appropriate banks. Moreover we should implement the function `grad_con_to_pri` to `BaseFluidElements` to transform from conservative to primitive gradients. I'll try to take a look to see if this is feasible without a lot of work

**Comment by FreddieWitherden** [comment]
> I understand your concerns regarding the lag between the gradients and the conservative variables values at the end of the RK step

A larger issue is that the solutions used by the individual stages of an RK scheme are typically of much lower order than the final solution generated by the scheme.  In the RK literature this is known as the 'stage order.'  The only good solution is the final solution yielded by the scheme; all of the intermediate solutions should be regarded as suspect.  Only when they are combined (right at the end) do you obtain a useful solution.  Hence the issue with taking the gradients picked up from one of these intermediate solutions.

**Comment by GonzaloSaez** [comment]
Hi @FreddieWitherden , let me know if this new implementation works for you. There might some more details to be discussed (the implementation is only compatible with the solver `navier-stokes` although it could be easily extended to `ac-navier-stokes`). I'm also not sure if the gradient computation method is compatible with other than `std` integrators (see the logic used to chose `regidx` in `integrate.__call__`)

**Comment by FreddieWitherden** [comment]
> Hi @FreddieWitherden , let me know if this new implementation works for you. There might some more details to be discussed (the implementation is only compatible with the solver `navier-stokes` although it could be easily extended to `ac-navier-stokes`). I'm also not sure if the gradient computation method is compatible with other than `std` integrators (see the logic used to chose `regidx` in `integrate.__call__`)

So I think we want a mechanism for plugins to get gradients generically like they can get the solution with `intg.soln`.  For example `intg.grad_soln` which they can then convert to primitive form by calling the relevant method.  This attribute can also adopt a similar caching mechanism to what we use with `intg.soln`.

File: pyfr/solvers/baseadvecdiff/system.py [hunk 1]
```diff
@@ -63,3 +63,38 @@ def rhs(self, t, uinbank, foutbank):
         q1.enqueue(kernels['eles', 'tdivtconf'])
         q1.enqueue(kernels['eles', 'negdivconf'], t=t)
         runall([q1])
+
+    def compute_grads(self, t, uinbank, foutbank):
```
FreddieWitherden, line 67: No need for the outbank argument.

pyfr/solvers/baseadvecdiff/system.py
```diff
@@ -63,3 +63,38 @@ def rhs(self, t, uinbank, foutbank):
...
+        runall = self.backend.runall
+        q1, q2 = self._queues
+        kernels = self._kernels
+
+        self._bc_inters.prepare(t)
+
+        self.eles_scal_upts_inb.active = uinbank
+        self.eles_scal_upts_outb.active = foutbank
```
FreddieWitherden, line 75: No need for this.

pyfr/solvers/baseadvecdiff/system.py
```diff
@@ -63,3 +63,38 @@ def rhs(self, t, uinbank, foutbank):
...
+
+        q1.enqueue(kernels['eles', 'disu'])
+        q1.enqueue(kernels['mpiint', 'scal_fpts_pack'])
+        runall([q1])
+        if ('eles', 'copy_soln') in kernels:
```
FreddieWitherden, line 80: No need for this.

pyfr/solvers/baseadvecdiff/system.py
```diff
@@ -63,3 +63,38 @@ def rhs(self, t, uinbank, foutbank):
...
+            q1.enqueue(kernels['eles', 'copy_soln'])
+        if ('iint', 'copy_fpts') in kernels:
+            q1.enqueue(kernels['iint', 'copy_fpts'])
+        q1.enqueue(kernels['iint', 'con_u'])
+        q1.enqueue(kernels['bcint', 'con_u'], t=t)
+        if ('eles', 'shocksensor') in kernels:
```
FreddieWitherden, line 86: No need for this.

pyfr/solvers/baseadvecdiff/system.py
```diff
@@ -63,3 +63,38 @@ def rhs(self, t, uinbank, foutbank):
...
+        q1.enqueue(kernels['eles', 'tgradpcoru_upts'])
+        q2.enqueue(kernels['mpiint', 'scal_fpts_send'])
+        q2.enqueue(kernels['mpiint', 'scal_fpts_recv'])
+        q2.enqueue(kernels['mpiint', 'scal_fpts_unpack'])
+
+        runall([q1, q2])
+
+        q1.enqueue(kernels['mpiint', 'con_u'])
+        q1.enqueue(kernels['eles', 'tgradcoru_upts'])
+        q1.enqueue(kernels['eles', 'gradcoru_upts'])
```
FreddieWitherden, line 98: This has changed with the most recent version of PyFR.

pyfr/solvers/navstokes/elements.py
```diff
@@ -8,6 +8,32 @@ class NavierStokesElements(BaseFluidElements, BaseAdvectionDiffusionElements):
...
+    @staticmethod
+    def grad_con_to_pri(cons, grad_cons, cfg):
+        rho = cons[0]
+        grad_rho = grad_cons[0]
+        grad_E = grad_cons[-1]
+        # Divide momentum components by rho
+        vs = [rhov/rho for rhov in cons[1:-1]]
+
+        # velocity gradients
+        grad_v = [(grad_cons[i + 1] - vel*grad_rho)/rho for i, vel in enumerate(vs)]
```
FreddieWitherden, line 24: zip rather than enumerate

pyfr/solvers/navstokes/elements.py
```diff
@@ -8,6 +8,32 @@ class NavierStokesElements(BaseFluidElements, BaseAdvectionDiffusionElements):
...
+
+        # pressure gradient
+        gamma = cfg.getfloat('constants', 'gamma')
+        vsqr = sum(v*v for v in vs)
+        grad_p = grad_E.copy()
```
FreddieWitherden, line 29: I think this can be cleaned up.

GonzaloSaez:
> > Hi @FreddieWitherden , let me know if this new implementation works for you. There might some more details to be discussed (the implementation is only compatible with the solver `navier-stokes` although it could be easily extended to `ac-navier-stokes`). I'm also not sure if the gradient computation method is compatible with other than `std` integrators (see the logic used to chose `regidx` in `integrate.__call__`)
> 
> So I think we want a mechanism for plugins to get gradients generically like they can get the solution with `intg.soln`. For example `intg.grad_soln` which they can then convert to primitive form by calling the relevant method. This attribute can also adopt a similar caching mechanism to what we use with `intg.soln`.

I have implemented the `intg.grad_soln` method to ensure that the gradients are only computed once if multiple plugins need to access the corrected gradient data. I suppose that the way in which I store `vect_upts_inb` might not be ideal, so please let me know if you think that there is another way of storing the banks of the gradients.

I also fixed all your comments (let me know if not), rebased the branch with develop and modified the `compute_grads` function to take into account linear and curved elements.

PD: the `intg.grad_soln` needs to also be implemented with other than `std` integrators, I'll do that after your review of the current implementation.

pyfr/plugins/integrate.py
```diff
@@ -21,10 +21,10 @@ def __init__(self, intg, cfgsect, suffix=None):
         comm, rank, root = get_comm_rank_root()
 
         # Underlying system
-        system = intg.system
+        self.system = intg.system
```
FreddieWitherden, line 24: Think these changes can be reverted.

pyfr/plugins/integrate.py
```diff
@@ -130,7 +130,7 @@ def _eval_exprs(self, intg):
         pnames = self.elementscls.privarmap[self.ndims]
 
         # Iterate over each element type in the simulation
-        for i, (soln, eleinfo) in enumerate(zip(intg.soln, self.eleinfo)):
+        for i, (soln, grad_soln, eleinfo) in enumerate(zip(intg.soln, intg.grad_soln, self.eleinfo)):
```
FreddieWitherden, line 133: I'd only pull the property inside of the gradpnames block; avoids computing the gradient if it is not needed.

pyfr/solvers/acnavstokes/elements.py
```diff
@@ -6,6 +6,13 @@
 
 class ACNavierStokesElements(BaseACFluidElements,
                              BaseAdvectionDiffusionElements):
+    gradconvarmap = {k : [f'grad_{var}' for var in v] for k, v in BaseACFluidElements.convarmap.items()}
```
FreddieWitherden, line 9: Unsure if we need these maps; they can always be synthesised from convarmap and privarmap.

pyfr/solvers/base/elements.py
```diff
@@ -212,6 +212,10 @@ def set_backend(self, backend, nscalupts, nonce, linoff):
         self.scal_upts_inb = inb = backend.matrix_bank(self._scal_upts)
         self.scal_upts_outb = backend.matrix_bank(self._scal_upts)
 
+        # Allocate and bank the storage required to store the solution grads
+        # which will be used by post-processing plugins
+        self.vect_upts_inb = backend.matrix_bank([self._vect_upts])
```
FreddieWitherden, line 217: No need to bank this.

pyfr/plugins/integrate.py
GonzaloSaez: True, I'll revert them

pyfr/plugins/integrate.py
GonzaloSaez: Yes you're right

FreddieWitherden:
This is shaping up nicely.

I think the next question we want to ask is if the gradient property should fall back to computing gradients the direct way (by differentiating the polynomials) if the system does not provide a proper method.  This would enable everything to work for Euler and friends (although the gradients are, of course, not as good).

pyfr/solvers/base/elements.py
GonzaloSaez: Do you then suggest to use `self.vect_upts_ib = self._vect_upts`? I'm sorry if these questions are a bit silly, I'm not fully used to the way in which data is accessed in PyFR

FreddieWitherden: Yes; this is on account of the fact that there is only one self._vect_upts array (unlike scalar arrays where due to the RK schemes there can be a bunch).

GonzaloSaez:
> This is shaping up nicely.
> 
> I think the next question we want to ask is if the gradient property should fall back to computing gradients the direct way (by differentiating the polynomials) if the system does not provide a proper method. This would enable everything to work for Euler and friends (although the gradients are, of course, not as good).

I think that this could be done by checking if `system` inherits of `BaseAdvecDiff`. In that case, we can rely on the corrected LDG gradients, else we rely on the old implementation.

pyfr/solvers/acnavstokes/elements.py
GonzaloSaez: True, I'll then check if the gradient to be computed is the gradient of a primitive variable or not

FreddieWitherden:
> > This is shaping up nicely.
> > I think the next question we want to ask is if the gradient property should fall back to computing gradients the direct way (by differentiating the polynomials) if the system does not provide a proper method. This would enable everything to work for Euler and friends (although the gradients are, of course, not as good).
> 
> I think that this could be done by checking if `system` inherits of `BaseAdvecDiff`. In that case, we can rely on the corrected LDG gradients, else we rely on the old implementation.

It might be a tiny bit more subtle as to do the untransformation some metric terms are needed.  And I believe by the time the function is called that these metric terms are no longer readily available.  (This is why things like the integrator plugin grab a reference to them when they're initialised.)  Doing this all the time might be okay, although it could also be a waste of memory.

GonzaloSaez:
> > > This is shaping up nicely.
> > > I think the next question we want to ask is if the gradient property should fall back to computing gradients the direct way (by differentiating the polynomials) if the system does not provide a proper method. This would enable everything to work for Euler and friends (although the gradients are, of course, not as good).
> > 
> > 
> > I think that this could be done by checking if `system` inherits of `BaseAdvecDiff`. In that case, we can rely on the corrected LDG gradients, else we rely on the old implementation.
> 
> It might be a tiny bit more subtle as to do the untransformation some metric terms are needed. And I believe by the time the function is called that these metric terms are no longer readily available. (This is why things like the integrator plugin grab a reference to them when they're initialised.) Doing this all the time might be okay, although it could also be a waste of memory.

Yes that's true, maybe this can be discussed in other PR? I have already introduced your suggestions in the implementation, although I still don't know how to access the data of `self._vect_upts` without a bank.

pyfr/solvers/base/system.py
```diff
@@ -42,6 +42,9 @@ def __init__(self, backend, rallocs, mesh, initsoln, nregs, cfg):
         self.ele_ndofs = [e.neles*e.nupts*e.nvars for e in eles]
         self.ele_shapes = [(e.nupts, e.nvars, e.neles) for e in eles]
 
+        # Get the gradient banks
+        self.ele_grad_banks = list(eles.vect_upts_inb)
```
FreddieWitherden, line 46: I do not think we need this, we can just call .get() on inb.

pyfr/integrators/std/base.py
```diff
@@ -47,6 +47,15 @@ def soln(self):
 
         return self._curr_soln
 
+    @property
+    def grad_soln(self):
+        # If we do not have the solution gradients cached then compute and fetch them
+        if not self._curr_grad_soln:
+            self.system.compute_grads(self.tcurr, self._idxcurr)
+            self._curr_grad_soln = self.system.ele_vect_upts()
```
FreddieWitherden, line 55: Can just fetch directly.

pyfr/solvers/base/system.py
GonzaloSaez: Done

pyfr/integrators/std/base.py
GonzaloSaez: Done

pyfr/plugins/integrate.py
```diff
@@ -152,12 +153,31 @@ def _eval_exprs(self, intg):
...
-                    tgradpn = gradop @ psoln
-                    tgradpn = tgradpn.reshape(self.ndims, nupts, -1)
-
-                    # Untransform this to get the physical gradient
-                    gradpn = np.einsum('ijkl,jkl->ikl', rcpjact, tgradpn)
+                    # In solvers deriving from BaseAdvecDiff, 
+                    # rely on the LDG corrected gradients to compute
+                    # the primitive gradients, instead of
+                    # evaluating them from the solution nodal basis
+                    if pname in self.elementscls.privarmap[self.ndims] \
```
FreddieWitherden, line 160: Ideally this code should not depend on the system but should be abstracted elsewhere.

GonzaloSaez: Abstracted in the `system` class maybe?

Should we also use these gradients in `fluidforce.py`?

FreddieWitherden: Fluid force is a bit of a special case as it needs gradients on the surfaces and only for a small subset of the elements.  So, for the moment I'd be inclined to keep fluid force the same.  The time averaging plugin, however, also uses gradients so it should be switched over as appropriate.

GonzaloSaez: I see, it seems reasonable to avoid the computation of corrected gradients for `fluidforce`. For the time averaging plugin, we could reuse the same logic as in `integrate` to use or not the  corrected primitive gradients although we need to find a sensible way of implementing this logic. Maybe we could add a function to `system` to decide whether a given gradient can be computed using the LDG method or not?

FreddieWitherden: Yeah, system or integrator seems like the right place.  This way plugins can just ask for gradients and do not need to concern themselves about where they are coming from.

pyfr/solvers/navstokes/elements.py
```diff
@@ -8,6 +8,26 @@ class NavierStokesElements(BaseFluidElements, BaseAdvectionDiffusionElements):
     # Use the density field for shock sensing
     shockvar = 'rho'
 
+    @staticmethod
+    def grad_con_to_pri(cons, grad_cons, cfg):
+        rho = cons[0]
+        grad_rho = grad_cons[0]
```
FreddieWitherden, line 14: grad_rho, *grad_uvw, grad_E = grad_cons

pyfr/solvers/navstokes/elements.py
```diff
@@ -8,6 +8,26 @@ class NavierStokesElements(BaseFluidElements, BaseAdvectionDiffusionElements):
...
+        # Divide momentum components by rho
+        vs = [rhov/rho for rhov in cons[1:-1]]
+
+        # Velocity gradients
+        grad_v = [(grad_rhov - v*grad_rho)/rho for grad_rhov, v in zip(grad_cons[1:-1], vs)]
+
+        # Pressure gradient
+        gamma = cfg.getfloat('constants', 'gamma')
+        grad_p = grad_E - 0.5*sum(v*v for v in vs)*grad_rho
+        for vel, grad_vel in zip(vs, grad_v):
```
FreddieWitherden, line 25: Can avoid multiply through by rho by using conservative grads

pyfr/solvers/base/system.py
```diff
@@ -186,6 +187,9 @@ def _gen_kernels(self, eles, iint, mpiint, bcint):
     def rhs(self, t, uinbank, foutbank):
         pass
 
+    def compute_grads(self, t, uinbank):
+        pass
```
FreddieWitherden, line 191: Maybe have this raise a not implemented error and have the integrator catch this exception to decide when to fallback.

pyfr/plugins/integrate.py
GonzaloSaez: I've come up with this implementation to compute the gradients in `intg`. `compute_grads` raises a `NotImplementedError` if the `system` does not explictly compute the gradients. In that case, the gradients are computed without LDG correction. However, too handle both cases we need to add a lot of parameters to the function (metric terms, gradient operator, element type index, etc.). I suppose that that's a bit ugly, do you have a better idea?

```python
def grad_pvar(self, pname, gradop, rcpjact, ele_index, eset):
        if pname not in self.system.elementscls.privarmap[self.system.ndims]:
            raise ValueError(f'Gradient of non-primitive variable {pname} can not be computed')
        # Subset and transpose the solution
        soln = self.soln[ele_index]
        soln = soln[..., eset].swapaxes(0, 1)
        try:
            # Fetch the corrected gradients only if needed
            grad_soln = self.grad_soln[ele_index]

            # Subset and transpose the solution gradients
            grad_soln = grad_soln[..., eset].swapaxes(0, 2).swapaxes(1, 2)

            # Transform from conservative to primitive gradients
            pgrads = self.system.elementscls.grad_con_to_pri(soln, grad_soln, self.cfg)
            idx = self.system.elementscls.privarmap[self.system.ndims].index(pname)
            return pgrads[idx]
        except NotImplementedError:
            # If the considered solver does not explictly compute gradients
            # of primitive variables, then compute the gradient from
            # the solution nodal basis

            # Number of solution points
            nupts = gradop.shape[1]

            # Convert from conservative to primitive variables
            psolns = self.system.elementscls.con_to_pri(soln, self.cfg)

            psoln = psolns[psolns.index(pname)]
            # Compute the transformed gradient
            tgradpn = gradop @ psoln
            tgradpn = tgradpn.reshape(self.ndims, nupts, -1)

            # Untransform this to get the physical gradient
            gradpn = np.einsum('ijkl,jkl->ikl', rcpjact, tgradpn)
            gradpn = gradpn.reshape(self.ndims, nupts, -1)
            return gradpn
```

FreddieWitherden: I agree this is quite a bit of work.  What are your thoughts on only allowing gradients for advection-diffusion derived systems?  This would greatly simplify the code and enable us to always provide `good' gradients.  I do not think many people make use of gradients in Euler simulations.

GonzaloSaez: I suppose that gradients might be important in Euler to tweak shock sensors, but I think that those values are used in a post-processing stage. You know much better this code than me so I can not really tell you if it is worth the effort. I agree that the implementation is a bit ugly but all plugins have direct access to those parameters so I really don't know, I'll let you decide!

FreddieWitherden: My inclination is to drop support.  If there are really important use cases we can consider adding it back in later.

GonzaloSaez: Let me know if the last commit works for you. The function `intg.grad_pvars` computes all the wanted gradients at once. I also erased the previous implementation from `integrate`. Maybe we should the same for `tavg`.

pyfr/integrators/base.py
```diff
@@ -181,3 +185,33 @@ def _add(self, *args):
 
         # Bind and run the axnpby kernels
         self._queue.enqueue_and_run(axnpby, *args[::2])
+
+    def grad_pvars(self, pnames, ele_index, eset):
```
FreddieWitherden, line 189: I would just have it return everything (no pnames/ele_index, or eset arg).

GonzaloSaez: Done,  I had move the function to `BaseIntegrator` to ensure compatibility with other than `std` integrators

pyfr/integrators/base.py
```diff
@@ -130,6 +134,38 @@ def cfgmeta(self):
         else:
             return {'config': cfg, 'config-0': cfg}
 
+    def grad_pvars(self, pnames):
```
FreddieWitherden, line 137: I'd also remove the pnames bit, too.  Let each plugin handle messing with the data as it sees fit.

GonzaloSaez: Done

pyfr/integrators/base.py
```diff
@@ -130,6 +134,32 @@ def cfgmeta(self):
         else:
             return {'config': cfg, 'config-0': cfg}
 
+    def grad_pvars(self):
+        try:
+            # Fetch the corrected gradients only if needed
+            grad_soln = self.grad_soln
+        except NotImplementedError:
```
FreddieWitherden, line 141: No need to try/catch; instead we can stick a better message in when we raise the NotImplementedError.

pyfr/integrators/base.py
```diff
@@ -130,6 +134,32 @@ def cfgmeta(self):
...
+            raise RuntimeError(f'Solver {self.system.name} not compatible with corrected gradients computation')
+
+        grads_eles = []
+        for soln, grad_soln in zip(self.soln, self.grad_soln):
+            grads_ele = []
+            # Subset and transpose the solution
+            soln = soln.swapaxes(0, 1)
+
+            # Subset and transpose the solution gradients
+            grad_soln = grad_soln.swapaxes(0, 2).swapaxes(1, 2)
```
FreddieWitherden, line 153: I think this might be equivalent to a rollaxis?

pyfr/integrators/base.py
```diff
@@ -130,6 +134,32 @@ def cfgmeta(self):
...
+
+            # Transform from conservative to primitive gradients
+            pgrads = self.system.elementscls.grad_con_to_pri(soln, grad_soln, self.cfg)
+
+            # Store the gradients
+            for grad in pgrads:
```
FreddieWitherden, line 159: .extend()

pyfr/solvers/base/elements.py
```diff
@@ -212,6 +212,10 @@ def set_backend(self, backend, nscalupts, nonce, linoff):
         self.scal_upts_inb = inb = backend.matrix_bank(self._scal_upts)
         self.scal_upts_outb = backend.matrix_bank(self._scal_upts)
 
+        # Allocate and bank the storage required to store the solution grads
+        # which will be used by post-processing plugins
+        self.vect_upts_inb = self._vect_upts
```
FreddieWitherden, line 217: Don't think we want to call this an `inb` (which is short for input bank, but it is not a bank per se and isn't used as an input).

GonzaloSaez: So maybe access directly `self._vect_upts` and rename `_vect_upts` to `vect_upts`?

pyfr/integrators/base.py
GonzaloSaez: Do you prefer raising an error in the NotImplementedError then?

pyfr/integrators/base.py
GonzaloSaez: I'll have a look at it, I'm not very familiar with `numpy` tbh

pyfr/integrators/base.py
FreddieWitherden: Yes, I think that is the most logical place.

pyfr/integrators/base.py
FreddieWitherden:
```
In [1]: import numpy as np

In [2]: x = np.random.randn(10, 20, 30)

In [3]: np.allclose(x.swapaxes(0, 2).swapaxes(1, 2), np.rollaxis(x, 2))
Out[3]: True
```

However, this code should probably be moved out to the plugins themselves (integrate and tavg).

pyfr/solvers/base/system.py
```diff
@@ -186,6 +187,9 @@ def _gen_kernels(self, eles, iint, mpiint, bcint):
     def rhs(self, t, uinbank, foutbank):
         pass
 
+    def compute_grads(self, t, uinbank):
+        raise NotImplementedError()
```
FreddieWitherden, line 191: We can improve the error message here to indicate the system in question.

GonzaloSaez: Done

pyfr/integrators/base.py
GonzaloSaez: Added the `rollaxis` call and also modified `tavg` to use these new gradients

pyfr/solvers/base/elements.py
GonzaloSaez: Removed this `_vect_upts_inb`, the implementation now accesses `_vect_upts` directly

pyfr/plugins/integrate.py
```diff
@@ -8,6 +8,7 @@
 from pyfr.nputil import npeval
 from pyfr.plugins.base import BasePlugin, init_csv
 from pyfr.quadrules import get_quadrule
+from pyfr.solvers.baseadvecdiff import BaseAdvectionDiffusionSystem
```
FreddieWitherden, line 11: Not sure if we need this import.

pyfr/solvers/acnavstokes/elements.py
```diff
@@ -6,6 +6,10 @@
 
 class ACNavierStokesElements(BaseACFluidElements,
                              BaseAdvectionDiffusionElements):
+    @staticmethod
+    def grad_con_to_pri(cons, grad_cons, cfg):
+        return [grad for grad in grad_cons]
```
FreddieWitherden, line 11: Can we just return grad_cons or list(grad_cons) here?

pyfr/integrators/base.py
```diff
@@ -130,6 +134,26 @@ def cfgmeta(self):
         else:
             return {'config': cfg, 'config-0': cfg}
 
+    def grad_pvars(self):
```
FreddieWitherden, line 137: I'd move this functionality down into each plugin.  This way that can fetch the gradients, subset them, and then only convert the subsetted parts to primitive.

pyfr/integrators/base.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from collections import deque
+from collections import Iterable
```
FreddieWitherden, line 4: Not sure if we need this import.

pyfr/solvers/acnavstokes/elements.py
GonzaloSaez: Done

pyfr/plugins/integrate.py
GonzaloSaez: Removed

pyfr/integrators/base.py
GonzaloSaez: Done for `integrate`, let me know if its ok for you and I'll move it also to `tavg`

pyfr/integrators/base.py
GonzaloSaez: Removed

pyfr/integrators/base.py
```diff
@@ -130,7 +133,6 @@ def cfgmeta(self):
         else:
             return {'config': cfg, 'config-0': cfg}
```
FreddieWitherden, line 135: Think this newline should be there.

pyfr/plugins/integrate.py
```diff
@@ -172,6 +165,29 @@ def _eval_exprs(self, intg):
...
         return intvals
 
+    def _grad_pvars(self, intg):
+        # Fetch the corrected gradients
+        grad_soln = intg.grad_soln
+
+        grads_eles = []
+        # Iterate over each element type in the simulation
+        for i, (soln, eleinfo) in enumerate(zip(intg.soln, self.eleinfo)):
+            plocs, wts, eset, emask = eleinfo
```
FreddieWitherden, line 175: eset, emask = eleinfo[2:] as I don't think we need the plocs of wts here.

pyfr/plugins/tavg.py
```diff
@@ -118,23 +122,12 @@ def _eval_acc_exprs(self, intg):
...
-                    # Compute the transformed gradient
-                    tgradpn = gradop @ psoln
-                    tgradpn = tgradpn.reshape(self.ndims, nupts, -1)
-
-                    # Untransform this to get the physical gradient
-                    gradpn = np.einsum('ijkl,jkl->ikl', rcpjact, tgradpn)
-                    gradpn = gradpn.reshape(self.ndims, nupts, -1)
-
+                    idx = self.elementscls.privarmap[self.ndims].index(pname)
+                    gradpn = gradps[idx][..., rgn]
```
FreddieWitherden, line 130: I think we've already subsetted the gradients here?

pyfr/plugins/integrate.py
```diff
@@ -172,6 +165,29 @@ def _eval_exprs(self, intg):
 
         return intvals
 
+    def _grad_pvars(self, intg):
+        # Fetch the corrected gradients
+        grad_soln = intg.grad_soln
+
+        grads_eles = []
+        # Iterate over each element type in the simulation
+        for i, (soln, eleinfo) in enumerate(zip(intg.soln, self.eleinfo)):
```
FreddieWitherden, line 174: It's also worth noting that the integrate and time averaging plugins use different region/subsetting logic and so I am unsure how much code they can share.  (Integrate can sometimes select specific points from specific elements, whereas tavg will also work in terms of full elements.)

pyfr/integrators/base.py
GonzaloSaez: Fixed

pyfr/plugins/integrate.py
GonzaloSaez: Fixed

pyfr/plugins/tavg.py
GonzaloSaez: I had to update the implementation in `tavg`. Now it is more consistent with `integrate`, let me know if you think it is OK

pyfr/plugins/integrate.py
GonzaloSaez: Yes, the last version includes `_grad_pvars` with the logic of `tavg`

FreddieWitherden: Okay, would you be able to test if everything still works as expected for regions which 'cut' through an element.  (Here PyFR should just include the quadrature points which are inside the region for boundary elements; which is a reasonable-ish assumption.)

pyfr/plugins/tavg.py
```diff
@@ -118,23 +122,12 @@ def _eval_acc_exprs(self, intg):
...
-
-                    # Compute the transformed gradient
-                    tgradpn = gradop @ psoln
-                    tgradpn = tgradpn.reshape(self.ndims, nupts, -1)
-
-                    # Untransform this to get the physical gradient
-                    gradpn = np.einsum('ijkl,jkl->ikl', rcpjact, tgradpn)
-                    gradpn = gradpn.reshape(self.ndims, nupts, -1)
-
+                    idx = self.elementscls.privarmap[self.ndims].index(pname)
```
FreddieWitherden, line 129: We might want to pre-compute these indices and store them in self._gradpnames (suitably renamed, of course).  This should enable us to simplify these two loops.

GonzaloSaez: `self_gradpidxs`?

pyfr/plugins/integrate.py
GonzaloSaez: There was a bug when meshes had more than one type of element due to the fact that `grad_soln` was being reused when calling `np.rollaxis`. This has been solved in the last commit.

I have tried the inc-cylinder-2d test case and it seems that `tavg` works okeysh even when using an arbitrary region box, I don't know if you have a better validation testcase for the `tavg` plugin. I'll leave attached a picture of the mean `u` velocity in the cylinder test case (the averaging starts from `t=0` so this result is not perfect but it should validate a bit the implementation)

image

GonzaloSaez: I'll try tomorrow to output the mean velocity gradients

FreddieWitherden: The other bug I watch out for is when a region only has a single type of element (but the partition itself has two or more element types).

pyfr/plugins/tavg.py
FreddieWitherden: Something like that, where the variable itself is a list of tuples (name, idx).

GonzaloSaez: Done

pyfr/plugins/integrate.py
```diff
@@ -106,6 +106,10 @@ def _init_gradients(self, intg, rinfo):
         for ex in self.exprs:
             gradpnames.update(re.findall(r'\bgrad_(.+?)_[xyz]\b', ex))
 
+        privarmap = self.elementscls.privarmap[self.ndims]
+        self._gradpinfo = [(pname, privarmap.index(pname)) 
+                            for pname in self._gradpnames]
```
FreddieWitherden, line 111: Can use gradpnames directly here (and we no longer need to store self._gradpnames, since all of the relevant info is now in self._gradpinfo).

pyfr/plugins/integrate.py
```diff
@@ -143,23 +151,11 @@ def _eval_exprs(self, intg):
...
-                    tgradpn = gradop @ psoln
-                    tgradpn = tgradpn.reshape(self.ndims, nupts, -1)
-
-                    # Untransform this to get the physical gradient
-                    gradpn = np.einsum('ijkl,jkl->ikl', rcpjact, tgradpn)
-                    gradpn = gradpn.reshape(self.ndims, nupts, -1)
-
+                gradps = grads_eles[i]
+                for pname, idx in self._gradpinfo:
+                    gradpn = gradps[idx]
```
FreddieWitherden, line 158:
```python
                for pname, idx in self._gradpinfo:
                    for dim, grad in zip('xyz', grads_eles[i][idx]):
                        subs[f'grad_{pname}_{dim}'] = grad
```

pyfr/plugins/integrate.py
```diff
@@ -143,23 +151,11 @@ def _eval_exprs(self, intg):
             subs = dict(zip(pnames, psolns))
             subs.update(zip('xyz', plocs))
 
-            # Compute any required gradients
+            # Prepare any required gradient
             if self._gradpnames:
```
FreddieWitherden, line 155: I think we can remove this check.

pyfr/plugins/tavg.py
```diff
@@ -118,23 +126,11 @@ def _eval_acc_exprs(self, intg):
...
-
-                    # Compute the transformed gradient
-                    tgradpn = gradop @ psoln
-                    tgradpn = tgradpn.reshape(self.ndims, nupts, -1)
-
-                    # Untransform this to get the physical gradient
-                    gradpn = np.einsum('ijkl,jkl->ikl', rcpjact, tgradpn)
-                    gradpn = gradpn.reshape(self.ndims, nupts, -1)
-
+                gradps = grads_eles[j]
```
FreddieWitherden, line 131: See proposed simplification for the integrate plugin.

pyfr/plugins/integrate.py
GonzaloSaez: I think that this is needed in order to compute all gradients before looping over the regions and elements. This could maybe be avoided if we passed the region information to `_grad_pvars`

pyfr/plugins/tavg.py
GonzaloSaez: Done

pyfr/plugins/integrate.py
GonzaloSaez: Done

pyfr/plugins/integrate.py
GonzaloSaez: Done

pyfr/plugins/integrate.py
```diff
@@ -143,24 +151,10 @@ def _eval_exprs(self, intg):
...
-                    tgradpn = gradop @ psoln
-                    tgradpn = tgradpn.reshape(self.ndims, nupts, -1)
-
-                    # Untransform this to get the physical gradient
-                    gradpn = np.einsum('ijkl,jkl->ikl', rcpjact, tgradpn)
-                    gradpn = gradpn.reshape(self.ndims, nupts, -1)
-
-                    for dim, grad in zip('xyz', gradpn):
+            # Prepare any required gradient
+            if self._gradpinfo:
```
FreddieWitherden, line 155: This if statement can go.

pyfr/plugins/tavg.py
```diff
@@ -118,24 +126,10 @@ def _eval_acc_exprs(self, intg):
...
-                    # Compute the transformed gradient
-                    tgradpn = gradop @ psoln
-                    tgradpn = tgradpn.reshape(self.ndims, nupts, -1)
-
-                    # Untransform this to get the physical gradient
-                    gradpn = np.einsum('ijkl,jkl->ikl', rcpjact, tgradpn)
-                    gradpn = gradpn.reshape(self.ndims, nupts, -1)
-
-                    for dim, grad in zip('xyz', gradpn):
+            # Prepare any required gradient
```
FreddieWitherden, line 129: gradients

GonzaloSaez: Done

pyfr/plugins/integrate.py
GonzaloSaez: Done

FreddieWitherden:
I think we're pretty much there; last bit to tweak is the grad_con_to_pri function for Navier-Stokes.

GonzaloSaez:
> I think we're pretty much there; last bit to tweak is the grad_con_to_pri function for Navier-Stokes.

So if i understood correctly you want to use the following expression?

FreddieWitherden:
> > I think we're pretty much there; last bit to tweak is the grad_con_to_pri function for Navier-Stokes.
> 
> So if i understood correctly you want to use the following expression?
> 
>  %2B \rho\vec{u} \cdot \nabla \vec{u} \right))

I think we can simply the implementation of that expression by taking advantage of various NumPy tricks.

GonzaloSaez:
> > > I think we're pretty much there; last bit to tweak is the grad_con_to_pri function for Navier-Stokes.
> > 
> > 
> > So if i understood correctly you want to use the following expression?
> >  p = \nabla E - \frac{1}{2}\left(\vec{u} \cdot \nabla (\rho \vec{u} ) %2B \rho\vec{u} \cdot \nabla \vec{u} \right))
> 
> I think we can simply the implementation of that expression by taking advantage of various NumPy tricks.

Yes sure, I'll work a bit with `np.einsum` to implement it

pyfr/solvers/navstokes/elements.py
FreddieWitherden: Just following up here.

pyfr/solvers/navstokes/elements.py
FreddieWitherden: I think np.dot here with the right set of arguments could be used here (so it performs a bunch of smaller dot products for us).

GonzaloSaez:
I have implemented the computation of `grad_p` using np.einsum, let me know if that works for you. The previous method computation was buggy. The new one should be OK

pyfr/solvers/navstokes/elements.py
```diff
@@ -8,6 +10,28 @@ class NavierStokesElements(BaseFluidElements, BaseAdvectionDiffusionElements):
     # Use the density field for shock sensing
     shockvar = 'rho'
 
+    @staticmethod
+    def grad_con_to_pri(cons, grad_cons, cfg):
+        rho = cons[0]
```
FreddieWitherden, line 15: rho, *rhouvw = cons[:-1]

pyfr/solvers/navstokes/elements.py
```diff
@@ -8,6 +10,28 @@ class NavierStokesElements(BaseFluidElements, BaseAdvectionDiffusionElements):
...
+        vs = [rhov/rho for rhov in cons[1:-1]]
+
+        # Velocity gradients
+        # ∇(\vec{u}) = 1/\rho*[∇(\rho\vec{u}) - \vec{u} \otimes ∇(\rho)]
+        grad_v = [(grad_rhov - v*grad_rho)/rho for grad_rhov, v in zip(grad_cons[1:-1], vs)]
+
+        # Pressure gradient
+        # ∇p = (gamma - 1)*[∇E - 1/2*(\vec{u}·∇(\rho\vec{u}) - \rho\vec{u}·∇(\vec{u}))]
+        gamma = cfg.getfloat('constants', 'gamma')
+        grad_p = grad_E - 0.5*(np.einsum('ijk, iljk -> ljk', vs, grad_rhouvw) \
```
FreddieWitherden, line 29: No need for \, see if the + will fit on the line above.

pyfr/solvers/navstokes/elements.py
```diff
@@ -8,6 +10,28 @@ class NavierStokesElements(BaseFluidElements, BaseAdvectionDiffusionElements):
...
+        
+        # Divide momentum components by rho
+        vs = [rhov/rho for rhov in cons[1:-1]]
+
+        # Velocity gradients
+        # ∇(\vec{u}) = 1/\rho*[∇(\rho\vec{u}) - \vec{u} \otimes ∇(\rho)]
+        grad_v = [(grad_rhov - v*grad_rho)/rho for grad_rhov, v in zip(grad_cons[1:-1], vs)]
+
+        # Pressure gradient
+        # ∇p = (gamma - 1)*[∇E - 1/2*(\vec{u}·∇(\rho\vec{u}) - \rho\vec{u}·∇(\vec{u}))]
```
FreddieWitherden, line 27: Can use actual rho.

pyfr/solvers/navstokes/elements.py
```diff
@@ -8,6 +10,28 @@ class NavierStokesElements(BaseFluidElements, BaseAdvectionDiffusionElements):
...
+    def grad_con_to_pri(cons, grad_cons, cfg):
+        rho = cons[0]
+        rhouvw = cons[1:-1]
+        grad_rho, *grad_rhouvw, grad_E = grad_cons
+        
+        # Divide momentum components by rho
+        vs = [rhov/rho for rhov in cons[1:-1]]
+
+        # Velocity gradients
+        # ∇(\vec{u}) = 1/\rho*[∇(\rho\vec{u}) - \vec{u} \otimes ∇(\rho)]
```
FreddieWitherden, line 23: Can use actual ρ and ·

pyfr/solvers/navstokes/elements.py
```diff
@@ -8,6 +10,28 @@ class NavierStokesElements(BaseFluidElements, BaseAdvectionDiffusionElements):
...
     shockvar = 'rho'
 
+    @staticmethod
+    def grad_con_to_pri(cons, grad_cons, cfg):
+        rho = cons[0]
+        rhouvw = cons[1:-1]
+        grad_rho, *grad_rhouvw, grad_E = grad_cons
+        
+        # Divide momentum components by rho
+        vs = [rhov/rho for rhov in cons[1:-1]]
```
FreddieWitherden, line 20: Call uvw for consistency?  Also can use rhouvw.

pyfr/solvers/navstokes/elements.py
```diff
@@ -8,6 +10,28 @@ class NavierStokesElements(BaseFluidElements, BaseAdvectionDiffusionElements):
...
+
+        # Velocity gradients
+        # ∇(\vec{u}) = 1/\rho*[∇(\rho\vec{u}) - \vec{u} \otimes ∇(\rho)]
+        grad_v = [(grad_rhov - v*grad_rho)/rho for grad_rhov, v in zip(grad_cons[1:-1], vs)]
```
FreddieWitherden, line 24: Use grad_rhouvw?

pyfr/solvers/navstokes/elements.py
GonzaloSaez: Not sure if I understood it correctly, let me know if the new version works for you

pyfr/solvers/navstokes/elements.py
```diff
@@ -8,6 +10,27 @@ class NavierStokesElements(BaseFluidElements, BaseAdvectionDiffusionElements):
...
+    def grad_con_to_pri(cons, grad_cons, cfg):
+        rho, *rhouvw = cons[:-1]
+        grad_rho, *grad_rhouvw, grad_E = grad_cons
+        
+        # Divide momentum components by ρ
+        uvw = [rhov/rho for rhov in rhouvw]
+
+        # Velocity gradients
+        # ∇(\vec{u}) = 1/ρ·[∇(ρ\vec{u}) - \vec{u} \otimes ∇ρ]
+        grad_uvw = [(grad_rhov - v*grad_rho)/rho for grad_rhov, v in zip(grad_rhouvw, uvw)]
```
FreddieWitherden, line 23: Line is a tad too long

pyfr/solvers/navstokes/elements.py
```diff
@@ -8,6 +10,27 @@ class NavierStokesElements(BaseFluidElements, BaseAdvectionDiffusionElements):
...
+
+        # Pressure gradient
+        # ∇p = (gamma - 1)·[∇E - 1/2*(\vec{u}·∇(ρ\vec{u}) - ρ\vec{u}·∇(\vec{u}))]
+        gamma = cfg.getfloat('constants', 'gamma')
+        grad_p = grad_E - 0.5*(np.einsum('ijk, iljk -> ljk', uvw, grad_rhouvw) +
+                              np.einsum('ijk, iljk -> ljk', rhouvw, grad_uvw))
```
FreddieWitherden, line 29: Add space before np.

pyfr/solvers/navstokes/elements.py
GonzaloSaez: Changed

FreddieWitherden:
Merged!  Thank you very much for this; it is greatly appreciated.

PR #192, opened 2021-02-09
ShockSensor: fix tplargs

GonzaloSaez:
The tplargs used by self.kernels['gradcoru_upts_curved']
and self.kernels['gradcoru_upts_linear'] was being overwritten
by the created for in self.kernels['shocksensor']

This regression issue didn't allow to run simulations with artificial viscosity


FreddieWitherden:
Looks good, although the shock sensor kernel is really in need of a rewrite.  Currently it works out to be an unnecessary dense matrix vector product where really the entire thing could be reduced down to just two dot products by pre-computing what's needed at the Python level.

PR #193, opened 2021-02-10
Rework _get_axnpby_kerns to avoid the use of matrix banks.

FreddieWitherden:


pyfr/plugins/integrate.py
```diff
@@ -107,26 +107,9 @@ def _init_gradients(self, intg, rinfo):
...
-                # Get the smats at the solution points and subset
-                smat = eles.smat_at_np('upts')[..., eset]
-
-                # Get |J|^-1 at the solution points and subset
-                rcpdjac = eles.rcpdjac_at_np('upts')[:, eset]
-
-                # Product to give J^-T at the solution points
-                self._rcpjact.append(rcpdjac*smat.transpose(2, 0, 1, 3))
-
     def _eval_exprs(self, intg):
```
WillTrojak, line 113: Should this change to the gradient handling be in a different PR?

pyfr/plugins/tavg.py
```diff
@@ -74,27 +74,9 @@ def _init_gradients(self, intg):
             gradpnames.update(re.findall(r'\bgrad_(.+?)_[xyz]\b', ex))
 
         privarmap = self.elementscls.privarmap[self.ndims]
-        self._gradpinfo = [(pname, privarmap.index(pname)) 
+        self._gradpinfo = [(pname, privarmap.index(pname))
                             for pname in gradpnames]
```
WillTrojak, line 79: Should this change to the gradient handling be in a different PR?

WillTrojak (commented):
Other than moving the plugin changes to a different PR, this looks fine.

pyfr/plugins/integrate.py
FreddieWitherden: It isn't a change per-se, just removing some dead code (the arrays are assigned to, but never read).

semi-h:
https://github.com/PyFR/PyFR/blob/9634d2aa7798262a0aa0420e2e03be437bea5857/pyfr/integrators/std/controllers.py#L125
https://github.com/PyFR/PyFR/blob/9634d2aa7798262a0aa0420e2e03be437bea5857/pyfr/integrators/dual/pseudo/pseudocontrollers.py#L28

Why didn't you get rid of these prepare_reg_banks calls as well?

FreddieWitherden:
> https://github.com/PyFR/PyFR/blob/9634d2aa7798262a0aa0420e2e03be437bea5857/pyfr/integrators/std/controllers.py#L125
> 
> https://github.com/PyFR/PyFR/blob/9634d2aa7798262a0aa0420e2e03be437bea5857/pyfr/integrators/dual/pseudo/pseudocontrollers.py#L28
> 
> Why didn't you get rid of these prepare_reg_banks calls as well?

These are for the errest kernels, rather than axnpby.  Their time will come, however.

semi-h (approved):

PR #194, opened 2021-02-11
Plugin: add stop conditions during runtime

GonzaloSaez:
When performin long simultions with PyFR, we have found that it is sometimes complicated to not waste cluster allocations due to the fact that solution data can not be generally written before the simulations are killed by the workload manager.

To avoid these issues, I have written a small plugin that checks different conditions such as the wall clock time, the physical time or the presence of an arbitrary file to write all solution data and stop the simulation.

One issue arising from this implementation is that the `writer` plugins will output data at intervals different than the ones imposed by the user when restarting simulations.
This happens due to the fact that the writing criteria is based on the time found within the restart file `self.tout_last = intg.tcurr` and not `self.tout_last = 0`. This could be easily changed by setting `self.tout_last = 0` in the `writer` plugin although I suppose that other undesired issues might arise when restarting simulations.

Example of use

```dosini
[soln-plugin-runtime]
max-phystime = 100 ; (max physical time in seconds)
max-walltime = 8 ; (max wall time in hours)
stop-file = stop ; (if a file called stop is found in basedir, stop the simulations)
nsteps = 50 ; (check each nsteps)
basedir = ./
basename = restart_{t:.15f} ; (file to be written at exit)
post-action = echo "Wrote file {soln} at time {t} for mesh {mesh}."
post-action-mode = blocking
```

Commit msg:

This plugin adds certain mechanisms to stop simulations
writing all solution data before exiting.

`max-phystime`:  if the physical time of the simulations
is higher than this parameter, all solution data will be
written to disk and the simulation will end.

`max-walltime`: if the wall clock time measured in hours 
from the start of the simulation is bigger than this parameter,
all solution data will be written to disk and the simulation
will end.

`stop-file`: if a file with the name associated to this parameter
is found, all solution data will be written to disk and the
simulation will end.


FreddieWitherden:
So in my experience with long running jobs the biggest issue is with writing out time averages rather than solutions.  That is to say, one needs to time the end of the simulation and cluster allocation period with the final time average file.

GonzaloSaez:
> So in my experience with long running jobs the biggest issue is with writing out time averages rather than solutions. That is to say, one needs to time the end of the simulation and cluster allocation period with the final time average file.

True, I didn't think about that. `tavg` can not read temporal averages from a restart file, so temporal averaging needs to be happen within a unique cluster allocation. 

One could consider storing temporal averages of all the domain inside restart solution files to ease out this issue, but that requires some thought I suppose. It is also possible to perform a-posteriori averaging of regions of interests with instantaneous data, but that might require a lot of hard drive space...

Anyhow, let me know if you think it is worth adding the plugin, I'll close the PR if not!

FreddieWitherden:
So one possibility is to extend the blocking post-action code to inspect the return value from the action and, if it is non-zero, terminate the simulation.  This would enable users to write little scripts which look at things such as the wall time, simulation time, and environment variables (which often contain information about the running cluster job) and use these to decide a course of action.  This would slot right into the existing framework and be somewhat more flexible (as the script can do anything, including asking the job scheduler for the remaining allocation time for the job).

GonzaloSaez:
The return value of a blocking post-action could be obtained in blocking mode using `subprocess.run(...)` with appropriate arguments. One could know the time left for a job with slurm using

```bash
seconds_left=`squeue -h -j $SLURM_JOBID -o %L | awk -F: '{ print ($1 * 3600) + ($2 * 60) + $3 }'``
```

However, I'm not sure if an if condition can be written in the post-action. Maybe using 'sh -c ' might help ?

FreddieWitherden:
> The return value of a blocking post-action could be obtained in blocking mode using `subprocess.run(...)` with appropriate arguments. One could know the time left for a job with slurm using
> 
> ```shell
> seconds_left=`squeue -h -j $SLURM_JOBID -o %L | awk -F: '{ print ($1 * 3600) + ($2 * 60) + $3 }'``
> ```
> 
> However, I'm not sure if an `if` condition can be written in the post-action. Maybe using `sh -c ` might help ?

Yes, so I would expect people to write a shell script and then as a post-action have `./myscript.sh args...`  We may want to provide some sample scripts in the documentation as appropriate.

GonzaloSaez:
I see, I think that that's doable, I'll try to implement it. Some more questions,

1. Should this be post-action be associated to a new plugin?  (this would be the easiest thing to implement I think, as we need treat the `post-action` differently than in the current implementation)
2. In that case, should the `end` signal be transmitted to other plugins? Or maybe should we ensure that this new plugin is the last plugin called?
3. What should the plugin do in case the `post-action` returns `true` (or `1`). Should it write all the solution data and exit?

FreddieWitherden:
I would wire the code into https://github.com/PyFR/PyFR/blob/develop/pyfr/plugins/base.py#L84 such that it works for any plugin which supports post-actions.  The user can then decide which plugin they want to attach the condition to.  The idea would be to abort early if it looks as if you will not have time to write out another solution/time average.  This avoids wasting time (which is key with time averages, where the last thing you want is a partial accumulation).

One option is just to abort as-is, which is probably fine.  Slightly more fancy is to set a flag in the integrator so that it terminates the simulation after all plugins have finished being called for this time step.

GonzaloSaez:
Let me know if the new changes work for you. All plugins can now store an `abort-action` which is called in `blocking` mode.
If the result of the action is different from `0`, the `intg` is informed and it will exit accordingly (only implemented for `std` integrators). Example:

```dosini
[soln-plugin-writer]
dt-out = 10.0
basedir = .
basename = euler_vortex_2d-{t:.1f}
post-action = echo "Writing"
post-action-mode = blocking
abort-action = sh ./abort.sh {t:.5f}
```

abort.sh (slurm seems pretty inconsistent with the format in which the time left of the allocation or the wall time is output. Maybe we can send more information through the action)

```bash
#!/bin/bash

max_phystime=1.4
max_walltime=83000
FILE=stop

curr_walltime=$(squeue -h -j $SLURM_JOBID -o %M | awk -F: '{ print ($1 * 3600) + ($2 * 60) + $3 }')
echo $curr_walltime

if (( $(echo "$1 > $max_phystime" | bc -l) )); then
    echo "Abort max_phystime"
    exit 1
fi

if (( $(echo "$curr_walltime > $max_walltime" | bc -l) )); then
    echo "Abort max_walltime"
    exit 1
fi

if test -f "$FILE"; then
    echo "Abort $FILE exists."
    exit 1
fi

exit 0;
```

FreddieWitherden:
So I am not sure if we need to add a new key; rather just add support for checking the return value to the existing post-action handler when in blocking mode.  This will simplify the code greatly.  Also, I do not think we need a new plugin, either, given that its functionality is handled by checking post-action return codes.

GonzaloSaez:
> So I am not sure if we need to add a new key; rather just add support for checking the return value to the existing post-action handler when in blocking mode. This will simplify the code greatly. Also, I do not think we need a new plugin, either, given that its functionality is handled by checking post-action return codes.

1. Not adding a new action might induce non-regression issues as some plugin's post action in some people's simulations might be returning other than 0, without the users knowing about it. Moreover, what if one wants to perform a `post-action` separately from a `abort-action`? I suppose that they could be separated with `;` ...  I'm just wondering about the consequences of the implementation I do not want to sound harsh at all! If you are OK with the previous observations I'll remove the `abort-action`.
2. The additional plugin will be deleted don't worry!

FreddieWitherden:
I would not worry too much about introducing regressions; very few people are using post-actions and, in instances where actions are returning non-zero they really should know about it (as it probably means they're broken).  This chance would help them become aware of it and act accordingly!  For multiple post actions the solution is to call a script which performs the actions in sequence; so your script does its usual work and then checks the abort condition all in one.

GonzaloSaez:
That sounds reasonable! I didn't think about using a script for the `post-action` thanks for that.

I implemented the check within the `post-action` logic.

pyfr/integrators/std/controllers.py
```diff
@@ -40,6 +41,12 @@ def _accept_step(self, dt, idxcurr, err=None):
         # Fire off any event handlers
         self.completed_step_handlers(self)
 
+        # Abort if plugins ask for it
```
FreddieWitherden, line 44: Also needs to be ported over to dual time stepping.

pyfr/integrators/std/controllers.py
```diff
@@ -40,6 +41,12 @@ def _accept_step(self, dt, idxcurr, err=None):
...
+        if self.abort:
+            # Ensure that the callbacks registered in atexit
+            # are called only once if stopping the computation
+            sys.exit()
```
FreddieWitherden, line 48: sys.exit(1) might be more appropriate.

pyfr/plugins/base.py
```diff
@@ -6,6 +6,7 @@
 
 import numpy as np
 from pytools import prefork
+import subprocess
```
FreddieWitherden, line 9: Must use prefork for calling scripts to avoid crashes with certain MPI libraries.

pyfr/plugins/base.py
```diff
@@ -77,13 +78,18 @@ def __init__(self, *args, **kwargs):
             if self.postactmode not in {'blocking', 'non-blocking'}:
                 raise ValueError('Invalid post action mode')
 
+        if self.cfg.hasopt(self.cfgsect, 'abort-action'):
```
FreddieWitherden, line 81: Not needed.

pyfr/plugins/base.py
```diff
@@ -95,10 +101,23 @@ def _invoke_postaction(self, **kwargs):
...
+        # Need to bcast the return code from the blocking post-action
+        # of the root rank
+        if rank != root and check_abort:
+            retcode = comm.bcast(None, root=root)
+
+        # If returing code of the blocking post-action
+        # is other than zero, then the computation will be aborted
+        if check_abort:
+            if retcode != 0:
+                intg.abort = True
```
FreddieWitherden, line 119: Assign intg.abort = comm.bcast(...) != 0

GonzaloSaez: Done

pyfr/plugins/base.py
GonzaloSaez: Done

pyfr/plugins/base.py
GonzaloSaez: Yes true, sorry for that I was just doing some tests

pyfr/integrators/std/controllers.py
GonzaloSaez: Done

pyfr/integrators/std/controllers.py
GonzaloSaez: Done

GonzaloSaez:
Added a check to ensure that `int.abort` is not overwritten by other plugins if it is `True`

FreddieWitherden:
In terms of handling the abort the most robust solution might be to do an allreduce in _accept_step (with | as the operator).  This opens up the abort functionality to all plugins and avoids them needing to be concerned about the MPI stuff.

GonzaloSaez:
> In terms of handling the abort the most robust solution might be to do an allreduce in _accept_step (with | as the operator). This opens up the abort functionality to all plugins and avoids them needing to be concerned about the MPI stuff.

Wouldn't that add one synchronization step each call to `accept_step`? That might be imply an _important_ performance penalty in large MPI computations, but I'll implement it there if that's OK for you

FreddieWitherden:
> > In terms of handling the abort the most robust solution might be to do an allreduce in _accept_step (with | as the operator). This opens up the abort functionality to all plugins and avoids them needing to be concerned about the MPI stuff.
> 
> Wouldn't that add one synchronization step each call to `accept_step`? That might be imply an _important_ performance penalty in large MPI computations, but I'll implement it there if that's OK for you

There will be an extra reduction per time step, although we already do such reductions with adaptive time stepping and have never observed them to contribute anything to the overall run time.  Nowadays with mpi4py the performance scales as log N where N is the number of ranks.

GonzaloSaez:
> > > In terms of handling the abort the most robust solution might be to do an allreduce in _accept_step (with | as the operator). This opens up the abort functionality to all plugins and avoids them needing to be concerned about the MPI stuff.
> > 
> > 
> > Wouldn't that add one synchronization step each call to `accept_step`? That might be imply an _important_ performance penalty in large MPI computations, but I'll implement it there if that's OK for you
> 
> There will be an extra reduction per time step, although we already do such reductions with adaptive time stepping and have never observed them to contribute anything to the overall run time. Nowadays with mpi4py the performance scales as log N where N is the number of ranks.

Perfect, sounds reasonable. I have added the `allreduce` call with `lor` operator

pyfr/integrators/base.py
```diff
@@ -133,6 +137,14 @@ def cfgmeta(self):
         else:
             return {'config': cfg, 'config-0': cfg}
 
+    def _check_abort(self):
+        comm, rank, root = get_comm_rank_root()
+        self.abort = comm.allreduce(self.abort, op=get_mpi('lor'))
```
FreddieWitherden, line 142: No need to assign, can just put it in the if clause.

pyfr/integrators/dual/phys/controllers.py
```diff
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+import sys
```
FreddieWitherden, line 3: Not needed.

pyfr/integrators/std/controllers.py
```diff
@@ -3,6 +3,7 @@
 import math
 
 import numpy as np
+import sys
```
FreddieWitherden, line 6: Not needed.

pyfr/plugins/base.py
```diff
@@ -95,7 +95,11 @@ def _invoke_postaction(self, **kwargs):
 
             # Invoke
             if self.postactmode == 'blocking':
-                prefork.call(cmdline)
+                # Store returning code of the post-action
+                # If it is different from zero
+                # request intg to abort the computation
+                rtcode = prefork.call(cmdline)
+                intg.abort = rtcode != 0 or intg.abort
```
FreddieWitherden, line 102: intg.abort |= not precall.call(cmdline)

FreddieWitherden:
The current recommendation is to have a post-action run after each file output and have this post-action decide if it is likely that the simulation will make it to the next file dump (for example by asking SLURM how much run-time is left and comparing it against the time difference between the two most recent output files).  If the chances are unfavourable, the post-action should ask the simulation to abort, ensuring that no compute is wasted.

PR #195, opened 2021-02-12
Writer: added high-order vtu output

GonzaloSaez:
This commit adds a new writer which outputs true
high-order VTK cells using the vtkLagrange* cell types.

The main benefits of this new writer are:

1) Visualization of the shape of non-linear cells within
Paraview.

2) The user may also modify the parameter
`Nonlinear subdivision Level` to increase or reduce the
interpolation and shape basis order used by VTK
or Paraview to represent the solution and cells' shape.

3) There is no need to output gradient data from
PyFR as VTK will use the same interpolation basis
than the one in PyFR, provided that the number of
subdivisions is consistent with the interpolation
order of the simulations.

4) File sizes are a smaller (40% in my tests) as less cells are output
and due to the fact that the connectivity arrays
have less entries. Moreover, as gradients are generally
not needed, it is supposed that this new writer
might reduce even further storage requirements

To use this writer one needs to add the argument
`-t vtkho` to the export command

WARNING: `pyr` type elements are not compatible as
there exists no `vtkLagrangePyramid` element in VTK.
It would be possible to use `vtkQuadraturePyramid`
for a quadratic `pyr` representation in VTK.
However, for the moment this writer will
subdivide `pyr` elements if found in the mesh.

INFO: the `nodemaps.py` file includes the mapping
of nodes from VTK8-9 to PyFR and viceversa.
A small note has been added to explain the
importance of using the VTK8 node connectivity
instead of that of VTK9

(My last PR for a while I promise 😅 )


GonzaloSaez:
BTW: the previous PRis bugged as the line `intg.abort |= not prefork.call(cmdline)` is doing the opposite of what it was supposed to do, I should have checked before, sorry for that. Do you want me to create a new PR to fix it?

WillTrojak:
Looks good, the only thing I would consider is changing from `-t vtkho` to maybe `-t vtk_ho` or similar, just a little clearer to me.

WillTrojak:
> BTW: the previous PR is bugged as the line `intg.abort |= not prefork.call(cmdline)` is doing the opposite of what it was supposed to do, I should have checked before, sorry for that. The fix is to change that statement with `intg.abort |= prefork.call(cmdline) != 0`.
> 
> Do you want me to create a new PR to fix it?

Yeah if you could make a separate PR fixing that that would be great.

FreddieWitherden:
> > BTW: the previous PR is bugged as the line `intg.abort |= not prefork.call(cmdline)` is doing the opposite of what it was supposed to do, I should have checked before, sorry for that. The fix is to change that statement with `intg.abort |= prefork.call(cmdline) != 0`.
> > Do you want me to create a new PR to fix it?
> 
> Yeah if you could make a separate PR fixing that that would be great.

I fixed it.

pyfr/writers/nodemaps.py
```diff
@@ -0,0 +1,386 @@
...
+
+class VTK9NodeMaps(object):
+    """
+    WARNING
+    VTK9 changed the connectivity mapping of vtkLagrangeHexahedron
+    cells. However, if the `xml version` token of a given `vtu` or
+    `pvtu` file is lower than 5.1, then VTK>=9 will suppose that the
+    connectivity is consistent with that of VTK8 and will reorder the
+    nodes without the user noticing it.
+    As `xml version` is hardcoded to 1.0 in PyFR, we shouldn't use
```
FreddieWitherden, line 270: Any reason we don't change the version we write out?

pyfr/writers/nodemaps.py
```diff
@@ -0,0 +1,386 @@
...
+                                12, 13, 16, 17, 18]),
+        ('quad', 36): np.array([0,  5, 35, 30,  1,
+                                2,  3,  4, 11, 17,
+                                23, 29, 31, 32, 33,
+                                34,  6, 12, 18, 24,
+                                7,  8,  9, 10, 13,
+                                14, 15, 16, 19, 20,
+                                21, 22, 25, 26, 27,
+                                28]),
+        ('hex', 8): np.array([0, 1, 3, 2, 4, 5,
```
FreddieWitherden, line 52: Can be a bit more aggressive with the line lengths.

pyfr/writers/vtkho.py
```diff
@@ -0,0 +1,117 @@
...
+
+        # Points inside of a standard element
+        svpts = self._get_std_ele(name, nspts)
+        nsvpts = len(svpts)
+
+        # Transform PyFR to VTK8 points
+        # Read nodemaps.py for more information on
+        # why we chose VTK8 instead of VTK9 points
+        svpts = np.array(svpts)[VTK8NodeMaps.from_pyfr[(name, nsvpts)]
+                                .astype(int)]
```
FreddieWitherden, line 75: Why is the astype needed?

pyfr/writers/vtkho.py
```diff
@@ -0,0 +1,117 @@
...
+    vtk_types = dict(tri=69, quad=70, tet=71, pri=73, hex=72)
+
+    def _get_npts_ncells_nnodes(self, sk):
+        etype, neles = self.soln_inf[sk][0], self.soln_inf[sk][1][2]
+
+        if etype == 'pyr':
+            # No Lagrange pyr cells in VTK
+            # Therefore, rely on the subdivision mechanism
+            # of the vtk writer
+            super()._get_npts_ncells_nnodes(sk)
```
FreddieWitherden, line 27: Move return up

pyfr/writers/vtkho.py
```diff
@@ -0,0 +1,117 @@
...
+import numpy as np
+
+from pyfr.shapes import BaseShape
+from pyfr.util import subclass_where
+from pyfr.writers.vtk import VTKWriter
+from pyfr.writers.nodemaps import VTK8NodeMaps
+
+
+class VTKHOWriter(VTKWriter):
+    name = 'vtkho'
```
FreddieWitherden, line 12: vtk-ho?

pyfr/writers/vtkho.py
```diff
@@ -0,0 +1,117 @@
...
+    extn = ['']
+
+    # Types of VTK_LAGRANGE_TRIANGLE, VTK_LAGRANGE_QUADRILATERAL
```
FreddieWitherden, line 15: Comment probably not needed

pyfr/writers/vtkho.py
```diff
@@ -0,0 +1,117 @@
...
+        shapecls = subclass_where(BaseShape, name=etype)
+
+        # Number of vis points
+        # This value coincides with the number of 
+        # nodes of the vtkLagrange* correspondent 
+        # objects
+        npts = shapecls.nspts_from_order(self.divisor + 1)*neles
+
+        # Number of cells and nodes
+        ncells = neles
```
FreddieWitherden, line 40: Not sure these assignments are needed.

pyfr/writers/vtkho.py
```diff
@@ -0,0 +1,117 @@
...
+        nnodes = npts
+
+        return npts, ncells, nnodes
+
+    def _write_data(self, vtuf, mk, sk):
+        name = self.mesh_inf[mk][0]
```
FreddieWitherden, line 46: etype for consistency?

pyfr/writers/vtkho.py
```diff
@@ -0,0 +1,117 @@
...
+        nspts, neles = mesh.shape[:2]
+
+        # Points inside of a standard element
+        svpts = self._get_std_ele(name, nspts)
+        nsvpts = len(svpts)
+
+        # Transform PyFR to VTK8 points
+        # Read nodemaps.py for more information on
+        # why we chose VTK8 instead of VTK9 points
+        svpts = np.array(svpts)[VTK8NodeMaps.from_pyfr[(name, nsvpts)]
```
FreddieWitherden, line 74: No need for ()

pyfr/writers/nodemaps.py
GonzaloSaez: I didn't want to introduce new regressions with this PR. Moreover, XML VTK files are usually poorly documented (see https://gitlab.kitware.com/vtk/vtk/-/issues/17675 or https://gitlab.kitware.com/vtk/vtk/-/issues/17989) so I do not know what may go wrong if one changes the value of `xml version` ... Anyway, I'll have a look at it soon and try to see if there are any regressions with different `xml version`

pyfr/writers/nodemaps.py
GonzaloSaez: Ok I'll try to reduce the line length. Do you think that this code should go to `readers.nodemaps` instead?

GonzaloSaez:
> Looks good, the only thing I would consider is changing from `-t vtkho` to maybe `-t vtk_ho` or similar, just a little clearer to me.

True, not the best name, I'll use `vtk-ho` as Freddie proposed

pyfr/writers/vtkho.py
GonzaloSaez: Maybe a leftover, I'll remove it

pyfr/writers/vtkho.py
GonzaloSaez: Sounds good

pyfr/writers/nodemaps.py
FreddieWitherden: I think the XML version should always be 1 as that is an XML standard thing.  It is the version="0.1" thing which might need to change.  If we want to be careful we can make the version a property of the class and have the high-order subclass bump it up.

pyfr/writers/nodemaps.py
FreddieWitherden: I think the line can be made longer and still be within 79 characters.  I do not have a strong preference as to where it goes; someday we might tabulate these things in files like we do with quadrature rules.

pyfr/writers/nodemaps.py
GonzaloSaez: Yes sorry I was referring to that token my bad. So it seems that for the `XML` like files (`vtu` for example) the switch is done for `version < 2.1` and for `Legacy` type files the switch is done for `version < 5.1`. I have tried using `version = "2.1"` and only the VTK9 connectivity gives correct results. So I'll add a property to each `vtk` writer with an appropriate `version` (0.1 for `vtk.py` and 2.1 for `vtkho.py` and I'll use VTK9 connectivity when needed (do you want to leave VTK8 connectivity within the files?)

FreddieWitherden: I think we can get rid of the VTK8 connectivity.

GonzaloSaez:
I added some logic to chose the `VTKFile version`.

Just a small warning, if one uses a `Paraview` installation with `VTK < 9` and opens a `vtu` file with  `VTKFile version = 2.1` I don't know which might be the consequences with the regards to the connectivity of `vtkLagrangeHexahedron`

pyfr/writers/nodemaps.py
```diff
@@ -0,0 +1,374 @@
...
+class VTK8NodeMaps(object):
+    """Mappings between the node ordering of PyFR and that of VTK
+    Lagrange cells
+
+    Node mappings are contained within two dictionaries; one maps from
+    VTK node ordering to PyFR, and the other provides the inverse.
+    Dictionary items are keyed by a tuple of element type (string) and
+    number of solution points per element (integer).
+
+    INFO: VTK updated the numbering of Lagrange hex elements in VTK9
```
FreddieWitherden, line 15: Check comment is up to date

pyfr/writers/vtk.py
```diff
@@ -57,6 +57,10 @@ def __init__(self, args):
                         (f'grad {var} {f}', nf(f)) for f in fields
                     )
 
+    @staticmethod
+    def _vtkfile_version():
```
FreddieWitherden, line 61: No need for a method here.  Can just assign an attribute in the class definition.

pyfr/writers/vtk.py
```diff
@@ -202,7 +206,8 @@ def write_out(self):
                 write_s_to_fh('<?xml version="1.0" ?>\n<VTKFile '
                               'byte_order="LittleEndian" '
                               'type="UnstructuredGrid" '
-                              'version="0.1">\n<UnstructuredGrid>\n')
+                              f'version=\"{self._vtkfile_version()}\">\n'
```
FreddieWitherden, line 209: Not sue if \ are needed

pyfr/writers/vtkho.py
```diff
@@ -0,0 +1,117 @@
+# -*- coding: utf-8 -*-
+
+import numpy as np
+
+from pyfr.shapes import BaseShape
+from pyfr.util import subclass_where
+from pyfr.writers.vtk import VTKWriter
+from pyfr.writers.nodemaps import VTK8NodeMaps, VTK9NodeMaps
```
FreddieWitherden, line 8: Move nodemaps import up so alphabetical.

pyfr/writers/vtkho.py
```diff
@@ -0,0 +1,117 @@
...
+        # Points inside of a standard element
+        svpts = self._get_std_ele(etype, nspts)
+        nsvpts = len(svpts)
+
+        # Transform PyFR to VTK8 points
+        # Read nodemaps.py for more information on
+        # why we chose VTK8 or VTK9 maps
+        vtk_from_pyfr = (VTK8NodeMaps.from_pyfr
+                         if self._vtkfile_version() == "0.1"
+                         else VTK9NodeMaps.from_pyfr)
```
FreddieWitherden, line 74: This is always 9, so we can remove VTK8 and simplify.

pyfr/writers/nodemaps.py
```diff
@@ -0,0 +1,374 @@
+# -*- coding: utf-8 -*-
+
+import numpy as np
+
+
+class VTK8NodeMaps(object):
```
FreddieWitherden, line 6: I think we can remove the VTK8 stuff and just call it VTKHONodeMaps

GonzaloSaez:
I'd appreciate if you could let me know if the implementation works on your side. I do not have access to meshes with `pyr` elements so I can not test if meshes with those elements are being correctly exported. I have tested grids with all other 2D and 3D elements and it seems to work correctly

pyfr/writers/vtk.py
```diff
@@ -16,6 +16,11 @@ class VTKWriter(BaseWriter):
     name = 'vtk'
     extn = ['.vtu', '.pvtu']
 
+    # With version >= 2.1, VTK9 supposes
```
FreddieWitherden, line 19: No need for the comment

FreddieWitherden:
> I'd appreciate if you could let me know if the implementation works on your side (I think that only Paraview 5.9.0 is compatible with the new file format). I do not have access to meshes with `pyr` elements so I can not test if meshes with those elements are being correctly exported. I have tested grids with all other 2D and 3D elements and it seems to work correctly.
> 
> This writer could also be enhanced by choosing the order of the elements based on the ratio of the modal coefficients of the output variables at these elements, in a similar way as naïve shock sensors work. But well we can leave this for a future development as it requires to fully rewrite the writer logic.

@WillTrojak @tarikdzanic Would you guys be able to test this PR with some of your pyramid meshes?

WillTrojak:
Tested this on a tet dominant mesh with some prisms and pyramids. Seems to work fine when subsequently visualising with Paraview, and seems to bring down the memory quite a bit. 79M points needed about 28GB based on `htop`

GonzaloSaez:
> Tested this on a tet dominant mesh with some prisms and pyramids. Seems to work fine when subsequently visualising with Paraview, and seems to bring down the memory quite a bit. 79M points needed about 28GB based on `htop`

That is expected as there are less number of cells to be processed and same number of points. Moreover, the `Gradient of UnstructuredGrid` filter should be more accurate I think, as VTK will use a Lagrange basis to compute them. You can also opt in to improve the visualization of the fields and curved elements with the `Nonlinear subdivision` option in the advanced options of Paraview, although this is computationally expensive. Out of curiosity, how much memory does it consume with the default `vtk` writer?

pyfr/writers/vtkho.py
```diff
@@ -0,0 +1,112 @@
...
+
+import numpy as np
+
+from pyfr.shapes import BaseShape
+from pyfr.util import subclass_where
+from pyfr.writers.nodemaps import VTKHONodeMaps
+from pyfr.writers.vtk import VTKWriter
+
+
+class VTKHOWriter(VTKWriter):
```
FreddieWitherden, line 11: Do we want to integrate this class in with the existing VTK writer, and have it controlled through a switch?

WillTrojak:
> > Tested this on a tet dominant mesh with some prisms and pyramids. Seems to work fine when subsequently visualising with Paraview, and seems to bring down the memory quite a bit. 79M points needed about 28GB based on `htop`
> 
> That is expected as there are less number of cells to be processed and same number of points. Moreover, the `Gradient of UnstructuredGrid` filter should be more accurate I think, as VTK will use a Lagrange basis to compute them. You can also opt in to improve the visualization of the fields and curved elements with the `Nonlinear subdivision` option in the advanced options of Paraview, although this is computationally expensive. Out of curiosity, how much memory does it consume with the default `vtk` writer?

With the regular vtk writer, it uses more than 64GB, after which I ran out of system memory. I didn't see the point in extending the swap memory allocation just to find out exactly how much more than 64GB.

FreddieWitherden:
> > > Tested this on a tet dominant mesh with some prisms and pyramids. Seems to work fine when subsequently visualising with Paraview, and seems to bring down the memory quite a bit. 79M points needed about 28GB based on `htop`
> > 
> > 
> > That is expected as there are less number of cells to be processed and same number of points. Moreover, the `Gradient of UnstructuredGrid` filter should be more accurate I think, as VTK will use a Lagrange basis to compute them. You can also opt in to improve the visualization of the fields and curved elements with the `Nonlinear subdivision` option in the advanced options of Paraview, although this is computationally expensive. Out of curiosity, how much memory does it consume with the default `vtk` writer?
> 
> With the regular vtk writer, it uses more than 64GB, after which I ran out of system memory. I didn't see the point in extending the swap memory allocation just to find out exactly how much more than 64GB.

So I think here it very much depends on the level of sub-division.  Using `-d 1` should produce an extremely small file, but with very little resolution.

pyfr/writers/vtkho.py
GonzaloSaez: This could be done yes, although the switch that needs to be done with `pyr` cells which render some logic a bit ugly. Anyway, how would you inform the `vtk` writer to activate this switch from the command line? If I'm not mistaken each `writer` is associated with a unique extension and name...

FreddieWitherden: I'd use the presence of the `-d` flag to decide if to output a high order representation or to sub-divide.  So if `-d` is present, we sub-divide, otherwise we output a native high order file (which should probably be the default).

GonzaloSaez: So `-d` to subdivide and if not present use the `self.divisor = order`? I don't know if VTK scales very well with very high-order cells but we can always think about a way to reduce `self.divisor`. I'll look at it soon

FreddieWitherden: So if a divisor is not present then output a high-order VTK file.  If a divisor is provided then fall back to linear sub-divison.  For pyramids (where we have to sub-divide) then use the order as the amount of sub-division.

pyfr/writers/nodemaps.py
```diff
@@ -0,0 +1,139 @@
+# -*- coding: utf-8 -*-
+
+import numpy as np
+
+
+class VTKHONodeMaps(object):
```
FreddieWitherden, line 6: Would it be possible to go up to 8th order for the mappings (assuming it is supported by ParaView)?  For the mesh node maps we go up to 6, but it is not unheard of for solution orders to be slightly higher than mesh orders.

GonzaloSaez: VTK supports arbitrary orders for non-linear cells but Paraview limits the `Nonlinear subdivision level` parameter to 4. The true meaning of this parameter if a bit unclear to me, but it is supposed to recursively divide non-linear cells in triangles `Nonlinear subdivision level` times. Then the Lagrange basis of the considered cells are utilized to interpolate to those internal points of the subdivision triangles. Anyhow, I could add compatibility for such high-order (>6) elements if you wish.

pyfr/writers/vtkho.py
GonzaloSaez: You can find the modified `vtk` writer in the `wip4` commit.  Let me know if you prefer to use new logic in the `vtk` writer or if it is better to use a new writer.

pyfr/writers/vtk.py
```diff
@@ -9,18 +9,36 @@
...
 from pyfr.writers import BaseWriter
+from pyfr.writers.nodemaps import VTKHONodeMaps
 
 
 class VTKWriter(BaseWriter):
     # Supported file types and extensions
     name = 'vtk'
     extn = ['.vtu', '.pvtu']
 
+    vtkfile_version = '0.1'
```
FreddieWitherden, line 20: Can remove.

pyfr/writers/vtk.py
```diff
@@ -9,18 +9,36 @@
...
         self.dtype = np.dtype(args.precision).type
-        self.divisor = args.divisor or self.cfg.getint('solver', 'order')
+
+        # Choose whether to output subdivided cells or high order VTK cells
+        # If args.divisor holds its default value, then use
+        # high-order VTK cells as output
+        # Else use cell subdivision
+        self.ho_output = args.divisor == 0
+        order = self.cfg.getint('solver', 'order')
+        self.divisor = order if self.ho_output else args.divisor
```
FreddieWitherden, line 34: self.divisor = args.divisor or self.cfg.getint('solver', 'order')

pyfr/writers/vtk.py
```diff
@@ -35,6 +53,11 @@ def __init__(self, args):
             self._soln_fields = self.stats.get('data', 'fields').split(',')
             self._vtk_vars = [(k, [k]) for k in self._soln_fields]
 
+        if self.ho_output:
+            self._get_npts_ncells_nnodes = self._get_npts_ncells_nnodes_ho
+        else:
+            self._get_npts_ncells_nnodes = self._get_npts_ncells_nnodes_div
```
FreddieWitherden, line 59: _lin?

pyfr/writers/vtk.py
```diff
@@ -138,6 +161,26 @@ def _get_npts_ncells_nnodes(self, sk):
...
         return npts, ncells, nnodes
 
+    def _get_npts_ncells_nnodes_ho(self, sk):
+        etype, neles = self.soln_inf[sk][0], self.soln_inf[sk][1][2]
+
+        if etype == 'pyr':
+            # No Lagrange pyr cells in VTK
+            # Therefore, rely on the subdivision mechanism
+            # of the vtk writer
+            return self._get_npts_ncells_nnodes(sk)
```
FreddieWitherden, line 171: I think we need to explicitly call the _div routine here.

pyfr/writers/vtk.py
```diff
@@ -333,19 +386,26 @@ def _write_data(self, vtuf, mk, sk):
         self._write_darray(vpts.swapaxes(0, 1), vtuf, self.dtype)
 
         # Perform the sub division
-        subdvcls = subclass_where(BaseShapeSubDiv, name=name)
-        nodes = subdvcls.subnodes(self.divisor)
+        if self.ho_output:
+            nodes = np.arange(0, nsvpts)
```
FreddieWitherden, line 390: No need for first argument to arange.

pyfr/writers/vtkho.py
FreddieWitherden: I prefer this new implementation.

pyfr/writers/vtk.py
GonzaloSaez: Good catch!

pyfr/writers/vtk.py
```diff
@@ -9,19 +9,36 @@
...
 from pyfr.writers import BaseWriter
+from pyfr.writers.nodemaps import VTKHONodeMaps
 
 
 class VTKWriter(BaseWriter):
     # Supported file types and extensions
     name = 'vtk'
     extn = ['.vtu', '.pvtu']
 
+    vtk_types = dict(tri=69, quad=70, tet=71, pri=73, hex=72)
```
FreddieWitherden, line 20: Might want to rename or define locally.

FreddieWitherden:
I'm happy with this.  @GiorgioGiangaspero Would appreciate your feedback, especially with regards to the defaults.

pyfr/writers/vtk.py
```diff
@@ -333,19 +385,26 @@ def _write_data(self, vtuf, mk, sk):
         self._write_darray(vpts.swapaxes(0, 1), vtuf, self.dtype)
 
         # Perform the sub division
-        subdvcls = subclass_where(BaseShapeSubDiv, name=name)
-        nodes = subdvcls.subnodes(self.divisor)
+        if self.ho_output:
```
FreddieWitherden, line 388: Do we need a pyr check here?

GonzaloSaez: Absolutely

pyfr/writers/nodemaps.py
FreddieWitherden: I think this is worthwhile; not because I expect it to be useful now but because I expect ParaView to improve its support for high-order elements in the future and so it is worth getting the support wired up now.

GonzaloSaez: Ok I'll regenerate the node maps. The current implementation is compatible with degree 5 hexahedrons. If I'm not mistaken, you want the implementation to be compatible with tensor-product elements up to degree 7 ?

pyfr/writers/vtk.py
```diff
@@ -9,19 +9,36 @@
...
         super().__init__(args)
 
         self.dtype = np.dtype(args.precision).type
+
+        # Choose whether to output subdivided cells or high order VTK cells
+        # If args.divisor holds its default value, then use
+        # high-order VTK cells as output
+        # Else use cell subdivision
+        self.ho_output = args.divisor == 0
+        order = self.cfg.getint('solver', 'order')
```
GiorgioGiangaspero, line 32: I think this `order` is unused

GiorgioGiangaspero:
I like very much the idea that the default output is a high-order vtk file, and the order is that of the solution we are exporting.

As as user, I still would like to have the possibility to downsample the solution for quick visualisation and post-processing.
At the moment this can be done by specifying a low number of linear subdivisions. However, forgetting the pyramids for a moment, do we still need the linear subdivisions or could we just think in terms of order?
What I am getting at is that I also like the idea that the user could pass the argument `--order` to `export`. Assuming we have a fourth order solution, the user would specify say, `--order 2` or 1 to have a downsampled solution (the default order being the one of the solution file of course). Would the overhead of low-order Lagrangian elements in Paraview be too much compared to 1 or 2 linear subdivisions?
I would still keep the linear subdivisions argument (`-d`) but I would make it mutually exclusive with `--order`.

And for pyramids, always assume the number of linear subdivisions equal to the given order.

FreddieWitherden:
> I like very much the idea that the default output is a high-order vtk file, and the order is that of the solution we are exporting.
> 
> As as user, I still would like to have the possibility to downsample the solution for quick visualisation and post-processing.
> At the moment this can be done by specifying a low number of linear subdivisions. However, forgetting the pyramids for a moment, do we still need the linear subdivisions or could we just think in terms of order?
> What I am getting at is that I also like the idea that the user could pass the argument `--order` to `export`. Assuming we have a fourth order solution, the user would specify say, `--order 2` or 1 to have a downsampled solution (the default order being the one of the solution file of course). Would the overhead of low-order Lagrangian elements in Paraview be too much compared to 1 or 2 linear subdivisions?
> I would still keep the linear subdivisions argument (`-d`) but I would make it mutually exclusive with `--order`, with `--order` being the default if none of them is passed.
> 
> And for pyramids, always assume the number of linear subdivisions equal to the given order.

This seems like a sensible idea to me.  Does the current PR work well for your larger cases?

GiorgioGiangaspero:
My big case (hex-dominant 2 million elements mesh with p=4) is still too big for my machine with 64Gb of memory, but I guess that was to be expected.
The high-order vtk solves the post-processing issue close to wall for my channel runs, which is very nice.
However,  probing the high-order solution takes ages: the spanwise averaging script that takes 2 minutes with 5 linear subdivisions was not done after 6 hours on the high-order vtk and I killed it.
Nevertheless, I think this is the way to go.

GonzaloSaez:
> My big case (hex-dominant 2 million elements mesh with p=4) is still too big for my machine with 64Gb of memory, but I guess that was to be expected.
> The high-order vtk solves the post-processing issue close to wall for my channel runs, which is very nice.
> However, probing the high-order solution takes ages: the spanwise averaging script that takes 2 minutes with 5 linear subdivisions was not done after 6 hours on the high-order vtk and I killed it.
> Nevertheless, I think this is the way to go.

The implementation of high-order cells in VTK is not very well optimized. The way the interpolate, compute the gradients and search for points inside high-order cells is not at all performance friendly.
Just FYI: one should be careful when using VTK's implementation to probe data with distorted non-linear cells, as the process used to transform from physical space coordinates to reference space coordinates is performed using some degree of linearization which does not yield the exact results (see https://gitlab.kitware.com/vtk/vtk/-/issues/18023). But well this error should be rather small.

So just to summarize the new logic to be implemented

1. The presence or not of the argument `-d` dictates whether to use subdivision or to output VTK high-order cells. However, this argument should not be accompanied with a integer. This will render old export scripts incompatible with the new implementation.
2. The token `--order` indicates the number of subdivisions or the order of the VTK Lagrange cells to be output. If not present, it is taken equal to the order of the simulation data.
3. Compatibility with high-order cells up to degree 9

Let me know if it sounds good to you and I'll implement the previous points

GonzaloSaez:
> My big case (hex-dominant 2 million elements mesh with p=4) is still too big for my machine with 64Gb of memory, but I guess that was to be expected.
> The high-order vtk solves the post-processing issue close to wall for my channel runs, which is very nice.
> However, probing the high-order solution takes ages: the spanwise averaging script that takes 2 minutes with 5 linear subdivisions was not done after 6 hours on the high-order vtk and I killed it.
> Nevertheless, I think this is the way to go.

You might be able to accelerate the interpolation process by extracting regions of interest before interpolating and using MPI. But I agree that VTK is much performant when treating linear cells

FreddieWitherden:
> So just to summarize the new logic to be implemented
> 
>     1. The presence or not of the argument `-d` dictates whether to use subdivision or to output VTK high-order cells. However, this argument should not be accompanied with a integer. This will render old export scripts incompatible with the new implementation.
> 
>     2. The token `--order` indicates the number of subdivisions or the order of the VTK Lagrange cells to be output. If not present, it is taken equal to the order of the simulation data.
> 
>     3. Compatibility with high-order cells up to degree 9
> 
> 
> Let me know if it sounds good to you and I'll implement the previous points

One alternative is to stick with accepting an integer after `-d` and if present employ subdivision.  This would maintain backwards compatibility in cases where an integer is currently provided.  High-order output can be controlled through a separate flag.  I like this as it enables us to adopt a blended approach in the future (for example some level of subdivision combined with high-order output, say quadratic, in the sub divided cells).

GonzaloSaez:
Ok I see. It seems that it is not so easy to have optional parameters with optional number of values with `argparse` as it is not straightforward, any recommendation to treat the case in which `--order` might appear but without value?

FreddieWitherden:
> Ok I see. It seems that it is not so easy to have optional parameters with optional number of values with `argparse`. How would you implement the case in which `--order` might appear but without value and distinguish it from the case in which `--order` does not appear?

I believe this is possible; see how we handle the `-t` flag (which is optional, but if you specify it a value is needed).  The `add_mutually_exclusive_group` method may also be useful, at least until sub-divided high order elements are supported.

GonzaloSaez:
I think that it is a bit more complicated as we need to support the following (-o is a shorthand of --order):

1. If no `-d` nor `-o` in cmd then suppose `-d 0`. Therefore no default values can be given to `-o`. This would ensure backwards compatibility with scripts that do not use `-d`.
2. If `-d present, `-o` must not be present and viceversa (`add_mutually_exclusive_group` works well here, but `-o` should not have any default, we can not use `required` in the group as condition 1. is needed for backwards compatibility)
3. If `-o` is present we would need to check if the parameter following is `int` or `string`. In the first case we would assign `order` to that `int`, in the last case then we should assign `order` to the simulation order. In my opinion, this is particular challenging as I don't see a straightforward way of checking the latter conditions in `argparse` (in a nice may without hacking into `argv` if I'm not mistaken). The fact that `o` can not have a default value, complicates even more the logic.

Is there an easier way of unrolling this logic?

FreddieWitherden:
> I think that it is a bit more complicated as we need to support the following (-o is a shorthand of --order):
> 
>     1. If no `-d` nor `-o` in cmd then suppose `-d 0`. Therefore no default values can be given to `-o`. This would ensure backwards compatibility with scripts that do not use `-d`.
> 
>     2. If `-d present, `-o` must not be present and viceversa (`add_mutually_exclusive_group`works well here, but`-o`should not have any default, we can not use`required` in the group as condition 1. is needed for backwards compatibility)
> 
>     3. If `-o` is present we would need to check if the parameter following is `int` or `string`. In the first case we would assign `order` to that `int`, in the last case then we should assign `order` to the simulation order. In my opinion, this is particular challenging as I don't see a straightforward way of checking the latter conditions in `argparse` (in a nice may without hacking into `argv` if I'm not mistaken). The fact that `o` can not have a default value, complicates even more the logic.
> 
> 
> Is there an easier way of unrolling this logic?

I would make both arguments optional (without defaults) and mutually exclusive, both being of type int.  If `-d` is provided then we output subdivided linear cells of the requested order.  If `-o` is provided (but we should not use `-o` as the flag) then we output high-order cells of the requested order, for `pyr` we use a subdivision order which is slightly higher than the `-o` order.  If neither is provided then we run with `-o = solution order`.

This preserves behaviour when `-d` is provided with things only changing when no specific sub-division amount is requested.

GonzaloSaez:
Sounds good, do you prefer this implementation

```python
        # Choose whether to output subdivided cells or high order VTK cells
        # If -o is given by the user then use high-order VTK cells as output
        # with order equal to the solution order or to the one provided by 
        # the user
        # Else if neither -o nor -d are found in the input then use high-order
        # VTK cells with order equal to the simulation order
        # Else use cell subdivision
        if args.order:
            self.ho_output = True
            self.divisor = args.order
        elif args.order is None and args.divisor is None:
            self.ho_output = True
            self.divisor = self.cfg.getint('solver', 'order')
        else:
            self.ho_output = False
            self.divisor = args.divisor
```

or rather this other one?

```python
        self.ho_output = args.order is not None \
                         or (args.order is None and args.divisor is None)
        if self.ho_output:
            order_soln = self.cfg.getint('solver', 'order')
            self.divisor = order_soln if args.order is None else args.order
        else:
            self.divisor = args.divisor
```

FreddieWitherden:
I think the first one is cleaner.

pyfr/__main__.py
```diff
@@ -70,10 +70,13 @@ def main():
...
-    ap_export.add_argument('-d', '--divisor', type=int, default=0,
-                           help='sets the level to which high order elements '
-                           'are divided; output is linear between nodes, so '
-                           'increased resolution may be required')
+    group = ap_export.add_mutually_exclusive_group(required=False)
+    group.add_argument('-d', '--divisor', type=int,
+                       help='sets the level to which high order elements '
+                       'are divided; output is linear between nodes, so '
+                       'increased resolution may be required')
+    group.add_argument('-o', '--order', type=int,
```
FreddieWitherden, line 78: Avoid -o as it is commonly used for output files.  Maybe -p?

pyfr/__main__.py
```diff
@@ -70,10 +70,13 @@ def main():
     ap_export.add_argument('-t', dest='type', choices=types, required=False,
                            help='output file type; this is usually inferred '
                            'from the extension of outf')
-    ap_export.add_argument('-d', '--divisor', type=int, default=0,
-                           help='sets the level to which high order elements '
-                           'are divided; output is linear between nodes, so '
-                           'increased resolution may be required')
+    group = ap_export.add_mutually_exclusive_group(required=False)
```
FreddieWitherden, line 73: Probably want a different name to group?

pyfr/writers/nodemaps.py
```diff
@@ -0,0 +1,489 @@
...
+                                235, 236, 237, 238, 239, 240, 243, 244, 245,
+                                246, 247, 250, 251, 252, 253, 256, 257, 258,
+                                261, 262, 265, 280, 281, 282, 283, 284, 285,
+                                288, 289, 290, 291, 292, 295, 296, 297, 298,
+                                301, 302, 303, 306, 307, 310, 325, 326, 327,
+                                328, 329, 330, 333, 334, 335, 336, 337, 340,
+                                341, 342, 343, 346, 347, 348, 351, 352, 355]),
+    }
+
+    to_pyfr = {k: np.argsort(v) for k, v in from_pyfr.items()}
```
FreddieWitherden, line 489: Not needed.

pyfr/writers/vtk.py
```diff
@@ -9,18 +9,44 @@
...
+        # If -o is given by the user then use high-order VTK cells as output
+        # with order equal to the solution order or to the one provided by 
+        # the user
+        # Else if neither -o nor -d are found in the input then use high-order
+        # VTK cells with order equal to the simulation order
+        # Else use cell subdivision
+        if args.order:
+            self.ho_output = True
+            self.divisor = args.order
+        elif args.order is None and args.divisor is None:
```
FreddieWitherden, line 37: Not sure if the first check is needed.

FreddieWitherden:
@GonzaloSaez Out of curiosity, how hard do you think it would be to support high-order elements with subdivision at the same time?

pyfr/writers/vtk.py
```diff
@@ -310,6 +363,14 @@ def _write_data(self, vtuf, mk, sk):
         svpts = self._get_std_ele(name, nspts)
         nsvpts = len(svpts)
 
+        if name != 'pyr' and self.ho_output:
+            # Transform PyFR to VTK9 points
+            # See nodemaps.py for more information on
+            # why VTK9 maps are chosen over those of VTK8
+            # High order `pyr` elements are not currently
+            # supported in VTK
+            svpts = np.array(svpts)[VTKHONodeMaps.from_pyfr[name, nsvpts]]
```
FreddieWitherden, line 372: Just so my understanding is correct this is going in the opposite direction (PyFR => VTK) than what we do in the reader (Gmsh => PyFR) but both are called `from_pyfr`.

GonzaloSaez: Changed it to `to_pyfr`. Let me know if its ok now

GonzaloSaez:
> @GonzaloSaez Out of curiosity, how hard do you think it would be to support high-order elements with subdivision at the same time?

Do you mean using linear subdivision and high-order elements in the same VTK file depending on the properties of the numerical solution or do you mean to use high-order elements to describe subdivided cells?
If its the latter it should be doable although some parts of the vtk<->pyfr connectivity would have to be reworked and the the additional point/nodes duplications arising from linear subdivision will become even worse with non-linear cells.
If it's the first option that could also be done but the code needs to be rearranged, as one might not know the number of points a priori. I think that the first choice is interesting, as one can decide based on the high-order modes energy of the primitive variables at each element whether the considered element should be output as a high-order element or not.

FreddieWitherden:
> > @GonzaloSaez Out of curiosity, how hard do you think it would be to support high-order elements with subdivision at the same time?
> 
> Do you mean using linear subdivision and high-order elements in the same VTK file depending on the properties of the numerical solution or do you mean to use high-order elements to describe subdivided cells?
> 
> If its the latter it should be doable although some parts of the vtk<->pyfr connectivity would have to be reworked and the the additional point/nodes duplications arising from linear subdivision will become even worse with non-linear cells.
> 
> If it's the first option that could also be done but the code needs to be rearranged, as one might not know the number of points a priori. I think that the first choice is interesting, as one can decide based on the high-order modes energy of the primitive variables at each element whether the considered element should be output as a high-order element or not.
> 
> As a user point of view I think that giving the possibility to avoid nodes duplication in VTU files should be the next step (in my opinion).

Yes, so for example you might specify two levels of sub-division and quadratic Lagrange polynomials inside of each element.  This allows once to get the best of both worlds and may prove useful as VTK  works to improve their support for high-order elements.

GonzaloSaez:
That might doable but it will incurr a lot node duplication between high
order subdivided elements, unless someone finds out a better way to
regenerate the elements connectivity at the faces... Do you think that
there are other ways to work around these issues?

FreddieWitherden:
> That might doable but it will incurr a lot node duplication between high order subdivided elements, unless someone finds out a better way to regenerate the elements connectivity at the faces... Do you think that there are other ways to work around these issues?

At least for the internal connectivity (between the interior faces of sub-divded elements) it should be possible to avoid the duplication without too much hassle.  There will of course be nodes doubling up on the faces of the actual element itself although we have those at the moment.

GonzaloSaez:
I think that one could use `divisor*order` as the real divisor, and then try to regenerate the connectivity of the subdivided elements of order `order` to avoid intraelement node duplication. Although I'm not going to lie, this seems like a bit of work and I don't think that I'll have much spare time in the near future. Are you OK with halting a bit the PR until I implement the new subdivision method or do you prefer to add the new functionality in a new PR?

FreddieWitherden:
> I think that one could use `divisor*order` as the real divisor, and then try to regenerate the connectivity of the subdivided elements of order `order` to avoid intraelement node duplication. Although I'm not going to lie, this seems like a bit of work and I don't think that I'll have much spare time in the near future. Are you OK with halting a bit the PR until I implement the new subdivision method or do you prefer to add the new functionality in a new PR?

This functionality should probably go into a new PR in the future.  I am happy to merge this PR so long as everyone is happy with the defaults/command line arguments.

pyfr/writers/vtk.py
```diff
@@ -9,18 +9,44 @@
...
+    vtk_types_ho = dict(tri=69, quad=70, tet=71, pri=73, hex=72)
+
     def __init__(self, args):
         super().__init__(args)
 
         self.dtype = np.dtype(args.precision).type
-        self.divisor = args.divisor or self.cfg.getint('solver', 'order')
+
+        # Choose whether to output subdivided cells or high order VTK cells
+        # If -k is given by the user then use high-order VTK cells as output
```
FreddieWitherden, line 28: This should probably be in the user guide.

GiorgioGiangaspero:
Did a simple test: computing and rendering the Qcriterion of a channel flow. Working with ho vtk is 3 times slower compared to 5 linear-subdivisions, and 60% slower compared to 6 linear subdivisions.
Despite this, I still think that the ho vtk should be our default, simply because we are using a high-order code and I think the output should be high-order too whenever possible. The user can always use linear subdivisions if required. And eventually, hopefully, Paraview will become faster with Lagrangian elements.

It seems to me that the last bit that needs to be decided is the number of linear subdivision for pyramids when a high-order export is requested. @FreddieWitherden you mentioned previously that you would like pyramids to be split with a number of linear subdivisions a bit higher than 'k'. Shall we say 'k+2'?

FreddieWitherden:
> Did a simple test: computing and rendering the Qcriterion of a channel flow. Working with ho vtk is 3 times slower compared to 5 linear-subdivisions, and 60% slower compared to 6 linear subdivisions.
> Despite this, I still think that the ho vtk should be our default, simply because we are using a high-order code and I think the output should be high-order too whenever possible. The user can always use linear subdivisions if required. And eventually, hopefully, Paraview will become faster with Lagrangian elements.
> 
> It seems to me that the last bit that needs to be decided is the number of linear subdivision for pyramids when a high-order export is requested. @FreddieWitherden you mentioned previously that you would like pyramids to be split with a number of linear subdivisions a bit higher than `k`. Shall we say `k+2`?

`k+2` works for me.

GonzaloSaez:
Changing the divisor to a different value only for pyramids produces some _ugly_ branching as you can see in the last commit. As I don't have access to grids with `pyr` elements, could anyone tell me if the last commit increases the number of subdivisions for `pyr` elements in high-order output mode? Ty

FreddieWitherden:
> Changing the divisor to a different value only for pyramids produces some _ugly_ branching as you can see in the last commit. As I don't have access to grids with `pyr` elements, could anyone tell me if the last commit increases the number of subdivisions for `pyr` elements in high-order output mode? Ty

Here it looks as if we might want to pass the divisor as a function argument rather than having it as a member variable.  This should clean up some of the issues.

GonzaloSaez:
> > Changing the divisor to a different value only for pyramids produces some _ugly_ branching as you can see in the last commit. As I don't have access to grids with `pyr` elements, could anyone tell me if the last commit increases the number of subdivisions for `pyr` elements in high-order output mode? Ty
> 
> Here it looks as if we might want to pass the divisor as a function argument rather than having it as a member variable. This should clean up some of the issues.

I've created a map `etypes_div` which links element types and number of sub-divisions. It should be cleaner now. Let me know if it works for you, if not I'll implement your approach.

I've also added the physical time as a FieldData in `vtu` files. This allows to visualize the physical time directly in Paraview and utilize its Statistical and Animation filters taking into account the physical times (one can choose to load or not this value directly in the GUI of Paraview >=5.9.0). Again, let me know if this sounds good to you, if not I'll remove this addition.

pyfr/writers/vtk.py
```diff
@@ -9,18 +9,57 @@
...
+        # of high order vtkLagrangeHexahedron cells
+        # See nodemaps.py for more information on
+        # why VTK9 maps are chosen over those of VTK8
+        self.vtkfile_version = '2.1' if self.ho_output else '0.1'
+
+        # Solution physical time
+        self.tcurr = self.stats.getfloat('solver-time-integrator', 'tcurr')
+
+        # Divisor for each type element
+        self.etypes_div = {etype: self.divisor
```
FreddieWitherden, line 55: self.etypes_div = defaultdict(lambda: self.divisor)

pyfr/writers/vtk.py
```diff
@@ -9,18 +9,57 @@
...
+                           for sk, (etype, shape) in self.soln_inf.items()}
+
+        # If using high-order VTK output mode
+        # set the number of sub-divisions of pyr
+        # elements to divisor + 2
+        if 'pyr' in self.etypes_div.keys() and self.ho_output:
```
FreddieWitherden, line 61: Can remove the first check

pyfr/writers/vtk.py
FreddieWitherden: Move up above the args.order if.

pyfr/writers/vtk.py
```diff
@@ -9,18 +9,56 @@
...
+            self.ho_output = True
+            self.divisor = self.cfg.getint('solver', 'order')
+        else:
+            self.ho_output = False
+            self.divisor = args.divisor
+
+        # If using high-order VTK output mode
+        # set the number of sub-divisions of pyr
+        # elements to divisor + 2
+        if self.ho_output:
```
FreddieWitherden, line 50: Can merge with the previous if statement

pyfr/writers/vtk.py
```diff
@@ -9,18 +9,56 @@
...
+    vtk_types_ho = dict(tri=69, quad=70, tet=71, pri=73, hex=72)
+
     def __init__(self, args):
         super().__init__(args)
 
         self.dtype = np.dtype(args.precision).type
-        self.divisor = args.divisor or self.cfg.getint('solver', 'order')
+
+        # Divisor for each type element
+        self.etypes_div = defaultdict(lambda: self.divisor)
```
FreddieWitherden, line 28: Need divisor.

pyfr/writers/vtk.py
```diff
@@ -35,6 +73,11 @@ def __init__(self, args):
             self._soln_fields = self.stats.get('data', 'fields').split(',')
             self._vtk_vars = [(k, [k]) for k in self._soln_fields]
 
+        if self.ho_output:
```
FreddieWitherden, line 76: See if we can merge with if block above.

pyfr/writers/vtk.py
GonzaloSaez: Didn't get this one. (see if the new modifications work for you). If not, we have if -> elif -> else and I would need to include the special `pyr` case in the if and elif.?

pyfr/writers/vtk.py
```diff
@@ -9,18 +9,60 @@
...
+        self.etypes_div = defaultdict(lambda: self.divisor)
+
+        # Choose whether to output subdivided cells or high order VTK cells
+        # If -k is given by the user then use high-order VTK cells as output
+        # with order equal to the solution order or to the one provided by
+        # the user
+        # Else if neither -o nor -d are found in the input then use high-order
+        # VTK cells with order equal to the simulation order
+        # Else use cell subdivision
+        if args.order:
```
FreddieWitherden, line 37: args.order or args.divisor is None

pyfr/writers/vtk.py
```diff
@@ -9,18 +9,60 @@
...
+            self.ho_output = True
+            self.divisor = args.order
```
FreddieWitherden, line 39: args.order or self.cfg.getint('solver', 'order')

pyfr/writers/vtk.py
```diff
@@ -9,18 +9,60 @@
...
+        elif args.divisor is None:
+            self.ho_output = True
```
FreddieWitherden, line 41: Can then remove this block

pyfr/writers/vtk.py
FreddieWitherden: Should now be able to merge the blocks

pyfr/writers/vtk.py
GonzaloSaez: I see now, yes it's true that it can be better organized. Let me know if the new commit works better for you.

pyfr/writers/vtk.py
```diff
@@ -9,18 +9,54 @@
...
+            self.ho_output = False
+            self.divisor = args.divisor
+            self._get_npts_ncells_nnodes = self._get_npts_ncells_nnodes_lin
+
+        # If outputting high-order VTK cells choose version 2.1
+        # to ensure consistency with VTK9 mappings
+        # of high order vtkLagrangeHexahedron cells
+        # See nodemaps.py for more information on
+        # why VTK9 maps are chosen over those of VTK8
+        self.vtkfile_version = '2.1' if self.ho_output else '0.1'
```
FreddieWitherden, line 56: Merge with above block

pyfr/writers/vtk.py
```diff
@@ -9,18 +9,54 @@
...
+
+        # Solution physical time
+        self.tcurr = self.stats.getfloat('solver-time-integrator', 'tcurr')
```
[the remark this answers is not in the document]
GonzaloSaez: Some options:

1) Do not ouput physical time for any export
2) Do not ouput physical time for time averages
3) Output `tend - tstart` as physical time for time averages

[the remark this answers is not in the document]
FreddieWitherden: I think we should output the time when exporting solutions.  @GiorgioGiangaspero Thoughts here?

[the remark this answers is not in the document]
GiorgioGiangaspero: I would write the time for instantaneous solutions as it has a clear meaning. For time averages, I think any value can easily be misinterpreted, so I would avoid it in this case.

pyfr/writers/vtk.py
```diff
@@ -292,6 +352,16 @@ def _write_parallel_header(self, vtuf):
 
         write_s('</PPointData>\n')
 
+    def _write_time_value(self, write_s):
+        write_s('<FieldData>\n')
+        write_s(f'<DataArray Name="TimeValue" type="Float64" '
+                f'NumberOfComponents="1" '
```
FreddieWitherden, line 358: Merge into a single write call.  (Python will concat the strings for you.)

PR #198, opened 2021-02-19
Partition: lower tolerance to select non-linear cells

GonzaloSaez:
## Short description

I was having issues with a SD7003 simulation in which the pressure coefficient was not consistent with reference values obtained with PyFR. To understand these issues I have tried to assess if the free-stream condition was assessed in this case. After some tests, I observed free-stream preservation issues after partitioning a mesh with non-linear cells. 

With some debugging I have found that this issue disappears if the `tol` parameter in `_find_curved_eles` is set to `0`. In that case all non-linear elements are tagged as non-linear cells. Moreover, in my tests a value of `tol=1e-15` ensures that non-linear cells with straight edges are tagged as linear cells, but I'm not sure if there exists an appropriate value of `tol` which allows to select non-linear cells with linear shapes.

## More details

I’m simulating the SD7003 test case provided by

B.C. Vermeire, F.D. Witherden, P.E. Vincent,
On the utility of GPU accelerated high-order methods for unsteady flow simulations: A comparison with industry-standard tools,
Journal of Computational Physics, Volume 334, 2017,
https://doi.org/10.1016/j.jcp.2016.12.049.

whose mesh consists of approximately 1.5E5 degree 2 hexahedrons (hex27) and p=4 (therefore there should not be any free-stream preservation issues theoretically even when using the true metrics). I’m having some issues when trying to replicate the results provided in the aforementioned study so I’ve tried to study if there could be an issue with non-linear cells.
I have changed the BCs definition so that they all behave as inlet conditions equal to the initial condition and I have modified the initial condition to ensure that it is constant over all the domain (see the attached ini)

With such test case definition, one expects the solution to remain constant over time (free-stream preservation condition). However, I have simulated this test case (after partitioning it with 2 ranks) using the current develop branch of PyFR and as one can observe in the attached figure, there exists generation of spurious velocities in cells close to the walls and at the interface between mesh blocks (zones where non-linear Jacobians are important I suppose). These regions are also located close to zones in which my simulations do not show the same results as those in the aforementioned article.

image
`sd7003.ini`
```ini
[backend]
precision = double

[backend-openmp]
cc = gcc
gimmik-max-nnz = 200000000

[backend-cuda]
device-id = local-rank
gimmik-max-nnz = 512000000
mpi-type = standard

[constants]
gamma = 1.4
mu    = 3.94405318873308E-6
Pr    = 0.72
M     = 0.2

[solver-time-integrator]
scheme = rk45
controller = pi
t0 = 0.0
dt = 0.0001
tend = 200
atol = 0.00001
rtol = 0.00001
safety-fact = 0.9
min-fact = 0.3
max-fact = 1.15

[solver]
system = navier-stokes
order  = 4

[solver-interfaces]
riemann-solver = rusanov
ldg-beta = 0.0
ldg-tau = 0.0

[solver-interfaces-line]
flux-pts = gauss-legendre

[solver-interfaces-quad]
flux-pts = gauss-legendre

[solver-elements-hex]
soln-pts = gauss-legendre

[soln-bcs-outlet]
type = char-riem-inv
rho = 1.0
u   = 0.2366431913
v   = 0.0
w   = 0.0
p   = 1.0

[soln-bcs-inlet]
type = char-riem-inv
rho = 1.0
u   = 0.2366431913
v   = 0.0
w   = 0.0
p   = 1.0

[soln-bcs-wall]
type = char-riem-inv
rho = 1.0
u   = 0.2366431913
v   = 0.0
w   = 0.0
p   = 1.0

[soln-ics]
rho  = 1.0
u    = 0.2366431913
v    = 0.0
w    = 0.0
p    = 1.0

[soln-plugin-writer]
dt-out = 0.001
basedir = ./
basename = slice_sd7003_{t:.3f}
region = [(-3, -3, -0.9), (3, 3, 1.1)]
post-action = echo "Wrote slice file {soln} at time {t} for mesh {mesh}."
post-action-mode = non-blocking

; [soln-plugin-integrate]
; nsteps = 1
; file = integrate
; int-rho = rho - 1.0
; int-u = u - 0.2366431913

```

### DIFF
--- a/pyfr/partitioners/base.py
+++ b/pyfr/partitioners/base.py
@@ -14,7 +14,7 @@ from pyfr.util import subclass_where
 Graph = namedtuple('Graph', ['vtab', 'etab', 'vwts', 'ewts'])
 
 
-def _find_curved_eles(etype, eles, tol=1e-5):
+def _find_curved_eles(etype, eles, tol=1e-15):
     nspts, neles, ndims = eles.shape
 
     shape = subclass_where(BaseShape, name=etype)


FreddieWitherden:
I think the underlying issue is that we do not normalise; hence it is difficult to find a single tol value that will always get the cut-off right.  (As for a physically larger cell you want a correspondingly larger tol and vice versa.)  Saying that, I agree that we should err on the side of caution.

FreddieWitherden:
Also as an FYI: the results for the aforementioned paper were done with PyFR v0.3.0 (+ a patch available as the ESM).  However, looking through git we did not add free stream preservation support to PyFR until v1.3.0.

GonzaloSaez:
> Also as an FYI: the results for the aforementioned paper were done with PyFR v0.3.0 (+ a patch available as the ESM). However, looking through git we did not add free stream preservation support to PyFR until v1.3.0.

Hi Freddie,

With $p=4$ and $N=2$ degree hexahedron cells, free stream preservation is ensured with all FR methods when using the true metric. This condition is a bit different (less constraint) with FR-DG as one only needs $p=2$ to ensure free-stream preservation with $N=2$ hexahedron.

By looking a bit at `shapes.py` of PyFR v0.3.0 I see that (if I'm not mistaken)  the true metric is used to compute the covariant vectors. Therefore, I think that the results of the aforementioned paper didn't have free stream preservation issues. 

The issue with the current implementation is that the imposed `tol` value produces a weird distribution of linear and non-linear cells near the airfoil wall and in the interface between mesh blocks. As one can not generally preserve the freestream condition at interfaces between linear and non-linear cells, then these spurious velocities appear in the flow.

FreddieWitherden:
So thinking about what kind of check we need we may want to consider looking at the magnitude of the largest non-linear mode over that of the largest (non-constant) linear mode.  I think this will get us a relative measure.

GonzaloSaez:
Mmm I don't know if that covers all edge cases. (Let me know if I'm wrong in the following).

Suppose that one is deciding whether a `N=3` degree hexahedron is linear or not. If the shape of the hexahedron can be exactly approximated with a `N'=2` shape function (which may be true depending on the mesher and domain triangulation), then the degree 3 modes will be `0`. So the ratio that you propose would be zero and you would tag the cell as linear, while it is not at all linear.

I think that normalization could be related to the volume of the considered elements. But I'm not sure if it is worth going down this route to distinguish non-linear cells with linear shapes from purely linear cells... Is the gain in terms of memory bandwidth worth it?

GonzaloSaez:
BTW, my results of SD7003 are much closer now to the reference ones. I'll continue taking some flow statistics and I'll comeback soon with a comparison.

FreddieWitherden:
> Mmm I don't know if that covers all edge cases. (Let me know if I'm wrong in the following).
> 
> Suppose that one is deciding whether a `N=3` degree hexahedron is linear or not. If the shape of the hexahedron can be exactly approximated with a `N'=2` shape function (which may be true depending on the mesher and domain triangulation), then the degree 3 modes will be `0`. So the ratio that you propose would be zero and you would tag the cell as linear, while it is not at all linear.
> 
> If by largest non-linear mode you mean the non-linear mode with the largest amplitude, that can also be sometimes prone to errors. If the cell centre is near zero, then the zeroth degree term will be close to zero and therefore the ratio might be misleading.
> 
> I think that the assessment of non-linear cells with linear shapes could be performed based on the contribution to the volume of all modes of the considered element (the zeroth degree mode does not contribute). But I'm not sure if it is worth going down this route to distinguish non-linear cells with linear shapes from purely linear cells... This is always prone to freestream issues if special care is not taken, see [1]. Is the gain in terms of memory bandwidth related to the metric terms worth it?
> 
> [1] Kopriva, D.A., Hindenlang, F.J., Bolemann, T. et al. Free-Stream Preservation for Curved Geometrically Non-conforming Discontinuous Galerkin Spectral Elements. J Sci Comput 79, 1389–1408 (2019). https://doi.org/10.1007/s10915-018-00897-9

By largest I meant in the `np.max(np.abs(...))` sense, so we partition the modes into three sets: constant mode (discard), linear modes, and non-linear high order modes.  Then we consider `np.max(np.abs(hmodes)) / np.max(np.abs(lmodes))` on an element-wise basis.

GonzaloSaez:
> > Mmm I don't know if that covers all edge cases. (Let me know if I'm wrong in the following).
> > Suppose that one is deciding whether a `N=3` degree hexahedron is linear or not. If the shape of the hexahedron can be exactly approximated with a `N'=2` shape function (which may be true depending on the mesher and domain triangulation), then the degree 3 modes will be `0`. So the ratio that you propose would be zero and you would tag the cell as linear, while it is not at all linear.
> > If by largest non-linear mode you mean the non-linear mode with the largest amplitude, that can also be sometimes prone to errors. If the cell centre is near zero, then the zeroth degree term will be close to zero and therefore the ratio might be misleading.
> > I think that the assessment of non-linear cells with linear shapes could be performed based on the contribution to the volume of all modes of the considered element (the zeroth degree mode does not contribute). But I'm not sure if it is worth going down this route to distinguish non-linear cells with linear shapes from purely linear cells... This is always prone to freestream issues if special care is not taken, see [1]. Is the gain in terms of memory bandwidth related to the metric terms worth it?
> > [1] Kopriva, D.A., Hindenlang, F.J., Bolemann, T. et al. Free-Stream Preservation for Curved Geometrically Non-conforming Discontinuous Galerkin Spectral Elements. J Sci Comput 79, 1389–1408 (2019). https://doi.org/10.1007/s10915-018-00897-9
> 
> By largest I meant in the `np.max(np.abs(...))` sense, so we partition the modes into three sets: constant mode (discard), linear modes, and non-linear high order modes. Then we consider `np.max(np.abs(hmodes)) / np.max(np.abs(lmodes))` on an element-wise basis.

I see. That could maybe work, but the ratio it is still not normalized I think. `lmodes` have and `hmodes` have different dimensions. But well if `tol` is low enough the method you propose should work in my opinion.

FreddieWitherden:
> I see. That could maybe work, but the ratio it is still not normalized I think. `lmodes` have and `hmodes` have different dimensions. But well if `tol` is low enough the method you propose should work in my opinion.

As the modes themselves are all orthonormal I think the expansion coefficients (which are what constitute lmodes and hmodes) should be comparable and even follow a decay law.

GonzaloSaez:
Something like this?

```python
    # Indices of linear modes
    lmodes = np.arange(1, len(lmodes))

    # Modal coeffs of linear shape functions for x, y, z
    elmodes = basis.invvdm.T[lmodes] @ eles.reshape(nspts, -1)
    elmodes = elmodes.reshape(-1, neles, ndims)

    # Modal coeffs of non-linear shape functions for x, y, z
    ehmodes = basis.invvdm.T[hmodes] @ eles.reshape(nspts, -1)
    ehmodes = ehmodes.reshape(-1, neles, ndims)

    # Energy content of linear shape functions
    energy_elmodes = np.linalg.norm(elmodes, axis=0)

    # Energy content of non-linear shape functions
    energy_ehmodes = np.linalg.norm(ehmodes, axis=0)

    # Energy ratio
    ratio = energy_ehmodes/energy_elmodes

    return np.any(np.abs(ratio) > tol, axis=1)
```

FreddieWitherden:
> Something along these lines?
> 
> ```python
>     # Indices of linear modes
>     lmodes = np.arange(1, len(lmodes))
> 
>     # Modal coeffs of linear shape functions for x, y, z
>     elmodes = basis.invvdm.T[lmodes] @ eles.reshape(nspts, -1)
>     elmodes = elmodes.reshape(-1, neles, ndims)
> 
>     # Modal coeffs of non-linear shape functions for x, y, z
>     ehmodes = basis.invvdm.T[hmodes] @ eles.reshape(nspts, -1)
>     ehmodes = ehmodes.reshape(-1, neles, ndims)
> 
>     # Energy content of linear shape functions
>     energy_elmodes = np.linalg.norm(elmodes, axis=0)
> 
>     # Energy content of non-linear shape functions
>     energy_ehmodes = np.linalg.norm(ehmodes, axis=0)
> 
>     # Energy ratio
>     ratio = energy_ehmodes/(energy_elmodes + 1e-10)
> 
>     return np.any(np.abs(ratio) > tol, axis=1)
> ```

Yes, although maybe with a slightly different means of identifying the low order modes so that we're not depending on any specific ordering of the basis.  I am also wondering if it is possible for a cell to be valid and have no energy in the low order modes, as it pertains to the factor of 1e-10 in the denominator.

GonzaloSaez:
In linear cells the linear modes will always have a non-zero energy. I don't know if there are cases of high-order elements with zero energy in the linear modes, maybe they exist? 

We could instead compare the energy of the non-linear modes with that of all modes (except the 0th)?

GonzaloSaez:
To show the importance of solving this bug, I'll attach two figures representing the time averaged `c_p` distributions of a SD7003 using the mesh of Vermeire et al. The first one is obtained my implementation of SDM in PyFR v1.11.0 without bugfix, and the second one is with the bugfix (I suppose that the same conclusions apply for the default PyFR FR-DG implementation). With the default v1.11.0 transition happens closer to the leading edge on the suction side than in the reference data, possibly due to the perturbations arising from the wall zones with free-stream preservation issues (see the first figure of this issue). Moreover there also exist high pressure oscillations on the lower side near zones with free-stream preservation issues.
Anyway, these figures show that the bug introduced when partitioning non-linear cells seems rather important for the accuracy of the simulations

image

image

FreddieWitherden:
> In linear cells the linear modes will always have a non-zero energy. I don't know if there are cases of high-order elements with zero energy in the linear modes, maybe they exist?
> 
> We could instead compare the energy of the non-linear modes with that of all modes (except the 0th)?

I think this should do it
```python
def _find_curved_eles(etype, eles, tol=1e-5):
    nspts, neles, ndims = eles.shape

    shape = subclass_where(BaseShape, name=etype)
    order = shape.order_from_nspts(nspts)
    basis = get_polybasis(etype, order, shape.std_ele(order - 1))

    lbasis = get_polybasis(etype, 2).degrees
    lmodes = [i for i, j in enumerate(basis.degrees) if j in lbasis and sum(j)]
    hmodes = [i for i, j in enumerate(basis.degrees) if j not in lbasis]

    emodes = basis.invvdm.T @ eles.reshape(nspts, -1)
    emodes = emodes.reshape(-1, neles, ndims)

    elnorm = np.linalg.norm(emodes[lmodes], axis=0)
    ehnorm = np.linalg.norm(emodes[hmodes], axis=0)

    return np.any(ehnorm / elnorm > tol, axis=1)
```
Let me know if this works for your cases.

GonzaloSaez:
So I've tested this approach in a uniform cube of size 2L case initially meshed with degree 2 hexa27 elements of edge size `h` and then transformed using the typical transform utilized to assess freestream preservation

`x' = x + A * h * sin(2 pi y/L) sin(2 pi z/L)` and similarly for `y'` and `z'`

In this case a value of `A = 5e-4` and `tol=1e-5` provides the threshold with which your proposed method distinguish between non-linear cells with linear shapes and non-linear cells with non-linear shapes. So now the question is whether `tol` should be lower or not.  I'll need to test this in the SD7003 test case, but there it is a bit more complicated to debug and distinguish the different non-linear elements...

FreddieWitherden:
> So I've tested this approach in a uniform cube of size 2L case initially meshed with degree 2 hexa27 elements of edge size `h`. One the initial uniform mesh is create, I modify the nodes positions using the typical transform utilized to assess freestream preservation
> 
> `x' = x + A * h * sin(2 pi y/L) sin(2 pi z/L)` and similarly for `y'` and `z'`
> 
> If `tol=1e-5`, then `A = 5e-4` provides the threshold with which your proposed method distinguishes between non-linear cells with linear shapes and non-linear cells with non-linear shapes if the aforementioned configuration.
> 
> So now the question is whether `tol` should be lower or not. I'll need to test this in the SD7003 test case, but this test case is not easy to debug as it is not trivial to know a priori if a non-linear element is curved or not. Moreover, there might be other of mesh configurations which might include edges cases which require lower values of `tol`.

Usually with Gmsh generated grids it is relatively easy to identify the curved elements (at least for extruded cases) since they appear on the boundary (first two layers) and then at the interface between the boundary layer and the far-field (one layer either side).

GonzaloSaez:
I've tried the SD7003 mesh with quadratic 137916 hex27 elements of Vermeire et. al. and your fixes. These are the findinds with different values of the tolerance. To me there doesnt seem to be an appropriate value of `tol` (apart from `tol = 1e-16` which tags all cells as non-linear...) as most elements present a certain degree of non-linearity. If those cells are tagged as linear cells, freestream preservation issues will arise. Of course the lower the `tol` the lower the spurious currents generated by the freestream preservation violation, but it is something to bare in mind in my opinion.

- tol=0.1 number of non-linear elements = 0
- tol=0.01 number of non-linear elements = 180
- tol=0.001 number of non-linear elements = 2412
- tol=0.0001 number of non-linear elements = 6972
- tol=1e-05 number of non-linear elements = 9576
- tol=1e-06 number of non-linear elements = 9588
- tol=1e-07 number of non-linear elements = 9588
- tol=1e-08 number of non-linear elements = 33456
- tol=1e-09 number of non-linear elements = 131176
- tol=1e-10 number of non-linear elements = 136680
- tol=1e-11 number of non-linear elements = 137740
- tol=1e-12 number of non-linear elements = 137740
- tol=1e-13 number of non-linear elements = 137740
- tol=1e-14 number of non-linear elements = 137892
- tol=1e-15 number of non-linear elements = 137915
- tol=1e-16 number of non-linear elements = 137916
- tol=1e-17 number of non-linear elements = 137916
- tol=1e-18 number of non-linear elements = 137916
- tol=1e-19 number of non-linear elements = 137916

FreddieWitherden:
> I've tried the SD7003 mesh with quadratic 137916 hex27 elements of Vermeire et. al. and your fixes. These are the findinds with different values of the tolerance. To me it is not straightforward to define an appropriate value of `tol` (apart from `tol = 1e-16` which tags all cells as non-linear...) as most elements present a certain degree of non-linearity. If those cells are tagged as linear cells, freestream preservation issues will arise. Of course, the lower the `tol` the lower the spurious currents generated by the freestream preservation violation, but it is something to bare in mind in my opinion. I don't know if you think that there are other types of test which can help us find an appropirate value of `tol`?
> 
>     * tol=0.1 number of non-linear elements = 0
> 
>     * tol=0.01 number of non-linear elements = 180
> 
>     * tol=0.001 number of non-linear elements = 2412
> 
>     * tol=0.0001 number of non-linear elements = 6972
> 
>     * tol=1e-05 number of non-linear elements = 9576
> 
>     * tol=1e-06 number of non-linear elements = 9588
> 
>     * tol=1e-07 number of non-linear elements = 9588
> 
>     * tol=1e-08 number of non-linear elements = 33456
> 
>     * tol=1e-09 number of non-linear elements = 131176
> 
>     * tol=1e-10 number of non-linear elements = 136680
> 
>     * tol=1e-11 number of non-linear elements = 137740
> 
>     * tol=1e-12 number of non-linear elements = 137740
> 
>     * tol=1e-13 number of non-linear elements = 137740
> 
>     * tol=1e-14 number of non-linear elements = 137892
> 
>     * tol=1e-15 number of non-linear elements = 137915
> 
>     * tol=1e-16 number of non-linear elements = 137916
> 
>     * tol=1e-17 number of non-linear elements = 137916
> 
>     * tol=1e-18 number of non-linear elements = 137916
> 
>     * tol=1e-19 number of non-linear elements = 137916

One thing we should probably be doing is actually linearising elements which are marked as being linear.  This should eliminate any possibility for trouble given that some of the code paths use the free-stream metric terms whilst others (like the flux kernel) will use the regular metric for elements which are tagged as linear.

To give a bit of background on why it is important to know which elements are linear.  For high-order overset a major cost is interpolating solutions between grids since you need to continually map from physical space to reference space in order to perform the lookup.  For linear elements this is trivial.  For non linear elements it involves costly Newton iterations.

GonzaloSaez:
Yes I understand the reasoning behind all this. What you are trying to say
is that once we have defined nonlinear cells with linear shapes, we should
then remove them from their original group of non linear cells and add them
to the linear cells group, removing the high order points from their
cell-to-node connectivity?

FreddieWitherden:
A preliminary fix is now in #200

FreddieWitherden:
Discussion continues in #200.

PR #200, opened 2021-02-25
 Rework how linear elements are identified.

FreddieWitherden:
The key change here is that nodepts is now a dense NumPy array as opposed to a dict.  This will be important for robustly resolving #198

### DIFF
--- a/pyfr/partitioners/base.py
+++ b/pyfr/partitioners/base.py
@@ -6,28 +6,10 @@ import uuid
 
 import numpy as np
 
-from pyfr.polys import get_polybasis
-from pyfr.shapes import BaseShape
-from pyfr.util import subclass_where
+from pyfr.inifile import Inifile
 
 
 Graph = namedtuple('Graph', ['vtab', 'etab', 'vwts', 'ewts'])
-
-
-def _find_curved_eles(etype, eles, tol=1e-5):
-    nspts, neles, ndims = eles.shape
-
-    shape = subclass_where(BaseShape, name=etype)
-    order = shape.order_from_nspts(nspts)
-    basis = get_polybasis(etype, order, shape.std_ele(order - 1))
-
-    lmodes = get_polybasis(etype, 2).degrees
-    hmodes = [i for i, j in enumerate(basis.degrees) if j not in lmodes]
-
-    ehmodes = basis.invvdm.T[hmodes] @ eles.reshape(nspts, -1)
-    ehmodes = ehmodes.reshape(-1, neles, ndims)
-
-    return np.any(np.abs(ehmodes) > tol, axis=(0, 2))
 
 
 class BasePartitioner(object):
@@ -66,8 +48,9 @@ class BasePartitioner(object):
         # Get the per-partition element counts
         pinf = mesh.partition_info('spt')
 
-        # Shape points, element number offsets, and remapping table
+        # Shape points, linear flags, element offsets, and remapping table
         spts = defaultdict(list)
+        linf = defaultdict(list)
         offs = defaultdict(dict)
         rnum = defaultdict(dict)
 
@@ -77,6 +60,11 @@ class BasePartitioner(object):
                     offs[en][i] = off = sum(s.shape[1] for s in spts[en])
                     spts[en].append(mesh[f'spt_{en}_p{i}'])
                     rnum[en].update(((i, j), (0, off + j)) for j in range(n))
+
+                    try:
+                        linf[en].append(mesh[f'spt_{en}_p{i}', 'linear'])
+                    except KeyError:
+                        linf[en].append(np.full(spts[en][-1].shape[1], False))
 
         def offset_con(con, pr):
             con = con.copy().astype('U4,i4,i1,i2')
@@ -116,19 +104,20 @@ class BasePartitioner(object):
         # Concatenate these arrays to from the new mesh
         newmesh = {'con_p0': np.hstack(intcon).astype(dtype)}
 
-        for k, v in spts.items():
-            newmesh[f'spt_{k}_p0'] = np.hstack(v)
+        for en in spts:
+            newmesh[f'spt_{en}_p0'] = np.hstack(spts[en])
+            newmesh[f'spt_{en}_p0', 'linear'] = np.hstack(linf[en])
 
         for k, v in bccon.items():
             newmesh[f'bcon_{k}_p0'] = np.hstack(v).astype(dtype)
 
         return newmesh, rnum
 
-    def _combine_soln_parts(self, soln):
+    def _combine_soln_parts(self, soln, prefix):
         newsoln = defaultdict(list)
 
-        for f, (en, shape) in soln.array_info('soln').items():
-            newsoln[f'soln_{en}_p0'].append(soln[f])
+        for f, (en, shape) in soln.array_info(prefix).items():
+            newsoln[f'{prefix}_{en}_p0'].append(soln[f])
 
         newsoln = {k: np.dstack(v) for k, v in newsoln.items()}
         newsoln['config'] = soln['config']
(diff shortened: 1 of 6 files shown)


FreddieWitherden:
Items to discuss: what should the default tolerance be and should we make it configurable?  Also, as this changes the mesh file format a fair bit of testing will be needed.

pyfr/partitioners/base.py
```diff
@@ -340,21 +299,21 @@ def partition(self, mesh):
...
         # Build the solution converter
         def partition_soln(soln):
             # Check the UUID
             if curruuid != soln['mesh_uuid']:
                 raise ValueError('Mismatched solution/mesh')
 
+            # Obtain the prefix
+            prefix = Inifile(soln['stats']).get('data', 'prefix')
+
             # Combine any pre-existing parititons
```
GonzaloSaez, line 311: parititons -> partitions (there is another similar mistake) in line 273

pyfr/partitioners/base.py
```diff
@@ -340,21 +299,21 @@ def partition(self, mesh):
...
-        self._prepare_attrs(newmesh)
-
         # Build the solution converter
         def partition_soln(soln):
             # Check the UUID
             if curruuid != soln['mesh_uuid']:
                 raise ValueError('Mismatched solution/mesh')
 
+            # Obtain the prefix
+            prefix = Inifile(soln['stats']).get('data', 'prefix')
```
GonzaloSaez, line 309: If I understood correctly, this should to allow repartition files with temporal average data isn't it? That's I nice addition

GonzaloSaez:
I tested the number of non-linear elements tagged as non-linear as a function of the tolerance in the SD7003 mesh of Vermeire et al.

```
tolerance=1e-05 Total of 137904 vs 137916 tagged as linear
tolerance=1e-06 Total of 137904 vs 137916 tagged as linear
tolerance=1e-07 Total of 137904 vs 137916 tagged as linear
tolerance=1e-08 Total of 137904 vs 137916 tagged as linear
tolerance=1e-09 Total of 72576 vs 137916 tagged as linear
tolerance=1e-10 Total of 58968 vs 137916 tagged as linear
tolerance=1e-11 Total of 56744 vs 137916 tagged as linear
tolerance=1e-12 Total of 56744 vs 137916 tagged as linear
tolerance=1e-13 Total of 56744 vs 137916 tagged as linear
tolerance=1e-14 Total of 54904 vs 137916 tagged as linear
tolerance=1e-15 Total of 50384 vs 137916 tagged as linear
```

So the new proposed method is less sensitive to the tolerance value (I suppose that this is because you are now checking based on node coordinates and not based on modal coeffs). I'll test freestream preservation vs tolerance value and I'll come back soon with a bit more info of the analysis

pyfr/partitioners/base.py
```diff
@@ -183,69 +168,43 @@ def _renumber_verts(self, mesh, vetimap, vparts):
...
-        # Partition the solutions
+            newmesh[f] = np.array(spt_px[etype, pn]).swapaxes(0, 1)
+            newmesh[f, 'linear'] = np.array(lin_px[etype, pn])
+
+        return newmesh
+
+    def _partition_soln(self, soln, prefix, vetimap, vparts):
         soln_px = defaultdict(list)
         for (etype, eidxg), part in zip(vetimap, vparts):
             soln_px[etype, part].append(soln_p0[etype][..., eidxg])
```
GonzaloSaez, line 205: `soln_p0` -> `soln` ?

pyfr/solvers/base/system.py
```diff
@@ -111,7 +113,8 @@ def _load_eles(self, rallocs, mesh, initsoln, nregs, nonce):
             k = f'spt_{etype}_p{rallocs.prank}'
 
             try:
```
GonzaloSaez, line 115: I have a question here, (sorry if I'm mistaken I think we have already talked abot this) but it seems that the computation of `linoff` supposes that the elements are ordered based on if they are `linear` or not. However, this ordering is only imposed when partitioning the mesh object. So if no partition is done, all elements will be treated as `curved` isn't it?

pyfr/solvers/base/system.py
```diff
@@ -111,7 +113,8 @@ def _load_eles(self, rallocs, mesh, initsoln, nregs, nonce):
...
-                linoff = mesh[k, 'lin_off']
+                curved = ~mesh[k, 'linear']
```
GonzaloSaez, line 116: I have a question here, (sorry if I'm mistaken I think we have already talked abot this) but it seems that the computation of `linoff` supposes that the elements are ordered based on if they are `linear` or not. However, this ordering is only imposed when partitioning the mesh object. So if no partition is done, all elements will be treated as `curved` isn't it?

pyfr/solvers/base/system.py
```diff
@@ -111,7 +113,8 @@ def _load_eles(self, rallocs, mesh, initsoln, nregs, nonce):
...
+                linoff = np.max(*np.nonzero(curved), initial=-1) + 1
             except KeyError:
```
GonzaloSaez, line 118: I have a question here, (sorry if I'm mistaken I think we have already talked abot this) but it seems that the computation of `linoff` supposes that the elements are ordered based on if they are `linear` or not. However, this ordering is only imposed when partitioning the mesh object. So if no partition is done, all elements will be treated as `curved` isn't it?

pyfr/partitioners/base.py
FreddieWitherden: Yes.

pyfr/solvers/base/system.py
FreddieWitherden: Correct, this avoids the need for the importer to output a remapping table (as it would if it went and reordered elements).

FreddieWitherden:
Thanks! Fixed.

semi-h:
Can we add an optional argument to the partitioner so that when it is present all elements are assumed curved?

FreddieWitherden:
> Can we add an optional argument to the partitioner so that when it is present all elements are assumed curved?

We could expose a tolerance parameter to the mesh importer (as this is where linear elements are now identified).

GonzaloSaez:
I like the idea of using a tolerance to linearize but I think that it is difficult for a user to understand which is the value of the tolerance to be set as there is no easy way of knowing how many and which cells are being linearized. This can be assessed with a simple a posteriori script so it shouldn't be a big deal.

At last, which should be the default value of `tol`? For example, with the default tolerance `1e-5`, in the SD7003 mesh of Vermeire et al., only 12 elements are tagged as non-linear elements. It is not easy to say if that's OK or not...

FreddieWitherden:
> I like the idea of using an input tolerance to linearize but I think that it is difficult for a user to understand which is the value of the tolerance to be set as there is no easy way of knowing how many and which cells are being linearized. This can be assessed with a simple a posteriori script so it shouldn't be a big deal.
> 
> At last, which should be the default value of `tol`? For example, with the default tolerance `1e-5`, in the SD7003 mesh of Vermeire et al., only 12 elements are tagged as non-linear elements. It is not easy to say if that's OK or not...

The current tolerance, assuming I've implemented everything correctly, works out to a relative difference of 1 part in 100,000.  This is pretty small, and at single precision, is not too far off the smallest difference you can resolve.  A key aspect of the new approach is that beyond tagging elements, it also shifts the nodes for elements which are deemed to be linear (and does so using the global node map).  Thus I do not think that there can be any free-stream issues (assuming there are no bugs elsewhere).

GonzaloSaez:
Sounds good! I have done several tests with highly deformed grids made up of high-order elements and I haven't found any freestream issues with the provided tolerance. Of course, if I increase the tolerance by some orders of magnitude, issues arise but that's expected. Anyway, the default `tol` seems OK!

FreddieWitherden:
Great.  Would be good to give this a little bit more testing before merging.

Also, at the moment it is not possible to repartition old solution files without first reimporting the mesh.  Old solutions will run, however.

GonzaloSaez:
> Great. Would be good to give this a little bit more testing before merging.
> 
> Also, at the moment it is not possible to repartition old solution files without first reimporting the mesh. Old solutions will run, however.

Repartition of old solution and mesh files could be performed supposing that all elements are curved? I have done some modifications to the partitioner to assess if the `linear` key exists in the mesh file and if not treat all elements as curved and it seems to work, at least in serial. I'm sure I'm missing something though.

Have you considered using Github Actions? It might be worth establishing some small tests (comparing solution/mesh files) to be sure that everything works as expected in each PR (although it might it may cost some money depending on the number of tests and number of PR per month).

pyfr/partitioners/base.py
```diff
@@ -76,6 +59,7 @@ def _combine_mesh_parts(self, mesh):
                 if n > 0:
                     offs[en][i] = off = sum(s.shape[1] for s in spts[en])
                     spts[en].append(mesh[f'spt_{en}_p{i}'])
+                    linf[en].append(mesh[f'spt_{en}_p{i}', 'linear'])
```
GiorgioGiangaspero, line 62: How about this to make sure we can still partition old meshes?
```
                    # Assume all non-linear elements if the attribute linear
                    # is not present. This ensure backwards compatibility.
                    try:
                        linf[en].append(mesh[f'spt_{en}_p{i}', 'linear'])
                    except KeyError:
                        flags = np.full(mesh[f'spt_{en}_p{i}'].shape[1], False)
                        linf[en].append(flags)
```

FreddieWitherden: I'll take this for 1.12, but pending removal in 1.13.

FreddieWitherden:
I think everything is in place.  Is everyone okay for me to merge?

PR #203, opened 2021-03-20
Improve interface sorting.

FreddieWitherden:


pyfr/backends/base/types.py
```diff
@@ -41,18 +41,14 @@ def __init__(self, backend, dtype, ioshape, initval, extent, aliases,
             leaddim = nvar*csubsz if backend.blocks else ncol
 
         # Assign
-        self.nrow, self.ncol, self.leaddim = int(nrow), int(ncol), int(leaddim)
+        self.nrow, self.ncol, self.leaddim = nrow, ncol, leaddim
 
-        self.datashape = aosoashape
-        self.ioshape = ioshape
```
semi-h, line 47: datashape is used later in _unpack, and ioshape is used in various places such as _unpack and blasext functions.

semi-h (approved):

PR #206, opened 2021-03-28
Cleanups.

FreddieWitherden:


semi-h:
Are you sure about merging into develop?

FreddieWitherden:
Yes, as I'd like to keep the upcoming release compatible with older versions of Python.

semi-h (approved):

PR #210, opened 2021-04-05
Cleanups.

FreddieWitherden:


pyfr/backends/openmp/kernels/axnpby.mako
```diff
@@ -11,31 +11,27 @@ axnpby(int nrow, int nblocks,
...
     for (int ib = 0; ib < nblocks; ib++)
     {
-        int idx, blksz = nrow*BLK_SZ*${ncola};
         #pragma omp simd
-        for (int i = 0; i < blksz; i++)
-        {
-            idx = i + ib*blksz;
-            x0[idx] = ${pyfr.dot('a{l}', 'x{l}[idx]', l=nv)};
-        }
+        for (int i = ib*nrow*BLK_SZ*${ncola}; i < (ib + 1)*nrow*BLK_SZ*${ncola}; i++)
```
WillTrojak, line 15: given that `ib` is always multiplied by nrow in this loop, could you change the in for loop stride?

pyfr/backends/openmp/kernels/axnpby.mako
```diff
@@ -11,31 +11,27 @@ axnpby(int nrow, int nblocks,
...
             {
                 #pragma omp simd
                 for (int _xj = 0; _xj < SOA_SZ; _xj++)
                 {
+                    int i;
+
                 % for k in subdims:
-                    idx = _y*BLK_SZ*${ncola} + ib*BLK_SZ*${ncola}*nrow + X_IDX_AOSOA(${k}, ${ncola});
-                    x0[idx] = ${pyfr.dot('a{l}', 'x{l}[idx]', l=nv)};
+                    i = _y*BLK_SZ*${ncola} + ib*BLK_SZ*${ncola}*nrow + X_IDX_AOSOA(${k}, ${ncola});
```
WillTrojak, line 33: I think you can do a similar thing with `ib` in here.

pyfr/backends/openmp/kernels/axnpby.mako
FreddieWitherden: I think that would work, although the variable would be something of an oddball as it would not represent the block number or the total number of elements in a block.

semi-h (approved):

PR #214, opened 2021-04-11
Avoid using non-temporal stores for small matrices in libxsmm.

FreddieWitherden:


semi-h (approved):
https://github.com/PyFR/PyFR/blob/develop/pyfr/backends/base/types.py#L175

Column sliced matrices do not have the nbytes attribute. Currently we don't call a mul kernel over column sliced matrices, but we used to in the past with _int _ext stuff. Just wanted to leave a note here for future reference.

PR #216, opened 2021-04-14
Python 3.10 support

FreddieWitherden:
This PR migrates away from various deprecated interfaces which are going away in Python 3.9 and future NumPy releases.


semi-h (approved):
works with Python 3.9.2

PR #217, opened 2021-04-16
Remove max matrix size limitation for libxsmm.

FreddieWitherden:


semi-h:
On my workstation (no AVX-512) gimmik performs better than the latest libxsmm master branch. And this ini file option make things easy for me when I want to switch between gimmik and libxsmm for testing.

FreddieWitherden:
> On my workstation (no AVX-512) gimmik performs better than the latest libxsmm master branch. And this ini file option make things easy for me when I want to switch between gimmik and libxsmm for testing.

What CPU do you have in your workstation?

semi-h:
> > On my workstation (no AVX-512) gimmik performs better than the latest libxsmm master branch. And this ini file option make things easy for me when I want to switch between gimmik and libxsmm for testing.
> 
> What CPU do you have in your workstation?

Intel(R) Core(TM) i7-4820K CPU @ 3.70GHz

FreddieWitherden:
> Intel(R) Core(TM) i7-4820K CPU @ 3.70GHz

Not only does that chip not support AVX-512 it also does not even support AVX2 (so no FMA or vector gather instructions).  The libxsmm support for sparse matrices only goes back to Haswell class CPUs (which are ~8 years old now!).  Performance on what are effectively museum pieces should not be a priority.

semi-h:
As we removed cblas support for openmp, now gimmik is the last resort for the mul kernel provider. But we still have a max-nnz criteria for gimmik. So if libxsmm is not available people running pyfr with openmp backend will see 'mul kernel is not available' error for some cases. I'm sure we'll get some questions on discourse on this issue in near future if we don't fix this.
The performance will be poor indeed for certain cases if we allow gimmik to do dense matrix multiplications, but otherwise pyfr won't run. What do you think we should do?

semi-h:
And with cblas there was this ini file option to show the location of the cblas installation. Can we do the same with libxsmm and avoid adding libxsmm library location to LD_LIBRARY_PATH? If we can do this then it will solve the problem I'd have about switching between libxsmm and gimmik.

FreddieWitherden:
> As we removed cblas support for openmp, now gimmik is the last resort for the mul kernel provider. But we still have a max-nnz criteria for gimmik. So if libxsmm is not available people running pyfr with openmp backend will see 'mul kernel is not available' error for some cases. I'm sure we'll get some questions on discourse on this issue in near future if we don't fix this.
> The performance will be poor indeed for certain cases if we allow gimmik to do dense matrix multiplications, but otherwise pyfr won't run. What do you think we should do?

Yes, this is the expected outcome.  If libxsmm is not available we'll try GiMMiK, but if the number of non zeros is too large for it to deliver good performance, then we'll error out.  Here the user can either manually increase the number of non-zeros allowed by GiMMiK (although the performance may be suspect) or get libxsmm installed.

Even with cblas as a last resort we often used to get questions along these lines as people would not set the .ini file correctly and so would see mul kernel not available errors.

FreddieWitherden:
> And with cblas there was this ini file option to show the location of the cblas installation. Can we do the same with libxsmm and avoid adding libxsmm library location to LD_LIBRARY_PATH? If we can do this then it will solve the problem I'd have about switching between libxsmm and gimmik.

We have this functionality already if PYFR_XSMM_LIBRARY_PATH is set we'll try and load libxsmm from the path given.  Indeed, this is not unique to libxsmm but for all libraries loaded by PyFR (metis, scotch, clblast, hip, cuda, ...) all of which can have their paths provided explicitly through an environmental variable.

semi-h:
> > And with cblas there was this ini file option to show the location of the cblas installation. Can we do the same with libxsmm and avoid adding libxsmm library location to LD_LIBRARY_PATH? If we can do this then it will solve the problem I'd have about switching between libxsmm and gimmik.
> 
> We have this functionality already if PYFR_XSMM_LIBRARY_PATH is set we'll try and load libxsmm from the path given. Indeed, this is not unique to libxsmm but for all libraries loaded by PyFR (metis, scotch, clblast, hip, cuda, ...) all of which can have their paths provided explicitly through an environmental variable.

Actually I mean this bit:
https://github.com/PyFR/PyFR/blob/7529c8a522db2b41334ed9e18d0d2909c89f8bda/pyfr/backends/openmp/cblas.py#L53
and
https://github.com/PyFR/PyFR/blob/7529c8a522db2b41334ed9e18d0d2909c89f8bda/pyfr/backends/openmp/cblas.py#L26
If I remember well if cblas path (L53) was not given in the ini file the library wouldn't load and alternative mul provider would be used.

FreddieWitherden:
So the cblas integration was a little different to other libraries as there is no standard name/location for such libraries.  Moreover, it is also important to know if the library is serial or parallel.  As such we had to use the .ini file, although this was always sub-optimal and could be a pain when restarting a simulation on a different system.

Medium term I'm looking to remove GiMMiK support entirely on CPUs and make libxsmm a required library.  The only blocker here is the current lack of support for sparse kernels on ARM CPUs (dense is already supported).  Once this is in a public release we can migrate entirely to libxsmm which should give a better experience all around.

semi-h (approved):

PR #218, opened 2021-04-18
Vectorise polys.py.

FreddieWitherden:


pyfr/polys.py
```diff
@@ -10,7 +9,7 @@
 
 
 def jacobi(n, a, b, z):
```
WillTrojak, line 11: should there not some checks to see if a and b are valid of Jacobi polynomials? Also, some defaults for a=0, and b=0 would probably be good.

pyfr/polys.py
```diff
@@ -10,7 +9,7 @@
...
-    j = [1]
+    j = [np.ones_like(z)]
 
     if n >= 1:
         j.append(((a + b + 2)*z + a - b) / 2)
```
WillTrojak, line 15: the remainder of this function is quite unclear. Is the single line saving of using `u,y = x,y` really worth the loss in clarity?
Also I think you calculate Jacobi polynomial without needing to use a three-term recession, so would remove some of the ifs. Here is a snippet from a Fortran implementation.
```
      J = 0.
      do m=0,n
         c = choose(n, m)*gamma(a + b + n + m + 1)/gamma(a + m + 1)
         J = J + c*((0.5*(x-1))**m)
      enddo
      J = J*gamma(a+n+1)/gamma(a+b+n+1)/fac(n)
```

pyfr/polys.py
```diff
@@ -32,7 +31,7 @@ def jacobi(n, a, b, z):
 
 
 def jacobi_diff(n, a, b, z):
```
WillTrojak, line 33: Again, I think you might need some checks on the validity of a and b, and some defaults would be good.

pyfr/polys.py
```diff
@@ -165,14 +162,14 @@ def degrees(self):
 class QuadPolyBasis(BasePolyBasis):
     name = 'quad'
 
-    def ortho_basis_at_py(self, p, q):
+    def _ortho_basis_at(self, p, q):
```
WillTrojak, line 165: Some comments on what p, q and r are would be good. It is not immediately clear what they are.

WillTrojak (commented):
What is the difference in performance this gives? Does it improve things?

pyfr/polys.py
FreddieWitherden: The function returns a list of all of the Jacobi polynomials from 0..n.  This works nicely as whenever we need the n'th we also need all of the lower degree polynomials too.  Thanks to three term recurrence relation these are effectively free.  It is also reasonably numerically stable and, as a minor benefit, avoids the need to evaluate any gamma functions.

However, due to the recurrence the conditionals are required to correctly handle start-up.

pyfr/polys.py
FreddieWitherden: Defaulting would be difficult as it would force a and b to be at the end which would be somewhat non-standard.

WillTrojak: Ok, I still think you should add tests for the validity of a and b though

semi-h: I think the below is valid;
`def jacobi_diff(n, a=0, b=0, *, z):`
but this forces 'z' to be a keyword only argument which might be inconvenient.

pyfr/polys.py
WillTrojak: Fair enough.

PR #219, opened 2021-04-25
Integrate plugin enhancements (quad-deg, quad-pts and access to tcurr)

GonzaloSaez:
1) Allow to access `tcurr` using `t` within the expressions of the integrate plugin (useful to compute errors with respect to analytical solutions)

2) Allow to choose the quadrature degree used to evaluate the expressions in the plugin integrate. To do so, the token `quad-deg` needs to be added to the input file in the integrate plugin section. The default quadrature type is taken from that used to set the solution points. If the user adds `quad-pts-{etype} = quad_name` to the integrate plugin section, then the the `quad_name quadrature will be used`.

Example:

```
[soln-plugin-integrate]
nsteps = 10
quad-deg = 10 ; quadrature degree
quad-pts-tet = witherden-vincent ; quadrature type
file = integrate
int-rho = rho
```

### DIFF
--- a/pyfr/plugins/integrate.py
+++ b/pyfr/plugins/integrate.py
@@ -129,6 +129,7 @@ class IntegratePlugin(BasePlugin):
             # Prepare the substitutions dictionary
             subs = dict(zip(pnames, psolns))
             subs.update(zip('xyz', plocs))
+            subs.update({'t': intg.tcurr})
 
             # Prepare any required gradients
             if self._gradpinfo:
--- a/pyfr/plugins/integrate.py
+++ b/pyfr/plugins/integrate.py
@@ -32,11 +32,6 @@ class IntegratePlugin(BasePlugin):
                       for k in self.cfg.items(cfgsect)
                       if k.startswith('int-')]
 
-        # Select a quadrature different from that of the
-        # solution points if using the token qdeg
-        qdeg = self.cfg.getint(cfgsect, 'quad-deg', -1)
-        self.qdeg = qdeg if qdeg != -1 else None
-
         # Integration region pre-processing
         rinfo = self._prepare_region_info(intg)
 
@@ -59,20 +54,20 @@ class IntegratePlugin(BasePlugin):
 
         # Prepare the per element-type info list
         self.eleinfo = []
-        self.m0s = []
         for (ename, eles), (eset, emask) in zip(system.ele_map.items(), rinfo):
             # Obtain quadrature info
             rname = self.cfg.get(f'solver-elements-{ename}', 'soln-pts')
-            if self.qdeg:
+            m0 = None
+            if self.cfg.hasopt(cfgsect, 'quad-deg'):
                 if self.cfg.hasopt(cfgsect, f'quad-pts-{ename}'):
                     rname = self.cfg.get(cfgsect, f'quad-pts-{ename}')
 
                 # Quadrature rule
-                r = get_quadrule(ename, rname, qdeg=self.qdeg)
+                qdeg = self.cfg.getint(cfgsect, 'quad-deg')
+                r = get_quadrule(ename, rname, qdeg=qdeg)
 
                 # Solution interpolation matrix to quadrature points
                 m0 = eles.basis.ubasis.nodal_basis_at(r.pts)
-                self.m0s.append(m0)
             else:
                 # Default to the quadrature rule of the solution points
                 r = get_quadrule(ename, rname, eles.nupts)
@@ -84,7 +79,8 @@ class IntegratePlugin(BasePlugin):
             rcpdjacs = eles.rcpdjac_at_np(r.pts)[:, eset]
 
             # Save
-            self.eleinfo.append((ploc, r.wts[:, None] / rcpdjacs, eset, emask))
+            self.eleinfo.append((ploc, r.wts[:, None] / rcpdjacs, m0, eset,
+                                 emask))
 
     def _prepare_region_info(self, intg):
         # All elements
@@ -135,14 +131,14 @@ class IntegratePlugin(BasePlugin):
 
         # Iterate over each element type in the simulation
         for i, (soln, eleinfo) in enumerate(zip(intg.soln, self.eleinfo)):
-            plocs, wts, eset, emask = eleinfo
+            plocs, wts, m0, eset, emask = eleinfo
 
             # Subset and transpose the solution
             soln = soln[..., eset].swapaxes(0, 1)
 
             # Interpolate the solution to the quadrature points
-            if self.qdeg:
-                soln = self.m0s[i] @ soln
+            if m0 is not None:
+                soln = m0 @ soln
 
             # Convert from conservative to primitive variables
             psolns = self.elementscls.con_to_pri(soln, self.cfg)
@@ -157,8 +153,8 @@ class IntegratePlugin(BasePlugin):
                 grad_soln = np.rollaxis(intg.grad_soln[i], 2)[..., eset]
 
                 # Interpolate the gradients to the quadrature points
-                if self.qdeg:
-                    grad_soln = self.m0s[i] @ grad_soln
+                if m0 is not None:
+                    grad_soln = m0 @ grad_soln
 
                 # Transform from conservative to primitive gradients
                 pgrads = self.elementscls.grad_con_to_pri(soln, grad_soln,
--- a/pyfr/plugins/integrate.py
+++ b/pyfr/plugins/integrate.py
@@ -4,6 +4,7 @@ import re
 
 import numpy as np
 
+from pyfr.inifile import NoOptionError
 from pyfr.mpiutil import get_comm_rank_root, get_mpi
 from pyfr.nputil import npeval
 from pyfr.plugins.base import BasePlugin, init_csv
@@ -52,36 +53,30 @@ class IntegratePlugin(BasePlugin):
             # Open
             self.outf = init_csv(self.cfg, cfgsect, ','.join(header))
 
-        # Whether to use a quadrature different to that of the solution points
-        quad = self.cfg.hasopt(cfgsect, 'quad-deg')
-
         # Prepare the per element-type info list
         self.eleinfo = eleinfo = []
         for (ename, eles), (eset, emask) in zip(system.ele_map.items(), rinfo):
             # Obtain quadrature info
             rname = self.cfg.get(f'solver-elements-{ename}', 'soln-pts')
 
-            # Raise error with edge cases
-            if not quad and (self.cfg.hasopt(cfgsect, f'quad-pts-{ename}')
-                             or self.cfg.hasopt(cfgsect, f'quad-deg-{ename}')):
-                raise ValueError('Option "quad-deg" is not found')
-
-            if quad:
+            try:
                 # Quadrature rule (default to that of the solution points)
                 qrule = self.cfg.get(cfgsect, f'quad-pts-{ename}', rname)
 
                 # Quadrature rule degree
-                qdeg = self.cfg.getint(cfgsect, 'quad-deg')
-                if self.cfg.hasopt(cfgsect, f'quad-deg-{ename}'):
+                try:
                     qdeg = self.cfg.getint(cfgsect, f'quad-deg-{ename}')
+                except NoOptionError:
+                    qdeg = self.cfg.getint(cfgsect, 'quad-deg')
 
                 r = get_quadrule(ename, qrule, qdeg=qdeg)
-            else:
+
+                # Interpolation to quadrature points matrix
+                m0 = eles.basis.ubasis.nodal_basis_at(r.pts)
+            except NoOptionError:
                 # Default to the quadrature rule of the solution points
                 r = get_quadrule(ename, rname, eles.nupts)
-
-            # Interpolation to quadrature points matrix
-            m0 = eles.basis.ubasis.nodal_basis_at(r.pts) if quad else None
+                m0 = None
 
             # Locations of each quadrature point
             ploc = eles.ploc_at_np(r.pts).swapaxes(0, 1)[..., eset]
--- a/pyfr/plugins/integrate.py
+++ b/pyfr/plugins/integrate.py
@@ -63,46 +63,28 @@ class IntegratePlugin(BasePlugin):
         for (ename, eles), (eset, emask) in zip(system.ele_map.items(), rinfo):
             # Obtain quadrature info
             rname = self.cfg.get(f'solver-elements-{ename}', 'soln-pts')
-
             if self.qdeg:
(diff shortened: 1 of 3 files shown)


FreddieWitherden:
I'd try and avoid configuring it through a single qdeg parameter as this makes it difficult to specify tensor product rules (whose order on paper is low, but because they can integrate all combinations of that order tend to be quite effective).  Perhaps we should use the existing per-element mechanism?

pyfr/plugins/integrate.py
```diff
@@ -54,17 +59,43 @@ def __init__(self, intg, cfgsect, suffix=None):
...
+                wts = r.wts
+
+                # Solution interpolation matrix to quadrature points
+                m0 = eles.basis.ubasis.nodal_basis_at(r.pts)
+                self.m0s.append(m0)
+
+                # Locations of each quadrature point
+                op = eles.basis.sbasis.nodal_basis_at(r.pts)
+                ploc = op @ eles.eles.reshape(eles.nspts, -1)
+                ploc = ploc.reshape(-1, eles.neles, eles.ndims).swapaxes(1, 2)
```
FreddieWitherden, line 82: .transpose can avoid a double swapaxes

pyfr/plugins/integrate.py
```diff
@@ -54,17 +59,43 @@ def __init__(self, intg, cfgsect, suffix=None):
...
         for (ename, eles), (eset, emask) in zip(system.ele_map.items(), rinfo):
-            # Locations of each solution point
-            ploc = eles.ploc_at_np('upts')[..., eset]
-            ploc = ploc.swapaxes(0, 1)
+            # Obtain quadrature info
+            rname = self.cfg.get(f'solver-elements-{ename}', 'soln-pts')
 
-            # Jacobian determinants
-            rcpdjacs = eles.rcpdjac_at_np('upts')[:, eset]
+            if self.qdeg:
```
FreddieWitherden, line 67: Does this branch handle element subsetting correctly?

pyfr/plugins/integrate.py
```diff
@@ -123,18 +154,27 @@ def _eval_exprs(self, intg):
...
 
+            # Interpolate the solution to the quadrature points
+            if self.qdeg:
+                soln = self.m0s[i] @ soln
+
             # Convert from conservative to primitive variables
             psolns = self.elementscls.con_to_pri(soln, self.cfg)
 
             # Prepare the substitutions dictionary
             subs = dict(zip(pnames, psolns))
```
FreddieWitherden, line 165: dict(zip(pnames, psolns), t=intg.tcurr)

pyfr/plugins/integrate.py
GonzaloSaez: No it didn't. Fixed it with the new commit.

GonzaloSaez:
> I'd try and avoid configuring it through a single qdeg parameter as this makes it difficult to specify tensor product rules (whose order on paper is low, but because they can integrate all combinations of that order tend to be quite effective). Perhaps we should use the existing per-element mechanism?

If the mesh contains different elements then the user would need to add to the `[solver-elements-*]` sections the quadrature degree of the integrate plugin, which does not seem for me too intuitive and it may be clash a bit with the qdeg option of anti-aliasing methods. Anyhow, I understand that it might be more interesting to select a quadrature degree for each element type, so if you want to rely on the latter approach I'll implement it. Which would be token that you want to use to select this quadrature degree? Would it also be interesting to use a token to specify the quadrature rule name in the `solver-elements` section?

FreddieWitherden:
A while ago I coded up a similar extension.  While I can not find the full patch, I was able to dig up one of the extensions I made.
```
diff --git a/pyfr/solvers/base/elements.py b/pyfr/solvers/base/elements.py
index 5fa4440..ffab78d 100644
--- a/pyfr/solvers/base/elements.py
+++ b/pyfr/solvers/base/elements.py
@@ -250,7 +250,8 @@ class BaseElements(object):
_, djacs_mpts = self._smats_djacs_mpts

# Interpolation matrix to pts
-        m0 = self.basis.mbasis.nodal_basis_at(getattr(self.basis, name))
+        pt = getattr(self.basis, name) if isinstance(name, str) else name
+        m0 = self.basis.mbasis.nodal_basis_at(pt)

# Interpolate the djacs
djac = m0 @ djacs_mpts
@@ -267,7 +268,8 @@ class BaseElements(object):

@memoize
def ploc_at_np(self, name):
-        op = self.basis.sbasis.nodal_basis_at(getattr(self.basis, name))
+        pt = getattr(self.basis, name) if isinstance(name, str) else name
+        op = self.basis.sbasis.nodal_basis_at(pt)

ploc = op @ self.eles.reshape(self.nspts, -1)
ploc = ploc.reshape(-1, self.neles, self.ndims).swapaxes(1, 2)
```
The purpose was to allow ploc_at_np and rcpjdac_at_np to accept arbitrary arrays of points in lieu of a named set.

FreddieWitherden:
I'd make m0 an extra member in the eleinfo tuple (defaulting to None if it is not needed).  This avoids the need for self.qdeg and self.m0.

pyfr/plugins/integrate.py
```diff
@@ -55,19 +55,32 @@ def __init__(self, intg, cfgsect, suffix=None):
         # Prepare the per element-type info list
         self.eleinfo = []
         for (ename, eles), (eset, emask) in zip(system.ele_map.items(), rinfo):
-            # Locations of each solution point
-            ploc = eles.ploc_at_np('upts')[..., eset]
-            ploc = ploc.swapaxes(0, 1)
+            # Obtain quadrature info
+            rname = self.cfg.get(f'solver-elements-{ename}', 'soln-pts')
+            m0 = None
```
FreddieWitherden, line 60: I'd move the assignment down to the else branch.

pyfr/plugins/integrate.py
```diff
@@ -55,19 +55,32 @@ def __init__(self, intg, cfgsect, suffix=None):
         # Prepare the per element-type info list
         self.eleinfo = []
```
FreddieWitherden, line 56: self.eleinfo = eleinfo = [] may enable you to avoid the line break in self.eleinfo.append

pyfr/plugins/integrate.py
```diff
@@ -53,21 +53,33 @@ def __init__(self, intg, cfgsect, suffix=None):
...
-        self.eleinfo = []
+        self.eleinfo = eleinfo = []
         for (ename, eles), (eset, emask) in zip(system.ele_map.items(), rinfo):
-            # Locations of each solution point
-            ploc = eles.ploc_at_np('upts')[..., eset]
-            ploc = ploc.swapaxes(0, 1)
+            # Obtain quadrature info
+            rname = self.cfg.get(f'solver-elements-{ename}', 'soln-pts')
+            if self.cfg.hasopt(cfgsect, 'quad-deg'):
+                if self.cfg.hasopt(cfgsect, f'quad-pts-{ename}'):
```
FreddieWitherden, line 61: and

FreddieWitherden: Or alternatively you can just try and

FreddieWitherden: Or alternatively you can put the entire fragment in a try/except block, which might be cleaner.

GonzaloSaez: Here there are two main *branches*

1) If `quad-deg` is given but `quad-pts-{ename}` is not found in the input file: then use the quadrature rule name of the solution points

2) If both `quad-deg` and `quad-pts-{ename}` are provided: then use the quadrature rule name of `quad-pts-{ename}`

That's why I did not merge the two ifs. If I use an `and` condition then it means that option 1) can not work, is this OK? I'll rewrite the ifs in a `try/except` block once we've figured out if we want to be compatible with options 1) and 2).

FreddieWitherden:
Okay, so I think the final thing to decide on is how we want to specify the quadrature rules for different elements in a way that is consistent, flexible, and intuitive.

@WillTrojak @semi-h Any preferences?

GonzaloSaez:
Some proposals:

1) For each mesh element type add the following tokens to the `[solver-elements-*]` section

```
[solver-elements-*]
...
integrate-quad-deg = 6
integrate-quad-pts = name
```

2) For each mesh element type add the following to the `[soln-plugin-integrate]` section

```
[soln-plugin-integrate]
...
quad-deg-etype = 6
quad-pts-etype = name
```

3) The current implementation: set a unique quadrature degree for all elements, and default to the quadrature rule name of the solution points (with an option to choose other of quadrature rules) 

```
[soln-plugin-integrate]
...
quad-deg = 6
quad-pts-etype = name ; optional
```

4) Combination of the previous options

semi-h:
Integration plugin is a separate process than anti-aliasing indeed but if we are happy to carry out the simulation with a given point set and quadrature degree for anti-aliasing, wouldn't the same settings be enough for integration purposes as well? 

If there is a need to define a different point set and quadrature degree for any reason, I think the number 2 @GonzaloSaez  proposed above is the most sensible.

pv101:
@GonzaloSaez - thanks for your contributions!

@semi-h - I think it is reasonable that people might want to set different quadrature rules for the solver and the integrate plugin (minimum required for stability at solver level, but then perhaps higher for accuracy at integrate level). There is also the scenario where no anti-aliasing is used at the solver level but you still want to use a specific quadrature rule for integrate

Of the options above I like option 2 the best. But one question. If you have elements etype in your mesh, but you don't specify:

quad-deg-etype = 6
quad-pts-etype = name

what happens? What does it default to for integrate? Or will it give an error?

GonzaloSaez:
Hi @pv101 ,

If those options are not found, then the `integrate` plugin will maintain compatibility with the current implementation, i.e. it will rely on the quadrature associated with the solution points. Let me know if you think that this is not optimal, we can maybe find better default options for the plugin (especially for other than tensor product elements, whose solution points quadratures are usually of much lower order). An open question would be whether the user only provides one of the two tokens. Should we raise an error in that case?

@pv101 @semi-h  Option 2 seems indeed like the correct option (although a bit verbose). Let's see if Freddie agrees and I'll implement this functionality.

pv101:
Regarding: "An open question would be whether the user only provides one of the two tokens. Should we raise an error in that case?":

What do we currently do if someone only supplies one of quad-deg and quad-pts in a [solver-] block?

semi-h:
Actually how about having a quad-deg parameter that applies all the elements, and optionally quad-deg-etype for specific elements that require a different quad degree? This way ini file can be cleaner I think.
This is similar to how we handle pseudo residual tolerances. There is a generic value applies to all but then individual state variables can have their own tolerance value;
https://github.com/PyFR/PyFR/blob/develop/pyfr/integrators/dual/pseudo/base.py#L69

GonzaloSaez:
@pv101 

I think that if one does not use `quad-pts` **and** `quad-deg` (for anti-aliasing) then it would result in an error.

@semi-h 

Sounds good, my proposal would be

```dos
[soln-plugin-integrate]
...
quad-deg = 6 ; applied to every element type, by default suppose that the quadrature points rule is the as that of the solution points
quad-deg-etype = 5 ; etype will use a quadrature of degree 5 instead of 6
quad-pts-etype = name ; etype2 will use a quadrature of degree 5 with quadrature points rule equal to name
quad-pts-etype2 = name2 ; etype2 will use a quadrature of degree 6 with quadrature points rule equal to name2
```

All those parameters would be optional (to ensure compatibility with older versions).

One extra rule: using `quad-deg-etype` and `quad-pts-etype` without providing `quad-deg` would raise an error. This avoids the need to treat edge cases and ensures that there exists a default option for all element types.

FreddieWitherden:
> @pv101
> 
> I think that if one does not use `quad-pts` **and** `quad-deg` (for anti-aliasing) then it would result in an error.
> 
> @semi-h
> 
> Sounds good, my proposal would be
> 
> ```
> [soln-plugin-integrate]
> ...
> quad-deg = 6 ; applied to every element type, by default suppose that the quadrature points rule is the as that of the solution points
> quad-deg-etype = 5 ; etype will use a quadrature of degree 5 instead of 6
> quad-pts-etype = name ; etype2 will use a quadrature of degree 5 with quadrature points rule equal to name
> quad-pts-etype2 = name2 ; etype2 will use a quadrature of degree 6 with quadrature points rule equal to name2
> ```
> 
> All those parameters would be optional (to ensure compatibility with older versions).
> 
> One extra rule: using `quad-deg-etype` and/or `quad-pts-etype` without providing `quad-deg` would raise an error. This avoids the need to treat edge cases and ensures that there exists a default option for all element types.

Sounds good to me.

pyfr/plugins/integrate.py
```diff
@@ -52,22 +52,45 @@ def __init__(self, intg, cfgsect, suffix=None):
             # Open
             self.outf = init_csv(self.cfg, cfgsect, ','.join(header))
 
+        # Whether to use a quadrature different to that of the solution points
+        quad = self.cfg.hasopt(cfgsect, 'quad-deg')
```
FreddieWitherden, line 56: Try to avoid using hasopt, it usually leads to convoluted code.  Try/except based solutions are very often cleaner here.

pyfr/plugins/integrate.py
```diff
@@ -52,22 +52,45 @@ def __init__(self, intg, cfgsect, suffix=None):
...
-            # Quadature weights
-            rname = self.cfg.get(f'solver-elements-{ename}', 'soln-pts')
-            wts = get_quadrule(ename, rname, eles.nupts).wts
+            if quad:
+                # Quadrature rule (default to that of the solution points)
+                qrule = self.cfg.get(cfgsect, f'quad-pts-{ename}', rname)
+
+                # Quadrature rule degree
+                qdeg = self.cfg.getint(cfgsect, 'quad-deg')
+                if self.cfg.hasopt(cfgsect, f'quad-deg-{ename}'):
```
FreddieWitherden, line 75: Does this mean that I need to specify quad-deg even if I explicitly specify the rule degree for each element type in my simulation?

GonzaloSaez: Yes, although this can be modified. What would happen if the user does not specify the quadrature degree of an element type? Do we want to force the user to set the quadrature degree of all element types?

GonzaloSaez: We could maybe change those lines to

```python
try:
    qdeg = self.cfg.getint(cfgsect, f'quad-deg-{ename}')
except:
    qdeg = self.cfg.getint(cfgsect, f'quad-deg')
```

This would give preference to `quad-deg-{ename}` over `quad-deg`

GonzaloSaez: Ok. I've uploaded the new solution

1) Try with `quad-deg-{ename}`.
2) Except 1. try with `quad-deg`.
3) Except 2. use the quadrature of the solution points.

Let me know what you think!

PR #222, opened 2021-05-23
Add a performance tuning guide to the documentation.

FreddieWitherden:


doc/src/index.rst
```diff
@@ -21,6 +21,7 @@ Contents:
    user_guide
    developer_guide
    examples
+   performance_tuning
```
WillTrojak, line 24: Maybe worth putting the performance tuning before the examples

doc/src/performance_tuning.rst
```diff
@@ -0,0 +1,151 @@
+.. highlight:: none
+
+******************
+Performance Tuning
+******************
+
```
WillTrojak, line 6: A sentence summarising the aims of the section would be good.

doc/src/performance_tuning.rst
```diff
@@ -0,0 +1,151 @@
...
+OpenMP Backend
+==============
+
+libxsmm
+-------
+
+Although PyFR will usually function without libxsmm it is nevertheless
```
WillTrojak, line 13: Maybe add a one-liner about what libsxmm does and add a link to the git repo

doc/src/performance_tuning.rst
```diff
@@ -0,0 +1,151 @@
...
+        cflags = -mprefer-vector-width=512
+
+Cores vs. threads
+-----------------
+
+PyFR does not typically derive any benefit from SMT.  As such the
+number of OpenMP threads should be chosen to be equal to the number of
+physical cores.
+
+MPI processes vs. OpenMP threads
```
WillTrojak, line 35: Nowhere in the docs do we say that the environment variable OMP_NUM_THREADS can be used to set the threads in pyfr. This might be useful information to either give here or somewhere else.

doc/src/performance_tuning.rst
```diff
@@ -0,0 +1,151 @@
...
+and MPI distribution.
+
+CUDA Backend
+============
+
+CUDA-aware MPI
+--------------
+
+PyFR is capable of taking advantage of CUDA-aware MPI.  This enables
+CUDA device pointers to be directly passed MPI routines.  Under the
```
WillTrojak, line 52: 'directly passed MPI routines' -> 'directly passed to MPI routines'

doc/src/performance_tuning.rst
```diff
@@ -0,0 +1,151 @@
...
+
+        $ export UCX_MEMTYPE_CACHE=n
+
+Partitioning
+============
+
+METIS vs SCOTCH
+---------------
+
+The partitioning module in PyFR includes support for both METIS and
```
WillTrojak, line 73: maybe add links to metis and scotch

doc/src/performance_tuning.rst
```diff
@@ -0,0 +1,151 @@
...
+SCOTCH.  Both usually result in high-quality decompositions.  However,
+for long running simulations on complex geometries it may be worth
+partitioning a grid with both and observing which decomposition
+performs best.
+
+Heterogeneous grids
+-------------------
+
+When running PyFR in parallel on heterogeneous grids it is necessary
```
WillTrojak, line 82: I'm not sure the term heterogenous grids, in this context, is widely used or understood. My first thought for example was to think of heterogeneity in terms of the element size. Maybe add a description of how you are using this term here.

doc/src/performance_tuning.rst
```diff
@@ -0,0 +1,151 @@
+.. highlight:: none
+
+******************
+Performance Tuning
+******************
+
+OpenMP Backend
```
WillTrojak, line 7: It might also be worth stressing that this section is about performance *tuning*, and that if your performance is a long way from what might be considered reasonable then these points might be so useful. A non-exhaustive list of common issues with performance might be a good idea, but equally, this can be a bit of a pandora's box.

WillTrojak (commented):
Generally a useful addition to the documentation, but I think it needs a few more bits. Also when I tried to compile it there was a sphinx dependency missing, this is not related to this PR specifically, so I will submit a new one to fix this

doc/src/performance_tuning.rst
FreddieWitherden: Fixed

doc/src/performance_tuning.rst
FreddieWitherden: Done

doc/src/performance_tuning.rst
FreddieWitherden: Updated to mixed grids

doc/src/performance_tuning.rst
FreddieWitherden: We have links in the installation guide; and the main pages for both are not particularly helpful (and people typically end up going to their package manager/module system for them)

doc/src/performance_tuning.rst
FreddieWitherden: It is a bit more subtle, as OMP_NUM_THREADS is not always the best way to go about it depending on the OpenMP runtime/MPI library.

doc/src/performance_tuning.rst
FreddieWitherden: Updated with description of libxsmm, link is already in the installation guide (along with the specifics about how it needs to be compiled)

doc/src/index.rst
FreddieWitherden: Moved

WillTrojak (approved):

PR #223, opened 2021-05-28
Bug fix in linearisation of elements.

GiorgioGiangaspero:
Define an element as linear when max deltas in all directions (not just any of them) are less than the given tolerance.


GiorgioGiangaspero:
@GonzaloSaez could you take a look at this? thanks a lot.

GonzaloSaez:
Hi @GiorgioGiangaspero sorry for the late answer, the bugfix seems reasonable to me (it also works in some of my tests). Beforehand, there were edge cases in which an element was linearized if at least **one** of the coordinates of the reference frame could be approximated by a linear function within the given the element. The bugfix will ensure that the linearization only happens if **all** the coordinates of the reference frame can be approximated by a linear function within the given element.

PR #230, opened 2021-06-25
Separated cuda runtime config from build kernel and added dynamic shared config

WillTrojak:
The unified cache preference can now be separately set for a kernel, as well as configuring the dynamic shared memory usage and shared carveout.

### DIFF
--- a/doc/src/user_guide.rst
+++ b/doc/src/user_guide.rst
@@ -128,7 +128,7 @@ Parameterises the CUDA backend with
 
     *int*
 
-5. ``cflags`` --- additional Nvidia realtime compiler (``nvrtc``) flags:
+5. ``cflags`` --- additional NVIDIA realtime compiler (``nvrtc``) flags:
     
     *string*
 
--- a/doc/src/user_guide.rst
+++ b/doc/src/user_guide.rst
@@ -128,8 +128,7 @@ Parameterises the CUDA backend with
 
     *int*
 
-5. ``opflags`` --- additional optimisation flags to be passed to Nvidia 
-   real-time compiler:
+5. ``cflags`` --- additional Nvidia realtime compiler (``nvrtc``) flags:
     
     *string*
 
--- a/doc/src/user_guide.rst
+++ b/doc/src/user_guide.rst
@@ -128,6 +128,11 @@ Parameterises the CUDA backend with
 
     *int*
 
+5. ``opflags`` --- additional optimisation flags to be passed to Nvidia 
+   real-time compiler:
+    
+    *string*
+
 Example::
 
     [backend-cuda]
--- a/pyfr/backends/cuda/blasext.py
+++ b/pyfr/backends/cuda/blasext.py
@@ -22,8 +22,7 @@ class CUDABlasExtKernels(CUDAKernelProvider):
 
         # Build the kernel
         kern = self._build_kernel('axnpby', src,
-                                  [np.int32]*3 + [np.intp]*nv + [dtype]*nv
-                                 )
+                                  [np.int32]*3 + [np.intp]*nv + [dtype]*nv)
 
         # Determine the grid/block
         block = (128, 1, 1)
--- a/pyfr/backends/cuda/blasext.py
+++ b/pyfr/backends/cuda/blasext.py
@@ -88,7 +88,10 @@ class CUDABlasExtKernels(CUDAKernelProvider):
             argt = [np.int32]*3 + [np.intp]*3 + [dtype]
 
         # Build the reduction kernel
-        rkern = self._build_kernel('reduction', src, argt, prefer_l1=True)
+        rkern = self._build_kernel('reduction', src, argt)
+
+        # Configure L1
+        rkern.set_cache_pref(prefer_l1=True)
 
         # Norm type
         reducer = np.max if norm == 'uniform' else np.sum
--- a/pyfr/backends/cuda/blasext.py
+++ b/pyfr/backends/cuda/blasext.py
@@ -22,8 +22,7 @@ class CUDABlasExtKernels(CUDAKernelProvider):
 
         # Build the kernel
         kern = self._build_kernel('axnpby', src,
-                                  [np.int32]*3 + [np.intp]*nv + [dtype]*nv,
-                                  prefer_l1=True
+                                  [np.int32]*3 + [np.intp]*nv + [dtype]*nv
                                  )
 
         # Determine the grid/block
@@ -89,9 +88,6 @@ class CUDABlasExtKernels(CUDAKernelProvider):
 
         # Build the reduction kernel
         rkern = self._build_kernel('reduction', src, argt)
-
-        # Configure L1
-        rkern.set_cache_pref(prefer_l1=True)
 
         # Norm type
         reducer = np.max if norm == 'uniform' else np.sum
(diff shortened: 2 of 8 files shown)


pyfr/backends/cuda/blasext.py
```diff
@@ -22,7 +22,9 @@ def axnpby(self, *arr, subdims=None):
 
         # Build the kernel
         kern = self._build_kernel('axnpby', src,
-                                  [np.int32]*3 + [np.intp]*nv + [dtype]*nv)
+                                  [np.int32]*3 + [np.intp]*nv + [dtype]*nv,
+                                  prefer_l1=True
```
FreddieWitherden, line 26: Can change/remove?

pyfr/backends/cuda/blasext.py
```diff
@@ -88,6 +90,9 @@ def reduction(self, *rs, method, norm, dt_mat=None):
         # Build the reduction kernel
         rkern = self._build_kernel('reduction', src, argt)
 
+        # Configure L1
+        rkern.set_cache_pref(prefer_l1=True)
```
FreddieWitherden, line 94: As this is a reduction kernel which does use some shared memory is it worth declaring a preference?

pyfr/backends/cuda/driver.py
```diff
@@ -198,11 +212,20 @@ def set_cache_pref(self, *, prefer_l1=None, prefer_shared=None):
         pref = self.cuda._get_cache_pref(prefer_l1, prefer_shared)
         self.cuda.lib.cuFuncSetCacheConfig(self, pref)
 
-    def exec_async(self, grid, block, stream, *args):
+    def set_shared_size(self, *, shared=0, carveout=None):
```
FreddieWitherden, line 215: Include some reference to dynamic here?

pyfr/backends/cuda/gimmik.py
```diff
@@ -35,7 +35,10 @@ def mul(self, a, b, out, alpha=1.0, beta=0.0):
 
         # Build
         fun = self._build_kernel('gimmik_mm', src,
-                                 [np.int32, np.intp]*2 + [np.int32])
+                                 [np.int32, np.intp]*2 + [np.int32]
+                                )
```
FreddieWitherden, line 39: Can move back up

pyfr/backends/cuda/blasext.py
```diff
@@ -22,7 +22,8 @@ def axnpby(self, *arr, subdims=None):
 
         # Build the kernel
         kern = self._build_kernel('axnpby', src,
-                                  [np.int32]*3 + [np.intp]*nv + [dtype]*nv)
+                                  [np.int32]*3 + [np.intp]*nv + [dtype]*nv
+                                 )
```
FreddieWitherden, line 26: Can move ) back up.

pyfr/backends/cuda/compiler.py
```diff
@@ -93,7 +93,7 @@ def compile(self, name, src, flags=[]):
 
 
 class SourceModule(object):
-    def __init__(self, backend, src):
+    def __init__(self, backend, src, additional_opt=False):
```
FreddieWitherden, line 96: Still not sure about this; how about we instead expose flags to the ini file like we do with OpenMP?  Makes it nice and accessible for those who want to poke around as new features are added to NVRTC.

pyfr/backends/cuda/compiler.py
```diff
@@ -106,18 +106,16 @@ def __init__(self, backend, src):
...
 
         # Compile to PTX
         ptx = backend.nvrtc.compile('kernel', src, flags)
 
         # Load it as a module
         self.mod = backend.cuda.load_module(ptx)
 
-    def get_function(self, name, argtypes, *, prefer_l1=None,
-                     prefer_shared=None):
+    def get_function(self, name, argtypes, *):
```
FreddieWitherden, line 118: No need for *.

pyfr/backends/cuda/provider.py
```diff
@@ -14,10 +14,10 @@ def get_grid_for_block(block, nrow, ncol=1):
 
 class CUDAKernelProvider(BaseKernelProvider):
     @memoize
-    def _build_kernel(self, name, src, argtypes):
+    def _build_kernel(self, name, src, argtypes, **kwargs):
```
FreddieWitherden, line 17: Not sure we need this?

doc/src/user_guide.rst
```diff
@@ -128,6 +128,10 @@ Parameterises the CUDA backend with
 
     *int*
 
+5. ``cflags`` --- additional Nvidia realtime compiler (``nvrtc``) flags:
```
FreddieWitherden, line 131: NVIDIA

pyfr/backends/cuda/gimmik.py
```diff
@@ -34,8 +34,10 @@ def mul(self, a, b, out, alpha=1.0, beta=0.0):
                           alpha=alpha, beta=beta)
 
         # Build
-        fun = self._build_kernel('gimmik_mm', src,
+        fun = self._build_kernel('gimmik_mm', src,
```
FreddieWitherden, line 37: Trailing space.

pyfr/backends/cuda/packing.py
```diff
@@ -18,7 +18,9 @@ def pack(self, mv):
         src = self.backend.lookup.get_template('pack').render()
 
         # Build
-        kern = self._build_kernel('pack_view', src, [np.int32]*3 + [np.intp]*4)
+        kern = self._build_kernel('pack_view', src, [np.int32]*3 + [np.intp]*4,
+                                  prefer_l1=True
```
FreddieWitherden, line 22: Is this still correct?

pyfr/backends/cuda/packing.py
```diff
@@ -18,7 +18,8 @@ def pack(self, mv):
         src = self.backend.lookup.get_template('pack').render()
 
         # Build
-        kern = self._build_kernel('pack_view', src, [np.int32]*3 + [np.intp]*4)
```
FreddieWitherden, line 21: No need to change this.

WillTrojak: line length

FreddieWitherden:
```
In [1]: len('''        kern = self._build_kernel('pack_view', src, [np.int32]*3 + [np.intp]*4)''')
Out[1]: 79
```

pyfr/backends/cuda/blasext.py
FreddieWitherden: Just following up here, and should we be setting the l1 pref? (Don't think it should make any difference, but would be interested to see.)

FreddieWitherden:
Can we do a quick profile to ensure that post-changes all of the kernels still perform the same (as some kernels have had their L1 preferences changed).

WillTrojak:
The performance was pretty much the smae, some kernel were a little faster, some a little slower. But all were within the varaition from one profiler run to the next. Just as a point for those using Nsight compute, it seems unable to attach to pyfr when using python3.9 at the moment. So in both instantces I used python 3.8.

Also HIP has similar features, so I should probably addd these features to the HIP backend as well, but I will wait until I have access to AMD hardware to do this.

FreddieWitherden:
If you can make the final couple of formatting changes I think we're good to merge.

PR #232, opened 2021-06-30
Fixed missing flux variable.

WillTrojak:


WillTrojak:
Tested on an incompressible version of the vortex test case and seemed to work fine.

PR #233, opened 2021-07-11
Synthetic turbulence generation

GiorgioGiangaspero:
Synthetic turbulence generation via source term


FreddieWitherden:
So the key thing for this PR will be separating the infrastructure changes from the functionality itself.  Broadly, we want a sequence of commits which add generic functionality to PyFR, and then a final commit which adds the turbulence generation support which only impacts files in the plugins/ directory.  This will be essential to maintaining orthogonality.

pyfr/plugins/turbulencegenerator.py
```diff
@@ -53,6 +53,11 @@ class TurbulenceGeneratorPlugin(BasePlugin):
     def __init__(self, intg, cfgsect, suffix):
         super().__init__(intg, cfgsect, suffix)
 
+        # do not allow multiple instances
+        if suffix:
```
FreddieWitherden, line 57: Think generically!  This should be handled by a mechanism by which plugins can declare themselves to be singletons.  For example 

```
class MyPlugin(BasePlugin):
    name = ...
    ...
    singleton = True
```

when plugins are being constructed this property can be queried.  Also note that the current code will break if you specify only one instance but opt to include a suffix.

pyfr/solvers/baseadvec/kernels/negdivconf.mako
```diff
@@ -11,4 +11,87 @@
...
+
+// Add density and energy fluctuations (compressible solver only, of course).
+% if system == 'compr':
+    // density
+    fpdtype_t rM2 = ${rhomeanex}*${Mmeanex}*${Mmeanex};
+    tdivtconf[0] += factor[${Ubulkdir}]*${rhofluctfactor}*rM2*utilde[${Ubulkdir}];
+
+    // energy equation
+    fpdtype_t udotu_fluct = ${pyfr.dot('utilde[{i}]', i=(0, ndims))};
+    tdivtconf[${ndims} + 1] += factor[${Ubulkdir}]*0.5*u[0]*udotu_fluct;
```
WillTrojak, line 93: move +1 inside ${}

pyfr/solvers/baseadvec/kernels/negdivconf.mako
```diff
@@ -11,4 +11,87 @@
...
+// Working variables
+fpdtype_t arg, arg2, pexp;
+fpdtype_t csi[${ndims}];
+
+// Compute csi_max[3][3] on the fly for this point as lturb/lturb ref.
+// make sure the minimum value is not less than cmm for stability reasons.
+fpdtype_t csimax[${ndims}][${ndims}] = {{1.0}};
+% for i in range(ndims):
+    % for j in range(ndims):
+        csimax[${i}][${j}] = min(max(${lturbex[i][j]}*lturbrefinv[${i}], ${cmm}), 1.0);
```
WillTrojak, line 32: can you use the `pyfr.ndrange(ndims, ndims)`?

PR #234, opened 2021-07-20
Add support for __launch_bounds__ in the HIP backend.

FreddieWitherden:
This fixes a performance regression with recent versions of HIP.


WillTrojak:
Which tests have you run and out of interest what was the effect on performance?

Also do you think its worth adding a way to optionally set the min warps per eu?

FreddieWitherden:
On my p = 4 TGV case on a Radeon VII the old version of HIP ran two time units in ~223s.  Newer version of HIP (which change the default launch bounds to 1024 if nothing is explicitly specified) caused this to regress to ~335s.  With the proposed PR this time goes down to ~230s.  AMD internally observed similar numbers on a related case.

Specifying min warps per EU is a bit like specifying a register count for CUDA; usually detrimental to performance.

WillTrojak (approved):
looks good to me

PR #235, opened 2021-07-20
Add support for hipEventDisableTiming in events.

FreddieWitherden:


pyfr/backends/hip/driver.py
```diff
@@ -119,7 +119,8 @@ class HIPWrappers(LibWrapper):
     }
 
     # Constants
-    MemcpyDefault = 4
+    EVENT_DISABLE_TIMING = 2
```
WillTrojak, line 122: Can you add hipEventDefault in case someone wants to diable this feature?

WillTrojak (commented):
Appart from my comment on event default, this looks good. But what is the effect on performance? And does disablling event timing effect profiling?

pyfr/backends/hip/driver.py
FreddieWitherden: Here we would probably want them to use hipEventCreate rather than calling hipEventCreateWithFlags with an unnecessary flags argument.  However, as there are no current consumers of this the function is unwrapped.

PR #237, opened 2021-07-28
Make the file writing code more flexible.

FreddieWitherden:
It is now possible for data sets to be specified at the time a file is being written as opposed to when the NativeWriter is constructed.

### DIFF
--- a/pyfr/plugins/base.py
+++ b/pyfr/plugins/base.py
@@ -8,7 +8,6 @@ import numpy as np
 from pytools import prefork
 
 from pyfr.mpiutil import get_comm_rank_root
-from pyfr.writers.native import NativeWriter
 
 
 def init_csv(cfg, cfgsect, header, *, filekey='file', headerkey='header'):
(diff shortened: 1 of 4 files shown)


pyfr/plugins/base.py
```diff
@@ -104,54 +104,29 @@ def _invoke_postaction(self, intg, **kwargs):
...
+            self._prepare_region_data_box(intg, *box)
         # All elements on a boundary
         else:
-            mdata = self._prepare_mdata_bcs(intg, fpdtype, nout, prefix,
-                                            region)
-            self._add_region_data = self._add_region_data_subset
-
-        # Construct the file writer
-        return NativeWriter(intg, mdata, basedir, basename)
+            self._prepare_region_data_bcs(intg, region)
```
GonzaloSaez, line 122: `from pyfr.writers.native import NativeWriter` should be removed I think

pyfr/plugins/writer.py
```diff
@@ -22,6 +28,9 @@ def __init__(self, intg, cfgsect, suffix=None):
         # Output field names
         self.fields = intg.system.elementscls.convarmap[self.ndims]
 
+        # Output data type
+        self.fpdtype = intg.backend.fpdtype
```
GonzaloSaez, line 32: Wouldn't it be interesting to give the users to output solution files in other than the default precision? (opt-in feature of course)

pyfr/plugins/base.py
FreddieWitherden: Fixed.

pyfr/plugins/writer.py
FreddieWitherden: We can certainly add that without too much hassle.

PR #239, opened 2021-08-13
Add moments computation to fluidforce

GonzaloSaez:
As requested in https://pyfr.discourse.group/t/the-question-about-the-moment-calculation-in-fluidforce/457. This PR is a draft and may contain bugs.

To activate the moments computation add the `morigin` token to the `fluidforce` plugin

```ini
[soln-plugin-fluidforce-bcwalllower]
header = true
...
morigin = [(0, 0)]
```

### DIFF
--- a/pyfr/plugins/fluidforce.py
+++ b/pyfr/plugins/fluidforce.py
@@ -37,7 +37,12 @@ class FluidForcePlugin(BasePlugin):
         self.elementscls = intg.system.elementscls
 
         # Boundary to integrate over
-        bc = f'bcon_{suffix}_p{intg.rallocs.prank}'
+        bc = 'bcon_{0}_p{1}'.format(suffix, intg.rallocs.prank)
+
+        # Moments origin
+        origin = self.cfg.get(cfgsect, 'morigin', '*')
+        self.morigin = (None if "*" in origin
+                        else self.cfg.getliteral(cfgsect, 'morigin'))
 
         # Get the mesh and elements
         mesh, elemap = intg.system.mesh, intg.system.ele_map
@@ -48,12 +53,25 @@ class FluidForcePlugin(BasePlugin):
         # The root rank needs to open the output file
         if rank == root:
             if not any(bcranks):
-                raise RuntimeError(f'Boundary {suffix} does not exist')
+                raise RuntimeError('Boundary {0} does not exist'
+                                   .format(suffix))
 
             # CSV header
             header = ['t', 'px', 'py', 'pz'][:self.ndims + 1]
             if self._viscous:
                 header += ['vx', 'vy', 'vz'][:self.ndims]
+            if self.morigin:
+                # Add headers of moments. Moments always have one and three components
+                # in 2D and 3D simulations, respectively
+                self._mcomponents = 3 if self.ndims == 3 else 1
+                if self.ndims == 2:
+                    header += ['mpz']
+                    if self._viscous:
+                        header += ['mvz']
+                else:
+                    header += ['mpx', 'mpy', 'mpz']
+                    if self._viscous:
+                        header += ['mvx', 'mvy', 'mvz']
 
             # Open
             self.outf = init_csv(self.cfg, cfgsect, ','.join(header))
@@ -71,6 +89,8 @@ class FluidForcePlugin(BasePlugin):
             # Element indices and associated face normals
             eidxs = defaultdict(list)
             norms = defaultdict(list)
+            if self.morigin:
+                rfpts = defaultdict(list)
 
             for etype, eidx, fidx, flags in mesh[bc].astype('U4,i4,i1,i2'):
                 eles = elemap[etype]
@@ -97,11 +117,21 @@ class FluidForcePlugin(BasePlugin):
                 npn = eles.get_norm_pnorms(eidx, fidx)
                 mpn = eles.get_mag_pnorms(eidx, fidx)
 
+                # Get the flux points position of the given face and element indices
+                # relative to the moment origin
+                if self.morigin:
+                    rfpt = eles.get_fpts(eidx, fidx) - np.array(self.morigin)
+
                 eidxs[etype, fidx].append(eidx)
                 norms[etype, fidx].append(mpn[:, None]*npn)
+                if self.morigin:
+                    rfpts[etype, fidx].append(rfpt)
 
             self._eidxs = {k: np.array(v) for k, v in eidxs.items()}
             self._norms = {k: np.array(v) for k, v in norms.items()}
+            if self.morigin:
+                self._rfpts = {k: np.array(v)
+                               for k, v in rfpts.items()}
 
             if self._viscous:
                 self._rcpjact = {k: rcpjact[k[0]][..., v]
@@ -121,6 +151,11 @@ class FluidForcePlugin(BasePlugin):
 
         # Force vector
         f = np.zeros(2*ndims if self._viscous else ndims)
+
+        # Moments vector
+        if self.morigin:
+            m = np.zeros(
+                2*self._mcomponents if self._viscous else self._mcomponents)
 
         for etype, fidx in self._m0:
             # Get the interpolation operator
@@ -145,6 +180,15 @@ class FluidForcePlugin(BasePlugin):
 
             # Do the quadrature
             f[:ndims] += np.einsum('i...,ij,jik', qwts, p, norms)
+            if self.morigin:
+                rfpts = self._rfpts[etype, fidx]
+                # Pressure force moments
+                if self.ndims == 2:
+                    m[:self._mcomponents] += np.einsum(
+                        'i...,ij,ji', qwts, p, np.cross(rfpts, norms))
+                else:
+                    m[:self._mcomponents] += np.einsum(
+                        'i...,ij,jik', qwts, p, np.cross(rfpts, norms))
 
             if self._viscous:
                 # Get operator and J^-T matrix
@@ -172,6 +216,13 @@ class FluidForcePlugin(BasePlugin):
 
                 # Do the quadrature
                 f[ndims:] += np.einsum('i...,klij,jil', qwts, vis, norms)
+                if self.morigin:
+                    # Viscous force at each flux point
+                    viscf = np.einsum('ijkl, lkj->lki', vis, norms)
+
+                    # Viscous force moments
+                    m[self._mcomponents:] += np.einsum(
+                        'i,ji', qwts, np.cross(rfpts, viscf)).sum()
 
         # Reduce and output if we're the root rank
         if rank != root:
@@ -179,8 +230,13 @@ class FluidForcePlugin(BasePlugin):
         else:
             comm.Reduce(get_mpi('in_place'), f, op=get_mpi('sum'), root=root)
 
+            # Build the row
+            row = [intg.tcurr] + f.tolist()
+            if self.morigin:
+                row += m.tolist()
+
             # Write
-            print(intg.tcurr, *f, sep=',', file=self.outf)
+            print(','.join(str(r) for r in row), file=self.outf)
 
             # Flush to disk
             self.outf.flush()
@@ -204,7 +260,7 @@ class FluidForcePlugin(BasePlugin):
         mu = c['mu']
 
         if self._viscorr == 'sutherland':
-            cpT = c['gamma']*(E/rho - 0.5*np.sum(u[1:-1]**2, axis=0)/rho**2)
+            cpT = c['gamma']*(E/rho - 0.5*np.sum(u[1:-1]**2, axis=0))
             Trat = cpT/c['cpTref']
             mu *= (c['cpTref'] + c['cpTs'])*Trat**1.5 / (cpT + c['cpTs'])
 
--- a/pyfr/solvers/base/elements.py
+++ b/pyfr/solvers/base/elements.py
@@ -177,11 +177,11 @@ class BaseElements(object):
         sbufs, abufs = self._scratch_bufs, []
 
         # Convenience functions for scalar/vector allocation
-        def alloc(ex, n): return abufs.append(
+        alloc = lambda ex, n: abufs.append(
             backend.matrix(n, extent=nonce + ex, tags={'align'})
         ) or abufs[-1]
-        def salloc(ex, n): return alloc(ex, (n, nvars, neles))
-        def valloc(ex, n): return alloc(ex, (ndims, n, nvars, neles))
+        salloc = lambda ex, n: alloc(ex, (n, nvars, neles))
+        valloc = lambda ex, n: alloc(ex, (ndims, n, nvars, neles))
 
         # Allocate required scalar scratch space
         if 'scal_fpts' in sbufs:
--- a/pyfr/solvers/base/elements.py
+++ b/pyfr/solvers/base/elements.py
@@ -177,11 +177,11 @@ class BaseElements(object):
         sbufs, abufs = self._scratch_bufs, []
 
         # Convenience functions for scalar/vector allocation
-        alloc = lambda ex, n: abufs.append(
+        def alloc(ex, n): return abufs.append(
             backend.matrix(n, extent=nonce + ex, tags={'align'})
         ) or abufs[-1]
-        salloc = lambda ex, n: alloc(ex, (n, nvars, neles))
-        valloc = lambda ex, n: alloc(ex, (ndims, n, nvars, neles))
+        def salloc(ex, n): return alloc(ex, (n, nvars, neles))
+        def valloc(ex, n): return alloc(ex, (ndims, n, nvars, neles))
 
         # Allocate required scalar scratch space
         if 'scal_fpts' in sbufs:
@@ -396,6 +396,10 @@ class BaseElements(object):
     def get_norm_pnorms(self, eidx, fidx):
         fpts_idx = self.basis.facefpts[fidx]
         return self._norm_pnorm_fpts[fpts_idx, eidx]
+
+    def get_fpts(self, eidx, fidx):
+        fpts_idx = self.basis.facefpts[fidx]
+        return self.plocfpts[fpts_idx, eidx]
 
     def get_scal_fpts_for_inter(self, eidx, fidx):
         nfp = self.nfacefpts[fidx]


pyfr/plugins/fluidforce.py
```diff
@@ -39,6 +39,11 @@ def __init__(self, intg, cfgsect, suffix):
         # Boundary to integrate over
         bc = f'bcon_{suffix}_p{intg.rallocs.prank}'
 
+        # Moments origin
+        origin = self.cfg.get(cfgsect, 'morigin', '*')
```
FreddieWitherden, line 43: Since the only valid value is a literal hasopt and getliteral may be cleaner.  Also, compute the number of moments up front.

pyfr/plugins/fluidforce.py
```diff
@@ -54,6 +59,18 @@ def __init__(self, intg, cfgsect, suffix):
...
+            if self.morigin:
+                # Add headers of moments. Moments always have one and three components
+                # in 2D and 3D simulations, respectively
+                self._mcomponents = 3 if self.ndims == 3 else 1
+                if self.ndims == 2:
+                    header += ['mpz']
+                    if self._viscous:
+                        header += ['mvz']
+                else:
+                    header += ['mpx', 'mpy', 'mpz']
```
FreddieWitherden, line 71: Given the number of moments we should be able to play a similar trick to `['vx', 'vy', 'vz'][:self.ndims]` to simplify this block.

pyfr/plugins/fluidforce.py
```diff
@@ -71,6 +88,8 @@ def __init__(self, intg, cfgsect, suffix):
             # Element indices and associated face normals
             eidxs = defaultdict(list)
             norms = defaultdict(list)
+            if self.morigin:
```
FreddieWitherden, line 91: Always assign.

pyfr/plugins/fluidforce.py
FreddieWitherden: Also cast to an np.array up front.

pyfr/plugins/fluidforce.py
```diff
@@ -99,9 +118,17 @@ def __init__(self, intg, cfgsect, suffix):
 
                 eidxs[etype, fidx].append(eidx)
                 norms[etype, fidx].append(mpn[:, None]*npn)
+                # Get the flux points position of the given face and element indices
+                # relative to the moment origin
+                if self.morigin:
```
FreddieWitherden, line 123: Switched based off of number of components (0 = no moment calculation).

pyfr/plugins/fluidforce.py
```diff
@@ -99,9 +118,17 @@ def __init__(self, intg, cfgsect, suffix):
...
+                    rfpt = eles.get_fpts(eidx, fidx) - np.array(self.morigin)
+                    rfpts[etype, fidx].append(rfpt)
 
             self._eidxs = {k: np.array(v) for k, v in eidxs.items()}
             self._norms = {k: np.array(v) for k, v in norms.items()}
+            if self.morigin:
```
FreddieWitherden, line 129: Always assign (effective no-op if the number of moment components is zero).

pyfr/plugins/fluidforce.py
```diff
@@ -145,6 +177,15 @@ def __call__(self, intg):
 
             # Do the quadrature
             f[:ndims] += np.einsum('i...,ij,jik', qwts, p, norms)
+            if self.morigin:
+                rfpts = self._rfpts[etype, fidx]
+                # Pressure force moments
+                if self.ndims == 2:
+                    m[:self._mcomponents] += np.einsum(
+                        'i...,ij,ji', qwts, p, np.cross(rfpts, norms))
```
FreddieWitherden, line 185: Factor out the string and assign this based off an if statement.  Can also move the cross into the einsum expression to avoid a temporary.

pyfr/plugins/fluidforce.py
```diff
@@ -172,6 +213,13 @@ def __call__(self, intg):
 
                 # Do the quadrature
                 f[ndims:] += np.einsum('i...,klij,jil', qwts, vis, norms)
+                if self.morigin:
+                    # Viscous force at each flux point
+                    viscf = np.einsum('ijkl, lkj->lki', vis, norms)
+
+                    # Viscous force moments
+                    m[self._mcomponents:] += np.einsum(
+                        'i,ji', qwts, np.cross(rfpts, viscf)).sum()
```
FreddieWitherden, line 222: Should be able to merge the sum into the einsum expression.  Can also move the cross into the einsum which enables viscf to be merged, too.

pyfr/solvers/base/elements.py
```diff
@@ -397,6 +397,10 @@ def get_norm_pnorms(self, eidx, fidx):
         fpts_idx = self.basis.facefpts[fidx]
         return self._norm_pnorm_fpts[fpts_idx, eidx]
 
+    def get_fpts(self, eidx, fidx):
```
FreddieWitherden, line 400: Is there a way to avoid adding this method?

pyfr/plugins/fluidforce.py
GonzaloSaez: I'll wait for a bit more validation before merging the cross product in the `einsum`. Do you know of a quick way to use `einsum` for cross products without calling `tensordot`?

pyfr/solvers/base/elements.py
GonzaloSaez: Well we need the access the flux points physical position. I could just write this function inside the `fluidforce` plugin if that simplifies things.

pyfr/plugins/fluidforce.py
GonzaloSaez: Done. I'll wait for a bit more validation before merging the cross product in the einsum (see the next comment)

pyfr/plugins/fluidforce.py
GonzaloSaez: This is tricky since for 2D cases it's only the z direction component which is non-zero which is different from the forces in which we have `ndims` non-zero components for both 2D and 3D cases.

FreddieWitherden: `['mpx', 'mpy', 'mpz'][3 - mcomp:]` I think will do it.

pyfr/plugins/fluidforce.py
FreddieWitherden: It's been a while, but I believe I've used it in the past for batch cross product calculations.

pyfr/plugins/fluidforce.py
GonzaloSaez: Makes sense, thanks!

pyfr/plugins/fluidforce.py
```diff
@@ -99,9 +111,15 @@ def __init__(self, intg, cfgsect, suffix):
 
                 eidxs[etype, fidx].append(eidx)
                 norms[etype, fidx].append(mpn[:, None]*npn)
+                # Get the flux points position of the given face and element
+                # indices relative to the moment origin
+                if self._mcomp:
+                    rfpt = eles.get_fpts(eidx, fidx) - morigin
+                    rfpts[etype, fidx].append(rfpt)
```
FreddieWitherden, line 118: Can we bake in the cross product with the normals here?

pyfr/solvers/base/elements.py
FreddieWitherden: I think that might be cleaner; get_fpts is also a bit ambiguous as it could also refer to the flux point array itself.

GonzaloSaez: Makes sense, I'll move this function to the plugin

pyfr/plugins/fluidforce.py
GonzaloSaez: I thought about that. This is perfectly fine for the pressure forces, since the pressure is a scalar. However, for the viscous forces we have `\vector{\rfpt} \cross ( \tensor{\tau} \cdot \vector{\normal})`. I'll have to simplify the latter with Einstein notation to see if the term `\vector{\rfpt} \cross \vector{\normal}` appears in it.

pyfr/__main__.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-
+import sys
+sys.path.append('/home/gonzalo/PyFR_fork/PyFR/')
```
Winxingtf, line 4: So the path added in this line of code is the home directory to the PyFR?

GonzaloSaez: That was a mistake on my end, I'll fix it. It's just a small band aid to help me debug stuff.

GonzaloSaez:
@FreddieWitherden I can store in memory the cross product of the relative flux points position and their _normals_ to compute the pressure forces. However, (if I'm not mistaken) this doesn't help when computing the viscous force moments since we cannot factor out the same cross product. Is it then worth increasing a bit the memory footprint of the plugin?

pyfr/plugins/fluidforce.py
```diff
@@ -172,6 +201,15 @@ def __call__(self, intg):
...
                 f[ndims:] += np.einsum('i...,klij,jil', qwts, vis, norms)
+                if self._mcomp:
+                    # Viscous force at each flux point
+                    viscf = np.einsum('ijkl,lkj->lki', vis, norms)
+
+                    # Viscous force moments
+                    mop = 'i,ji' if self.ndims == 2 else 'i,jik'
+                    m[self._mcomp:] += np.einsum(mop, qwts,
+                                                 np.cross(rfpts, viscf)
+                                                 ).sum(axis=0)
```
FreddieWitherden, line 212: Can einsum not handle the .sum(axis=0) part?

pyfr/plugins/fluidforce.py
```diff
@@ -39,6 +39,14 @@ def __init__(self, intg, cfgsect, suffix):
         # Boundary to integrate over
         bc = f'bcon_{suffix}_p{intg.rallocs.prank}'
 
+        # Moments
+        mcomp = 3 if self.ndims == 3 else 1
+        self._mcomp = mcomp if self.cfg.hasopt(cfgsect, 'morigin') else 0
+        if self._mcomp:
+            morigin = np.array(self.cfg.getliteral(cfgsect, 'morigin'))
+            if morigin.shape[1] != self.ndims:
```
FreddieWitherden, line 47: morigin should be a 1D array (so specified as (x, y, z) or [x, y, z]).

pyfr/plugins/fluidforce.py
GonzaloSaez: Fixed it.

pyfr/plugins/fluidforce.py
GonzaloSaez: True. This is fixed now.

pyfr/plugins/fluidforce.py
FreddieWitherden: If there are no tricks we can play down the line, it might be easier to not bake in the cross products.

pyfr/plugins/fluidforce.py
```diff
@@ -122,6 +147,9 @@ def __call__(self, intg):
         # Force vector
         f = np.zeros(2*ndims if self._viscous else ndims)
 
+        # Moments vector
+        m = np.zeros(2*self._mcomp if self._viscous else self._mcomp)
```
FreddieWitherden, line 151: Merge with f so we only need a single MPI reduction.

pyfr/plugins/fluidforce.py
```diff
@@ -145,6 +173,12 @@ def __call__(self, intg):
 
             # Do the quadrature
             f[:ndims] += np.einsum('i...,ij,jik', qwts, p, norms)
+            if self._mcomp:
+                rfpts_c_norms = self._rfpts_c_norms[etype, fidx]
+
+                # Pressure force moments
+                mop = 'i...,ij,ji' if self.ndims == 2 else 'i...,ij,jik'
```
FreddieWitherden, line 180: Can we unify these through broadcasting?

pyfr/plugins/fluidforce.py
```diff
@@ -172,15 +206,28 @@ def __call__(self, intg):
...
+                if self._mcomp:
+                    # Get the relative flux points position
+                    rfpts = self._rfpts[etype, fidx]
+
+                    # Viscous force at each flux point
+                    viscf = np.einsum('ijkl,lkj->lki', vis, norms)
+
+                    # Viscous force moments
+                    mop = 'i,ji->' if self.ndims == 2 else 'i,jik->k'
+                    m[self._mcomp:] += np.einsum(mop, qwts,
```
FreddieWitherden, line 218: What does the non-einsum (broadcast-based) solution look like?

pyfr/plugins/fluidforce.py
```diff
@@ -172,15 +206,28 @@ def __call__(self, intg):
...
+                                                 np.cross(rfpts, viscf))
 
         # Reduce and output if we're the root rank
+        red = np.array([*f, *m])
```
FreddieWitherden, line 222: See above comment about constructing as one big array straight off.

pyfr/plugins/fluidforce.py
```diff
@@ -215,3 +262,7 @@ def ac_stress_tensor(self, du):
         gradu, nu = du[:, 1:], self._constants['nu']
 
         return -nu*(gradu + gradu.swapaxes(0, 1))
+
+    def get_fpts(self, eles, eidx, fidx):
```
FreddieWitherden, line 266: I'd move this inline as it probably should not be a public function.

pyfr/plugins/fluidforce.py
GonzaloSaez: I used `np.atleast_3d` to unify, let me know if that's OK

pyfr/plugins/fluidforce.py
GonzaloSaez: You mean for the cross product or to multiply `qwts` by the cross product?

I've changed this part to avoid the if in `mop`.

pyfr/plugins/fluidforce.py
```diff
@@ -117,10 +143,12 @@ def __call__(self, intg):
...
         # Solution matrices indexed by element type
         solns = dict(zip(intg.system.ele_types, intg.soln))
-        ndims, nvars = self.ndims, self.nvars
+        ndims, nvars, mcomp = self.ndims, self.nvars, self._mcomp
 
-        # Force vector
-        f = np.zeros(2*ndims if self._viscous else ndims)
+        # Force and moment vector
+        n = 2 if self._viscous else 1
+        fm = np.zeros(n*(ndims + mcomp))
```
FreddieWitherden, line 150: We may want to consider using a 2D array here, with n being the second dimension (one or two).  May result in cleaner/more natural code.

doc/src/user_guide.rst
```diff
@@ -1276,12 +1276,15 @@ file. Parameterised with
 
     *boolean*
 
+4. ``morigin`` --- origin used to compute the moments (optional)
```
FreddieWitherden, line 1279: @pv101 Thoughts on what we should call this?

pyfr/plugins/fluidforce.py
```diff
@@ -171,16 +211,26 @@ def __call__(self, intg):
...
         if rank != root:
-            comm.Reduce(f, None, op=get_mpi('sum'), root=root)
+            comm.Reduce(fm, None, op=get_mpi('sum'), root=root)
         else:
-            comm.Reduce(get_mpi('in_place'), f, op=get_mpi('sum'), root=root)
+            comm.Reduce(get_mpi('in_place'), fm, op=get_mpi('sum'), root=root)
 
             # Write
-            print(intg.tcurr, *f, sep=',', file=self.outf)
+            print(intg.tcurr, *fp, *fv, *mp, *mv, sep=',', file=self.outf)
```
FreddieWitherden, line 233: `f.ravel` might be able to help us out here (may need to refine the shape of f to be its transpose, though).

pyfr/plugins/fluidforce.py
```diff
@@ -117,10 +143,18 @@ def __call__(self, intg):
 
         # Solution matrices indexed by element type
         solns = dict(zip(intg.system.ele_types, intg.soln))
-        ndims, nvars = self.ndims, self.nvars
+        ndims, nvars, mcomp = self.ndims, self.nvars, self._mcomp
+
+        # Force and moment vectors
+        n = 2 if self._viscous else 1
+        fm = np.zeros((ndims + mcomp, n))
```
FreddieWitherden, line 150: Thinking out loud some more, would a 3d array work better (ndims, 1 + moments, 1 + visc) or some rearrangement thereof?

GonzaloSaez: This does not help much to clean the this code since it's difficult to unpack when writing the files. The main issue is that the moments may have different components than the force and therefore 2D/3D arrays are tricky. I think that the following may be cleaner...

```
fm = np.zeros((ndims + mcomp)*(1 + self._viscous))
f, m = np.split(fm, [ndims*(1 + self._viscous)])
fp, fv = np.split(f, 2)
mp, mv = np.split(m, 2)
...
print(intg.tcurr, *fm, sep=',', file=self.outf)
```

pyfr/plugins/fluidforce.py
```diff
@@ -71,6 +84,8 @@ def __init__(self, intg, cfgsect, suffix):
             # Element indices and associated face normals
             eidxs = defaultdict(list)
             norms = defaultdict(list)
+            rfpts = defaultdict(list)
+            rfpts_c_norms = defaultdict(list)
```
FreddieWitherden, line 88: Not sure if pre-computing the cross products is necessarily worth it having seen both.

pyfr/plugins/fluidforce.py
```diff
@@ -144,7 +174,17 @@ def __call__(self, intg):
             norms = self._norms[etype, fidx]
 
             # Do the quadrature
-            f[:ndims] += np.einsum('i...,ij,jik', qwts, p, norms)
+            fp += np.einsum('i...,ij,jik', qwts, p, norms)
+            if self._mcomp:
```
FreddieWitherden, line 178: Is it cleaner to group all of the moment stuff together?

pyfr/plugins/fluidforce.py
```diff
@@ -96,12 +110,20 @@ def __init__(self, intg, cfgsect, suffix):
                 # Unit physical normals and their magnitudes (including |J|)
                 npn = eles.get_norm_pnorms(eidx, fidx)
                 mpn = eles.get_mag_pnorms(eidx, fidx)
+                wnorm = mpn[:, None]*npn
```
FreddieWitherden, line 113: Do we need this variable now?

pyfr/plugins/fluidforce.py
FreddieWitherden: The comment suggests these points are relative, but they look absolute to me?

pyfr/plugins/fluidforce.py
GonzaloSaez: Fixed.

pyfr/plugins/fluidforce.py
GonzaloSaez: Removed it.

pyfr/plugins/fluidforce.py
GonzaloSaez: My bad, I lost `morigin`  at some point. I fixed it

doc/src/user_guide.rst
```diff
@@ -1276,12 +1276,15 @@ file. Parameterised with
...
+
 Example::
 
     [soln-plugin-fluidforce-wing]
     nsteps = 10
     file = wing-forces.csv
     header = true
+    morigin = [0.0, 0.0, 0.5]
```
FreddieWitherden, line 1287: For consistency go with a tuple here.

pyfr/plugins/fluidforce.py
```diff
@@ -71,6 +84,7 @@ def __init__(self, intg, cfgsect, suffix):
             # Element indices and associated face normals
```
FreddieWitherden, line 84: Update comment.

pyfr/plugins/fluidforce.py
```diff
@@ -100,8 +114,16 @@ def __init__(self, intg, cfgsect, suffix):
                 eidxs[etype, fidx].append(eidx)
                 norms[etype, fidx].append(mpn[:, None]*npn)
 
+                # Get the flux points position of the given face and element
+                # indices relative to the moment origin
+                if self._mcomp:
+                    fpts_idx = eles.basis.facefpts[fidx]
+                    rfpt = eles.plocfpts[fpts_idx, eidx] - morigin
+                    rfpts[etype, fidx].append(rfpt)
```
FreddieWitherden, line 122: Will this all go on one line?

pyfr/plugins/fluidforce.py
```diff
@@ -117,10 +139,18 @@ def __call__(self, intg):
...
+
+        # Force and moment vectors
+        n = 2 if self._viscous else 1
+        fm = np.zeros((ndims + mcomp, n))
 
-        # Force vector
-        f = np.zeros(2*ndims if self._viscous else ndims)
+        # Pressure and viscous components
+        fp, mp = fm[:ndims, 0], fm[ndims:, 0]
+        if self._viscous:
```
FreddieWitherden, line 150: Can we avoid defining fv and mv?

pyfr/plugins/fluidforce.py
GonzaloSaez: This is the one-liner respecting the 79 column limit
```python
                    rfpts[etype, fidx].append(
                        eles.plocfpts[eles.basis.facefpts[fidx], eidx] -
                        morigin)
```

I could chop `morigin` to `mo` to fit it in one line.

pyfr/plugins/fluidforce.py
GonzaloSaez: As it was previously with just `f` and `m` for both viscous and pressure forces/moments?

FreddieWitherden: Accumulating into fm[...] directly.

pyfr/plugins/fluidforce.py
```diff
@@ -171,16 +202,37 @@ def __call__(self, intg):
...
+
+                # Do the cross product with the normal vectors
+                rfpts_c_norms = np.atleast_3d(np.cross(rfpts, norms))
+
+                # Pressure force moments
+                mp += np.einsum('i...,ij,jik->k', qwts, p,
+                                rfpts_c_norms)
+
+                if self._viscous:
+                    # Viscous force at each flux point
```
FreddieWitherden, line 219: Normal viscous force?

pyfr/plugins/fluidforce.py
```diff
@@ -117,10 +140,12 @@ def __call__(self, intg):
...
         solns = dict(zip(intg.system.ele_types, intg.soln))
-        ndims, nvars = self.ndims, self.nvars
+        ndims, nvars, mcomp = self.ndims, self.nvars, self._mcomp
 
-        # Force vector
-        f = np.zeros(2*ndims if self._viscous else ndims)
+        # Force and moment vectors
+        n = 2 if self._viscous else 1
+        fm = np.zeros((ndims + mcomp)*n)
+        f, m = np.split(fm, [ndims*n])
```
FreddieWitherden, line 148: No need for split or defining f or m.

pyfr/plugins/fluidforce.py
```diff
@@ -117,10 +140,11 @@ def __call__(self, intg):
...
         # Solution matrices indexed by element type
         solns = dict(zip(intg.system.ele_types, intg.soln))
-        ndims, nvars = self.ndims, self.nvars
+        ndims, nvars, mcomp = self.ndims, self.nvars, self._mcomp
 
-        # Force vector
-        f = np.zeros(2*ndims if self._viscous else ndims)
+        # Force and moment vectors
+        n = 2 if self._viscous else 1
+        fm = np.zeros((ndims + mcomp)*n)
```
FreddieWitherden, line 147: Can we not keep this as a 2D array to simplify the indexing when we use fm?

GonzaloSaez: No that I'm aware since `mcomp` may be different than `ndims`. I tried using lists of arrays but then `mpi4py` may fail because if `mcomp == 0` then one of the arrays has zero length. That's why I tried to use the array splitting or `fp, fv, mp, mv` in the previous commits. It's nicest way to deal with this that I know of. I'm open to suggestions here.

FreddieWitherden: So if `fm = np.zeros(ndims + mcomp, 2 if self._viscous else 1)` then:
 - `fm[:ndims, 0] += np.einsum('i...,ij,jik', qwts, p, norms)`
 - `fm[:ndims, 1] += np.einsum('i...,klij,jil', qwts, vis, norms)`
 - `fm[ndims:, 0] += np.einsum(mop, qwts, p, rfpts_c_norms)`
 - `fm[ndims:, 1] += np.einsum('i,jik->k', qwts, rcf)`

The MPI stuff should work unchanged, and all that needs to change is the output routine.  Although here we can simply do `..., *fm.ravel, ` and correct for the different arrangement by changing the headers.

pyfr/plugins/fluidforce.py
```diff
@@ -171,16 +193,37 @@ def __call__(self, intg):
...
         if rank != root:
-            comm.Reduce(f, None, op=get_mpi('sum'), root=root)
+            comm.Reduce(fm, None, op=get_mpi('sum'), root=root)
         else:
-            comm.Reduce(get_mpi('in_place'), f, op=get_mpi('sum'), root=root)
+            comm.Reduce(get_mpi('in_place'), fm, op=get_mpi('sum'), root=root)
 
             # Write
-            print(intg.tcurr, *f, sep=',', file=self.outf)
+            print(intg.tcurr, *fm.T.ravel(), sep=',', file=self.outf)
```
FreddieWitherden, line 226: Would it be easier to redefine fm as its transpose?

PR #241, opened 2021-08-25
All plugins are called for the first time outside their init function

GiorgioGiangaspero:
This separates the initialisation from the first call so that all plugins are initialised before any of them is actually called. This will allow plugins to interact with each other, if necessary.

### DIFF
--- a/pyfr/integrators/dual/phys/controllers.py
+++ b/pyfr/integrators/dual/phys/controllers.py
@@ -10,8 +10,9 @@ class BaseDualController(BaseDualIntegrator):
         # Solution filtering frequency
         self._fnsteps = self.cfg.getint('soln-filter', 'nsteps', '0')
 
-        # Fire off any event handlers
-        self.completed_step_handlers(self)
+        # Fire off any event handlers if not restarting
+        if not self.isrestart:
+            self.completed_step_handlers(self)
 
     def _accept_step(self, idxcurr):
         self.tcurr += self._dt
--- a/pyfr/integrators/dual/phys/controllers.py
+++ b/pyfr/integrators/dual/phys/controllers.py
@@ -10,6 +10,9 @@ class BaseDualController(BaseDualIntegrator):
         # Solution filtering frequency
         self._fnsteps = self.cfg.getint('soln-filter', 'nsteps', '0')
 
+        # Fire off any event handlers
+        self.completed_step_handlers(self)
+
     def _accept_step(self, idxcurr):
         self.tcurr += self._dt
         self.nacptsteps += 1
--- a/pyfr/integrators/std/controllers.py
+++ b/pyfr/integrators/std/controllers.py
@@ -18,8 +18,9 @@ class BaseStdController(BaseStdIntegrator):
         # Stats on the most recent step
         self.stepinfo = []
 
-        # Fire off any event handlers
-        self.completed_step_handlers(self)
+        # Fire off any event handlers if not restarting
+        if not self.isrestart:
+            self.completed_step_handlers(self)
 
     def _accept_step(self, dt, idxcurr, err=None):
         self.tcurr += dt
--- a/pyfr/integrators/std/controllers.py
+++ b/pyfr/integrators/std/controllers.py
@@ -18,6 +18,9 @@ class BaseStdController(BaseStdIntegrator):
         # Stats on the most recent step
         self.stepinfo = []
 
+        # Fire off any event handlers
+        self.completed_step_handlers(self)
+
     def _accept_step(self, dt, idxcurr, err=None):
         self.tcurr += dt
         self.nacptsteps += 1
--- a/pyfr/plugins/residual.py
+++ b/pyfr/plugins/residual.py
@@ -26,9 +26,6 @@ class ResidualPlugin(BasePlugin):
             # Open
             self.outf = init_csv(self.cfg, cfgsect, ','.join(header))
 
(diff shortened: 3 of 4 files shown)


FreddieWitherden:
Do all the usual tests past in terms of things happening as expected on restart etc?

GonzaloSaez:
> This separates the initialisation from the first call so that all plugins are initialised before any of them is actually called. This will allow plugins to interact with each other, if necessary.

Do you mean interaction during in the initialization process? Interaction of plugins during runtime would require a bit more thought I think.

FreddieWitherden:
> > This separates the initialisation from the first call so that all plugins are initialised before any of them is actually called. This will allow plugins to interact with each other, if necessary.
> 
> Do you mean interaction during in the initialization process? Interaction of plugins during runtime would require a bit more thought I think.

The medium term goal is to allow plugins to save state.  The most natural means of doing this is through adding a method on each plugin which writers can call to get per-plugin state when writing a solution file.  The issue, however, is how this works when you're writing a file at the first time step.  Currently this is accomplished by plugins calling themselves in their constructor.  However, the issue here is that your plugin my be called before all of the other plugins have been constructed (since you are calling yourself from your constructor).  This is a problem as it means we could miss some state.  The idea with this PR is to wait until all plugins have been constructed before calling them for the first time.

GiorgioGiangaspero:
yep

GonzaloSaez:
> The medium term goal is to allow plugins to save state. The most natural means of doing this is through adding a method on each plugin which writers can call to get per-plugin state when writing a solution file. The issue, however, is how this works when you're writing a file at the first time step. Currently this is accomplished by plugins calling themselves in their constructor. However, the issue here is that your plugin my be called before all of the other plugins have been constructed (since you are calling yourself from your constructor). This is a problem as it means we could miss some state. The idea with this PR is to wait until all plugins have been constructed before calling them for the first time.

Makes sense to me, thanks for the explanation. My only issue with this PR would be that some plugins are not prepared to be called before finishing the first step. This may lead to double entries in some files (e.g. the `fluidforce` output). I'd need to double check this though.

GiorgioGiangaspero:
> Do all the usual tests past in terms of things happening as expected on restart etc?

yep

GonzaloSaez:
> > > The medium term goal is to allow plugins to save state. The most natural means of doing this is through adding a method on each plugin which writers can call to get per-plugin state when writing a solution file. The issue, however, is how this works when you're writing a file at the first time step. Currently this is accomplished by plugins calling themselves in their constructor. However, the issue here is that your plugin my be called before all of the other plugins have been constructed (since you are calling yourself from your constructor). This is a problem as it means we could miss some state. The idea with this PR is to wait until all plugins have been constructed before calling them for the first time.
> > 
> > 
> > Makes sense to me, thanks for the explanation. My only issue with this PR would be that some plugins are not prepared to be called before finishing the first step. This may lead to double entries in some files (e.g. the `fluidforce` output). I'd need to double check this though.
> 
> `fluidforce` does not call itself in its `__init__` function so this PR should not make a difference in this regard.

Ok, I thought that `completed_step_handlers` would call the `__call__` function of each plugin.

GiorgioGiangaspero:
> > > > The medium term goal is to allow plugins to save state. The most natural means of doing this is through adding a method on each plugin which writers can call to get per-plugin state when writing a solution file. The issue, however, is how this works when you're writing a file at the first time step. Currently this is accomplished by plugins calling themselves in their constructor. However, the issue here is that your plugin my be called before all of the other plugins have been constructed (since you are calling yourself from your constructor). This is a problem as it means we could miss some state. The idea with this PR is to wait until all plugins have been constructed before calling them for the first time.
> > > 
> > > 
> > > Makes sense to me, thanks for the explanation. My only issue with this PR would be that some plugins are not prepared to be called before finishing the first step. This may lead to double entries in some files (e.g. the `fluidforce` output). I'd need to double check this though.
> > 
> > 
> > `fluidforce` does not call itself in its `__init__` function so this PR should not make a difference in this regard.
> 
> Ok, I thought that `completed_step_handlers` would call the `__call__` function of each plugin.

it does when you call it like this: `self.completed_step_handlers(self)`. It's tricky as it is a `proxylist`

GonzaloSaez:
> > > > > The medium term goal is to allow plugins to save state. The most natural means of doing this is through adding a method on each plugin which writers can call to get per-plugin state when writing a solution file. The issue, however, is how this works when you're writing a file at the first time step. Currently this is accomplished by plugins calling themselves in their constructor. However, the issue here is that your plugin my be called before all of the other plugins have been constructed (since you are calling yourself from your constructor). This is a problem as it means we could miss some state. The idea with this PR is to wait until all plugins have been constructed before calling them for the first time.
> > > > 
> > > > 
> > > > Makes sense to me, thanks for the explanation. My only issue with this PR would be that some plugins are not prepared to be called before finishing the first step. This may lead to double entries in some files (e.g. the `fluidforce` output). I'd need to double check this though.
> > > 
> > > 
> > > `fluidforce` does not call itself in its `__init__` function so this PR should not make a difference in this regard.
> > 
> > 
> > Ok, I thought that `completed_step_handlers` would call the `__call__` function of each plugin.
> 
> it does when you call it like this: `self.completed_step_handlers(self)`. It's tricky as it is a `proxylist`

Makes sense, thanks for the insight!

FreddieWitherden:
So long as no one can observe any unwanted side effects for plugins such as the residual calculation and force integration I am happy to merge.

PR #242, opened 2021-08-25
Allow plugins to save data in the solution file.

GiorgioGiangaspero:
Plugins can now interact with the writer plugin to save data in solution files. Note that each plugin is responsible for its own MPI communication and that the function get_data should return a dictionary. The keys of this dictionary will be the names of the datasets written in the solution file. Still have to figure out how to restore data.

### DIFF
--- a/pyfr/partitioners/base.py
+++ b/pyfr/partitioners/base.py
@@ -316,13 +316,8 @@ class BasePartitioner(object):
 
             # Copy over the metadata
             for f in soln:
-                if re.match('stats|config', f):
+                if re.match('stats|config|plugins.*', f):
                     newsoln[f] = soln[f]
-
-                if re.match('plugins', f):
-                    # this is a group so dive in recursively to get a
-                    # dictionary of datasets
-                    newsoln.update(soln.get_datasets({}, f))
 
             # Apply the new UUID
             newsoln['mesh_uuid'] = newuuid
--- a/pyfr/partitioners/base.py
+++ b/pyfr/partitioners/base.py
@@ -316,7 +316,7 @@ class BasePartitioner(object):
 
             # Copy over the metadata
             for f in soln:
-                if re.match('stats|config|plugins.*', f):
+                if re.match('stats|config|plugins', f):
                     newsoln[f] = soln[f]
 
             # Apply the new UUID
--- a/pyfr/partitioners/base.py
+++ b/pyfr/partitioners/base.py
@@ -264,6 +264,15 @@ class BasePartitioner(object):
 
         return con, eleglmap
 
+    def _get_datasets(self, data, aname, soln):
+        try:
+            data[aname] = soln[aname]
+        except AttributeError:
+            for key in soln._file[aname].keys():
+                nname = f'{aname}/{key}'
+                data = self._get_datasets(data, nname, soln)
+        return data
+
     def partition(self, mesh):
         # Extract the current UUID from the mesh
         curruuid = mesh['mesh_uuid']
@@ -318,6 +327,12 @@ class BasePartitioner(object):
             for f in soln:
                 if re.match('stats|config', f):
                     newsoln[f] = soln[f]
+
+                if re.match('plugins', f):
+                    # this is a group so dive in recursively to get the datasets
+                    data = self._get_datasets({}, f, soln)
+                    for k, v in data.items():
+                        newsoln[k] = v
 
             # Apply the new UUID
             newsoln['mesh_uuid'] = newuuid
--- a/pyfr/partitioners/base.py
+++ b/pyfr/partitioners/base.py
@@ -264,15 +264,6 @@ class BasePartitioner(object):
 
         return con, eleglmap
 
-    def _get_datasets(self, data, aname, soln):
-        try:
-            data[aname] = soln[aname]
-        except AttributeError:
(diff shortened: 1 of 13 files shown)


pyfr/plugins/base.py
```diff
@@ -59,6 +59,9 @@ def __init__(self, intg, cfgsect, suffix=None):
     def __call__(self, intg):
         pass
 
+    def get_data(self):
```
FreddieWitherden, line 62: Maybe something like serialise is a better name?  Also, is there any benefit to passing intg?

pyfr/plugins/writer.py
```diff
@@ -58,6 +65,18 @@ def __call__(self, intg):
         else:
             metadata = None
 
+        # Fetch data from other plugins and add it to metadata with ad-hoc keys
+        for csh in intg.completed_step_handlers:
+            try:
+                opdata = csh.get_data()
+
+                if rank == root:
+                    prefix = self._get_plugin_data_prefix(csh.name, csh.suffix)
```
FreddieWitherden, line 74: I'd get rid of this method as it is only called in one place.

FreddieWitherden:
Don't forget to rebase off of develop.

pyfr/partitioners/base.py
```diff
@@ -316,7 +316,7 @@ def partition_soln(soln):
 
             # Copy over the metadata
             for f in soln:
-                if re.match('stats|config', f):
+                if re.match('stats|config|plugins.*', f):
```
FreddieWitherden, line 319: No need for the .*

pyfr/plugins/writer.py
GiorgioGiangaspero: I'd keep it as we might need it when restoring data

FreddieWitherden: I'm not sure how it could be used, as the method is attached to the writer rather than the base plugin/integrator?

pyfr/plugins/writer.py
```diff
@@ -58,6 +58,18 @@ def __call__(self, intg):
...
             metadata = None
 
+        # Fetch data from other plugins and add it to metadata with ad-hoc keys
+        for csh in intg.completed_step_handlers:
+            try:
+                opdata = csh.serialise(intg)
+
+                if rank == root:
+                    prefix = intg.get_plugin_data_prefix(csh.name, csh.suffix)
+                    metadata.update(
```
GonzaloSaez, line 68: Just asking, will plugin data always belong to `metadata`? It could get confusing if we store more than _metadata_ (i.e solution fields)

pyfr/plugins/writer.py
```diff
@@ -58,6 +58,18 @@ def __call__(self, intg):
         else:
             metadata = None
 
+        # Fetch data from other plugins and add it to metadata with ad-hoc keys
```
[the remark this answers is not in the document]
FreddieWitherden: 79 as per PEP8

pyfr/plugins/writer.py
FreddieWitherden: Yes, if we want to store more substantial amounts of data an additional mechanism (beyond what the writers already have access to) will be needed, as the metadata variable is written out by the root rank and so does not scale well with large amounts of data.

pyfr/plugins/writer.py
```diff
@@ -58,6 +58,18 @@ def __call__(self, intg):
...
+        for csh in intg.completed_step_handlers:
+            try:
+                opdata = csh.serialise(intg)
+
+                if rank == root:
+                    prefix = intg.get_plugin_data_prefix(csh.name, csh.suffix)
+                    metadata.update(
+                        {f'{prefix}/{k}': v for k, v in opdata.items()})
```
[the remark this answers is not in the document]
FreddieWitherden: The `)` should probably be on its own line.  Or prefix/opdata should be shortened.

pyfr/readers/native.py
```diff
@@ -7,7 +7,7 @@
 import h5py
 import numpy as np
 
-from pyfr.util import memoize
+from pyfr.util import memoize, lazyprop
```
FreddieWitherden, line 10: Sort alphabetically.

pyfr/plugins/writer.py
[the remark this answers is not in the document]
GonzaloLuminary: Thanks! My bad on this one @GiorgioGiangaspero this line seems to have 79 characters

pyfr/plugins/writer.py
```diff
@@ -58,6 +58,18 @@ def __call__(self, intg):
...
+            except AttributeError:
```
FreddieWitherden, line 70: Move the except up and then in the failover block do `opdata = {}` this ensures that our try/except block is as small as possible.

pyfr/plugins/writer.py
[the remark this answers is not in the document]
GonzaloSaez: Thanks! I was wrong with the number of chars, it has 79! my bad @GiorgioGiangaspero

FreddieWitherden:
For restoring the data we have a few options.  Might be worth prototyping a couple of ideas and seeing what is cleanest.

FreddieWitherden:
Superseded by 243.

PR #243, opened 2021-09-06
Allow plugins to save/restore data to/from the solution file.

GiorgioGiangaspero:


pyfr/plugins/writer.py
```diff
@@ -58,6 +58,18 @@ def __call__(self, intg):
         else:
             metadata = None
 
+        # Fetch data from other plugins and add it to metadata with ad-hoc keys
+        for csh in intg.completed_step_handlers:
+            try:
+                pdata = csh.serialise(intg)
+                prefix = intg.get_plugin_data_prefix(csh.name, csh.suffix)
```
FreddieWitherden, line 65: This can be moved out of the try as it should always succeed.  Also may want to have the method take `csh` as an argument and let it figure out what info it wants to use.

GiorgioGiangaspero: This belongs to PR242. It is not going to always succeed as `csh` may not be a plugin. And yep, I changed the method to take `csh` as an argument.

pyfr/integrators/base.py
```diff
@@ -64,18 +64,33 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
...
-    def _get_plugins(self):
+    def _get_plugins(self, initsoln):
         plugins = []
 
         for s in self.cfg.sections():
             if (m := re.match('soln-plugin-(.+?)(?:-(.+))?$', s)):
                 cfgsect, name, suffix = m.group(0), m.group(1), m.group(2)
 
+                data = {}
+                if initsoln:
```
FreddieWitherden, line 75: is not None

pyfr/integrators/base.py
```diff
@@ -64,18 +64,33 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
...
+                    # get the plugin data stored in the solution, if any
+                    prefix = self.get_plugin_data_prefix(name, suffix)
+                    for f in initsoln:
+                        if re.match(prefix + '/', f):
```
FreddieWitherden, line 79: Should be able to use f.startswith(f'plugins/{prefix}/')

pyfr/integrators/base.py
```diff
@@ -64,18 +64,33 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
...
+                            data[f.split('/')[-1]] = initsoln[f]
```
FreddieWitherden, line 80: Use a positive index to be robust against plugins which want to use sub-groups.

pyfr/integrators/base.py
```diff
@@ -64,18 +64,33 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
...
-                plugins.append(get_plugin(name, self, cfgsect, suffix))
+                plugins.append(get_plugin(name, self, cfgsect, suffix, **data))
 
         return plugins
 
+    @staticmethod
+    def get_plugin_data_prefix(plugin):
+        prefix = f'plugins/{plugin.name}'
+        if plugin.suffix:
+            prefix += f'-{plugin.suffix}'
```
FreddieWitherden, line 91: Avoid using += on strings.  if/else two returns is probably easier.

pyfr/plugins/writer.py
```diff
@@ -58,6 +58,18 @@ def __call__(self, intg):
         else:
             metadata = None
 
+        # Fetch data from other plugins and add it to metadata with ad-hoc keys
+        for csh in intg.completed_step_handlers:
+            try:
+                pdata = csh.serialise(intg)
+                prefix = intg.get_plugin_data_prefix(csh)
```
FreddieWitherden, line 65: Define prefix first

pyfr/plugins/writer.py
```diff
@@ -58,6 +58,18 @@ def __call__(self, intg):
...
+            except AttributeError:
+                pdata = {}
+                prefix = ''
```
FreddieWitherden, line 68: No need to define prefix, as if pdata is empty the offending block which uses prefix will never execute.

pyfr/readers/native.py
```diff
@@ -36,10 +36,22 @@ def __getitem__(self, aname):
...
 
     def __len__(self):
-        return len(self._file)
+        return len(self._keys)
+
+    @lazyprop
+    def _keys(self):
+        keys = set()
+
+        def visitor(name, item):
```
FreddieWitherden, line 48: Worth verifying that this does not cause us to read the entire data set in.  Can test easily enough with a large solution and seeing how long it takes to get the keys.

pyfr/plugins/writer.py
GiorgioGiangaspero: fixed in PR242

pyfr/plugins/writer.py
GiorgioGiangaspero: fixed in PR242

pyfr/integrators/base.py
GiorgioGiangaspero: fixed in PR242

pyfr/integrators/base.py
```diff
@@ -64,18 +64,34 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
...
-                plugins.append(get_plugin(name, self, cfgsect, suffix))
+                plugins.append(get_plugin(name, self, cfgsect, suffix, **data))
 
         return plugins
 
+    @staticmethod
+    def get_plugin_data_prefix(name, suffix):
+        prefix = f'plugins/{name}'
+        if suffix:
+            return prefix + f'-{suffix}'
```
FreddieWitherden, line 91: No need to define prefix, saves a line and avoids the +.

pyfr/readers/native.py
GiorgioGiangaspero: `print([key for key in sol.keys()])` with a 10Gb solution file returned almost instantly

FreddieWitherden:
Code looks good to me.  Am happy to merge.  However, first are you okay to update your turbulent inflow PR to use this new code to verify that it adds all the necessary functionality?

GiorgioGiangaspero:
I tested it with a dummy plugin, but yes, it's worth trying it with the turbulent inflow as I have to do it anyway. Bear with me.

pyfr/integrators/base.py
```diff
@@ -64,18 +64,33 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
...
+    def _get_plugins(self, initsoln):
         plugins = []
 
         for s in self.cfg.sections():
             if (m := re.match('soln-plugin-(.+?)(?:-(.+))?$', s)):
                 cfgsect, name, suffix = m.group(0), m.group(1), m.group(2)
 
+                data = {}
+                if initsoln is not None:
+                    # get the plugin data stored in the solution, if any
```
FreddieWitherden, line 76: Get

pyfr/plugins/writer.py
FreddieWitherden: I'd be tempted to force all plugins to be subclasses of BasePlugin.  I think the only one which isn't is the progress bar, but this is easy enough to update.

GiorgioGiangaspero:
Tested this PR with my turbulent inflow. All the functionalities I need are there and makes the code much cleaner. I think this PR is ready for merging.

PR #245, opened 2021-09-14
Enable 2D thread blocks for 2D kernels.

FreddieWitherden:


semi-h (approved):

WillTrojak:
I can test this on the case we discussed in the next couple of days.

WillTrojak:
Tested this on the SD7003 that Niki performed and the run time on 32 V100s for one flow over chord is ~01:38:00, compared to ~01:44:00. I'm happy if you merge.

PR #247, opened 2021-10-06
fix a metis partitioner error when mesh has only one partition

mhawwary:
Changes to account for the case when the mesh has very few elements (not any multiple of 64, min n subelements). In this case the mesh is not partitioned and there is an exception to handle the case of Metis returning wrong indecies for one partition.

### DIFF
--- a/pyfr/partitioners/base.py
+++ b/pyfr/partitioners/base.py
@@ -158,8 +158,8 @@ class BasePartitioner(object):
             nsp = len(svetimap) // self.nsubeles + 1
 
             # Partition the graph
-            if nsp==1:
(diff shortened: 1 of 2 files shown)


pyfr/partitioners/base.py
```diff
@@ -158,7 +158,10 @@ def _renumber_verts(self, mesh, vetimap, vparts):
             nsp = len(svetimap) // self.nsubeles + 1
 
             # Partition the graph
-            svparts = self._partition_graph(sgraph, [1]*nsp)
+            if nsp==1:
```
FreddieWitherden, line 161: Spaces around ==

PR #248, opened 2021-10-08
Delete multi-p elements maps explicitly after plugins have been constructed.

GiorgioGiangaspero:
This allows plugins to interact with the elements maps (and the corresponding kernels) of the lower multi-p levels.


FreddieWitherden:
I do not think that having plugins directly probe the inner workings of multi-p is a good design.  Rather, we should try and come up with a more formal API for handling this (plugins needing to interact with all of the systems) which is likely to be less fragile.

PR #253, opened 2021-12-05
Fix an issue with the tetrahedral alpha optimised points.

FreddieWitherden:


pv101:
Do the experiments that you were doing all check out now i.e. do face solutions now depend on the expected number of (and indeed expected) solution points?

FreddieWitherden:
Yes.

PR #255, opened 2021-12-21
Simplify stream/queue management for GPU backends.

FreddieWitherden:
By no longer attempting to run copies in parallel with packing kernels we are able to greatly simplify the queue management for GPU backends.  This overlap never did much for performance due to the packing kernels being an order of magnitude faster than the copies.  These simplifications are an important first step in enabling PyFR to migrate to a task graph API in the future.


FreddieWitherden:
Benchmarks by Semih and Tarik show a ~10-15% slowdown for the CUDA backend when MPI is not CUDA aware.

FreddieWitherden:
@tarikdzanic Would you be able to test this with a case which uses AV?

tarikdzanic:
Benchmark performance or just test if it works as expected?

FreddieWitherden:
> Benchmark performance or just test if it works as expected?

Just checking if it all still works as expected.

tarikdzanic:
Looks like it works normally

PR #262, opened 2022-01-02
Enhance the point sampler plugin.

FreddieWitherden:
The plugin now makes use of Newton's method to invert the mapping function, enabling precise sampling (compared with the current approach of simply selecting the closest solution point).

### DIFF
--- a/pyfr/plugins/sampler.py
+++ b/pyfr/plugins/sampler.py
@@ -4,12 +4,13 @@ import numpy as np
 
 from pyfr.mpiutil import get_comm_rank_root, get_mpi
 from pyfr.plugins.base import BasePlugin, init_csv
+from pyfr.quadrules import get_quadrule
 
 
-def _closest_upts_bf(eupts, pts):
+def _closest_pts_bf(epts, pts):
     for p in pts:
         # Compute the distances between each point and p
-        dists = [np.linalg.norm(e - p, axis=2) for e in eupts]
+        dists = [np.linalg.norm(e - p, axis=2) for e in epts]
 
         # Get the index of the closest point to p for each element type
         amins = [np.unravel_index(np.argmin(d), d.shape) for d in dists]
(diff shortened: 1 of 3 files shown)


pyfr/plugins/sampler.py
```diff
@@ -121,7 +214,7 @@ def _process_samples(self, samps):
             samps = self.elementscls.con_to_pri(samps.T, self.cfg)
             samps = np.array(samps).T
 
-        return samps.tolist()
+        return np.ascontiguousarray(samps, dtype=float)
```
GonzaloLuminary, line 217: Just asking, is the use of float intentional here to reduce the file size?

GonzaloLuminary (approved):

pyfr/plugins/sampler.py
FreddieWitherden: `dtype=float` refers to a Python float which is double (`np.float` is now deprecated).  Ensuring everything is a double is important as we're now using Gatherv which depends on the precise type of the array.

pyfr/plugins/sampler.py
```diff
@@ -76,35 +102,69 @@ def __init__(self, intg, cfgsect, suffix):
...
+
+            # Note what rank is responsible for the point
+            if rank == root:
+                ptsrank.append(mrank)
+
+        # Refine
+        self._ourpts = ourpts = self._refine_pts(intg, elepts)
+
+        # Send the refined sample locations to the root rank
+        ptsplocs = comm.gather([pl for et, ei, pl, op in ourpts], root=root)
```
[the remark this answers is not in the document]
FreddieWitherden: Maybe.  The other option was not to unpack and do `[pt[2] for pt in ourpts]`. I am undecided.

FreddieWitherden:
One final fix I should probably make is ensure that the points used for closest upts search do not contain any points on the element boundary.  In cases where the solution point closest to the requested sample point is a boundary point it is possible for the wrong rank to be chosen, which in turn will cause the Newton iteration to go out of bounds.

Practically, this means hard-coding a set of solution point names for each element type which are known to be good and then explicitly finding their physical locations rather than using those associated with the actual solution points.

GonzaloLuminary:
Would starting the Newton iteration using the element center as initial guess simplify the issue that you are describing?

FreddieWitherden:
> Would starting the Newton iteration using the element center as initial guess simplify the issue that you are describing?

The issue occurs when the closest solution point to a sample point is inside of a different element to the sample point itself.  The easiest way to trigger this is to use solution points which go out to the boundary and then pick a  sample point very close to an element boundary.  Here, the two closest points will be on top of each other (boundary solution points from different elements).  Thus, there is a 50/50 chance that the solution point inside the wrong element will be found to be closet (triggering the issue).

FreddieWitherden:
This should now be resolved.  There are still some edge cases where the wrong element can be identified, but they should be few and far between.

PR #266, opened 2022-01-18
Switch to a task graph interface.

FreddieWitherden:
This is a WIP PR to migrate PyFR to a task graph interface for scheduling kernels.  Using a task graph opens up opportunities for reducing Python overhead (CUDA, HIP, OpenMP) and enabling parallel execution of kernels (CUDA, HIP, OpenCL).  All of this improves strong scaling.

Most of the code is in place, with the PR itself being blocked on ROCm 5 which has the required graph embedding API (until then only GiMMiK is functional on HIP).

### DIFF
--- a/pyfr/backends/base/__init__.py
+++ b/pyfr/backends/base/__init__.py
@@ -6,4 +6,5 @@ from pyfr.backends.base.kernels import (BaseKernelProvider,
                                         Kernel, MetaKernel, NotSuitableError,
                                         NullKernel)
 from pyfr.backends.base.types import (ConstMatrix, Matrix, MatrixBase,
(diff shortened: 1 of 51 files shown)


FreddieWitherden:
A few points on the structuring of the tasks graphs.

Firstly, if a kernel overwrites data which is used by another kernel, then it is necessary for these other kernels to be listed as dependencies of the kernel which is doing the overwriting.

Secondly, some kernels have false dependencies on various MPI packing kernels.  This is done to ensure that these kernels are executed sooner rather than later; something which is important if we are to overlap communication with computation.

FreddieWitherden:
Blocked on https://github.com/ROCm-Developer-Tools/HIP/issues/2450

PR #269, opened 2022-02-21
Added mod to nputils

WillTrojak:
This is to help with this question: https://pyfr.discourse.group/t/output-euler-vortex-super-accuracy/529/7


FreddieWitherden:
Does % not work?

FreddieWitherden:
Just following up here.

PR #272, opened 2022-03-09
change advance_to step condition

mhawwary:
Force "advance_to" to only advance if the required time is dtmin away from tcurr


FreddieWitherden:
@WillTrojak @semi-h This will be a fiddly critter, would appreciate your input.

semi-h (approved):
I rarely encountered the advance time is in the past error. And the reason was always this edge case where tcurr is smaller than target t, but tcurr+dtmin is bigger than that. I think this PR solves the issue.

WillTrojak:
I'm not quite sure what problem this is solving? We now go from overshooting tend by [0, dt_min] to undershooting it by [-dt_min, 0]. This will probably effect the solution writer for many cases as the final solution probably won't get output.

mhawwary:
This change is to prevent pyfr from taking dtmin steps if the requested time 't' is not dtmin or more away from the tcurr.

For writers, this will mean they will still write their outputs at this tcurr and not miss any steps. 

Imagine a situation where t-tcurr = 1e-16 and dtmin = 1e-12, in this case with this new change, pyfr will ignore the request to advance and will not do a new step. Instead it will just write at tcurr as if it is the requested t. This should not affect any plugin since anyway you were going to overshoot the time point requested and tcurr is closer to t than if you take tcurr+dtmin. Basically dtmin become your time tolerance.

WillTrojak:
Sure, but it will also stop cases such, tend=2, dtmin=1e-3, then the run could stop at t=1.9991. This case could when using adaptive time stepping. If the data output frequency is every t=1, then you won't get a solution at t=2 as the current time won't surpass t=2. Obviously this is an extreme example, but you could contrive the same thing with more reasonable number. 

Ultimately this change will turn the time stepper from guaranteed overshoot to guaranteed undershoot of tend. This can impact any plugin that has a fixed frequency output.

mhawwary:
I see, but in these cases the user should expect an error in advance time that is at most of dtmin, this is the time tolerance he agreed on. If he wants a tighter tolerance he can simply adjust dtmin. Whereas the current situation and bad edge case provide the user with no solution or control to prevent it. I encountered that case where a t-tcurr = 1e-16 and the solver has taken dtmin=1e-12. 

If the same dtmin is used in adaptive time stepping, then we may introduce a t_tol that is the same as dtmin for non controllers and different for adaptive time. In adaptive time, the logic behind dtmin I believe is to prevent the controller from taking very small time-steps and to have a lower bound on dt.

After all, needing to increase tend to catch that last time step is less severe and can be noticed and corrected by the user easily by increasing his tend a little bit or doing one more step.

WillTrojak:
Ok, in that case isn't an error at startup when tend-tcurr<dtmin preferable? In all other cases I think overshoot is better than undershoot.

mhawwary:
The solution to undershoot vs overshoot problem is that the user can now control his tolerance which in case of none controller is dtmin and in case of adaptive time-step we can make it t_tol and again in this case the user has control over overshoot and undershoot by adjusting his tolerances. 

So, probably in the example you provided the user will restart from another t = 1.9 or so and then by adjusting his tolerance he can still make it to the time step he desires for writers. And if the default t_tol  is 1e-12 then your case will always pass as usual with no change. 

This is my proposal here and will leave it to you and Freddie to decide.

FreddieWitherden:
So if I understand everything correctly with this PR we will go from a guaranteed overshoot to a guaranteed undershoot with any discrepancy being within dtmin.  The advantage of this approach is that we can sometimes avoid the need for a small time step?

mhawwary:
@FreddieWitherden , yes, but I think that for more safety with adaptive time-stepping we may need to introduce `t_tol` and keep `dtmin` only as a lower bound for the controller otherwise we may just accept undershoots.

FreddieWitherden:
The benefits seem marginal.  Does this fix any actual bugs we've observed?

mhawwary:
In my case, if I am using none controller with dt = 3.125e-4 and writer dtout = 0.0125 you can experience this edge case of dt=1e-16 despite the fact that your dtout = 40 dt but round-off errors kick in. Also, if you are doing interpolation between t and tcurr with 1e-16 difference the situation can be bad. 

Anyway, if you think it is marginal that is ok, I can withdraw it for now.

WillTrojak:
Ok, so the real issue isn't with dtmin, but epsilon?

FreddieWitherden:
> In my case, if I am using none controller with dt = 3.125e-4 and writer dtout = 0.0125 you can experience this edge case of dt=1e-16

Given `dt-min` defaults to 1e-12 I am unsure how the described edge case can occur.

mhawwary:
> Ok, so the real issue isn't with dtmin, but epsilon?

yes sort of.

mhawwary:
> > In my case, if I am using none controller with dt = 3.125e-4 and writer dtout = 0.0125 you can experience this edge case of dt=1e-16
> 
> Given `dt-min` defaults to 1e-12 I am unsure how the described edge case can occur.

Because of round of errors. In the above comment I meant writer requested to advance_to `t = tcurr + 1.e-16`, and of course **pyfr** will do `dt=dtmin=1e-12` and compute at `t = tcurr + 1e-12`.

FreddieWitherden:
So I think the underlying issue is that we are making very large and dramatic changes to the time step in order to hit specific target times.  Here, it would appear as if the correct solution is for the controllers to gradually modify dt many steps in advance of the proposed target time.  Slight changes to dt made well in advance should enable us to hit any target time without any sudden changes (either increasing or decreasing).

PR #274, opened 2022-03-11
[WIP] FWH solver plugins using frequency domain formulations

mhawwary:
This is WIP with more testing underway we may need to revisit the surface identification method.
Can handle both solid and permeable surface formulations in frequency domain.


FreddieWitherden:
Please reformat in accordance with PEP8.

pyfr/nputil.py
```diff
@@ -126,3 +127,187 @@ def npdtype_to_ctypestype(dtype):
...
+            zip(c[::-1], d[:0:-1], u[:0:-1]), start=1):
+        dd[i] -= ci*dj
+        uu[i] -= ci*uj
+
+    #compute the final solution
+    fact = (d[0] + alpha*d[-1]/gam)/(1.0 + u[0] + alpha*u[-1]/gam)
+    d -= fact*u[:, None]
+
+    
+class CurveFit:
```
FreddieWitherden, line 203: This class is pointless.

pyfr/nputil.py
```diff
@@ -126,3 +127,187 @@ def npdtype_to_ctypestype(dtype):
...
+        pass
+
+    def intrp(self, *args):
+        pass
+
+class LinearFit(CurveFit):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self._interp = np.vectorize(np.interp, signature='(r),(n),(m)->(r)')
```
FreddieWitherden, line 214: Why is this needed?  Why not call the relevant function in a loop inside interp?

pyfr/nputil.py
```diff
@@ -126,3 +127,187 @@ def npdtype_to_ctypestype(dtype):
...
+        if compute:
+            if not np.any(xs):
+                xs, ys, _ = self.spl
+            self._construct(xs, ys, self._bctype)
+
+    def _construct(self, xs, ys,  bctype='not-a-knot'):
+        self._xs = xs
+        self._ys = ys
+        a, b, c, d = self._spl_matrix_coeffs()
+        if bctype == 'periodic':
```
FreddieWitherden, line 240: If statements like this are generally considered bad form.  Use a dict to dispatch (or a full subclass if the type never changes).

pyfr/nputil.py
```diff
@@ -126,3 +127,187 @@ def npdtype_to_ctypestype(dtype):
...
+        self._zs = d
+
+    def intrp(self, x_intrp, xs=None, ys=None):
+        if np.any(xs):
+            self.update(xs, ys, self._bctype)
+        return self._spleval(x_intrp)
+
+    @property
+    def spl(self):
+        return tuple((self._xs, self._ys, self._zs))
```
FreddieWitherden, line 253: This line is pointless; `(self._xs, self._ys, self._zs)` is a tuple so passing it to `tuple` does nothing.  Probably better just to have `self.spl = (x, y, z)` or something along those lines.

pyfr/nputil.py
```diff
@@ -126,3 +127,187 @@ def npdtype_to_ctypestype(dtype):
...
+  
+    def _spl_matrix_coeffs(self):
+        t, y = self._xs, self._ys
+        n = np.asarray(t).shape[0]
+        b = np.zeros(n, dtype=t.dtype)
```
FreddieWitherden, line 258: Why are we passing dtype here?  Highly unusual.

pyfr/nputil.py
```diff
@@ -126,3 +127,187 @@ def npdtype_to_ctypestype(dtype):
...
+
+        #preparing the core part of the matrix
+        b[1: -1] = 2*(dx[: -1] + dx[1: ])  #diag
+        a = dx.copy() #lower diag
+        c = dx.copy() #upper diag
+        d[1: -1] = 6*(slope[1: ] - slope[ :-1]) #rhs
+        d[0] = slope[0]
+        d[-1] = slope[-1]
+
+        return [a, b, c, d]
```
FreddieWitherden, line 275: Why a list here?

pyfr/nputil.py
```diff
@@ -126,3 +127,187 @@ def npdtype_to_ctypestype(dtype):
...
+
+    @property
+    def spl(self):
+        return tuple((self._xs, self._ys, self._zs))
+  
+    def _spl_matrix_coeffs(self):
+        t, y = self._xs, self._ys
+        n = np.asarray(t).shape[0]
+        b = np.zeros(n, dtype=t.dtype)
+        d = np.zeros(y.shape, dtype=y.dtype)
```
FreddieWitherden, line 259: np.zeros_like

pyfr/nputil.py
```diff
@@ -126,3 +127,187 @@ def npdtype_to_ctypestype(dtype):
...
+        idxs = np.array([np.max(np.nonzero(xx >= t[:-1] - 1e-12)[0]) 
+                                                        for xx in xarr])
+    else:
+        idxs = np.max(np.nonzero(xarr >= t[:-1] - 1e-12)[0])
+    return idxs
+
+#Modified Gauss Elimination/Thompson algorithm for tridiag systems
+#All changes are done in place and the algorithm 
+#is applied to the lhs and rhs together
+def trislv(a, b, c, d):
```
FreddieWitherden, line 144: Probably cleaner to rework the arguments as `trislv(abc, rhs)` as the first three go together.

pyfr/nputil.py
```diff
@@ -126,3 +127,187 @@ def npdtype_to_ctypestype(dtype):
...
+    dd = d[::-1]
+    for i, (ci, dj) in enumerate(zip(c[::-1], d[:0:-1]), start=1):
+        dd[i] -= ci*dj
+
+
+#tridiagonal solver for cyclic/periodic systems 
+#Using Sherman-Morrison formula and Thompson algorithm
+# A = A + u ⊗ v
+# u, v are vectors
+def trislvcyc(a, b, c, d, alpha, beta):
```
FreddieWitherden, line 166: It's been a while, but you should be able to have this call the regular `trislv` internally.

pyfr/plugins/__init__.py
```diff
@@ -10,6 +10,7 @@
 from pyfr.plugins.sampler import SamplerPlugin
 from pyfr.plugins.tavg import TavgPlugin
 from pyfr.plugins.writer import WriterPlugin
+from pyfr.plugins.acoustics import FwhSolverPlugin
```
FreddieWitherden, line 13: Order alphabetically.

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
+# -*- coding: utf-8 -*-
+
+from collections import defaultdict, namedtuple
+from mpi4py import MPI
```
FreddieWitherden, line 4: Can not safely import anything from mpi4py at the global namespace. Make all sub imports local (get_mpi can help).

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+from pyfr.nputil import CubicSplineFit, fuzzysort, LinearFit, npeval
+from pyfr.plugins.base import BasePlugin
+from pyfr.writers.native import NativeWriter
+
+
+def open_csv(fname, header=None, mode='w'):
+    # Open for mode
+    outf = open(fname, mode)
+    # Output a header if required
+    if os.path.getsize(fname) == 0 and header:
```
FreddieWitherden, line 21: Reuse `init_csv`, modifying if required to suit your needs.  Code duplication is bad.

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+    return outf
+
+def write_fftdata(fname, *indata, header='', mode='a'):
+    #open the file
+    outf = open_csv(fname, ''.join(header), mode)
+    #prepare the data array
+    data = np.array(indata)
+    dshape = data.shape
+    #do write
+    for i in range(dshape[1]):
```
FreddieWitherden, line 32: `for d in data.T:`

Nine times out of ten, if you are explicitly looking at the shapes of arrays, or using range based for loops, your solution is not optimal.

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+    # Open for mode
+    outf = open(fname, mode)
+    # Output a header if required
+    if os.path.getsize(fname) == 0 and header:
+        print(header, file=outf)
+    return outf
+
+def write_fftdata(fname, *indata, header='', mode='a'):
+    #open the file
+    outf = open_csv(fname, ''.join(header), mode)
```
FreddieWitherden, line 27: Bad form.  If the lifetime of a file can be bounded you should use `with` to ensure it is always closed.

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+    #prepare the data array
+    data = np.array(indata)
+    dshape = data.shape
+    #do write
+    for i in range(dshape[1]):
+        print(*data[:, i].ravel(), sep=' ', file=outf)
+    outf.flush()
+    return
```
FreddieWitherden, line 35: None return is implicit.

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+
+def realfft(udata_):
+        dsize = udata_.shape[-1]
+        # freqsize: size of half the spectra with positive frequencies
+        freqsize = int(dsize/2) if dsize%2 == 0 else int(dsize-1)/2
```
FreddieWitherden, line 40: `freqsize = dsize // 2` (Python has truncation division as part of the language.)

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+    #prepare the data array
+    data = np.array(indata)
+    dshape = data.shape
+    #do write
+    for i in range(dshape[1]):
+        print(*data[:, i].ravel(), sep=' ', file=outf)
+    outf.flush()
+    return
+
+def realfft(udata_):
```
FreddieWitherden, line 37: Function is mostly pointless and can likely be inlined.

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+        dsize = udata_.shape[-1]
+        # freqsize: size of half the spectra with positive frequencies
+        freqsize = int(dsize/2) if dsize%2 == 0 else int(dsize-1)/2
+        ufft = np.fft.rfft(udata_)/freqsize 
+        return ufft
+
+def compute_spl(pref, amp, df=None):
+    sqrt2 = 1./np.sqrt(2.)
```
FreddieWitherden, line 45: 2**-0.5.  Python has built in support for exponentiation and NumPy scalars should be avoided.

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+def realfft(udata_):
+        dsize = udata_.shape[-1]
+        # freqsize: size of half the spectra with positive frequencies
+        freqsize = int(dsize/2) if dsize%2 == 0 else int(dsize-1)/2
+        ufft = np.fft.rfft(udata_)/freqsize 
+        return ufft
+
+def compute_spl(pref, amp, df=None):
+    sqrt2 = 1./np.sqrt(2.) 
+    lgsqrt2 = 20.*np.log10(sqrt2)
```
FreddieWitherden, line 46: No need for `.`

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+    factor = sqrt2/pref if not df else sqrt2/(pref*np.sqrt(df))
+    spl = 20.*np.log10(factor*amp) 
+    spl[0] -= lgsqrt2
+    spl[-1] -= lgsqrt2
+    #overall sound pressure level
+    ospl_sum = sum(pow(10., spl/10.)) if not df else df*sum(pow(10., spl/10.))
```
FreddieWitherden, line 52: `10**(spl / 10)`.  Use `np.sum`, with the `axis=` directive if you need to sum a NumPy array rather than generic sum.

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+    return
+
+def realfft(udata_):
+        dsize = udata_.shape[-1]
+        # freqsize: size of half the spectra with positive frequencies
+        freqsize = int(dsize/2) if dsize%2 == 0 else int(dsize-1)/2
+        ufft = np.fft.rfft(udata_)/freqsize 
+        return ufft
+
+def compute_spl(pref, amp, df=None):
```
FreddieWitherden, line 44: I think the function could be made quite a bit simpler if `df` defaulted to 1.

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+    oaspl = 10.* np.log10(ospl_sum) 
+    return spl, oaspl
+
+def compute_psd(amp, df):
+    psd = 0.5*amp*amp/df
+    psd[0] *= 2.
+    psd[-1] *= 2.
+    return psd
+
+def Gatherv_data_arr(comm, rank, root, data_arr):
```
FreddieWitherden, line 62: This function is silly.  The purpose of the capital letter methods in mpi4py is to reduce overhead at the cost of additional complexity.  But here, we get the worse of both worlds: the inefficiency of the lower case method with the complexity of the upper case method.  Just use `comm.gather(data_arr, root=root)` unless you're going to have the root rank cache the lengths in advance.

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+    if rank == root:
+        count = np.array(count, dtype='i')
+        recvbuf = np.empty(sum(count), dtype=float)
+        displ = np.array([sum(count[:p]) for p in range(len(count))])
+
+    comm.Gatherv(sendbuf, [recvbuf, count, displ, MPI.DOUBLE], root=root)
+
+    return recvbuf
+
+def Allgatherv_data_arr(comm, data_arr):
```
FreddieWitherden, line 79: See above.

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+        self.basename = self.cfg.get(self.cfgsect, 'basename')
+        
+        # read time and fft parameters
+        self.tstart = self.cfg.getfloat(cfgsect, 'tstart', intg.tstart)
+        self.dtsub = self.cfg.getfloat(cfgsect, 'dt', intg._dt)
+        # read the window length, default is up to simulation tend
+        ltsub_def = intg.tend - self.tstart
+        self.ltsub = self.cfg.getfloat(cfgsect, 'Lt', ltsub_def)        
+        shift = self.cfg.getfloat(cfgsect, 'time-shift', 0.5)
+        window_func = self.cfg.get(cfgsect, 'window', 'hanning')
```
FreddieWitherden, line 166: Check valid, or better still don't expose the option.

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+        kk = 1j*kwv*nR + np.einsum('i,ij->ij', invRs, nRs)
+        dphy = - np.einsum('i,ij->ji', phy, kk) 
+        mdotdphy = sum((Mo*dphy.T).T)
+        return phy, dphy, mdotdphy
+
+class DipoleSrc(PointAcousticSrc):
+    def __init__(self, name, srclocs, flowinfo, *srcdata):
+        super().__init__(name, srclocs, flowinfo, *srcdata)
+
+class VTUSurfWriter(object):
```
FreddieWitherden, line 1780: I'm guessing this is for debugging purposes?

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+        pfft =  np.einsum('ijk,k->ij', pfreq, self.qdA)
+        pfft /= (4.*np.pi)
+        if self.window:
+            pfft *= self.windscale
+        
+        return pfft
+
+
+class PointAcousticSrc(object):
+    def __init__(self, name, srclocs, flowinfo, *srcdata):
```
FreddieWitherden, line 1701: `*srcdata` pointless.

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+        self.nfreq = len(self.freq)
+        
+    def acoustic_psoln(self, xyz_ob):
+        nobserv = xyz_ob.shape[0]
+        ptime = np.empty((self.nt, nobserv))
+        pfft = np.empty((nobserv, self.nfreq), dtype=np.complex64)
+
+        tarr = np.arange(0, self.nt)*self.dt
+        for i, ti in enumerate(tarr):
+            ptime[i] = self._comput_exact_psoln(xyz_ob, ti, ptime[i])
```
FreddieWitherden, line 1739: This smells funny (mutable on LHS and RHS).

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+
+        ptime = np.moveaxis(ptime, 0, 1)
+        pfft = realfft(ptime)
+
+        return self.freq, pfft
+
+    def _comput_exact_psoln(self, xyz_ob, tcurr, psoln):
```
FreddieWitherden, line 1746: Unused argument.

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+        usoln[-1] = p + self.pinf
+        usoln[0] = p/(co*co) + self.rhoinf  #rho
+        usoln[1: self.ndims+1] = np.array([ui + uo for ui, uo in
+                                    zip(np.real(dphy), self.uinf)]) #u
+
+    def _comput_vel_potentials(self, xyz_ob, tcurr):
+        Mo = self.uinf/self.co
+        kwv = self.omega/self.co
+        xyz_src = self.srclocs
+        mR, mRs, nR, nRs = FwhFreqDomainSolver.compute_distance_vars(xyz_src,
```
FreddieWitherden, line 1765: Fragile.  Prefer super().

pyfr/nputil.py
mhawwary: both do the same thing, either a loop or using vectorize.

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+                                                                    xyz_ob, Mo)
+        # phy_potential of the source and its derivatives
+        amp_4piRs = self.ampl/(4.*np.pi*mRs)
+        phy = amp_4piRs*np.exp(1j*(self.omega*tcurr - kwv*mR)) 
+        invRs = 1./mRs
+        kk = 1j*kwv*nR + np.einsum('i,ij->ij', invRs, nRs)
```
FreddieWitherden, line 1771: Is einsum needed here?  There do not appear to be any contractions?

pyfr/nputil.py
mhawwary: The point is to encapsulate different types of interpolation, linear and spline, etc.

pyfr/nputil.py
mhawwary: just a typo in the last commit.

pyfr/nputil.py
mhawwary: I tried it but It is complicated and this is faster. I can do two calls to trislv but then it is not as faster as solving both systems in one set of loops.

pyfr/plugins/acoustics.py
mhawwary: used several times and this is better wrapper to suit our needs. In the future you can be performing more rfft's in another acoustic solver such as time domain or so. Less code writing.

pyfr/plugins/acoustics.py
mhawwary: sure.

pyfr/plugins/acoustics.py
mhawwary: sure

pyfr/plugins/acoustics.py
mhawwary: sure

pyfr/plugins/acoustics.py
mhawwary: Yes

pyfr/plugins/acoustics.py
mhawwary: Debug analytical src and different sources could be initialized with different inputs.

FreddieWitherden: Sure, but why pass a load of junk to the base class which it is going to do nothing with!?!

pyfr/plugins/acoustics.py
```diff
@@ -0,0 +1,2106 @@
...
+
+    def acoustic_psoln(self, *args):
+        pass
+    def update_usoln(self, *args):
+        pass
+
+class MonopoleSrc(PointAcousticSrc):
+    def __init__(self, name, srclocs, flowinfo, ampl, srcfreq, ntime, nperiods,
+                 gamma):
+        super().__init__(name, srclocs, flowinfo, ampl, srcfreq, ntime,
```
FreddieWitherden, line 1718: Why pass a load of unused arguments to the base class?

pyfr/plugins/acoustics.py
mhawwary: This function Gatherv is used in multiple places and it is more compact to have a method for it here. Also, the arrays I send and receive are variable so "v" is needed and they are numpy arrays and I prefer to work with them and be able to manipulate the shape easier.

pyfr/nputil.py
FreddieWitherden: But it encapsulates nothing!  There is no common code factored out.

pyfr/plugins/acoustics.py
FreddieWitherden: Read my comment again.  This entire function is just a poor reimplementaton of `comm.gather`.  Every use of it can be replaced with `comm.gather(data_arr, root=root)` and will deliver similar performance.  And the lower case `.gather` works with everything (Python objects, NumPy arrays, whatever).

pyfr/plugins/acoustics.py
mhawwary: Just initializing common vars for any point source which are its location and the flowinfo. 
Also, this is just for ease of testing and will probably be removed at the end so no point of modifying it right now.

pyfr/plugins/acoustics.py
mhawwary: can be improved for sure.

pyfr/plugins/acoustics.py
mhawwary: You mean make it a subclass of the frequency domain solver. In this case it is not, just a static method that can be used by others such as analytical sources but analytical sources are not subclasses of the fwh solver.

pyfr/nputil.py
mhawwary: It makes it easy to always just call Class.interp without worrying about which interp type is used in the fwh solver. It is just decided at initialization if linear or others. The CurveFit is also made available for future development if needed to add more and have common parts. Since I am just offering linear and cubic spline for now there is no common code in base class.

pyfr/nputil.py
```diff
@@ -6,6 +6,7 @@
 import re
 
 import numpy as np
+from pyfr.util import memoize
```
FreddieWitherden, line 9: Unused import.

pyfr/nputil.py
FreddieWitherden: Your system is A*x = b = (A' + u*v^T)*x = b where A' is tridiagonal and u and v are vectors whose outer product gives the periodic components.  Whence solve A'*[y z ] = [u b] which is a tridiagonal system and should be *one* call to `trislv` with a stacked RHS.  Then x = z - (y*v^T*z)/(1 + v^T*y).  Entire function should only be four to five lines of code with roughly similar performance  to what your hand unrolled version delivers.  Moreover, this approach also enables the cyclic routine to automatically benefit from any fixes/improvements in `trislv`. 

As a general point any more complex solution to a problem must justify is existence by either providing additional functionality to the user, or demonstrably improved performance for the user.  Given this function is not in the critical path it is highly unlikely that even a five-fold increase in runtime would be noticeable.  Thus go with the simpler solution.

pyfr/nputil.py
FreddieWitherden: Sure, but that does not require a class hierarchy.  Python is duck typed.  You are free to call any method on any object you like and so long as a particular object implements the method in question everything will work.

Developing something now with the view that it will be of utility in the future is seldom a sensible proposition.  You're imposing additional complexity now (for everyone reading the code) in the hope that it will be somehow beneficial in the future.  It is much better to keep things lightweight and only later, when more specifics are known, try and factor out common code.

pyfr/nputil.py
mhawwary: Tried it before and I know it can be done by stacking but I just tried to use the same in-place concept. I do not think you still can keep trislv with in-place but I will retry again. Otherwise, I can modify trislv not to be inplace. Let me know if you have an idea to do it in-place or I can just modify trislv not to be in-place. I am open to suggestions here.

pyfr/nputil.py
mhawwary: I started with it as a bunch of functions but then thought this can be helpful. Anyway, I will convert all to just functions unless you have another suggestion.

pyfr/plugins/acoustics.py
mhawwary: Thought about using init_csv but it will need to be changed from using filekey to fname.

pyfr/nputil.py
FreddieWitherden: You can still have it work in place easily enough.  Something like
```
def trislvcyc(a, b, c, d, alpha, beta):
    ud = np.hstack([d, np.zeros((len(d), 1))])
    ud[[0, -1], -1] = -b[0], alpha
    trislv(a, b, c, ud)
    
    d[:] = ud[:, :-1] - ...
```

pyfr/nputil.py
FreddieWitherden: If you need state it is best to have an object which overloads `__call__`.  Then if no state is needed a bare function can work, and if state is needed then you can create an instance of an object which can be used as if it were a function.

pyfr/plugins/acoustics.py
mhawwary: Removed most of the capital letter mpi communications. Can change to capital letter after testing if needed with the help of cached counts and displs.

pyfr/nputil.py
mhawwary: done

pyfr/nputil.py
mhawwary: Done, let me know if you have other ideas.

pyfr/nputil.py
mhawwary: Done, let me know if this is what you meant.

pyfr/nputil.py
```diff
@@ -126,3 +128,160 @@ def npdtype_to_ctypestype(dtype):
...
+    trislv(abc, ud)
+    d[:] = ud[:, :-1]
+    u = ud[:, -1]
+
+    #compute the final solution
+    fact = (d[0] - alpha*d[-1]/b[0])/(1.0 + u[0] - alpha*u[-1]/b[0])
+    d -= fact*u[:, None]
+
+
+def linintrp(x_intrp, xs, ys):
```
FreddieWitherden, line 184:
```
def interpv(x, xp, yp):
    idx = np.interp(x, xp, np.arange(len(xp)))
    lidx = np.floor(idx).astype(int)
    ridx = np.minimum(lidx + 1, len(xp) - 1)
    
    t = np.reshape(lidx + 1 - idx, (-1,) + (1,)*(yp.ndim - 1))
    return t*yp[lidx] + (1 - t)*yp[ridx]
```

PR #278, opened 2022-04-10
Elide some kernels at p = 0.

FreddieWitherden:
This avoids some matrix multiplication operators where A = 0, which causes issues for libxsmm.  It also results in a very slight reduction in overhead, as we now run fewer kernels at p = 0.


WillTrojak (commented):
I think there is some more abstraction that can be used to handle this more cleanly.

PR #283, opened 2022-05-10
Added tags for periodic boundary faces.

tarikdzanic:


pyfr/partitioners/base.py
```diff
@@ -89,6 +92,9 @@ def offset_con(con, pr):
         for k, v in bccon.items():
             newmesh[f'bcon_{k}_p0'] = np.hstack(v).astype(dtype)
 
+        for k, v in pccon.items():
```
FreddieWitherden, line 95: Can merge with above loop as:
```python
    for pfx, con in [('bcon', bcon), ('pccon', pccon)]:
        for k, v in con.items():
            newmesh[f'{pfx}_{k}_p0'] = np.hstack(v).astype(dtype)
```

pyfr/partitioners/base.py
```diff
@@ -215,6 +221,7 @@ def _partition_con(self, mesh, vetimap, vparts):
         con_px = defaultdict(list)
         con_pxpy = defaultdict(list)
         bcon_px = defaultdict(list)
```
FreddieWitherden, line 223: bpccon = {'bcon': defaultdict(list), 'pccon': defaultdict(list)}

pyfr/partitioners/base.py
```diff
@@ -243,12 +250,15 @@ def _partition_con(self, mesh, vetimap, vparts):
 
         # Generate boundary conditions
         for f in filter(lambda f: isinstance(f, str), mesh):
-            if (m := re.match('bcon_(.+?)_p0$', f)):
+            if (m := re.match('[bp]con_(.+?)_p0$', f)):
```
FreddieWitherden, line 253: ([bp]con)_

pyfr/partitioners/base.py
```diff
@@ -243,12 +250,15 @@ def _partition_con(self, mesh, vetimap, vparts):
...
                 for lpetype, leidxg, lfidx, lflags in mesh[f].tolist():
                     lpart, leidxl = eleglmap[lpetype, leidxg]
                     conl = (lpetype, leidxl, lfidx, lflags)
 
-                    bcon_px[m[1], lpart].append(conl)
+                    if 'bcon' in m.string:
+                        bcon_px[m[1], lpart].append(conl)
```
FreddieWitherden, line 259: ```bpcon[m[1]][m[2], lpart].append(conl)```

pyfr/partitioners/base.py
```diff
@@ -265,6 +275,9 @@ def _partition_con(self, mesh, vetimap, vparts):
         for (etype, px), v in bcon_px.items():
             con[f'bcon_{etype}_p{px}'] = np.array(v, dtype=dtype)
 
+        for (etype, px), v in pcon_px.items():
```
FreddieWitherden, line 278: Can merge with above.

pyfr/readers/base.py
```diff
@@ -184,7 +180,25 @@ def get_connectivity(self):
...
+        pbc, pcon = {}, defaultdict(list)
+
+        for k, v in self._pfacespents.items():
+            pbc[f'periodic_{k}_l'] = bpart[v[0]] 
+            pbc[f'periodic_{k}_r'] = bpart[v[1]] 
+
+        for lpent in pbc:
+            for pftype in pbc[lpent]:
+                lfnodes = pbc[lpent][pftype]
+                lfpts = np.array([[self._nodepts[n] for n in fn] for fn in lfnodes])
```
FreddieWitherden, line 193: Check line length

FreddieWitherden:
Made a few minor remarks, but otherwise looks good.

FreddieWitherden:
Thinking out loud one of the things I would like to add in the future is support for axi-symmetric BCs.  Here, the periodic faces are not simple translations, which adds some extra issues in terms of processing (as LHS and RHS have different normal vectors).

My key idea for resolving this is to process such interfaces with separate kernels.  This effort is greatly simplified if periodic boundaries are always internal (never straddling partitions).  We can accomplish this by merging nodes in the graph which have periodic edges.  In turn, this would allow the periodic arrays to be stored as simple attributes on `con_p<x>` which list the indices of periodic face pairs.  This also avoids confusion around what is included in pcon.

pyfr/readers/base.py
```diff
@@ -184,7 +180,25 @@ def get_connectivity(self):
...
+        # Tag periodic boundaries
+        pbc, pcon = {}, defaultdict(list)
+
+        for k, v in self._pfacespents.items():
+            pbc[f'periodic_{k}_l'] = bpart[v[0]] 
+            pbc[f'periodic_{k}_r'] = bpart[v[1]] 
+
+        for lpent in pbc:
+            for pftype in pbc[lpent]:
+                lfnodes = pbc[lpent][pftype]
```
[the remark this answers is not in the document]
tarikdzanic: Was following syntax from [L140](https://github.com/PyFR/PyFR/blob/2b3af04e12cc8730d12ace045a37f8ddf46b9fcf/pyfr/readers/base.py#L140), updated now

FreddieWitherden:
Does #285 satisfy your needs here?

tarikdzanic:
> Does #285 satisfy your needs here?

It should work with some modifications of the recirculation plugin

FreddieWitherden:
Okay.  In terms of the plugin itself please follow up with @pv101 who has been working on a new class of plugin which can inject code and the such like into kernels.  This is probably what you want to use for your recirculation plugin.

PR #285, opened 2022-05-14
Tag periodic connectivity in the mesh file.

FreddieWitherden:
Also, ensure that periodic faces are never split across partition boundaries.


mhawwary (approved):
I think this is a very helpful PR, especially the idea of having periodic pairs on the same mpi rank.

PR #293, opened 2022-07-13
Modified controller parameters

rishit2307:
New alpha is 0.58, new beta is 0.42

### DIFF
--- a/pyfr/integrators/std/controllers.py
+++ b/pyfr/integrators/std/controllers.py
@@ -105,8 +105,8 @@ class StdPIController(BaseStdController):
             raise ValueError('Invalid error norm')
 
         # PI control values
-        self._alpha = self.cfg.getfloat(sect, 'pi-alpha', 0.7)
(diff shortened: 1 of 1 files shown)


rishit2307:
I haven't tested the local error estimator with the new parameters. I can do that. 

Attached are the benchmarks.


[controller_benchmark.xlsx](https://github.com/PyFR/PyFR/files/9098428/controller_benchmark.xlsx)

FreddieWitherden:
Please check with an incompressible case which uses the local-pi controller to see if these new parameters work there, too.  Also squish everything down into a single commit.

rishit2307:
I tested a few different values of the safety factor for TGV case, but haven't observed any substantial improvements when compared to the default values. The sweet spot there seems to lie in the range of 0.81~0.82. I am attaching a file which compares different values of the safety-factor.

[Safety-fact-tgv-comparison.xlsx](https://github.com/PyFR/PyFR/files/9133591/Safety-fact-tgv-comparison.xlsx)

PR #294, opened 2022-07-18
Change controller parameters for std time stepping

rishit2307:
Parameters alpha, beta are changed for the standard controller. New default for alpha is 0.58, and beta is 0.42. Previous defaults for alpha and beta were 0.7 and 0.4 respectively. I am attaching a file that compares the improvements observed by changing values of alpha, beta

[controller_benchmark.xlsx](https://github.com/PyFR/PyFR/files/9133434/controller_benchmark.xlsx)

### DIFF
--- a/pyfr/integrators/std/controllers.py
+++ b/pyfr/integrators/std/controllers.py
@@ -105,8 +105,8 @@ class StdPIController(BaseStdController):
             raise ValueError('Invalid error norm')
 
         # PI control values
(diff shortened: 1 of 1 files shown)


rishit2307:
For dual time stepping, [https://github.com/PyFR/PyFR/blob/develop/pyfr/integrators/dual/pseudo/pseudocontrollers.py#L99](url) I tried one of cases with modified values of alpha and beta. I didn't find any improvements there by modifying alpha and beta.

PR #296, opened 2022-07-27
Cubin over PTX

NickLeaf:
Defaults CUDA JIT compilation to cubin instead of PTX. The compiled PTX can sometimes fail to load when the CUDA version used to JIT compile it doesn't match the driver--e.g. when running inside a container. Compiling directly to cubin avoids this issue.

PTX compilation is still available, and can be selected by changing the prefix of the `--gpu-architecture` flag to `compute_` instead of `sm_`.


pyfr/backends/cuda/compiler.py
```diff
@@ -79,18 +81,27 @@ def compile(self, name, src, flags=[]):
 
                 raise RuntimeError(log.value.decode())
 
-            # Fetch the program size
-            ptxsz = c_size_t()
-            self.lib.nvrtcGetPTXSize(prog, ptxsz)
+            codesz = c_size_t()
+            if len([f for f in flags if 'sm_' in f]) > 0:
```
FreddieWitherden, line 85: Does it make sense to always try and get the cubin (size) first, and if that fails (returns 0) ask for the PTX?  Saves going through the flags and having to think about if later flags can supersede pervious ones.

pyfr/backends/cuda/compiler.py
```diff
@@ -103,18 +114,18 @@ def __init__(self, backend, src):
...
         flags += shlex.split(backend.cfg.get('backend-cuda', 'cflags', ''))
 
-        # Compile to PTX
-        ptx = backend.nvrtc.compile('kernel', src, flags)
+        # Compile to CUDA code (either PTX or CUBIN, depending on arch flag)
+        cucode = backend.nvrtc.compile('kernel', src, flags)
 
         # Load it as a module
-        self.mod = backend.cuda.load_module(ptx)
+        self.mod = backend.cuda.load_module(cucode)
```
FreddieWitherden, line 128: Check if the `load_module` method needs its first argument changing from `ptx`.

NickLeaf: The docs for [cuModuleLoadDataEx](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MODULE.html#group__CUDA__MODULE_1g9e8047e9dbf725f0cd7cafd18bfd4d12) indicate that it should take cubin or PTX interchangeably, but I think you're referring to just the variable naming convention, here. I'll refactor the references to PTX in `load_module` in order to avoid confusion.

pyfr/backends/cuda/compiler.py
NickLeaf: Yeah, defaulting to cubin and falling back to PTX makes sense to me, and should be a bit more robust. I was trying to be as minimally invasive as possible in case there was a reason to favor PTX that I wasn't aware of. I'll change it to default to cubin/fallback to PTX.

FreddieWitherden: I do not recall any specific reason why I went for PTX when I first wrote the code, so assuming it does not cause any issues lets go for requesting a cubin by default.

PR #299, opened 2022-08-09
Enable automatic reuse of constant matrices.

FreddieWitherden:


pyfr/backends/base/backend.py
```diff
@@ -120,8 +122,17 @@ def _malloc_impl(self, nbytes):
...
-    def const_matrix(self, initval, extent=None, tags=set()):
-        return self.const_matrix_cls(self, initval, extent, tags)
+    def const_matrix(self, initval, dtype=None, tags=set()):
+        dtype = dtype or self.fpdtype
+
+        # See if we have previously allocated an identical matrix
+        for m in self.mats.values():
+            if (isinstance(m, self.const_matrix_cls) and
+                m.dtype == dtype and m.ioshape == initval.shape and
+                tags.issubset(m.tags) and (m.get() == initval).all()):
```
semi-h, line 132: Why are tags a consideration here and why is having a subset of them enough?

FreddieWitherden: If you pass some tags in, you should expect to get them back out as .tags.  Subset is there since const matrices add the const tag automatically.

PR #308, opened 2022-09-16
Cleanups and documentation updates.

FreddieWitherden:


doc/src/performance_tuning.rst
```diff
@@ -175,7 +207,16 @@ prerequisites must be satisfied:
...
+`h5py parallel HDF5 example <https://docs.h5py.org/en/stable/mpi.html#using-parallel-hdf5-from-h5py>`_.
+
+Plugins
+=======
+
+A common source of performance issues is running plugins too
+frequently.  Given the time steps taken by PyFR are typically much
+smaller than those associated with the underlying physics there is
+seldom any benefit to running integration and/or time average
+accumulation plugins more frequently than once every 50 steps.
```
WillTrojak, line 219: Maybe also add a comment about the nan check plugin having significant overhead in the current implementation as well. It can be a silent killer to performance.

WillTrojak (commented):
Generally fine, maybe just a comment about the Nan checker as we've been burnt by that before.

FreddieWitherden:
Updated.

PR #312, opened 2022-10-05
HIP Graph Support

FreddieWitherden:
This PR adds full support for HIP graphs, bringing the HIP backend on par with the CUDA backend.  However, whereas for simple test cases overhead decreased by a factor of ~2 when moving to graphs, for HIP overhead increases by ~2.  The performance impact for non-trivial cases still needs to be assessed, however.


FreddieWitherden:
Still broken with ROCm 6.

PR #307, opened 2022-09-14
Added in situ visulisation plugin using Ascent

WillTrojak:

PR #313, opened 2022-10-10
Feature/global libs

WillTrojak:
Added support for library load config, splitting functionality from #307.

### DIFF
--- a/pyfr/ctypesutil.py
+++ b/pyfr/ctypesutil.py
@@ -10,9 +10,10 @@ class LibWrapper:
     _libname = None
     _statuses = None
     _functions = None
+    _mode = ctypes.DEFAULT_MODE
 
     def __init__(self):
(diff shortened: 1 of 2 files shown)


pyfr/ctypesutil.py
```diff
@@ -44,24 +45,28 @@ def get_libc_function(fn):
...
     # If an explicit override has been given then use it
     lpath = os.environ.get(f'PYFR_{name.upper()}_LIBRARY_PATH')
     if lpath:
-        return ctypes.CDLL(lpath)
+        return ctypes.CDLL(lpath, mode=mode)
 
     # Otherwise synthesise the library name and start searching
     lname = platform_libname(name)
 
-    # Check our search paths
```
FreddieWitherden, line 56: Why is this code being reverted?  It was a recent commit to fix a Spack issue.

PR #320, opened 2022-11-21
Added command line interface to plugins

WillTrojak:
Plugins can now expose functions to be called from the commadn line. This is to allow for existing plugin infrastucture to be used to perform addition analysis tasks in pre- and post-processing, rather than relying on adhoc tools.

A new cli section has been added to the documentation. Currently the exposed plugin functions are:

* tavg merge --- allowing multiple time average files to be merged. 
* writer export --- this moves the export funtionality from a major command to a subcommand of the writer.

### DIFF
--- a/doc/src/cli.rst
+++ b/doc/src/cli.rst
@@ -65,3 +65,11 @@ Plugin Commands
 Plugins are able to define tools that are accessible from the command line,
 below are listed the the various options currently available.
 
+Time Average
+^^^^^^^^^^^^
+
+1. ``pyfr tavg merge`` --- merge together multiple time average files into a
+   single time average file. The averaging times are read from the file and do
+   not need to be evenly spaced in time. Example::
+
+        pyfr tavg merge avg-1.00.pyfrs avg-2.00.pyfrs avg-10.00.pyfrs merged_avg.pyfrs
--- a/pyfr/plugins/__init__.py
+++ b/pyfr/plugins/__init__.py
@@ -9,7 +9,7 @@ from pyfr.plugins.pseudostats import PseudoStatsPlugin
 from pyfr.plugins.residual import ResidualPlugin
 from pyfr.plugins.sampler import SamplerPlugin
 from pyfr.plugins.source import SourcePlugin
-from pyfr.plugins.tavg import TavgPlugin
+from pyfr.plugins.tavg import TavgCLIPlugin, TavgPlugin
 from pyfr.plugins.writer import WriterPlugin
 from pyfr.util import subclass_where
 
--- a/pyfr/plugins/base.py
+++ b/pyfr/plugins/base.py
@@ -1,3 +1,4 @@
+import functools as ft
 import re
 import shlex
 
@@ -6,6 +7,15 @@ from pytools import prefork
 
 from pyfr.mpiutil import get_comm_rank_root
 from pyfr.regions import BoundaryRegion, ConstructiveRegion
+
+
+def cli_external(meth):
+    @ft.wraps(meth)
+    def newmeth(cls, args):
+        obj = cls(args)
+        return meth(obj)
+
+    return classmethod(newmeth)
 
 
 def init_csv(cfg, cfgsect, header, *, filekey='file', headerkey='header'):
--- a/pyfr/plugins/tavg.py
+++ b/pyfr/plugins/tavg.py
@@ -1,12 +1,17 @@
+from collections import defaultdict
 import re
 
+import h5py
 import numpy as np
 
 from pyfr.inifile import Inifile
 from pyfr.mpiutil import get_comm_rank_root, mpi
 from pyfr.nputil import npeval
-from pyfr.plugins.base import BaseSolnPlugin, PostactionMixin, RegionMixin
+from pyfr.plugins.base import (BaseSolnPlugin, BaseCLIPlugin, cli_external,
+                               PostactionMixin, RegionMixin)
+from pyfr.readers import NativeReader
 from pyfr.writers.native import NativeWriter
+from pyfr.util import merge_intervals
 
 
 class TavgPlugin(PostactionMixin, RegionMixin, BaseSolnPlugin):
@@ -104,7 +109,7 @@ class TavgPlugin(PostactionMixin, RegionMixin, BaseSolnPlugin):
                 self.outfields.append(f'std-{k[4:]}')
 
             for k in cfg.items(cfgsect, prefix='fun-avg-'):
-                self.outfields.append(f'fun-std-{k[4:]}')
+                self.outfields.append(f'std-fun-{k[4:]}')
 
     def _init_gradients(self):
         # Determine what gradients, if any, are required
@@ -357,3 +362,250 @@ class TavgPlugin(PostactionMixin, RegionMixin, BaseSolnPlugin):
                     self.tstart_acc = intg.tcurr
 
                 self.tout_last = intg.tcurr
+
+
+class TavgCLIPlugin(BaseCLIPlugin):
+    name = 'tavg'
+
+    @classmethod
+    def add_cli(cls, parser):
+        cmd = cls.name
+        cp = parser.add_parser(cmd, help=f'{cmd} --help')
+        sp = cp.add_subparsers(dest='sub_cmd')
+
+        # Merge command
+        ssp = sp.add_parser('merge', help=f'{cmd} merge --help')
+        ssp.set_defaults(process=cls.merge_cli)
+        ssp.add_argument('solns', nargs='*', help='Averages to merge')
+        ssp.add_argument('output', help='Output file name')
+
+    @cli_external
+    def merge_cli(self):
+        # Open all the solution files
+        self._prepare_files(self.args.solns)
+
+        # Initialise things needed for the merge
+        self._init_tavg_merge()
+
+        with h5py.File(self.args.output, 'w') as outf:
+            # Merge the averages and stds
+            self._merge_data(outf)
+
+            # Merge the metadata
+            self._merge_meta(outf)
+
+    def _eval_fun_exprs(self, avars):
+        # Prepare the substitution dictionary
+        subs = dict(zip(self.anames, avars.swapaxes(0, 1)))
+
+        # Evaluate the function and return
+        return np.stack([npeval(v, subs) for v in self.fexprs], axis=1)
+
+    def _eval_fun_var(self, std, avars):
+        dfexpr = []
+
+        # Evaluate the function
+        fexprs = self._eval_fun_exprs(avars)
+
+        for avi in avars.swapaxes(0, 1):
+            # Calculate step size for finite difference
+            h = self.delta_h*np.abs(avi)
+            h[np.where(h == 0)] = self.delta_h
+
+            # Calculate derivatives for functional averages
+            avi += h
+            dfexpr.append((self._eval_fun_exprs(avars) - fexprs) / h[:,None])
+            avi -= h
+
+        # Multiply by variance and take RMS value
+        fv = np.linalg.norm(np.array(dfexpr)*std.swapaxes(0, 1)[:,:,None],
+                            axis=0)
+        return fexprs, fv
+
+    def _init_tavg_merge(self):
+        self.delta_h = np.finfo(np.float64).eps**0.5
+
+        for i, (file, stats, cfg, dt) in enumerate(self.files):
+            fields = stats.get('data', 'fields').split(',')
+            cfgsect = stats.get('tavg', 'cfg-section')
+
+            # Mapping for expression to sorted list
+            aidx = [i for i, v in enumerate(fields) if v.startswith('avg-')]
+            anames = [fields[i].removeprefix('avg-') for i in aidx]
+            amap = [i for n, i in sorted(zip(anames, aidx),
+                                         key=lambda x: x[0])]
+
+            sidx = [i for i, v in enumerate(fields) if
+                    re.match(r'\bstd-(?!fun-avg-)', v)]
+            snames = [fields[i].removeprefix('std-') for i in sidx]
+            smap = [i for n, i in sorted(zip(snames, sidx),
+                                         key=lambda x: x[0])]
+
+            if i == 0:
+                self.stats, self.cfg = stats, cfg
+
+                prec = cfg.get('backend', 'precision')
+                self.dtype = np.float32() if prec == 'single' else np.float64()
+
+                self.fields = fields
+                self.cfgsect = cfgsect
+                self.region = cfg.get(cfgsect, 'region')
+                self.uuid = file['mesh_uuid']
+
+                self.idxs = [k for k in file if re.search(r'_idxs_p\d+$', k)]
+                self.dkeys = [k for k in file if
+                              re.match(r'tavg_[^\W_]+_p\d+$', k)]
+
+                self.mapping = [(amap, smap)]
+                fnames = [x.removeprefix('fun-avg-') for x in fields if
+                          x.startswith('fun-avg-')]
+                self.anames = sorted(anames)
+                self.fnames = sorted(fnames)
+
+                # Initialise std max and avg registers
+                if self.std_all:
+                    self.std_max,self.std_avg  = np.zeros((2, len(anames)))
+                    self.fstd_max, self.fstd_avg = np.zeros((2, len(fnames)))
+
+                # Build fun avg expressions
+                c = cfg.items_as('constants', float)
+                self.fexprs = [cfg.getexpr(cfgsect, f'fun-avg-{k}', subs=c)
+                               for k in fnames]
+            else:
+                self.mapping.append((amap, smap))
+
+            # Check for compatibility of files
+            if self.uuid != file['mesh_uuid']:
+                raise RuntimeError('Average files computed on different '
+                                   'meshes')
+            if self.region != cfg.get(cfgsect, 'region'):
+                raise RuntimeError('Average files computed on different '
+                                   'regions')
+
+            for k in cfg.items(cfgsect, prefix='avg-'):
+                if cfg.get(cfgsect, k) != self.cfg.get(self.cfgsect, k):
+                    raise RuntimeError('Different average field definitions')
+
+    def _merge_data(self, outf):
+        file0, stats, cfg, dt = self.files[0]
+        amap0, smap0 = self.mapping[0]
+
+        for key in self.dkeys:
+            # Initialise accumulators
+            data = file0[key]
+            avg_acc = np.zeros_like(data[:, amap0])
+            var_acc = np.zeros_like(data[:, smap0])
+
+            # Perform the rest of the accumulations
+            t = 0
+            for files, mapping in zip(self.files, self.mapping):
+                file, stats, cfg, dt = files
+                amap, smap = mapping
+                data = file[key]
+
+                if self.std_all:
+                    var_acc = ((t*var_acc + dt*data[:, smap]**2)/(t + dt) +
+                               dt/(t + dt)**2*(avg_acc - t*data[:, amap])**2)
+                avg_acc += dt*data[:, amap]
+                t += dt
+
+            # Get standard deviation
+            if self.std_all:
+                var_acc = var_acc**0.5
+                np.maximum(self.std_max, np.amax(var_acc, axis=(0, 2)),
+                           out=self.std_max)
+                self.std_avg += np.mean(var_acc, axis=(0, 2))
+
+            # Evaluate function expression and write out
+            if self.fexprs and self.std_all:
+                fun_avg, fun_std = self._eval_fun_var(var_acc, avg_acc)
+                np.maximum(self.fstd_max, np.amax(fun_std, axis=(0, 2)),
+                           out=self.fstd_max)
+                self.fstd_avg += np.mean(fun_std, axis=(0, 2))
+
+                outstack = (avg_acc, fun_avg, var_acc, fun_std)
+            elif self.fexprs:
+                fun_avg = self._eval_fun_exprs(avg_acc)
+                outstack = (avg_acc, fun_avg, var_acc)
+            else:
+                outstack = (avg_acc, var_acc)
+
+            outf[key] = np.hstack(outstack, dtype=self.dtype)
+
+    def _merge_meta(self, outf):
+        cfgsect = self.stats.get('tavg', 'cfg-section')
+        self.stats.set('tavg', 'merged-from', self.merged_from)
+        self.stats.set('tavg', 'range', self.merged_range)
+
+        # If all files have full std stats then these can be writen
+        if self.std_all:
+            self.std_avg = self.std_avg.astype(self.dtype, copy=False)
+            self.std_max = self.std_max.astype(self.dtype, copy=False)
+            for n, avg, max in zip(self.anames, self.std_avg, self.std_max):
+                self.stats.set('tavg', f'avg-std-{n}', avg)
+                self.stats.set('tavg', f'max-std-{n}', max)
+
+            for n, avg, max in zip(self.fnames, self.fstd_avg, self.fstd_max):
+                self.stats.set('tavg', f'avg-std-fun-{n}', avg)
+                self.stats.set('tavg', f'max-std-fun-{n}', max)
+        # Otherwise the std summary has to be removed
+        else:
+            self.cfg.set(cfgsect, 'std-mode', 'none')
+            for opt in self.stats.items('tavg'):
+                if opt.startswith(('avg-std', 'max-std')):
+                    self.stats.remove_option('tavg', opt)
+
+        # Write out the region index data
+        file0, stats, cfg, dt = self.files[0]
+        for k in self.idxs:
+            outf[k] = file0[k]
+
+        # Write re-ordered data fields
+        fields = [f'avg-{x}' for x in self.anames]
+        fields.extend(f'fun-avg-{x}' for x in self.fnames)
+        if self.std_all:
+            fields.extend(f'std-{x}' for x in self.anames)
+            fields.extend(f'std-fun-avg-{x}' for x in self.fnames)
+        self.stats.set('data', 'fields', ','.join(fields))
+
+        # Merge runtime stats
+        sect = 'solver-time-integrator'
+        rt_stats = defaultdict(lambda: [])
+
+        for file, stats, cfg, dt in self.files:
+            for k in self.stats.items(sect):
+                rt_stats[k].append(stats.getliteral(sect, k))
+
+        for k, v in rt_stats.items():
+            self.stats.set(sect, k, v)
+
+        # Write out
+        outf['config'] = self.cfg.tostr()
+        outf['stats'] = self.stats.tostr()
+        outf['mesh_uuid'] = self.uuid
+
+    def _prepare_files(self, filenames):
+        self.files = files = []
+        twindows, std = [], []
+        for filename in filenames:
+            f = NativeReader(filename)
+            stats, cfg = Inifile(f['stats']), Inifile(f['config'])
+            cfgsect = stats.get('tavg', 'cfg-section')
+            std.append(cfg.get(cfgsect, 'std-mode'))
+
+            twind = stats.getliteral('tavg', 'range')
+            twindows.extend(twind)
+            dt = sum(tend - tstart for tstart, tend in twind)
+
+            files.append((f, stats, cfg, dt))
+
+        self.merged_from = twindows
+        self.std_all = all(s == 'all' for s in std)
+
+        try:
+            self.merged_range = merge_intervals(twindows)
+        except ValueError:
+            raise RuntimeError('Overlapping averge time ranges in files')
+
+        self.avg_time = at = sum(t[1] - t[0] for t in self.merged_range)
+        self.files = [(*fsc, dt / at) for *fsc, dt in files]
--- a/pyfr/util.py
+++ b/pyfr/util.py
@@ -96,6 +96,25 @@ class silence:
         os.close(self.saved_fds[1])
 
 
+def merge_intervals(ivals, tol=1e-5):
+    ivals = sorted(ivals, reverse=True)
+    mivals = [ivals.pop()]
+
+    while ivals:
+        lstart, lend = mivals[-1]
+        cstart, cend = ivals.pop()
+
+        if cstart < lend:
+            raise ValueError('Overlapping range')
+
+        if abs(cstart - lend) < tol:
+            mivals[-1] = (lstart, cend)
+        else:
+            mivals.append((cstart, cend))
+
+    return mivals
+
+
 @contextmanager
 def setenv(**kwargs):
     _env = os.environ.copy()
--- a/pyfr/inifile.py
+++ b/pyfr/inifile.py
@@ -120,6 +120,9 @@ class Inifile:
 
         return iv
 
+    def remove_option(self, section, option):
+        self._cp.remove_option(section, option)
+
     def sections(self):
         return self._cp.sections()
 
--- a/pyfr/readers/__init__.py
+++ b/pyfr/readers/__init__.py
@@ -1,5 +1,6 @@
 from pyfr.readers.base import BaseReader, NodalMeshAssembler
 from pyfr.readers.gmsh import GmshReader
+from pyfr.readers.native import NativeReader
 
 from pyfr.util import subclasses, subclass_where
 


pyfr/__main__.py
```diff
@@ -108,6 +87,14 @@ def main():
         p.add_argument('--progress', '-p', action='store_true',
                        help='show a progress bar')
 
+    # Plugin commands
+    for scls in subclasses(BasePlugin):
```
FreddieWitherden, line 91: I would have a separate class hierarchy for this such as `BaseCLIPlugin` as this avoids awkwardness with `__init__`.

pyfr/__main__.py
```diff
@@ -118,6 +105,16 @@ def main():
         ap.print_help()
 
 
+def plugin_cli_parser(sp, plugin):
+    cmd = plugin.name
+    for scmd_cfg in plugin.cli:
+        scmd, proc = scmd_cfg[0], getattr(plugin , scmd_cfg[1])
+        ssp = sp.add_parser(scmd, help=f'{cmd} {scmd} --help')
+        ssp.set_defaults(process=proc)
+        for args in scmd_cfg[2:]:
```
FreddieWitherden, line 114: I'd let the plugin be responsible for configuring this object.  Gives it more control.

pyfr/plugins/tavg.py
```diff
@@ -220,6 +226,46 @@ def _acc_avg_var(self, intg, currex):
             if self.tstart_acc != self.prevt:
                 v +=  (Wmp1mpn / Wp * (a - W1mpn * ppc)**2)
 
+    @staticmethod
+    def merge_cli(args):
+        a_out = h5py.File(args.output, 'w')
```
FreddieWitherden, line 231: Use the NativeReader here.

pyfr/plugins/tavg.py
```diff
@@ -220,6 +226,46 @@ def _acc_avg_var(self, intg, currex):
...
+        start_time = np.empty((0, 1))
+        end_time = np.empty((0, 1))
+
+        for i, filename in enumerate(args.solns):
```
FreddieWitherden, line 235: Also need to handle functional averages (they need re-evaluating based off of the new data) and also code to handle the variance calculations (both summary and full outputs).  @rishit2307 can help you here.

pyfr/plugins/tavg.py
```diff
@@ -220,6 +226,46 @@ def _acc_avg_var(self, intg, currex):
...
+            else:
+                if uuid != a_in['mesh_uuid'][()]:
+                    raise RuntimeError('Average files computed on different meshes')
+                for k in data_keys:
+                    a_out[k][:] += a_in[k][:]*dt
+            a_in.close()
+
+        stat_base.set('tavg', 'tstart', min(start_time))
+        stat_base.set('tavg', 'tend', max(end_time))
+        dout = a_out.create_dataset('stats', data=stat_base.tostr())
```
FreddieWitherden, line 261: Usually avoid `create_dataset` except when we need low-level control.

pyfr/plugins/tavg.py
```diff
@@ -220,6 +226,46 @@ def _acc_avg_var(self, intg, currex):
...
+
+            stat = Inifile(a_in['stats'][()].decode('UTF-8'))
+            end_time = np.append(end_time, stat.getfloat('tavg', 'tend'))
+            start_time = np.append(start_time, stat.getfloat('tavg', 'tstart'))
+            dt = end_time[-1] - start_time[-1]
+
+            if i == 0:
+                uuid = a_in['mesh_uuid'][()]
+                stat_base = stat
+                data_keys = [k for k in a_in.keys() if k.startswith('tavg_')]
```
FreddieWitherden, line 246: Some extra sanity checks are probably worthwhile here (disallow overlapping time ranges unless a flag is passed, check the field names are all the same).

pyfr/plugins/base.py
```diff
@@ -62,6 +62,25 @@ def serialise(self, intg):
         return {}
 
 
+class CLIMixin:
```
FreddieWitherden, line 65: I would not use a mixin here.  It causes issues since we can never construct what it is mixed in with (as that requires an integrator).  Thus, I'd go for an entirely separate class hierarchy.

pyfr/plugins/tavg.py
```diff
@@ -220,6 +237,70 @@ def _acc_avg_var(self, intg, currex):
...
+                data_keys = [k for k in f.keys() if k.startswith('tavg_')]
+                
+                avg_idx, anames = [], []
+                for afield in fields:
+                    avg_idx.append(afield.startswith('avg-'))
+                    if afield.startswith('avg-'):
+                        anames.append(afield.removeprefix('avg-'))
+
+                # Build fun avg expressions
+                suffix = stat.get('tavg', 'suffix', '')
```
FreddieWitherden, line 273: Might be easier to just store the full cfgsect in the stats file.

pyfr/plugins/writer.py
```diff
@@ -1,14 +1,39 @@
...
+from pyfr.util import subclasses
 
-
-class WriterPlugin(PostactionMixin, RegionMixin, BasePlugin):
+class WriterPlugin(CLIMixin, PostactionMixin, RegionMixin, BasePlugin):
     name = 'writer'
     systems = ['*']
     formulations = ['dual', 'std']
 
+    cli = [
```
FreddieWitherden, line 15: I'd leave the writer as-is for the moment.  As the new mechanism is a little bit confusing as the writer plugin is being used to export tavg files (which are produced by the tavg plugin).

pyfr/plugins/tavg.py
```diff
@@ -338,4 +346,147 @@ def __call__(self, intg):
...
+
+                data_keys = [k for k in f.keys() if k.startswith('tavg_')]
+
+                aidx = []
+                sidx = []
+                self.anames = anames = []
+                self.snames = snames = []
+                for field in fields:
+                    # Truth table to extract collumns later
+                    aidx.append(field.startswith('avg-'))
```
FreddieWitherden, line 397: Construct these using comprehensions.  So `aidx = [f for f in fields if f.startswith('avg-')]`

pyfr/plugins/tavg.py
```diff
@@ -338,4 +346,147 @@ def __call__(self, intg):
...
+
+            if i == 0:
+                uuid = f['mesh_uuid']
+                stat_base = stat
+                fields_base = fields
+                region_base = stat.get('tavg', 'region')
+                
+                cfgs = {k: Inifile(f[k]) for k in f.keys() if k.startswith('config')}
+
+                data_keys = [k for k in f.keys() if k.startswith('tavg_')]
```
FreddieWitherden, line 389: No need for keys (implied when you iterate).

pyfr/plugins/tavg.py
```diff
@@ -338,4 +346,147 @@ def __call__(self, intg):
...
+                if uuid != f['mesh_uuid']:
+                    raise RuntimeError('Average files computed on different meshes')
+                elif region_base != stat.get('tavg', 'region'):
+                    raise RuntimeError('Average files computed on different regions')
+                elif fields != fields_base:
+                    raise RuntimeError('Average files computed on different fields')
+                for k in data_keys:
+                    avg_acc[k] += f[k][:, aidx]*dt
+                    if std_all:
+                        var_acc[k] += pow(f[k][:, sidx], 2)*dt
```
FreddieWitherden, line 424: **2

pyfr/plugins/tavg.py
```diff
@@ -338,4 +346,147 @@ def __call__(self, intg):
...
+
+        # Write out the result to a pyfrs file
+        out = h5py.File(args.output, 'w')
+        std_max = np.zeros(len(snames))
```
FreddieWitherden, line 428: These only make sense if we have the raw variance data.

pyfr/plugins/__init__.py
```diff
@@ -1,4 +1,4 @@
-from pyfr.plugins.base import BasePlugin
+from pyfr.plugins.base import BasePlugin, CLIBasePlugin
```
FreddieWitherden, line 1: BaseCLIPlugin (may want to rename BasePlugin to something different, but check with @pv101 as he has similar changes in this direction which his invasive turbulent inflow plugin).

pyfr/util.py
```diff
@@ -96,6 +96,50 @@ def __exit__(self, *args):
...
+class Interval:
+    def __init__(self, a, b):
+        self.start = min(a, b)
+        self.end = max(a, b)
+        self.len = self.end - self.start
+
+    def __contains__(self, other):
+        return other.start >= self.start and other.end <= self.end
+
+    # set intersection without boundaries
```
FreddieWitherden, line 108: o_intersection and c_intersection are probably better names (open/closed).

pyfr/util.py
```diff
@@ -96,6 +96,50 @@ def __exit__(self, *args):
...
+    # set intersection with boundaries
+    def sintersection(self, other):
+        return (self.start <= other.start <= self.end or 
+                self.start <= other.end <= self.end)
+
+    def merge(self, other):
+        return Interval(min(self.start, other.start), max(self.end, other.end))
+
+
+def merge_intervals(intervals: list[Interval]):
```
FreddieWitherden, line 122: This seems a little bit more complex than it needs to be.  I think a sorting based approach will clean up a lot here.

pyfr/plugins/tavg.py
```diff
@@ -338,4 +346,147 @@ def __call__(self, intg):
...
+        for i, (rng, filename, std_mode) in enumerate(active):
+            f = NativeReader(filename)
+
+            stat = Inifile(f['stats'])
+            fields = stat.get('data', 'fields').split(',')
+
+            # Averager time window for current file
+            dt = rng.len
+
+            if i == 0:
```
FreddieWitherden, line 381: I'd break this block out into its own function.

pyfr/plugins/tavg.py
```diff
@@ -338,4 +346,147 @@ def __call__(self, intg):
...
+    ]
+
+    def merge_cli(self, args):
+        self.args = args
+        start_time = np.empty((0, 1))
+        end_time = np.empty((0, 1))
+
+        active = self._merge_active_files(args.solns)
+        merged_times = [(r.start, r.end) for r, _, _ in active]
+        std_all = all([s == 'all' for _, _, s in active])
```
FreddieWitherden, line 368: For this kind of thing I'd be inclined to require that every section in the config file other than nsteps, averaging mode, and dt out need to be the same.  (So for each file grab the cfg items for our plugins section, pop the aforementioned keys, and then require them all to be the same.)

pyfr/util.py
WillTrojak: I don't think sorting will simplifiy this much, there are quite a few edge cases that you have to take care of. This seemed like the most robust and cleanest method I could come up with, but would be more than happy to consider more concrete ideas.

pyfr/plugins/tavg.py
WillTrojak: Yep, I broke this into two and init and an accumulate.

pyfr/plugins/tavg.py
```diff
@@ -338,4 +346,157 @@ def __call__(self, intg):
...
+        active = self._merge_active(args.solns)
+        merged_times = [(r.start, r.end) for r, f, s in active]
+        avg_time_inv = 1/sum([t.end - t.start for t in merged_times])
+        window = merge_intervals(merged_times)
+
+        self.std_all = std_all = all([s == 'all' for r, f, s in active])
+
+        # Merge the Data
+        for i, (rng, filename, std_mode) in enumerate(active):
+            file = NativeReader(filename)
```
FreddieWitherden, line 374: Open all of the files up front and the iterate over these.

pyfr/plugins/tavg.py
```diff
@@ -338,4 +346,157 @@ def __call__(self, intg):
...
+            dt = rng.len
+
+            if i == 0:
+                self._init_tavg_merge(file, dt)
+            else:
+                self._acc_tavg_merge(file, dt)
```
FreddieWitherden, line 380: Aim for an arrangement where you call acc on everything (even i = 0).

pyfr/plugins/tavg.py
```diff
@@ -338,4 +346,157 @@ def __call__(self, intg):
...
+
+        # Write out the result to a pyfrs file
+        out = h5py.File(args.output, 'w')
```
FreddieWitherden, line 383: With statement.

pyfr/plugins/tavg.py
```diff
@@ -338,4 +346,157 @@ def __call__(self, intg):
...
+        for k in self.dkeys:
+            if self.fexprs:
+                avg = np.transpose(self.avg_acc[k]*avg_time_inv, (1, 0, 2))
```
FreddieWitherden, line 386: .tranpose(1, 0, 2) is shorter.

pyfr/plugins/tavg.py
```diff
@@ -338,4 +346,157 @@ def __call__(self, intg):
...
+                fun_avg = self._eval_fun_exprs(avg)
+                out[k] = np.transpose(np.vstack((avg, fun_avg)), (1, 0, 2))
+            else:
+                out[k] = self.avg_acc[k]*avg_time_inv
```
FreddieWitherden, line 390: See if you can refactor such that fexprs means doing some extra to avoid the `self.avg_acc[k]*avg_time_inv` duplication.

pyfr/plugins/tavg.py
```diff
@@ -338,4 +346,157 @@ def __call__(self, intg):
...
+
+            if std_all:
+                std = np.transpose(np.sqrt(self.var_acc[k]*avg_time_inv), 
+                                   (1, 0, 2))
+                self.std_max = np.maximum(self.std_max,
```
FreddieWitherden, line 395: Use out= to avoid reallocating.

pyfr/plugins/tavg.py
```diff
@@ -338,4 +346,157 @@ def __call__(self, intg):
...
+        
+        self.cfgs = cfgs = {k: Inifile(file[k]) for k in file 
+                            if k.startswith('config')}
+
+        self.dkeys = dkeys = [k for k in file if k.startswith('tavg_')]
+
+        # Truth table to extract collumns later
+        self.aidx = aidx = [f.startswith('avg-') for f in fields]
+        self.sidx = sidx = [f.startswith('std-') for f in fields]
+        self.anames = [f.removeprefix('avg-') for f in fields
```
FreddieWitherden, line 437: Can't we iterate through aidx here and avoid the if?

pyfr/util.py
```diff
@@ -96,6 +96,48 @@ def __exit__(self, *args):
         os.close(self.saved_fds[1])
 
 
+class Interval:
```
FreddieWitherden, line 99: Can probably drop this class now we have `merge_intervals` and just work with lists of tuples.

FreddieWitherden:
I pushed a change to improve the stats record that should make this easier to implement.

pyfr/plugins/__init__.py
WillTrojak: I checked with @pv101 and he didn't have anything in mind. I reckon we crack on and can just change it later.

doc/src/cli.rst
```diff
@@ -0,0 +1,81 @@
...
+
+1. ``pyfr tavg merge`` --- merge together multiple windowed time average files 
+   into a single time average file. The time windows are read from the files 
+   and do no need to be even spaced in time.
+   Example::
+
+        pyfr tavg merge avg-1.00.pyfrs avg-2.00.pyfrs merged_avg.pyfrs
+
+
+Writer
```
FreddieWitherden, line 60: Are you okay to shelve this for the moment? It is a bit confusing to use the writer plugin to export time average files.

pyfr/plugins/tavg.py
```diff
@@ -338,4 +346,157 @@ def __call__(self, intg):
...
-                self.tout_last = intg.tcurr
\ No newline at end of file
+                self.tout_last = intg.tcurr
+
+
+class TavgCLIPlugin(CLIBasePlugin):
+    name = 'tavg'
+    systems = ['*']
+    formulations = ['dual', 'std']
+    cli = [
```
FreddieWitherden, line 356: Move this to a function.

doc/src/cli.rst
WillTrojak: Yep, sorry I forgot to update the documentation

pyfr/plugins/base.py
```diff
@@ -62,6 +62,26 @@ def serialise(self, intg):
...
 
 
+class BaseCLIPlugin:
+    name = None
+    systems = None
+    formulations = None
+    cli = None
+
+    @classmethod
+    def parse_cli(cls, parser):
```
FreddieWitherden, line 72: I'd force subclasses to implement this.  Makes the code more explicit and makes it easier for them to set a process method.

pyfr/plugins/tavg.py
```diff
@@ -104,7 +110,7 @@ def _prepare_exprs(self):
                 self.outfields.append(f'std-{k[4:]}')
 
             for k in cfg.items(cfgsect, prefix='fun-avg-'):
-                self.outfields.append(f'fun-std-{k[4:]}')
+                self.outfields.append(f'std-fun-{k[4:]}')
```
FreddieWitherden, line 113: Revert.

FreddieWitherden:
I think some care is needed around how we do the time averaging in term of keeping memory use down.  Specifically, we can not keep the entire time average in memory.  Rather, it needs to be done piece-by-piece so we only ever need at most one partition of data in memory.

pyfr/plugins/tavg.py
WillTrojak: This was delibrate, it makes filtering the fields much easier (and it makes more sense). std of the function rather than function of the std.

FreddieWitherden: Makes sense.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,200 @@ def __call__(self, intg):
...
+
+        # Merge command
+        scmd = 'merge'
+        ssp = sp.add_parser(scmd, help=f'{cmd} {scmd} --help')
+        ssp.set_defaults(process=TavgCLIPlugin.merge_cli)
+        ssp.add_argument('solns', nargs='*', help='Averages to merge')
+        ssp.add_argument('output', help='Output file name')
+
+    @classmethod
+    def merge_cli(cls, args):
```
FreddieWitherden, line 381: Can probably handle this with a decorator so you don't need both the class and regular method to be explicitly defined.

```
@cli_plugin_method
def merge(self, args):
```

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,200 @@ def __call__(self, intg):
...
+            cfgsect = stat.get('tavg', 'cfg-section')
+            fields = stat.get('data', 'fields').split(',')
+
+            if self.uuid != file['mesh_uuid']:
+                raise RuntimeError('Average files computed on different meshes')
+            if self.region != cfg.get(cfgsect, 'region'):
+                raise RuntimeError('Average files computed on different regions')
+            if self.fields != fields:
+                raise RuntimeError('Average files computed on different fields')
+            if not all(cfg.get(cfgsect, k) == self.cfg.get(self.cfgsect, k)
```
FreddieWitherden, line 444: I'd probably just use a for loop here.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,200 @@ def __call__(self, intg):
...
+                       for k in cfg.items(cfgsect, prefix='avg-')):
+                raise RuntimeError('Different average field definitions')
+
+    def _eval_fun_exprs(self, avars):
+
```
FreddieWitherden, line 449: Remove newline.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,200 @@ def __call__(self, intg):
...
+            for t_rng, (file, stat, cfg), std_mode in active:
+                dt = t_rng.len
+                avg_acc += file[key][:, self.aidx]*dt
+                if self.std_all:
+                    var_acc += file[key][:, self.sidx]**2*dt
+
+            if self.fexprs:
+                avg = avg_acc.transpose(1, 0, 2)*avg_time_inv
+                fun_avg = self._eval_fun_exprs(avg)
+                out[key] = np.transpose(np.vstack((avg, fun_avg)), (1, 0, 2))
```
FreddieWitherden, line 505: .tranpose

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,200 @@ def __call__(self, intg):
...
+        else:
+            self.cfg.set(cfgsect, 'std-mode', 'None')
+            std_opt = (self.stat.items('tavg', 'avg-std-') |
+                       self.stat.items('tavg', 'max-std-'))
+            for opt in std_opt:
+                self.stat.remove_option('tavg', opt)
+
+        out['config'] = self.cfg.tostr()
+        out['stats'] = self.stat.tostr()
+
```
FreddieWitherden, line 538: Remove newline.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,200 @@ def __call__(self, intg):
...
+        # Otherwise the std summary has to be removed
+        else:
+            self.cfg.set(cfgsect, 'std-mode', 'None')
+            std_opt = (self.stat.items('tavg', 'avg-std-') |
+                       self.stat.items('tavg', 'max-std-'))
+            for opt in std_opt:
+                self.stat.remove_option('tavg', opt)
+
+        out['config'] = self.cfg.tostr()
+        out['stats'] = self.stat.tostr()
```
FreddieWitherden, line 537: Rename to `self.stats` for consistency?

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,200 @@ def __call__(self, intg):
...
+
+        out['mesh_uuid'] = self.uuid
+
+    def _merge_runtime_stats(self, active):
+        sect = 'solver-time-integrator'
+        rt_stats = defaultdict(lambda: [])
+
+        for _, (_, stat, _), _ in active:
```
FreddieWitherden, line 545: Avoid multiple `_`

pyfr/__main__.py
```diff
@@ -108,6 +109,10 @@ def main():
         p.add_argument('--progress', '-p', action='store_true',
                        help='show a progress bar')
 
+    # Plugin commands
+    for scls in subclasses(BaseCLIPlugin):
+        scls.parse_cli(sp)
```
FreddieWitherden, line 114: Maybe rename from parse, as we actually do the parsing later (L117).

pyfr/plugins/__init__.py
```diff
@@ -1,13 +1,13 @@
 from pyfr.plugins.ascent import AscentPlugin
-from pyfr.plugins.base import BasePlugin
+from pyfr.plugins.base import BasePlugin, BaseCLIPlugin
```
FreddieWitherden, line 2: CLI should probably come first.

pyfr/plugins/base.py
```diff
@@ -62,6 +62,17 @@ def serialise(self, intg):
         return {}
 
 
+class BaseCLIPlugin:
+    name = None
+    systems = None
+    formulations = None
+    cli = None
```
FreddieWitherden, line 69: Is this needed?

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,200 @@ def __call__(self, intg):
...
+        ssp.add_argument('solns', nargs='*', help='Averages to merge')
+        ssp.add_argument('output', help='Output file name')
+
+    @classmethod
+    def merge_cli(cls, args):
+        merge = cls()
+        merge._merge_cli(args)
+
+    def _merge_cli(self, args):
+        self.args = args
```
FreddieWitherden, line 386: Can have the constructor take care of this.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,200 @@ def __call__(self, intg):
...
+    def _eval_fun_exprs(self, avars):
+
+        # Prepare the substitution dictionary
+        subs = dict(zip(self.anames, avars))
+
+        # Evaluate the function and return
+        return np.array([npeval(v, subs) for v in self.fexprs])
+
+    def _init_tavg_merge(self, files):
+        file, stat, cfg = files
```
FreddieWitherden, line 457: Probably worth just making these arguments.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,200 @@ def __call__(self, intg):
...
+            fexprs.append(cfg.getexpr(cfgsect, k, subs=c))
+
+        # Initialise std max and avg registers
+        if self.std_all:
+            self.std_max = np.zeros(nsnames)
+            self.std_avg = np.zeros(nsnames)
+
+    def _merge_data(self, out, active, avg_time):
+        avg_time_inv = 1/avg_time
+        file0, _, _ = active[0][1]
```
FreddieWitherden, line 488: This is clunky, try and unpack active as just extra regular arguments to remove the first level of indexing.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,200 @@ def __call__(self, intg):
...
+
+        for key in self.dkeys:
+            # New accumulation registers
+            avg_acc = np.zeros_like(file0[key][:, self.aidx])
```
FreddieWitherden, line 492: This is expensive, avoid.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,200 @@ def __call__(self, intg):
...
+    def _compatibility_merge(self, active):
+        for t_rng, (file, stat, cfg), std_mode in active:
+            cfgsect = stat.get('tavg', 'cfg-section')
+            fields = stat.get('data', 'fields').split(',')
+
+            if self.uuid != file['mesh_uuid']:
+                raise RuntimeError('Average files computed on different meshes')
+            if self.region != cfg.get(cfgsect, 'region'):
+                raise RuntimeError('Average files computed on different regions')
+            if self.fields != fields:
```
FreddieWitherden, line 442: No need to check this (if you verify the average fields match up you're good). Also enables functional expressions to change (which is something we should allow).  Although do allow that this does mean indexes can change, although this should be easy enough to deal with.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,192 @@ def __call__(self, intg):
...
+        # Merge command
+        scmd = 'merge'
+        ssp = sp.add_parser(scmd, help=f'{cmd} {scmd} --help')
+        ssp.set_defaults(process=TavgCLIPlugin.merge_cli)
+        ssp.add_argument('solns', nargs='*', help='Averages to merge')
+        ssp.add_argument('output', help='Output file name')
+
+    @cli_external
+    def merge_cli(self):
+
```
FreddieWitherden, line 382: Remove newline.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,192 @@ def __call__(self, intg):
...
+        # Open all the solition files
```
FreddieWitherden, line 383: Spelling.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,192 @@ def __call__(self, intg):
...
+                rt_stats[k].append(stats.getliteral(sect, k))
+
+        for k, v in rt_stats.items():
+            self.stats.set(sect, k, v)
+
+    def _prepare_files(self, filenames):
+        self.sfiles = sfiles = {}
+        for filename in filenames:
+            f = NativeReader(filename)
+            sfiles[filename] = (f, Inifile(f['stats']), Inifile(f['config']))
```
FreddieWitherden, line 548: No need for ()

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,192 @@ def __call__(self, intg):
...
+        for k in cfg.items(cfgsect, prefix='fun-avg-'):
+            fexprs.append(cfg.getexpr(cfgsect, k, subs=c))
+
+        # Initialise std max and avg registers
+        if self.std_all:
+            self.std_max = np.zeros(nsnames)
+            self.std_avg = np.zeros(nsnames)
+
+    def _merge_data(self, out, active, avg_time):
+        a = active
```
FreddieWitherden, line 479: Make a copy as the caller probably does not expect the active list to change after the function has been called.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,192 @@ def __call__(self, intg):
...
+            for t_rng, (file, stats, cfg), std_mode in a:
+                dt = t_rng.len
+                avg_acc += file[key][:, self.aidx]*dt
+                if self.std_all:
+                    var_acc += file[key][:, self.sidx]**2*dt
+
+            if self.fexprs:
+                avg = avg_acc.transpose(1, 0, 2)*avg_time_inv
+                fun_avg = self._eval_fun_exprs(avg)
+                out[key] = np.vstack((avg, fun_avg)).transpose(1, 0, 2)
```
FreddieWitherden, line 498: Should be able to use a different static here to avoid the two transposes.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,192 @@ def __call__(self, intg):
...
+    def _merge_meta(self, out, merged_times):
+        t_window = merge_intervals(merged_times)
+
+        cfgsect = self.stats.get('tavg', 'cfg-section')
+        self.stats.set('tavg', 'merged-from', merged_times)
+        self.cfg.set(cfgsect, 'range', t_window)
+
+        # If all files have full std stats then these can be writen
+        if self.std_all:
+            self.std_avg /= len(self.dkeys)
```
FreddieWitherden, line 517: Why are we dividing through by the number of time average quantities (which is what I think `len(self.dkeys)` corresponds to).

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,192 @@ def __call__(self, intg):
...
+            for i, sn in enumerate(self.snames):
+                self.stats.set('tavg', f'avg-std-{sn}', self.std_avg[i])
+                self.stats.set('tavg', f'max-std-{sn}', self.std_max[i])
+        # Otherwise the std summary has to be removed
+        else:
+            self.cfg.set(cfgsect, 'std-mode', 'None')
```
FreddieWitherden, line 523: Lower case none.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,192 @@ def __call__(self, intg):
...
+            std_opt = (self.stats.items('tavg', 'avg-std-') |
+                       self.stats.items('tavg', 'max-std-'))
+            for opt in std_opt:
```
FreddieWitherden, line 526: Might be easier to loop and then do `if opt.startswith(('avg-std', 'max-std'))`

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,192 @@ def __call__(self, intg):
...
+        for i, (r, f, s) in enumerate(times):
+            if any([(r in R and f != F) for R, F, S in active]):
+                active.pop(i)
+            elif any([r.o_intersection(R) for R, F, S in active]):
+                raise ValueError('Clash of average file time windows')
+
+        return active
+
+    def _compatibility_merge(self, active):
+        for t_rng, (file, stats, cfg), std_mode in active:
```
FreddieWitherden, line 429: `t_rng` may want to change the suffix as `rng` is often used for random number generators.  Maybe ival or seg or wind?

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,192 @@ def __call__(self, intg):
...
+            cfgsect = stats.get('tavg', 'cfg-section')
+            fields = stats.get('data', 'fields').split(',')
+
+            if self.uuid != file['mesh_uuid']:
+                raise RuntimeError('Average files computed on different meshes')
+            if self.region != cfg.get(cfgsect, 'region'):
+                raise RuntimeError('Average files computed on different regions')
+
+            for k in cfg.items(cfgsect, prefix='avg-'):
```
FreddieWitherden, line 438: You need to pick out indexes.  So a field can appear in a different order in one file to another.  Thus for each file a mapping is required from field name to index.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,192 @@ def __call__(self, intg):
...
+        self.std_all = all([s == 'all' for r, f, s in active])
+
+        # Initialise things needed for the merge
+        r0, f0, s0 = active[0]
+        self._init_tavg_merge(*f0)
+
+        # Check all files are compatible for merge
+        self._compatibility_merge(active)
+
+        with h5py.File(self.args.output, 'w') as out:
```
FreddieWitherden, line 399: f or outf are probably better names to make it clear they are file objects rather than arrays/dicts.

FreddieWitherden:
Worth a cross check on data types.  Idea is we want all calculations to be performed in double precision, but write out in whatever the format is of the first file (probably single precision by default).

pyfr/__main__.py
```diff
@@ -112,6 +113,10 @@ def main():
         p.add_argument('--progress', '-p', action='store_true',
                        help='show a progress bar')
 
+    # Plugin commands
+    for scls in subclasses(BaseCLIPlugin):
```
FreddieWitherden, line 117: just_leaf=True

pyfr/plugins/base.py
```diff
@@ -62,6 +70,19 @@ def serialise(self, intg):
         return {}
 
 
+class BaseCLIPlugin:
+    name = None
+    systems = None
```
FreddieWitherden, line 75: Not sure this or the below are needed.

pyfr/util.py
FreddieWitherden, line 99: I'm still not sure this class is needed.  Merge intervals will cause duplicate time regions.

pyfr/plugins/tavg.py
WillTrojak: Yep, that should be the number of active windows.

pyfr/plugins/tavg.py
WillTrojak: I did quite a bit of refactoring to handle this in a nice way. Let me knw if you can see any cleaner ways of doing this.

pyfr/plugins/base.py
WillTrojak: They aren't needed at the present but in the future they probably will. But I can remove.

FreddieWitherden: I'm not quite sure how they can ever be used; in regular plugins they work as the integrator has a system which has a name, and this can be used to check compatibility.  However, for a plugin which is free to pick its own set of command line arguments there really is no analogue.  Indeed, it may not always be the case that the thing a plugin is working on is a solution!

pyfr/util.py
WillTrojak: It is used in `_active_files` to reduce the list of files to be read to just those needed. For example continuous mode time averaging followed by one windowed mode at a later time could give files with the intervals [(0,1), (0,2), (3,4)].
merge_intervals will fail.

FreddieWitherden: That should probably be an error. Silently trying to correct our users mistakes typically never ends well.  We should error, and then let the user retry the command without the duplicate file range (or realise that they've done something else wrong and rethink their course of action).

FreddieWitherden:
Also, I think we are forgetting to copy region index data over (if present).  This can be tested by trying to merge some time average files which come from a region and seeing if the result can be exported.

pyfr/plugins/__init__.py
```diff
@@ -1,13 +1,13 @@
...
+from pyfr.plugins.base import BaseCLIPlugin, BasePlugin
 from pyfr.plugins.dtstats import DtStatsPlugin
 from pyfr.plugins.fluidforce import FluidForcePlugin
 from pyfr.plugins.integrate import IntegratePlugin
 from pyfr.plugins.nancheck import NaNCheckPlugin
 from pyfr.plugins.pseudostats import PseudoStatsPlugin
 from pyfr.plugins.residual import ResidualPlugin
 from pyfr.plugins.sampler import SamplerPlugin
-from pyfr.plugins.tavg import TavgPlugin
+from pyfr.plugins.tavg import TavgPlugin, TavgCLIPlugin
```
FreddieWitherden, line 10: C before P.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,198 @@ def __call__(self, intg):
...
+        self.uuid = file0['mesh_uuid']
+
+        self.dkeys = [k for k in file0 if k.startswith('tavg_')]
+
+        # Mapping for expression to sorted list
+        aidx = [f.startswith('avg-') for f in fields]
+        sidx = [f.startswith('std-') for f in fields]
+        self.anames = [f.removeprefix('avg-') for f in compress(fields, aidx)]
+        self.snames = [f.removeprefix('std-') for f in compress(fields, sidx)]
+        self.mapping = [(aidx, np.argsort(self.anames),
```
FreddieWitherden, line 426: Asking NumPy to sort lists is usually asking for trouble. `sorted(range(...), key=lambda i: ...)` is what you want.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,198 @@ def __call__(self, intg):
...
+    def _merge_data(self, outf):
+        avg_time_inv = 1/self.avg_time
+
+        (file0, stats0, cfg0, dt0), f = self._copy_pop(self.files)
+        (aidx0, amap0, sidx0, smap0), mapping = self._copy_pop(self.mapping)
+
+        for key in self.dkeys:
+            # Initialise accumulators
+            data = file0[key]
+            avg_acc = (data[:, aidx0]*dt0)[:, amap0]
```
FreddieWitherden, line 473: Odd indexing?

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,198 @@ def __call__(self, intg):
...
+        twindows, std = [], []
+        self.avg_time = 0
+        for filename in filenames:
+            f = NativeReader(filename)
+            stats, cfg = Inifile(f['stats']), Inifile(f['config'])
+            cfgsect = stats.get('tavg', 'cfg-section')
+            std.append(cfg.get(cfgsect, 'std-mode'))
+
+            twind = stats.getliteral('tavg', 'range')
+            twindows += twind
```
FreddieWitherden, line 543: Use .extend here

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,198 @@ def __call__(self, intg):
...
+            dt = sum(tend - tstart for tstart, tend in twind)
+            self.avg_time += dt
```
FreddieWitherden, line 545: Can we not compute this right at the end with a single sum pass over the times?

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,198 @@ def __call__(self, intg):
...
+
+    def _merge_meta(self, outf):
+        cfgsect = self.stats.get('tavg', 'cfg-section')
+        self.stats.set('tavg', 'merged-from', self.merged_from)
+        self.cfg.set(cfgsect, 'range', self.merged_range)
+
+        # If all files have full std stats then these can be writen
+        if self.std_all:
+            self.std_avg /= len(self.files)
+            for i, sn in enumerate(self.snames):
```
FreddieWitherden, line 507: zip

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,198 @@ def __call__(self, intg):
...
+            if self.std_all:
+                std = np.sqrt(var_acc*avg_time_inv).transpose(1, 0, 2)
+                np.maximum(self.std_max, np.amax(std, axis=(1, 2)),
+                           out=self.std_max)
+                self.std_avg += np.mean(std, axis=(1, 2))
+
+    def _merge_meta(self, outf):
+        cfgsect = self.stats.get('tavg', 'cfg-section')
+        self.stats.set('tavg', 'merged-from', self.merged_from)
+        self.cfg.set(cfgsect, 'range', self.merged_range)
```
FreddieWitherden, line 502: Not sure if we should be changing the ini file.

doc/src/cli.rst
```diff
@@ -0,0 +1,75 @@
...
+Plugin Commands
+===============
+
+Plugins are able to define tools that are accessible from the command line,
+below are listed the the various options currently available.
+
+Time Average
+^^^^^^^^^^^^
+
+1. ``pyfr tavg merge`` --- merge together multiple windowed time average files
```
FreddieWitherden, line 71: Don't think they need to be windowed, a continuous average can be merged too so long as it does not overlap with anything else.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,198 @@ def __call__(self, intg):
...
+
+    @classmethod
+    def add_cli(cls, parser):
+        cmd = cls.name
+        cp = parser.add_parser(cmd, help=f'{cmd} --help')
+        sp = cp.add_subparsers(dest='sub_cmd')
+
+        # Merge command
+        scmd = 'merge'
+        ssp = sp.add_parser(scmd, help=f'{cmd} {scmd} --help')
```
FreddieWitherden, line 375: Probably clearer to just write merge here.

pyfr/plugins/tavg.py
```diff
@@ -351,3 +357,198 @@ def __call__(self, intg):
...
+        ssp.set_defaults(process=TavgCLIPlugin.merge_cli)
```
FreddieWitherden, line 376: cls.merge_cli

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,200 @@ def __call__(self, intg):
...
+            if i == 0:
+                self.stats, self.cfg = stats, cfg
+
+                self.fields = fields
+                self.cfgsect = cfgsect
+                self.region = cfg.get(cfgsect, 'region')
+                self.uuid = file['mesh_uuid']
+
+                self.idxs  = [k for k in file if re.search(r'_idxs_p\d+$', k)]
+                self.dkeys = [k for k in file if k.startswith('tavg_') and k not
```
FreddieWitherden, line 436: `re.match(r'tavg_\w+_p\d+$', k)`

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,200 @@ def __call__(self, intg):
...
+
+    def _merge_data(self, outf):
+        avg_time_inv = 1/self.avg_time
+
+        (file0, stats0, cfg0, dt0), f = self._copy_pop(self.files)
+        (amap0, smap0), mapping = self._copy_pop(self.mapping)
+
+        for key in self.dkeys:
+            # Initialise accumulators
+            data = file0[key]
```
FreddieWitherden, line 475: Avoid making the first file special.  Just allocate some zeros up front and then all files become the same.

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,200 @@ def __call__(self, intg):
...
+                data = file[key]
+
+                avg_acc += data[:, amap]*dt
+                if self.std_all:
+                    var_acc += data[:, smap]**2*dt
+
+            if self.fexprs:
+                avg = avg_acc.transpose(1, 0, 2)*avg_time_inv
+                fun_avg = self._eval_fun_exprs(avg)
+                outf[key] = np.vstack((avg, fun_avg)).transpose(1, 0, 2)
```
FreddieWitherden, line 491: Can transposes (different stacking function will help, can also nudge the eval fun exprs to use a different layout).

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,200 @@ def __call__(self, intg):
...
+            else:
+                outf[key] = avg_acc*avg_time_inv
+
+            if self.std_all:
+                std = np.sqrt(var_acc*avg_time_inv).transpose(1, 0, 2)
```
FreddieWitherden, line 496: Can avoid transpose by changing axis arguments further down.

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,200 @@ def __call__(self, intg):
...
+
+            if i == 0:
+                self.stats, self.cfg = stats, cfg
+
+                self.fields = fields
+                self.cfgsect = cfgsect
+                self.region = cfg.get(cfgsect, 'region')
+                self.uuid = file['mesh_uuid']
+
+                self.idxs  = [k for k in file if re.search(r'_idxs_p\d+$', k)]
```
FreddieWitherden, line 435: Probably not worth assigning to. Instead, I'd keep a list of file names we want to write and then loop over everything in the first file and `if k not in ...` then copy it.

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,200 @@ def __call__(self, intg):
...
+                           out=self.std_max)
+                self.std_avg += np.mean(std, axis=(1, 2))
+
+    def _merge_meta(self, outf):
+        cfgsect = self.stats.get('tavg', 'cfg-section')
+        self.stats.set('tavg', 'merged-from', self.merged_from)
+        self.stats.set('tavg', 'range', self.merged_range)
+
+        # If all files have full std stats then these can be writen
+        if self.std_all:
```
FreddieWitherden, line 507: Check these stats by running a toy simulation with two plugin instances: one continuous writing every 2*dt and another windowed writing every dt and verify that merged windowed ~ continuous.

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,196 @@ def __call__(self, intg):
...
+
+        # Initialise things needed for the merge
+        self._init_tavg_merge()
+
+        with h5py.File(self.args.output, 'w') as outf:
+            # Merge the averages and stds
+            self._merge_data(outf)
+
+            # Merge runtime stats
+            self._merge_runtime_stats()
```
FreddieWitherden, line 396: I'd have `_merge_meta` call this, as it is basically metadata related.

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,196 @@ def __call__(self, intg):
...
+
+            # Merge the metadata
+            self._merge_meta(outf)
+
+    def _eval_fun_exprs(self, avars):
+        # Prepare the substitution dictionary
+        subs = {n: avars[:, i] for i, n in enumerate(self.anames)}
+
+        # Evaluate the function and return
+        f = np.array([npeval(v, subs) for v in self.fexprs])
```
FreddieWitherden, line 406: I think we can just use a stacking command here as opposed to `np.array` to avoid the transpose. Can never remember which one we want.

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,196 @@ def __call__(self, intg):
...
+            for (file, stats, cfg, dt), (amap, smap) in zip(self.files,
+                                                            self.mapping):
+                data = file[key]
+
+                avg_acc += data[:, amap]*dt
+                if self.std_all:
+                    var_acc += data[:, smap]**2*dt
+
+            if self.fexprs:
+                avg = avg_acc*avg_time_inv
```
FreddieWitherden, line 485: Can pull this out.

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,196 @@ def __call__(self, intg):
...
+        amap0, smap0 = self.mapping[0]
+
+        for key in self.dkeys:
+            # Initialise accumulators
+            data = file0[key]
+            avg_acc = np.zeros_like(data[:, amap0])
+            var_acc = np.zeros_like(data[:, smap0])
+
+            # Perform the rest of the accumulations
+            for (file, stats, cfg, dt), (amap, smap) in zip(self.files,
```
FreddieWitherden, line 476: Try and avoid newline breaks in these kinds of loops.  If needed unpack on the first line.

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,196 @@ def __call__(self, intg):
...
+                                                            self.mapping):
+                data = file[key]
+
+                avg_acc += data[:, amap]*dt
+                if self.std_all:
+                    var_acc += data[:, smap]**2*dt
```
FreddieWitherden, line 482: More typical to have the scalar*vector (think in algebra `2x` vs `x2`), plus a little bit easier to parse given `dt*x**2` vs `x**2*dt` where you need to think more carefully about precedence.

pyfr/plugins/tavg.py
FreddieWitherden: `return np.stack([npeval(v, subs) for v in self.fexprs], axis=1)` might do it

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,196 @@ def __call__(self, intg):
...
+
+            # Merge runtime stats
+            self._merge_runtime_stats()
+
+            # Merge the metadata
+            self._merge_meta(outf)
+
+    def _eval_fun_exprs(self, avars):
+        # Prepare the substitution dictionary
+        subs = {n: avars[:, i] for i, n in enumerate(self.anames)}
```
FreddieWitherden, line 403: `subs = dict(zip(self.anames, avars.swapaxis(0, 1)))` or something close to that

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,248 @@ def __call__(self, intg):
...
+                    raise RuntimeError('Different average field definitions')
+
+    def _merge_data(self, outf):
+        file0, stats, cfg, dt = self.files[0]
+        amap0, smap0 = self.mapping[0]
+
+        for key in self.dkeys:
+            # Initialise accumulators
+            data = file0[key]
+            avg_acc = np.zeros_like(data[:, amap0], dtype=np.float64())
```
FreddieWitherden, line 496: No need to call the dtype.

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,248 @@ def __call__(self, intg):
...
+                self.fstd_avg += np.mean(fun_std, axis=(0, 2))
+
+                outf[key] = np.hstack((avg_acc, fun_avg, var_acc, fun_std),
+                                       dtype=self.dtype)
+            elif self.fexprs:
+                fun_avg = self._eval_fun_exprs(avg_acc)
+                outf[key] = np.hstack((avg_acc, fun_avg, var_acc),
+                                      dtype=self.dtype)
+            else:
+                outf[key] = np.hstack((avg_acc, var_acc), dtype=self.dtype)
```
FreddieWitherden, line 530: What happens when `std_all` is False?

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,248 @@ def __call__(self, intg):
...
+
+    def _merge_meta(self, outf):
+        cfgsect = self.stats.get('tavg', 'cfg-section')
+        self.stats.set('tavg', 'merged-from', self.merged_from)
+        self.stats.set('tavg', 'range', self.merged_range)
+
+        # If all files have full std stats then these can be writen
+        if self.std_all:
+            self.std_avg /= len(self.files)
+            self.std_avg.astype(self.dtype, copy=False)
```
FreddieWitherden, line 540: Not sure what we're doing here?

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,248 @@ def __call__(self, intg):
...
+        self.merged_from = twindows
+        self.std_all = all(s == 'all' for s in std)
+
+        try:
+            self.merged_range = merge_intervals(twindows)
+        except ValueError:
+            raise RuntimeError('Overlapping averge time ranges in files')
+
+        avg_time = sum(t[1] - t[0] for t in self.merged_range)
+        for i, (f, stats, cfg, dt) in enumerate(files):
```
FreddieWitherden, line 609: Just create a new list with the corrected time and assign this to self.files.  `self.files = [(*fsc, dt / avg_time) for *fsc, dt in files]`

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,248 @@ def __call__(self, intg):
...
+            self._merge_meta(outf)
+
+    def _eval_fun_exprs(self, avars):
+        # Prepare the substitution dictionary
+        subs = dict(zip(self.anames, avars.swapaxes(0, 1)))
+
+        # Evaluate the function and return
+        return np.stack([npeval(v, subs) for v in self.fexprs], axis=1)
+
+    def _eval_fun_var(self, std, avars):
```
FreddieWitherden, line 405: There's a lot of code duplication here with the main plugin.  Worth refactoring to either a mixin or standalone function.

FreddieWitherden:
Are you okay to get this tidied away so we can merge?

FreddieWitherden:
Are you okay to get this tidied away so we can merge?

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,248 @@ def __call__(self, intg):
...
+
+                self.mapping = [(amap, smap)]
+                fnames = [x.removeprefix('fun-avg-') for x in fields if
+                          x.startswith('fun-avg-')]
+                self.anames = sorted(anames)
+                self.fnames = sorted(fnames)
+
+                # Initialise std max and avg registers
+                if self.std_all:
+                    self.std_max = np.zeros(len(anames), dtype=self.dtype)
```
FreddieWitherden, line 466: No dtype here? Just cast at the end when we write to ensure accumulation is done at the highest possible precision.

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,248 @@ def __call__(self, intg):
...
+                    self.std_avg = np.zeros(len(anames), dtype=self.dtype)
+                    self.fstd_max = np.zeros(len(fnames), dtype=self.dtype)
+                    self.fstd_avg = np.zeros(len(fnames), dtype=self.dtype)
+
+                # Build fun avg expressions
+                c = cfg.items_as('constants', float)
+                self.fexprs = fexprs = []
```
FreddieWitherden, line 473: List comprehension?

pyfr/plugins/tavg.py
WillTrojak: then var_acc is empty and it all works out fine.

pyfr/plugins/tavg.py
WillTrojak: This is inresponse to your request that accumulation is performed in double precision and then cast to dtype

pyfr/plugins/tavg.py
WillTrojak: Both version of this function do slightly different things. The main plugin is given region information and then calculates the fun var, whereas in the CLI version doens't have access region information and also works on a tranposed version of the data.

pyfr/plugins/tavg.py
FreddieWitherden: But this statement is a no-op; `astype` returns a new array which is never assigned anywhere.

pyfr/plugins/tavg.py
FreddieWitherden: Just following up here.

pyfr/plugins/tavg.py
FreddieWitherden: I think this body:
```
            # Evaluate the function
            fx = self._eval_fun_exprs(av)
            exprs.append(fx)

            for avi in av:
                # Calculate step size for finite difference
                h = self.delta_h*np.abs(avi)
                h[np.where(h == 0)] = self.delta_h

                # Calculate derivatives for functional averages
                avi += h
                df.append((self._eval_fun_exprs(av) - fx) / h)
                avi -= h
```
Can still be reasonably factored out into a function, even if it means one of the callers having to do a bit of pre/post transposing.

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,247 @@ def __call__(self, intg):
...
+                prec = cfg.get('backend', 'precision')
+                self.dtype = np.float32() if prec == 'single' else np.float64()
+
+                self.fields = fields
+                self.cfgsect = cfgsect
+                self.region = cfg.get(cfgsect, 'region')
+                self.uuid = file['mesh_uuid']
+
+                self.idxs = [k for k in file if re.search(r'_idxs_p\d+$', k)]
+                self.dkeys = [k for k in file if
```
FreddieWitherden, line 455: Move if down.

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,247 @@ def __call__(self, intg):
...
+                amap, smap = mapping
+                data = file[key]
+
+                avg_acc += dt*data[:, amap]
+                if self.std_all:
+                    var_acc += dt*(data[:, smap]**2 + data[:, amap]**2)
+
+            # Shift variance and get standard deviation
+            if self.std_all:
+                var_acc = np.sqrt(var_acc - avg_acc**2)
```
FreddieWitherden, line 511: Can you check if this approach is numerically stable (at least on paper)? For some reason I think it is the 'bad' way to do it.

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,247 @@ def __call__(self, intg):
...
+                self.std_avg += np.mean(var_acc, axis=(0, 2))
+
+            # Evaluate function expression and write out
+            if self.fexprs and self.std_all:
+                fun_avg, fun_std = self._eval_fun_var(var_acc, avg_acc)
+                np.maximum(self.fstd_max, np.amax(fun_std, axis=(0, 2)),
+                           out=self.fstd_max)
+                self.fstd_avg += np.mean(fun_std, axis=(0, 2))
+
+                outf[key] = np.hstack((avg_acc, fun_avg, var_acc, fun_std),
```
FreddieWitherden, line 523: Assign a tuple.  Then on a line outside the block do the `outf[key] = np.hstack(mytuple, dtype=self.dtype)`

pyfr/plugins/tavg.py
```diff
@@ -358,3 +363,247 @@ def __call__(self, intg):
...
+                                       dtype=self.dtype)
+            elif self.fexprs:
+                fun_avg = self._eval_fun_exprs(avg_acc)
+                outf[key] = np.hstack((avg_acc, fun_avg, var_acc),
```
FreddieWitherden, line 527: Do we want the variance here? I think it'll be an empty array?

FreddieWitherden:
Not sure how/why this got closed.

FreddieWitherden:
@WillTrojak Are you okay to rebase?  I think it is one of the big new features for the next release.

pyfr/plugins/tavg.py
```diff
@@ -357,3 +362,248 @@ def __call__(self, intg):
...
+            h = self.delta_h*np.abs(avi)
+            h[np.where(h == 0)] = self.delta_h
+
+            # Calculate derivatives for functional averages
+            avi += h
+            dfexpr.append((self._eval_fun_exprs(avars) - fexprs) / h[:,None])
+            avi -= h
+
+        # Multiply by variance and take RMS value
+        fv = np.linalg.norm(np.array(dfexpr)*std.swapaxes(0, 1)[:, :, None],
```
FreddieWitherden, line 421: Spacing

pyfr/plugins/tavg.py
```diff
@@ -357,3 +362,248 @@ def __call__(self, intg):
...
+
+                self.mapping = [(amap, smap)]
+                fnames = [x.removeprefix('fun-avg-') for x in fields if
+                          x.startswith('fun-avg-')]
+                self.anames = sorted(anames)
+                self.fnames = sorted(fnames)
+
+                # Initialise std max and avg registers
+                if self.std_all:
+                    self.std_max = np.zeros(len(anames))
```
FreddieWitherden, line 465: Can sometimes group these by doing `a,b = np.zeros((2, len(anames))`

pyfr/plugins/tavg.py
```diff
@@ -357,3 +362,248 @@ def __call__(self, intg):
...
+            # Initialise accumulators
+            data = file0[key]
+            avg_acc = np.zeros_like(data[:, amap0])
+            var_acc = np.zeros_like(data[:, smap0])
+
+            # Perform the rest of the accumulations
+            t = 0
+            for files, mapping in zip(self.files, self.mapping):
+                file, stats, cfg, dt = files
+                amap, smap = mapping
```
FreddieWitherden, line 501: Can avoid these by saying `adata, sdata = data[:, mapping]` or something close to that

pyfr/plugins/tavg.py
```diff
@@ -357,3 +362,248 @@ def __call__(self, intg):
...
+
+                if self.std_all:
+                    var_acc = ((t*var_acc + dt*data[:, smap]**2)/(t + dt) +
+                               (dt/(t + dt)**2)*(avg_acc - t*data[:, amap])**2)
+                avg_acc += dt*data[:, amap]
+                t += dt
+
+            # Get standard deviation
+            if self.std_all:
+                var_acc = np.sqrt(var_acc)
```
FreddieWitherden, line 512: **0.5

pyfr/plugins/tavg.py
FreddieWitherden: Is this above still an issue?

pyfr/plugins/tavg.py
```diff
@@ -357,3 +362,248 @@ def __call__(self, intg):
...
 
                 self.tout_last = intg.tcurr
+
+
+class TavgCLIPlugin(BaseCLIPlugin):
+    name = 'tavg'
+
+    @classmethod
+    def add_cli(cls, parser):
+        cmd = cls.name
```
FreddieWitherden, line 372: Maybe have this bit done by the calling code and just pass in sp.

pyfr/plugins/tavg.py
```diff
@@ -357,3 +362,248 @@ def __call__(self, intg):
...
+        cp = parser.add_parser(cmd, help=f'{cmd} --help')
+        sp = cp.add_subparsers(dest='sub_cmd')
```
FreddieWitherden, line 374: Is this dest needed?

WillTrojak: yeah, it allows us to have the command as `pyfr tavg merge ...`

pyfr/plugins/tavg.py
WillTrojak: It seems to me best to have this done here in the object that defines the cli plugin.

pyfr/plugins/tavg.py
WillTrojak: The files and mapping a created in different places, I think this is the cleanest way to do it.

pyfr/plugins/tavg.py
WillTrojak: No, if the variance isn't used then it is var_acc.shape is (n,0) and so doesn't get added

pyfr/plugins/tavg.py
FreddieWitherden: The lines are going to be generic for all CLI plugins: get our name, add a parser, add a sub parser.  I'd rather plugins have no choice about their command name (same way plugins don't get to decide their section names).

pyfr/plugins/tavg.py
FreddieWitherden: But is the `dest` bit needed?  As we never access `sub_cmd` as far as I can tell.

pyfr/plugins/tavg.py
WillTrojak: However, every CLI plugin might have different requirements in terms of inputs and outputs. Previously I had a method where the plugin defined an attribute list that was used to automatically define the parser for the cli. But you asked for the current method instead.

FreddieWitherden: It is always going to be `parser.add_parser(cmd, help=f'{cmd} --help')`.  What the plugin does beyond this though is it's call (in terms of subcommands, flags, etc).  But we can certainly factor out the parser creation.

PR #322, opened 2022-12-24
Bug fix in entropy filter MPI graph.

tarikdzanic:


FreddieWitherden:
Add on line 14 as:
```
g1.add_mpi_reqs(m['scal_fpts_recv'] + m['ent_fpts_recv'])
```
(We want to start recv requests as soon as possible .)

FreddieWitherden:
Can also move the request on L103 up (although this is just stylistic).

FreddieWitherden:
Also, this looks just to be for `baseadvec` but were we not also having problems with Navier–Stokes (which is `baseadvecdiff`)?

tarikdzanic:
> Also, this looks just to be for `baseadvec` but were we not also having problems with Navier–Stokes (which is `baseadvecdiff`)?

The issue was only on the preproc graphs which baseadvecdiff inherits from baseadvec

PR #324, opened 2023-01-03
Changed filename generator

WillTrojak:
This removes the need for a `next` call when the generator is created.


pyfr/util.py
```diff
@@ -169,7 +169,7 @@ def match_paired_paren(delim, n=5):
     return lft*n + mid + rgt*n
 
 
-def file_path_gen(basedir, basename, restore=False):
+def _file_path_gen(basedir, basename, restore=False):
```
FreddieWitherden, line 172: Can probably use an inner function to keep the namespace clean.

FreddieWitherden:
All works okay?

WillTrojak:
Yep, it works fine with the ascent plugin.

FreddieWitherden:
Also, there seem to be conflicts.

WillTrojak:
The conflicts are just from the function being very similar to what was there before and git getting confused.

PR #326, opened 2023-01-08
Allow selective exporting of fields.

FreddieWitherden:


FreddieWitherden:
I am unsure if that will make it into the next release or if it is something post v2.  Certainly, the functionality is useful now (especially if you have jumbo time average files that you just want to render one field of).

PR #327, opened 2023-01-11
Pass performance information to GiMMiK.

FreddieWitherden:


WillTrojak (commented):
Do you want to add documentation of `gimmik-nkerns` to the performance tuning section? Other than that, this looks fine. It'll be a great feature when it's fully working on the gimmik side.

FreddieWitherden:
It's one of those parameters that's mostly useful for those testing GiMMiK (or variants thereof that can offer up 100s of kernels), but should never need tweaking by a user.  Lower values may give a shorter start-up time at the cost of possibly lower run-time performance.

PR #329, opened 2023-01-19
Changed the name of region index datasets

WillTrojak:
A small change to the name of region index datasets to `{pre}_idxs_{etype}_{part}` from `{pre}_{etype}_idxs_{part}`. I think this makes a bit more sense and makes filtering the dataset keys easier.


FreddieWitherden:
It will also break the export of all existing region solutions.  We typically try and keep compatibility for at least two releases after a change to the file format (such as with lineles and going forwards with the periodic tagging).

PR #330, opened 2023-01-20
Added exact Riemann solver for Euler

WillTrojak:
Added the exact riemann solver to the Euler solver. This also adds the option for riemann solves to be performed in the 1D transformed reference space, which can also simplify the methods for complex Riemann solvers like HLLC and Roe.

### DIFF
--- a/pyfr/solvers/baseadvec/kernels/transform.mako
+++ b/pyfr/solvers/baseadvec/kernels/transform.mako
@@ -1,0 +1,80 @@
+<%namespace module='pyfr.backends.base.makoutil' name='pyfr'/>
+
+## Transforms to [1, 0, 0]^T from n
+## See Moler and Hughes 1999
+<%pyfr:macro name='transform_to' params='n, u, t, off'>
+% if ndims == 2:
+    t[off + 0] = n[0]*u[off + 0] + n[1]*u[off + 1];
+    t[off + 1] = n[0]*u[off + 1] - n[1]*u[off + 0];
+% elif ndims == 3:
+    if (fabs(n[0]) < 0.99)
+    {
+        fpdtype_t h = 1/(1 + n[0]);
+
+        t[off + 0] = n[0]*u[off + 0] + n[1]*u[off + 1] + n[2]*u[off + 2];
+        t[off + 1] = (n[0] + h*n[2]*n[2])*u[off + 1] - n[1]*u[off + 0]
+                     - h*n[1]*n[2]*u[off + 2];
+        t[off + 2] = (n[0] + h*n[1]*n[1])*u[off + 2] - n[2]*u[off + 0]
+                        - h*n[1]*n[2]*u[off + 1];
+    }
+    else if (fabs(n[1]) < fabs(n[2]))
+    {
+        fpdtype_t h = 1/(1 - n[1]);
+
+        t[off + 0] = n[0]*u[off + 0] + n[1]*u[off + 1] + n[2]*u[off + 2];
+        t[off + 1] = (1 - h*n[0]*n[0])*u[off + 0] + n[0]*u[off + 1]
+                     - h*n[0]*n[2]*u[off + 2];
+        t[off + 2] = n[2]*u[off + 1] - h*n[0]*n[2]*u[off + 0]
+                     + (1 - h*n[2]*n[2])*u[off + 2];
+    }
+    else
+    {
+        fpdtype_t h = 1/(1 - n[2]);
+
+        t[off + 0] = n[0]*u[off + 0] + n[1]*u[off + 1] + n[2]*u[off + 2];
+        t[off + 1] = (1 - h*n[1]*n[1])*u[off + 1] - h*n[0]*n[1]*u[off + 0]
+                     + n[1]*u[off + 2];
+        t[off + 2] = (1 - h*n[0]*n[0])*u[off + 0] - h*n[0]*n[1]*u[off + 1]
+                     + n[0]*u[off + 2];
+    }
+% endif
+</%pyfr:macro>
+
+## Transforms from [1, 0, 0]^T to n
+<%pyfr:macro name='transform_from' params='n, t, u, off'>
+% if ndims == 2:
+    u[off + 0] = n[0]*t[off + 0] - n[1]*t[off + 1];
+    u[off + 1] = n[1]*t[off + 0] + n[0]*t[off + 1];
+% elif ndims == 3:
+    if (fabs(n[0]) < 0.99)
+    {
+        fpdtype_t h = 1/(1 + n[0]);
+
+        u[off + 0] = n[0]*t[off + 0] - n[1]*t[off + 1] - n[2]*t[off + 2];
+        u[off + 1] = n[1]*t[off + 0] + (n[0] + h*n[2]*n[2])*t[off + 1]
+                     - h*n[1]*n[2]*t[off + 2];
+        u[off + 2] = n[2]*t[off + 0] - h*n[1]*n[2]*t[off + 1]
+                     + (n[0] + h*n[1]*n[1])*t[off + 2];
+    }
+    else if (fabs(n[1]) < fabs(n[2]))
+    {
+        fpdtype_t h = 1/(1 - n[1]);
+
+        u[off + 0] = n[0]*t[off + 0] +  (1 - h*n[0]*n[0])*t[off + 1]
+                     - h*n[0]*n[2]*t[off + 2];
+        u[off + 1] = n[1]*t[off + 0] + n[0]*t[off + 1] + n[2]*t[off + 2];
+        u[off + 2] = n[2]*t[off + 0] - h*n[0]*n[2]*t[off + 1]
+                     + (1 - h*n[2]*n[2])*t[off + 2];
+    }
+    else
+    {
+        fpdtype_t h = 1/(1 - n[2]);
+
+        u[off + 0] = n[0]*t[off + 0] - h*n[0]*n[1]*t[off + 1]
+                     + (1 - h*n[0]*n[0])*t[off + 2];
+        u[off + 1] = n[1]*t[off + 0] + (1 - h*n[1]*n[1])*t[off + 1]
+                     - h*n[0]*n[1]*t[off + 2];
+        u[off + 2] = n[2]*t[off + 0] + n[1]*t[off + 1] + n[0]*t[off + 2];
+    }
+% endif
+</%pyfr:macro>
--- a/pyfr/solvers/euler/inters.py
+++ b/pyfr/solvers/euler/inters.py
@@ -15,6 +15,12 @@ class FluidIntIntersMixin:
                 entmin_lhs=self._entmin_lhs, entmin_rhs=self._entmin_rhs
             )
 
+            self.p_min = self.cfg.getfloat('solver-entropy-filter', 'p_min',
+                                           1e-6)
+        else:
+            self.p_min = self.cfg.getfloat('solver-interfaces', 'p_min',
+                                           5*self._be.fpdtype_eps)
+
 
 class FluidMPIIntersMixin:
     def __init__(self, *args, **kwargs):
@@ -37,7 +43,7 @@ class EulerIntInters(FluidIntIntersMixin, BaseAdvectionIntInters):
 
         rsolver = self.cfg.get('solver-interfaces', 'riemann-solver')
         tplargs = dict(ndims=self.ndims, nvars=self.nvars, rsolver=rsolver,
-                       c=self.c)
+                       c=self.c, p_min=self.p_min)
 
         self.kernels['comm_flux'] = lambda: self._be.kernel(
             'intcflux', tplargs=tplargs, dims=[self.ninterfpts],
@@ -53,7 +59,7 @@ class EulerMPIInters(FluidMPIIntersMixin, BaseAdvectionMPIInters):
 
         rsolver = self.cfg.get('solver-interfaces', 'riemann-solver')
         tplargs = dict(ndims=self.ndims, nvars=self.nvars, rsolver=rsolver,
-                       c=self.c)
+                       c=self.c, p_min=self.p_min)
 
         self.kernels['comm_flux'] = lambda: self._be.kernel(
             'mpicflux', tplargs, dims=[self.ninterfpts],
@@ -69,7 +75,8 @@ class EulerBaseBCInters(BaseAdvectionBCInters):
 
         rsolver = self.cfg.get('solver-interfaces', 'riemann-solver')
         tplargs = dict(ndims=self.ndims, nvars=self.nvars, rsolver=rsolver,
-                       c=self.c, bctype=self.type, ninters=self.ninters)
+                       c=self.c, p_min=self.p_min, bctype=self.type,
+                       ninters=self.ninters)
 
         self.kernels['comm_flux'] = lambda: self._be.kernel(
             'bccflux', tplargs=tplargs, dims=[self.ninterfpts],
--- a/pyfr/solvers/euler/kernels/flux.mako
+++ b/pyfr/solvers/euler/kernels/flux.mako
@@ -24,3 +24,25 @@
     f[${i}][${j + 1}] = rhov[${i}]*v[${j}]${' + p' if i == j else ''};
 % endfor
 </%pyfr:macro>
+
+<%pyfr:macro name='inviscid_flux_1d' params='s, f, p, v'>
+    fpdtype_t invrho = 1.0/s[0], E = s[${nvars - 1}];
+
+    // Compute the velocities
+% for i in range(ndims):
+    v[${i}] = invrho*s[${i + 1}];
+% endfor
+
+    // Compute the pressure
+    p = ${c['gamma'] - 1}*(E - 0.5*invrho*${pyfr.dot('s[{i}]', i=(1, ndims + 1))});
+
+    // Density and energy fluxes
+    f[0] = s[1];
+    f[${nvars - 1}] = (E + p)*v[0];
+
+    // Momentum fluxes
+    f[1] = s[1]*v[0] + p;
+% for j in range(1, ndims):
+    f[${j + 1}] = s[1]*v[${j}];
+% endfor
+</%pyfr:macro>
--- a/pyfr/solvers/euler/kernels/rsolvers/exact.mako
+++ b/pyfr/solvers/euler/kernels/rsolvers/exact.mako
@@ -1,0 +1,211 @@
+<%namespace module='pyfr.backends.base.makoutil' name='pyfr'/>
+<%include file='pyfr.solvers.euler.kernels.flux'/>
+
+<% gamma = c['gamma'] %>
+<% hgm = 0.5*(gamma - 1) %>
+<% grgm = gamma/(gamma - 1) %>
+<% trgm = 2/(gamma - 1) %>
+<% trgp = 2/(gamma + 1) %>
+<% gmrtg = (gamma - 1)/(2*gamma) %>
+<% gprtg = (gamma + 1)/(2*gamma) %>
+<% tgrgm = (2*gamma)/(gamma - 1) %>
+<% gmrgp = (gamma - 1)/(gamma + 1) %>
+
+<%pyfr:macro name='star_flux' params='p, ps, rs, cs, f, fd'>
+    // Star Flux, assuming covolume = 0. See Toro 2009 Eq.(4.86 - 4.87)
+    if (p <= ps)
+    {
+       fpdtype_t pr = p / ps;
+       fpdtype_t pr_g = pow(pr, ${-gprtg});
+       fd = pr_g / (rs*cs);
+       f = ${trgm}*cs*(pr_g*pr - 1);
+    }
+    else
+    {
+       fpdtype_t pbs_inv = 1 / (p + ${gmrgp}*ps);
+       fpdtype_t sapb = sqrt(${trgp} * pbs_inv / rs);
+       f = (p - ps)*sapb;
+       fd = sapb - 0.5*f*pbs_inv;
+    }
+</%pyfr:macro>
+
+<%pyfr:macro name='primitives' params='s, p, v, c'>
+    fpdtype_t invrho = 1/s[0], E = s[${nvars - 1}];
+
+    // Compute the velocities
+% for i in range(ndims):
+    v[${i}] = invrho*s[${i + 1}];
+% endfor
+
+    // Compute the pressure
+    p = ${gamma - 1}*(E - 0.5*invrho*${pyfr.dot('s[{i}]', i=(1, ndims + 1))});
+
+    // Compute the local sound speed
+    c = sqrt(${gamma}*p*invrho);
+</%pyfr:macro>
+
+<% kmax = 3 %>
+
+<%pyfr:macro name='rsolve_1d' params='ul, ur, nf'>
+    fpdtype_t vl[${ndims}], pl, cl, fsl, fdl;
+    fpdtype_t vr[${ndims}], pr, cr, fsr, fdr;
+    fpdtype_t p0, p1;
+    fpdtype_t w0[${nvars}];
+
+    // Compute the left/right primitives
+    ${pyfr.expand('primitives', 'ul', 'pl', 'vl', 'cl')};
+    ${pyfr.expand('primitives', 'ur', 'pr', 'vr', 'cr')};
+
+    // Inital pressure guess
+    fpdtype_t rl = ul[0];
+    fpdtype_t rr = ur[0];
+    fpdtype_t bpv = fmax(0, 0.125*(vl[0] - vr[0])*(rl + rr)*(cl + cr) +
+                         0.5*(pl + pr));
+    fpdtype_t pmin = fmin(pl, pr);
+    fpdtype_t pmax = fmax(pl, pr);
+    if (pmax <= 2*pmin && pmin <= bpv && bpv <= pmax)
+    {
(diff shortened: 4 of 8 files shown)


FreddieWitherden:
I think the structure is mostly fine, although we should look to see if anything can be done about the `p_min` argument.  It is going to be a nightmare to document and seems like we're asking for trouble.

pyfr/solvers/baseadvec/kernels/transform.mako
```diff
@@ -0,0 +1,83 @@
+<%namespace module='pyfr.backends.base.makoutil' name='pyfr'/>
+
+<% t_tol = 0.99 %>
```
FreddieWitherden, line 3: Try to avoid declaring variables in these helper includes.

pyfr/solvers/baseadvec/kernels/transform.mako
```diff
@@ -0,0 +1,83 @@
...
+% if ndims == 2:
+    t[off + 0] =  n[0]*u[off + 0] + n[1]*u[off + 1];
+    t[off + 1] = -n[1]*u[off + 0] + n[0]*u[off + 1];
+% elif ndims == 3:
+    if (fabs(n[0]) < ${t_tol})
+    {
+        fpdtype_t h = 1/(1 + n[0]);
+
+        t[off + 0] = n[0]*u[off + 0] + n[1]*u[off + 1] + n[2]*u[off + 2];
+        t[off + 1] = - n[1]*u[off + 0] + (n[0] + h*n[2]*n[2])*u[off + 1]
```
FreddieWitherden, line 17: Avoid having negatives out front (so `-a + b`  => `b - a`). Saves an instruction or two for compilers which do not rearrange math expressions.

pyfr/solvers/euler/kernels/flux.mako
```diff
@@ -24,3 +24,40 @@
...
+<%pyfr:macro name='inviscid_pri' params='s, p, v, c'>
+    fpdtype_t invrho = 1/s[0], E = s[${nvars - 1}];
+
+    // Compute the velocities
+% for i in range(ndims):
+    v[${i}] = invrho*s[${i + 1}];
+% endfor
+
+    // Compute the pressure
+    p = ${c['gamma'] - 1}*(E - 0.5*invrho*${pyfr.dot('s[{i} + 1]', i=ndims)});
```
FreddieWitherden, line 59: {i + 1}

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,224 @@
...
+    if ((rpmax <= 2) && (pmin <= bpv) && (bpv <= pmax))
+    {
+        p0 = bpv;
+    }
+    else if (bpv < pmin)
+    {
+        // Two-rarefaction case
+        fpdtype_t pre = pow(pl/pr, ${gmrtg});
+        fpdtype_t um  = (pre*vl[0]/cl + vr[0]/cr + ${trgm}*(pre - 1)) /
+                        (pre/cl + 1/cr);
```
FreddieWitherden, line 31: Crunch `1/cl` and `1/cr` up front and reuse.

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,224 @@
...
+       fpdtype_t bs = ${gmrgp}*ps;
+       fpdtype_t sapb = sqrt(${trgp} / (rs*(p + bs)));
+       f  = (p - ps)*sapb;
+       fd = sapb - 0.5*sapb*(p - ps) / (p + bs);
+    }
+</%pyfr:macro>
+
+// Primitive to 1D inviscid flux
+// w = [density, v_1, ..., v_ndims, p]^T
+<%pyfr:macro name='primitive_1dflux' params='w, f'>
```
FreddieWitherden, line 66: Seems like code duplication with what we have in flux.

pyfr/solvers/baseadvec/kernels/transform.mako
```diff
@@ -0,0 +1,83 @@
+<%namespace module='pyfr.backends.base.makoutil' name='pyfr'/>
+
+<% t_tol = 0.99 %>
+
+// Transforms to m = [1, 0, 0]^T
+// See Moler and Hughes 1999
+<%pyfr:macro name='transform_to' params='n, u, t, off'>
```
FreddieWitherden, line 7: I think we can avoid `off`.  With good use of () we should just be able to pass u as `'u + 1'` as an argument.

pyfr/solvers/euler/kernels/flux.mako
WillTrojak: I don't think you can do that here. At least it gives an error. This is also the syntax used here: https://github.com/PyFR/PyFR/blob/78884175acf239b588e7b812634ef8bb1adb8094/pyfr/solvers/euler/kernels/entropy.mako#L12

pyfr/solvers/baseadvec/kernels/transform.mako
WillTrojak: I tried a few permutation for this but couldn't find something that worked. Ultimately the offset it required for Compressible and incompressilbe support. As well as other systems of equations. (I remember at the time when I first wrote this we tried a number of different things to tidy this up)

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,211 @@
...
+       fpdtype_t bs = ${gmrgp}*ps;
+       fpdtype_t pbs_inv = 1 / (p + bs);
+       fpdtype_t sapb = sqrt(${trgp} * pbs_inv / rs);
+       f  = (p - ps)*sapb;
+       fd = sapb - 0.5*sapb*(p - ps) * pbs_inv;
+    }
+</%pyfr:macro>
+
+// Exact solve solution decision tree
+<% switch = 0 %>
```
FreddieWitherden, line 64: ?

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,211 @@
...
+
+    if ((pmax <= 2*pmin) && (pmin <= bpv) && (bpv <= pmax))
+    {
+        p0 = bpv;
+    }
+    else if (bpv < pmin)
+    {
+        // Two-rarefaction case
+        fpdtype_t pre = pow(pl / pr, ${gmrtg});
+        fpdtype_t um  = (pre*vl[0]*cr + vr[0]*cl + ${trgm}*(pre - 1)*cl*cr) /
```
FreddieWitherden, line 29: Move / down.

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,211 @@
...
+            {
+                w0[0] = rl*(p0p + ${gmrgp}) / (p0p*${gmrgp} + 1);
+                w0[1] = us;
+                w0[${nvars - 1}] = p0;
+            }
+        }
+    }
+    else
+    {
+% for i in range(ndims - 1):
```
FreddieWitherden, line 117: `range(1, ndims)` might make things easier?

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,211 @@
...
+
+// Exact Riemann solver
+<% kmax = 3 %>
+<%pyfr:macro name='rsolve_1d' params='ul, ur, nf'>
+    fpdtype_t vl[${ndims}], pl, cl, fsl, fdl;
+    fpdtype_t vr[${ndims}], pr, cr, fsr, fdr;
+    fpdtype_t p0, p1;
+    fpdtype_t w0[${nvars}];
+
+    // Compute the left/right srimitives
```
FreddieWitherden, line 175: Spelling.

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,211 @@
...
+    // Primitive to conservative
+    w0[${nvars - 1}] += ${tmgrgm}*w0[${nvars - 1}]
+                        + 0.5*w0[0]*(${pyfr.dot('w0[{i} + 1]', i=ndims)});
+% for i in range(ndims):
+    w0[${i + 1}] *= w0[0];
+% endfor
+
+    // 1D flux
+    fpdtype_t ps, vs[${ndims}];
+    ${pyfr.expand('inviscid_flux_1d', 'w0', 'nf', 'ps', 'vs')};
```
FreddieWitherden, line 208: Easier to inline this.

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,211 @@
...
+                        (pre*cr + cl);
+        fpdtype_t ptl = 1 - ${hgm}*(um - vl[0]) / cl;
+        fpdtype_t ptr = 1 + ${hgm}*(um - vr[0]) / cr;
+
+        p0 = 0.5*(pl*pow(ptl, ${tgrgm}) + pr*pow(ptr, ${tgrgm}));
+    }
+    else
+    {
+        // Two-shock case
+        fpdtype_t gl = sqrt(${trgp} / (rl*(${gmrgp}*pl + bpv)));
```
FreddieWitherden, line 39: Move comment up.

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,211 @@
...
+        fpdtype_t gr = sqrt(${trgp} / (rr*(${gmrgp}*pr + bpv)));
+        p0 = (gl*pl + gr*pr - vr[0] + vl[0]) / (gl + gr);
+    }
+</%pyfr:macro>
+
+// Star Flux, assuming covolume = 0. See Toro 2009 Eq.(4.86 - 4.87)
```
FreddieWitherden, line 45: This will look weird in generated code as the macros disappear.

pyfr/solvers/baseadvec/kernels/transform.mako
FreddieWitherden: Okay. We should at least allow off to be passed as an int rather than a string.  This should be relatively simple.  Better still would be to allow calling as `off=1` but that would require a tiny bit more code in the expand function.  Probably worth doing though.

pyfr/solvers/euler/kernels/rsolvers/exact.mako
WillTrojak: See toro. This is the point on x/t to sample, this is what people normally do, but I don't see a reason wy we would ever want it to be anything other than 0.

pyfr/solvers/euler/kernels/flux.mako
tarikdzanic: I think you can use the syntax `p = ${c['gamma'] - 1}*(E - 0.5*invrho*${pyfr.dot('s[{i} + 1]', i=(1, ndims+1))});` but I'm not 100% sure

pyfr/solvers/euler/kernels/rsolvers/exact.mako
tarikdzanic: I think you would only use 0 for interface Riemann solvers and other values if you want to compute exact solutions, right?

pyfr/solvers/euler/inters.py
```diff
@@ -15,6 +15,11 @@ def __init__(self, *args, **kwargs):
                 entmin_lhs=self._entmin_lhs, entmin_rhs=self._entmin_rhs
             )
 
+            self.p_min = self.cfg.getfloat('solver-entropy-filter', 'p_min',
+                                           1e-6)
+        else:
+            self.p_min = self._be.fpdtype_eps
```
tarikdzanic, line 21: Two things: 1) I think this should be a parameter open in the solver-interfaces section (but not documented kind of like the entropy filter iterative solver parameters) so that it can be modified by people who know what they're doing if you're not using the entropy filter. 2) I think the default should be at least 2*eps or some factor of eps (maybe 10?), otherwise it seems too prone to roundoff error.

FreddieWitherden:
I've pushed some improvements to how we handle macro arguments that should enable you to pass the offset as `off=1` which is cleaner than `'1'`.

pyfr/solvers/euler/kernels/rsolvers/exact.mako
WillTrojak: Yep, I've removed this setting

pyfr/solvers/euler/kernels/flux.mako
```diff
@@ -24,3 +24,40 @@
     f[${i}][${j + 1}] = rhov[${i}]*v[${j}]${' + p' if i == j else ''};
 % endfor
 </%pyfr:macro>
+
+<%pyfr:macro name='inviscid_flux_1d' params='s, f, p, v'>
```
FreddieWitherden, line 28: Lets move this to the exact solver file as it is only used in one place.

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,211 @@
...
+<% gamma = c['gamma'] %>
+
+// Initial guess for pressure
+<%pyfr:macro name='init_p' params='rl, vl, pl, cl, rr, vr, pr, cr, p0'>
+    fpdtype_t bpv = 0.5*(pl + pr) + 0.125*(vl[0] - vr[0])*(rl + rr)*(cl + cr);
+    bpv = max(0, bpv);
+    fpdtype_t pmin = min(pl, pr);
+    fpdtype_t pmax = max(pl, pr);
+
+    if ((pmax <= 2*pmin) && (pmin <= bpv) && (bpv <= pmax))
```
FreddieWitherden, line 21: No need for ()

pyfr/solvers/euler/kernels/rsolvers/exact.mako
FreddieWitherden: It should allow some of the conditions involving `${switch}` to become more natural.

pyfr/solvers/baseadvec/kernels/transform.mako
```diff
@@ -0,0 +1,81 @@
...
+        t[off + 0] = n[0]*u[off + 0] + n[1]*u[off + 1] + n[2]*u[off + 2];
+        t[off + 1] = (1 - h*n[1]*n[1])*u[off + 1] - h*n[0]*n[1]*u[off + 0]
+                     + n[1]*u[off + 2];
+        t[off + 2] = (1 - h*n[0]*n[0])*u[off + 0] - h*n[0]*n[1]*u[off + 1]
+                     + n[0]*u[off + 2];
+    }
+% endif
+</%pyfr:macro>
+
+// Transforms from m = [1, 0, 0]^T
```
FreddieWitherden, line 43: Mako comment.

pyfr/solvers/baseadvec/kernels/transform.mako
```diff
@@ -0,0 +1,81 @@
+<%namespace module='pyfr.backends.base.makoutil' name='pyfr'/>
+
+// Transforms to m = [1, 0, 0]^T
```
FreddieWitherden, line 3: Mako comment.  Also not sure what m is.

pyfr/solvers/euler/kernels/rsolvers/rsolve1d.mako
```diff
@@ -0,0 +1,21 @@
+<%namespace module='pyfr.backends.base.makoutil' name='pyfr'/>
```
FreddieWitherden, line 1: I'd ditch this and just inline the solver since it is the only one of its kind.

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,211 @@
...
+% for i in range(ndims - 1):
+        w0[${i + 2}] = vr[${i + 1}];
+% endfor
+        if (p0 > pr)
+        {
+            fpdtype_t p0p = p0 / pr;
+            fpdtype_t sr = vr[0] + cr*sqrt(${gprtg}*p0p + ${gmrtg});
+            if (${switch} >= sr)
+            {
+                w0[0] = rr;
```
FreddieWitherden, line 126: I'd ternary operator these assignments.

pyfr/solvers/euler/kernels/rsolvers/exact.mako
FreddieWitherden, line 45: Mako comment. Or move inside macro.

pyfr/solvers/euler/kernels/flux.mako
WillTrojak: I didn't do it here, but if you were to add other 1D Riemann solvers they would need this, hence why I put it here. 
Once this is merged, my plan was to refactor the HLLC and Roe solvers as they are much cleaner when implemented in this format.

pyfr/solvers/baseadvec/kernels/transform.mako
WillTrojak: Reference to the paper that it came from I think

pyfr/solvers/euler/kernels/rsolvers/rsolve1d.mako
WillTrojak: As above, this framework makes many of the other Riemann solvers easier to implement. This is especially true for other systems, so I think it should stay here for utility.

FreddieWitherden: Okay.  I think the best approach is to port one of them over, and then make that a PR.  The exact Riemann solver can then bolt on later as a separate PR (as it is somewhat more complicated).

WillTrojak:
Was there anything you wanted here?

FreddieWitherden:
I think it is almost ready to go.  Although there are likely some algebraic simplifications possible.  Do a quick pass and see if anything stands out.

WillTrojak:
Will do

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,219 @@
...
+
+<% gmrtg = (c['gamma'] - 1)/(2*c['gamma']) %>
+<% gprtg = (c['gamma'] + 1)/(2*c['gamma']) %>
+<% tgrgm = (2*c['gamma'])/(c['gamma'] - 1) %>
+<% grgm = c['gamma']/(c['gamma'] - 1) %>
+<% trgm = 2/(c['gamma'] - 1) %>
+<% trgp = 2/(c['gamma'] + 1) %>
+<% gmrgp = (c['gamma'] - 1)/(c['gamma'] + 1) %>
+<% hgm = 0.5*(c['gamma'] - 1) %>
+<% gamma = c['gamma'] %>
```
FreddieWitherden, line 12: Move this up to the top so you can just use `gamma` rather than `c['gamma']`.

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,219 @@
...
+
+<%pyfr:macro name='init_p' params='rl, vl, pl, cl, rr, vr, pr, cr, p0'>
```
FreddieWitherden, line 14: Inline as only expanded once.

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,219 @@
...
+    bpv = max(0, bpv);
+    fpdtype_t pmin = min(pl, pr);
+    fpdtype_t pmax = max(pl, pr);
+
+    if (pmax <= 2*pmin && pmin <= bpv && bpv <= pmax)
+    {
+        p0 = bpv;
+    }
+    else if (bpv < pmin)
+    {   // Two-rarefaction case
```
FreddieWitherden, line 25: Move comment up above `else if`.

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,219 @@
...
+    else
+    {   // Two-shock case
+        fpdtype_t gl = sqrt(${trgp} / (rl*(${gmrgp}*pl + bpv)));
+        fpdtype_t gr = sqrt(${trgp} / (rr*(${gmrgp}*pr + bpv)));
+        p0 = (gl*pl + gr*pr - vr[0] + vl[0]) / (gl + gr);
+    }
+</%pyfr:macro>
+
+<%pyfr:macro name='star_flux' params='p, ps, rs, cs, f, fd'>
+    ## Star Flux, assuming covolume = 0. See Toro 2009 Eq.(4.86 - 4.87)
```
FreddieWitherden, line 43: May want to use a C comment here so it will make it into the generated code.

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,219 @@
...
+    if (p <= ps)
+    {
+       fpdtype_t pr = p / ps;
+       f = ${trgm}*cs*(pow(pr, ${gmrtg}) - 1);
+       fd = pow(pr, ${-gprtg}) / (rs*cs);
```
FreddieWitherden, line 48: Does this simplify given the definitions of p, ps, rs and cs?

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,219 @@
...
+    else
+    {
+       fpdtype_t pbs_inv = 1 / (p + ${gmrgp}*ps);
+       fpdtype_t sapb = sqrt(${trgp} * pbs_inv / rs);
+       f = (p - ps)*sapb;
+       fd = sapb - 0.5*f*pbs_inv;
+    }
+</%pyfr:macro>
+
+<%pyfr:macro name='riemann_decision' params='rl, vl, pl, cl, rr, vr, pr, cr, us, p0, w0'>
```
FreddieWitherden, line 59: Inline as only called once and has lots of arguments.

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,219 @@
...
+            {
+                w0[0] = rr;
+                w0[1] = vr[0];
+                w0[${nvars - 1}] = pr;
+            }
+            else
+            {
+                fpdtype_t p0p = p0 / pr;
+                fpdtype_t cmr = cr*pow(p0p, ${gmrtg});
+                if (us >= -cmr)
```
FreddieWitherden, line 143: Eliminate `-` by changing comparison operator.

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,219 @@
...
+                }
+            }
+        }
+    }
+</%pyfr:macro>
+
+<%pyfr:macro name='primitives' params='s, p, v, c'>
+    fpdtype_t invrho = 1/s[0], E = s[${nvars - 1}];
+
+    ## Compute the velocities
```
FreddieWitherden, line 164: Use a C comment here so it makes it into the generated source code.

pyfr/solvers/euler/kernels/rsolvers/hllc.mako
```diff
@@ -1,60 +1,51 @@
 <%namespace module='pyfr.backends.base.makoutil' name='pyfr'/>
 <%include file='pyfr.solvers.euler.kernels.flux'/>
 
-<%pyfr:macro name='rsolve' params='ul, ur, n, nf'>
+<%pyfr:macro name='rsolve_1d' params='ul, ur, nf'>
```
FreddieWitherden, line 4: Can you check if this kernel previously vectorized for AVX-512/NEON that it still does so?

pyfr/solvers/euler/kernels/rsolvers/exact.mako
```diff
@@ -0,0 +1,211 @@
...
+    // Compute the left/right primitives
+    ${pyfr.expand('primitives', 'ul', 'pl', 'vl', 'cl')};
+    ${pyfr.expand('primitives', 'ur', 'pr', 'vr', 'cr')};
+
+    // Inital pressure guess
+    fpdtype_t rl = ul[0];
+    fpdtype_t rr = ur[0];
+    fpdtype_t bpv = fmax(0, 0.125*(vl[0] - vr[0])*(rl + rr)*(cl + cr) +
+                         0.5*(pl + pr));
+    fpdtype_t pmin = fmin(pl, pr);
```
FreddieWitherden, line 64: Check convention for if we use min/max or fmin/fmax.

WillTrojak: I think fmin/fmax are more robust than min/max, I've had some issues with rounding on the openmp backend.

pyfr/solvers/euler/kernels/rsolvers/hllc.mako
FreddieWitherden: Just following up here.

WillTrojak: Using godbolt with gcc 13.2 x86 and adding `-mavx2` (as there isn't access to `-march=native`) in addition to the standard pyfr flags, the HLLC kernel is vectorising. iei using `ymm` registers and using commands `vmulpd` etc.

PR #335, opened 2023-02-07
Support for variable physical time stepping

sambitmishra98:
This PR is followed by work done in PR #273 and PR #317.

Major changes added to code:
- Physical time `dt` is allowed to be variable.
- Pseudo-time `dtau` is updated for any change in `dt` such that the ratio `dt/dtau` is kept constant. 
    - If `[solver-dual-time-integrator-multip]`  is enabled the entire field is updated at every multi-p level.
    - If local-PI controller is enabled, `pseudo-dt-min-mult` and `pseudo-dt-max-mult` are both updated.

With the above changes, the code is now ready for us to add adaptive physical time-stepping to the dual-time integrator.

Support is also added for the physical time to forecast if solution is needed at a specific time `t` within the next `3*dt` from the current time `tcurr`, and physical time is appropriately modified to exactly hit the desired time.

The code was tested with a Taylor green vortex simulation with 64^3 hex elements and polynomial order 3. To test the robustness of the code, I used dt = 0.0123 and outputted the solution file at `dt-out = 2.34`. I replaced line 60 in `pyfr/integrators/dual/phys/controllers.py` with `dt = self._dt* np.random.rand()` to find that the solution still converges and matches with the solution obtained with `develop` branch.


pyfr/integrators/dual/phys/controllers.py
```diff
@@ -47,8 +51,21 @@ def advance_to(self, t):
...
         while self.tcurr < t:
+
+            if self.tcurr + 2*self._dt < t < self.tcurr + 3*self._dt:
+                dt = 0.5*(t - self.tcurr)
+            elif self.tcurr + self._dt < t < self.tcurr + 2*self._dt:
+                dt = t - self.tcurr
+            else:
+                dt = self._dt
+
+            if self.pseudointegrator.dt != dt:
```
semi-h, line 62: I 'd just overwrite without the if check

pyfr/integrators/std/controllers.py
```diff
@@ -77,7 +77,12 @@ def advance_to(self, t):
 
         while self.tcurr < t:
             # Decide on the time step
-            dt = max(min(t - self.tcurr, self._dt), self.dtmin)
+            if self.tcurr + 2*self._dt < t < self.tcurr + 3*self._dt:
```
semi-h, line 80: I don't think we ever had any problems with std controller and the existing timestep strategy. std integrators can take a very small step and it doesn't cause any issues as far as I know

semi-h: Unless you had any issues I think it is better to stick with the current approach

pyfr/integrators/dual/phys/controllers.py
```diff
@@ -13,8 +15,10 @@ def __init__(self, *args, **kwargs):
             for csh in self.completed_step_handlers:
                 csh(self)
 
-    def _accept_step(self, idxcurr):
-        self.tcurr += self._dt
+        np.random.seed(0)
```
semi-h, line 18: I think you forgot to remove this line

pyfr/integrators/dual/phys/controllers.py
```diff
@@ -47,8 +51,21 @@ def advance_to(self, t):
             raise ValueError('Advance time is in the past')
 
         while self.tcurr < t:
+
+            if self.tcurr + 2*self._dt < t < self.tcurr + 3*self._dt:
```
semi-h, line 55: I think this is a way to ensure somewhat smoother transition down to the required level, but not sure if this is the best way. It looks like this will decrease the dt first and then increase a bit in some cases. Can you tabulate how the dt changes for a few different scenarios?

semi-h: And if your physical time stepper can handle this level of dt and the subsequent dtau change without introducing any error it means that the rest of stuff you're doing with the dtau works very well!

semi-h (changes requested):

pyfr/integrators/dual/phys/controllers.py
sambitmishra98: Overwriting dt every step  also means modifying dt/dtau ratio every step. The `dtau` fields and/or (`pseudo-dt-min-mult` and `pseudo-dt-max-mult`) will be changed every time-step. I would guess that will be expensive.

pyfr/integrators/std/controllers.py
sambitmishra98: I forgot to remove this, I was testing the code for `std` before moving to `dual`.

pyfr/integrators/dual/pseudo/pseudosteppers.py
```diff
@@ -203,6 +203,10 @@ def __init__(self, *args, **kwargs):
             'pyfr.integrators.dual.pseudo.kernels.rkvdh2pseudo'
         )
 
+    def dtau_mats_multiplied(self, y):
```
FreddieWitherden, line 206: This is going to be expensive.  Best resolved with a runtime scalar passed to the kernel telling it how to pre-scale the field.  Usually it can be 1.0 (so do nothing), but when we want to change the matrix we can set it, call the regular kernel.

pyfr/integrators/dual/phys/controllers.py
sambitmishra98: > Can you tabulate how the dt changes for a few different scenarios?

That question actually made me get a bug. I was over-shooting the `dt` by about 1.5 times to hit the desired time. This is how dt varied with time for fixed case. We want this to be lesser than the set `dt` value. 

variable-dt_fixed

This is how the dt varied when I added random dt at every step.

variable-dt_random

I just fixed the bug where `dt` is overshooting the config-set value.

pyfr/integrators/dual/pseudo/pseudosteppers.py
sambitmishra98: I could add the scalar factor in `pyfr/integrators/dual/pseudo/kernels/localerrest.mako#L15`: `vfac = alpha * min(${maxf}, max(${minf}, ${saff}*ufac));`.

FreddieWitherden:
@WillTrojak @semi-h What is the current review status?

pyfr/integrators/dual/pseudo/pseudosteppers.py
sambitmishra98: I added this change. We now modify the dtau field in the first pseudo-iteration, if the scaling multiplier alpha is not equal to 1.

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -103,24 +112,28 @@ def __init__(self, *args, **kwargs):
...
+        if not dtau_minf <= 1 < dtau_maxf:
+            raise ValueError('Invalid pseudo-dt-max-mult, pseudo-dt-min-mult')
 
         # Register a kernel to compute local error
         self.backend.pointwise.register(
             'pyfr.integrators.dual.pseudo.kernels.localerrest'
         )
 
+        # Storage for kernels that require dtau-min and dtau-max at runtime
+        self.ele_scal_upts_locs = []
```
FreddieWitherden, line 135: Not sure this should be needed.

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -137,8 +151,31 @@ def __init__(self, *args, **kwargs):
...
+        self.alpha = y
+        self._dtau_min *= y
+        self._dtau_max *= y
+
+        for i in self.ele_scal_upts_locs:
+            for k in self.pintgkernels['localerrest', i]:
+                k.bind(dtau_min = self._dtau_min, dtau_max = self._dtau_max,
+                       alpha = self.alpha)
+
+    def dtau_multiplier_reset(self):
```
FreddieWitherden, line 172: Why can this not just call dtau_multiplied with y = 1?

pyfr/integrators/dual/phys/controllers.py
```diff
@@ -47,8 +49,21 @@ def advance_to(self, t):
             raise ValueError('Advance time is in the past')
 
         while self.tcurr < t:
+
+            if self.tcurr + self._dt < t < self.tcurr + 2*self._dt:
```
semi-h, line 53: I don't really feel like this is the best strategy. If target t is something like n*dt+0.01dt, I think it will take a while to go down to that level. And it will be even worse if t is n*dt+0.001dt and so on. You actually provided a plot for the previous strategy (tcurr+2dt<t<tcurr+3dt), but not for the current one. Can you do a plot where t%dt varies from 0.01dt to 0.99dt and show how the dt changes and how many steps it takes until getting to the target t?

What I had in mind was more like limiting the variation in t with a min and max multiplier such as 0.9 and 1.1. And as you march in time and get close to target t, say when you're ~10dt away from it, you just fix dt by multiplying a coefficient between 0.9 and 1, so that target t can be achieved without changing dt again.

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -103,24 +112,28 @@ def __init__(self, *args, **kwargs):
...
         tplargs['saff'] = self.cfg.getfloat(sect, 'safety-fact', 0.8)
-        tplargs['dtau_maxf'] = self.cfg.getfloat(sect, 'pseudo-dt-max-mult',
-                                                 3.0)
+        
+        dtau_minf = self.cfg.getfloat(sect, 'pseudo-dt-min-mult', 1.0)
+        dtau_maxf = self.cfg.getfloat(sect, 'pseudo-dt-max-mult', 3.0)
+
+        self._dtau_min = dtau_minf * self._dtau
+        self._dtau_max = dtau_maxf * self._dtau
+        self.alpha = 1.0
```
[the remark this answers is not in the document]
sambitmishra98: `dtau_fieldf` fine?

pyfr/integrators/dual/pseudo/pseudocontrollers.py
sambitmishra98: We are changing all three variables `self.alpha`, `self._dtau_min` and `self._dtau_max` in the function `dtau_multiplied`. We are only resetting `self.alpha = 1.0` in `dtau_multiplier_reset` and not changing the min and max limits.

pyfr/integrators/dual/pseudo/pseudocontrollers.py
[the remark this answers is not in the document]
sambitmishra98: switched from `alpha` to `dtau_fieldf`.

pyfr/integrators/dual/pseudo/pseudocontrollers.py
FreddieWitherden: But all it does is multiply them by one, so in effect they don't change.  Saves a lot of code duplication.

sambitmishra98: You're right. I had the wrong idea, I thought it resets min/max limits too. I will add this change.

pyfr/integrators/dual/phys/controllers.py
```diff
@@ -47,8 +47,21 @@ def advance_to(self, t):
...
+                dt = 0.5*(t - self.tcurr)
+            elif self.tcurr < t < self.tcurr + self._dt:
+                dt = t - self.tcurr
+            else:
+                dt = self._dt
+
+            if self.pseudointegrator.dt != dt:
+                # Change dt in pseudo-integrator (and multi-p levels)
+
```
FreddieWitherden, line 60: Remove newline.

pyfr/integrators/dual/phys/controllers.py
```diff
@@ -47,8 +47,21 @@ def advance_to(self, t):
             raise ValueError('Advance time is in the past')
 
         while self.tcurr < t:
+
```
FreddieWitherden, line 50: Remove newline.

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -137,8 +151,30 @@ def __init__(self, *args, **kwargs):
                     )
                 )
 
+            for i in self.ele_scal_upts_locs:
```
FreddieWitherden, line 154: This should be accomplished by calling dtau_multiplied.

FreddieWitherden (changes requested):

pyfr/integrators/dual/pseudo/pseudocontrollers.py
sambitmishra98: Are the required kernels stored in another list?

I think I tried to separate out the for loop I am using it in. `self.system.ele_map` is deleted later on, so I couldn't loop over `self.system.ele_map.values()`.

FreddieWitherden: You should not need to store this at all.  Just iterate through self.pintgkernels and look for those that start with `localerrest`.

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -129,16 +142,28 @@ def __init__(self, *args, **kwargs):
...
             for i, err in enumerate(ele.scal_upts):
+                self.ele_scal_upts_locs.append(i)
                 self.pintgkernels['localerrest', i].append(
                     self.backend.kernel(
                         'localerrest', tplargs=tplargs,
                         dims=[ele.nupts, ele.neles], err=err,
                         errprev=err_prev, dtau_upts=dtaumat
                     )
                 )
+            self.dtau_multiplied(1.0)
```
FreddieWitherden, line 153: De-indent one level.

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -129,16 +142,28 @@ def __init__(self, *args, **kwargs):
...
         self.backend.commit()
 
+    def dtau_multiplied(self, y):
+        self.dtau_fieldf = y
+        self._dtau_min *= y
+        self._dtau_max *= y
+
+        for i in self.ele_scal_upts_locs:
+            for k in self.pintgkernels['localerrest', i]:
+                k.bind(dtau_min = self._dtau_min, dtau_max = self._dtau_max,
```
FreddieWitherden, line 164: Remove spaces around = (PEP8).

pyfr/integrators/dual/phys/controllers.py
sambitmishra98: Instead of changing dt within `[dt,2*dt]` prior to desired time, I am now changing within `[dt,i*dt]` prior to the desired time. `i` here is set by the config file. 
In total `dt` is updated thrice, once when it is near the desired time, next when it needs to exactly hit the desired time in a single step, and finally when it is equal to or greater than the desired time.

pyfr/integrators/dual/pseudo/multip.py
```diff
@@ -280,6 +280,18 @@ def _mg_regidx(self):
...
         return self.pintg._aux_regidx
 
+    @property
+    def dt(self):
+        return self.pintg._dt
+
+    @dt.setter
+    def dt(self, y):
+        if self.pintg.cfg.get('solver-time-integrator', 'pseudo-controller') == 'local-pi':
+            self.pintg.dtau_multiplied(y/self.pintg._dt)
```
FreddieWitherden, line 290: Spacing.

pyfr/integrators/dual/phys/steppers.py
```diff
@@ -27,11 +27,11 @@ def step(self, t, dt):
...
             self.pseudointegrator.pseudo_advance(t + dt*tc)
 
-        self._finalize_step()
+        self._finalize_step(dt)
 
-    def _finalize_step(self):
+    def _finalize_step(self, dt):
         if not self.fsal:
-            bcoeffs = [bt*self._dt for bt in self.b]
+            bcoeffs = [bt*dt for bt in self.b]
```
FreddieWitherden, line 34: Why can `self.dt` not be used here?

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -59,6 +59,15 @@ def _resid(self, rcurr, rold, dt_fac):
     def _update_pseudostepinfo(self, niters, resid):
         self.pseudostepinfo.append((self.ntotiters, niters, resid))
 
+    @property
+    def dt(self):
+        return self._dt
+
+    @dt.setter
+    def dt(self, y):
+        if self.cfg.get('solver-time-integrator', 'pseudo-controller') == 'local-pi':
```
FreddieWitherden, line 68: This is bad form.  Implement in the appropriate child class to avoid the check.  (The base class should not need to know anything about its subclasses.)

sambitmishra98: Shifted this to `DualPIPseudoController`

pyfr/integrators/dual/phys/steppers.py
sambitmishra98: `self.dt` does not exist for `BaseDIRKStepper` and `self._dt` is as per the value given in the config file. We do have `dt = self.pseudointegrator.dt`.

I think we should rename `self._dt` --> `self._dt_base` to differentiate between what we are updating the dt to and what is given in the config file. We are no longer using self._dt anywhere outside the class. This could later be misleading.

I am setting `self.dt` as `self.pseudointegrator.dt` for now.

pyfr/integrators/std/controllers.py
sambitmishra98: Apart from a small bug here, I guess this stays as it is? 

If we do want to gradually change dt here, since `self._dt` is changes with local controller here, we cannot copy-paste the same idea for implicit here. One option would be to assume the self._dt about 10 timesteps prior to desired time to be set as a constant, untill the desired time is reached.

pyfr/integrators/dual/phys/controllers.py
```diff
@@ -47,8 +55,23 @@ def advance_to(self, t):
...
+                    self._if_near = True
+                dt = self._dt_near
+            elif self.tcurr < t <= self.tcurr + self._dt:
+                dt = t - self.tcurr
+            else:
+                dt = self._dt
+                self._if_near = False
+
+            if self.pseudointegrator.dt != dt:
+                # Change dt in pseudo-integrator (and multi-p levels)
```
FreddieWitherden, line 70: Just set it.

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -59,7 +59,6 @@ def _resid(self, rcurr, rold, dt_fac):
     def _update_pseudostepinfo(self, niters, resid):
         self.pseudostepinfo.append((self.ntotiters, niters, resid))
 
-
```
FreddieWitherden, line 62: Keep newline (PEP8, two newlines between classes).

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -103,24 +102,27 @@ def __init__(self, *args, **kwargs):
...
-        tplargs['dtau_max'] = tplargs['dtau_maxf'] * self._dtau
+        if not dtau_minf <= 1 < dtau_maxf:
+            raise ValueError('Invalid pseudo-dt-max-mult, pseudo-dt-min-mult')
 
         # Register a kernel to compute local error
         self.backend.pointwise.register(
             'pyfr.integrators.dual.pseudo.kernels.localerrest'
         )
 
+        # Storage for kernels that require dtau-min and dtau-max at runtime
```
FreddieWitherden, line 124: Remove comment.

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -136,12 +138,33 @@ def __init__(self, *args, **kwargs):
...
+        self.dtau_multiplied(1.0)
 
         self.backend.commit()
 
+    def dtau_multiplied(self, y):
+        self.dtau_fieldf = y
+        self._dtau_min *= y
+        self._dtau_max *= y
+
+        for key in self.pintgkernels:
```
FreddieWitherden, line 150: `for k, idx in self.pintgkernels`

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -136,12 +138,33 @@ def __init__(self, *args, **kwargs):
...
     def localerrest(self, errbank):
         self.backend.run_kernels(self.pintgkernels['localerrest', errbank])
 
+    @property
+    def dt(self):
+        return self._dt
+
+    @dt.setter
+    def dt(self, y):
+        self.dtau_multiplied(y/self.dt)
```
FreddieWitherden, line 165: Space around /

pyfr/integrators/std/controllers.py
```diff
@@ -77,7 +77,12 @@ def advance_to(self, t):
 
         while self.tcurr < t:
             # Decide on the time step
-            dt = max(min(t - self.tcurr, self._dt), self.dtmin)
+            if self.tcurr + self._dt < t < self.tcurr + 2*self._dt:
```
FreddieWitherden, line 80: We should be consistent for std and dual controllers about what we do here.

pyfr/integrators/dual/pseudo/multip.py
```diff
@@ -280,6 +280,18 @@ def _mg_regidx(self):
...
+    def dt(self):
+        return self.pintg._dt
+
+    @dt.setter
+    def dt(self, y):
+        if self.pintg.cfg.get('solver-time-integrator', 'pseudo-controller')=='local-pi':
+            self.pintg.dtau_multiplied(y/self.dt)
+
+        for l in self.levels:
+            self.pintgs[l]._dt = y
```
FreddieWitherden, line 293: We should not be setting private member variables.

pyfr/integrators/dual/phys/controllers.py
```diff
@@ -47,8 +55,23 @@ def advance_to(self, t):
             raise ValueError('Advance time is in the past')
 
         while self.tcurr < t:
+            if self.tcurr + self._dt <= t < self.tcurr + self.i*self._dt:
```
FreddieWitherden, line 58: @semi-h Can you inspect this algorithm?

pyfr/integrators/dual/phys/controllers.py
```diff
@@ -47,8 +55,23 @@ def advance_to(self, t):
...
+            else:
+                dt = self._dt
+                self._if_near = False
+
+            if self.pseudointegrator.dt != dt:
+                # Change dt in pseudo-integrator (and multi-p levels)
+                self.pseudointegrator.dt = dt
+
             # Take the physical step
-            self.step(self.tcurr, self._dt)
```
FreddieWitherden, line 51: I still think it is worth passing dt here as it gives a nicer interface.

pyfr/integrators/dual/phys/controllers.py
```diff
@@ -8,13 +8,21 @@ def __init__(self, *args, **kwargs):
         # Solution filtering frequency
         self._fnsteps = self.cfg.getint('soln-filter', 'nsteps', '0')
 
+        self.i = self.cfg.getint('solver-time-integrator', 'dt-switch', 10)
+
+        if self.i>1:
+            raise ValueError('dt-switch must be greater than 1')
+
+        self._if_near = False
```
FreddieWitherden, line 16: We should not need this state variable.

pyfr/integrators/dual/phys/steppers.py
```diff
@@ -22,16 +22,16 @@ def __init__(self, *args, **kwargs):
...
+    def step(self, t):
         for s, (sc, tc) in enumerate(zip(self.a, self.c)):
-            self.pseudointegrator.init_stage(s, sc, dt)
-            self.pseudointegrator.pseudo_advance(t + dt*tc)
+            self.pseudointegrator.init_stage(s, sc, self.dt)
+            self.pseudointegrator.pseudo_advance(t + self.dt*tc)
 
         self._finalize_step()
 
     def _finalize_step(self):
```
FreddieWitherden, line 32: Maybe pass dt in here, too.

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -136,12 +138,33 @@ def __init__(self, *args, **kwargs):
                         errprev=err_prev, dtau_upts=dtaumat
                     )
                 )
+        self.dtau_multiplied(1.0)
```
FreddieWitherden, line 141: Add newline above.

pyfr/integrators/dual/phys/controllers.py
semi-h: We had a chat with @sambitmishra98 on how to best manage the changes in dt to meet the target t.

I think there are two important things
- We should aim to minimize the number of adjustments for the dt.
- And we should ideally stay as close as to the original dt as possible.

My idea is that we take an input from the ini file for the minimum factor we multiply dt to meet the target t. By default we can set it to 0.9 for example. Then, we'll need to monitor how far we are from the target and in case of factor 0.9 we should look to adjust dt when we're roughly 10 steps away from the target. What do you think about this @FreddieWitherden?

And on top of this Sambit mentioned that he wants to solve the machine precision related errors in t as well. So maybe we adjust the dt as I described above to roughly hit the target t, and then at the very last step we adjust dt one more time to get to the target t as accurately as possible.

I think the current implementation takes the input from ini file to determine how many time steps before we should adjust the dt. It'll be better if the input is the minimum factor that we multiply the dt, and then we figure out how many time steps before we need to take action.

FreddieWitherden: That seems sensible. Edge cases will need to be handled (for example when someone wants a plugin called so frequently that we do not have 10 steps to wok with).  Further, the code should be shared between std and dual integrators.

pyfr/integrators/std/controllers.py
sambitmishra98: I added a function in the base integrator class to be consistent

pyfr/integrators/dual/pseudo/multip.py
```diff
@@ -280,6 +280,18 @@ def _mg_regidx(self):
 
         return self.pintg._aux_regidx
 
+    @property
+    def dt(self):
+        return self.pintg._dt
+
+    @dt.setter
+    def dt(self, y):
+        if self.pintg.cfg.get('solver-time-integrator', 'pseudo-controller')=='local-pi':
```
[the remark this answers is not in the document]
sambitmishra98: removed

pyfr/integrators/dual/phys/controllers.py
sambitmishra98: Could you explain more on that? I am not sure what you mean by a nicer interface. I have used self._dt in many places, so I will change accordingly.

pyfr/integrators/base.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
         self.nacptchain = 0
 
         # Current and minimum time steps
-        self._dt = cfg.getfloat('solver-time-integrator', 'dt')
+        self._dt = self.__dt = cfg.getfloat('solver-time-integrator', 'dt')
```
FreddieWitherden, line 44: This `__dt` is messy and confusing.

pyfr/integrators/base.py
```diff
@@ -62,6 +62,27 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
...
+        self.near_t = self.cfg.getint('solver-time-integrator', 'dt-switch', 10)
+        self._dt_near = None
+
+    def adjust_step(self, t):
+
+        if self.tcurr + self.near_t*self.__dt <= t:
+            # Default, target time is not near
+            self._dt = self.__dt
+        else:
+            if self.tcurr + self.__dt <= t:
```
FreddieWitherden, line 75: Merge as an `elif`

pyfr/integrators/base.py
```diff
@@ -114,7 +135,7 @@ def get_plugin_data_prefix(name, suffix):
 
     def call_plugin_dt(self, dt):
         ta = self.tlist
-        tb = deque(np.arange(self.tcurr, self.tend, dt).tolist())
+        tb = deque(np.arange(self.tend-dt, self.tcurr, -dt).tolist()[::-1])
```
FreddieWitherden, line 138: Spacing around `-`

FreddieWitherden: Also, I think `arange` can give a reverse array directly.

pyfr/integrators/dual/phys/controllers.py
FreddieWitherden: But it is confusing.  We have an argument `dt` to `_accept_step` but instead are using `self._dt`. What if these values are different?

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -106,24 +111,24 @@ def __init__(self, *args, **kwargs):
...
-        tplargs['dtau_max'] = tplargs['dtau_maxf'] * self._dtau
+        if not dtau_minf <= 1 < dtau_maxf:
+            raise ValueError('Invalid pseudo-dt-max-mult, pseudo-dt-min-mult')
 
         # Register a kernel to compute local error
         self.backend.pointwise.register(
             'pyfr.integrators.dual.pseudo.kernels.localerrest'
         )
-
+
```
FreddieWitherden, line 131: Indentation.

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -136,12 +141,29 @@ def __init__(self, *args, **kwargs):
...
+    def pseudo_step_multiplied(self, y):
+        self.dtau_min *= y
+        self.dtau_max *= y
+
+        for k, idx in self.pintgkernels:
+            if k == 'localerrest':
+                for kk in self.pintgkernels[k, idx]:
+                    kk.bind(dtau_min=self.dtau_min, 
+                        dtau_max=self.dtau_max, dtau_fieldf=y)
```
FreddieWitherden, line 165: Indentation.

pyfr/integrators/base.py
sambitmishra98: numpy arange function does not seem to have anything to reverse the array.

pyfr/integrators/dual/pseudo/pseudocontrollers.py
sambitmishra98: You mean like this? I tried copying the format in lines 141-144.

pyfr/integrators/base.py
sambitmishra98: I changed the name to _dt_far, similar to _dt_near

pyfr/integrators/dual/phys/controllers.py
sambitmishra98: I had first tried to use a local variable `dt`. Since `self._dt` also occurs in multip.py as `self.pintgs[l]._dt = self._dt`, I deicided to keep the `self._dt` for consistency.

sambitmishra98: The case of differing `dt` and `self._dt` would occur when we are adding adaptive physical time stepping, in which case we will have to rewind to a few time-steps before `tcurr`. Errors slowly build up, so there wouldn't be much point to rejecting one step the way we do for explicit time stepping.
So maybe we could decide how to handle the variables then?

pyfr/integrators/base.py
```diff
@@ -62,6 +62,27 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
...
         self.abort = False
 
+        # Smoothly step to target time in the last near_t steps
+        self.near_t = self.cfg.getint('solver-time-integrator', 'dt-switch', 10)
+        self._dt_far = cfg.getfloat('solver-time-integrator', 'dt')
+        self._dt_near = None
+
+    def adjust_step(self, t):
+
+        if self.tcurr + self.near_t*self._dt_far <= t:
```
semi-h, line 72: I think we need to investigate this carefully. For example lets say 
tcurr = 3.0
dt_far = 1.0
target t = 15.0
Then there shouldn't be any adjustments to the dt as we approach target t because we can get there by doing nothing, but the current algorithm fiddles with dt unnecessarily.

semi-h (changes requested):
I think the PR is in good shape in general. I like the way pseudo integrator keeps the dt/pseudo_dt ratio constant and the work around this.

It is only the adjust_step function that needs more work.
I think it is not a good idea to input near_t (number of timesteps before starting to adjust dt). I think it is better to take an input from the user to determine how small they would like dt to go down to match the target t's. For example, if the user is happy to march in time with a timestep = 0.1dt then most cases we can wait until the very last step to do the necessary adjustment in dt. If the the minimum timestep user wants is around 0.75dt, then it is a bit more restrictive, and we should look to start playing with dt a bit earlier depending on the target t.
I think you can write a standalone python function just to explore how the dt changes depending on a few inputs. And it is important to pay attention to edge cases.

pyfr/integrators/base.py
```diff
@@ -62,6 +62,27 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
         # Abort computation
         self.abort = False
 
+        # Smoothly step to target time in the last near_t steps
+        self.near_t = self.cfg.getint('solver-time-integrator', 'dt-switch', 10)
+        self._dt_far = cfg.getfloat('solver-time-integrator', 'dt')
```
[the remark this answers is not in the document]
sambitmishra98: How about using `self._dt` to store input dt? 
I had this idea of using `self._dt` and `self._dtau` as inputs from the config file, whose values do not change. We could use self.dt and self.dtau as modified values.
This change would need to be done in every other code using it though.

pyfr/integrators/base.py
sambitmishra98: The algorithm changes `self._dt` once to `self.dt_near` and once when less than a `self._dt_far` step away. By default, it uses `self.dt_far`. 
So between one target `t` and the next, the value of `self._dt` changes 3 times.

The same will happen for the example you gave. There may be two adjustments to `self._dt` as we approach to target t to fix machine precision errors.

pyfr/integrators/base.py
```diff
@@ -62,6 +62,27 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
...
+        self._dt_near = None
+
+    def adjust_step(self, t):
+
+        if self.tcurr + self.near_t*self._dt_far <= t:
+            # Default, target time is not near
+            self._dt = self._dt_far
+        elif (self.tcurr + self.near_t*self._dt_far > t) and (self.tcurr + self._dt_far <= t):
```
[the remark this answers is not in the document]
sambitmishra98: Shortened the line.

pyfr/integrators/base.py
sambitmishra98: @semi-h I have tested a few cases in a [Google Colab notebook](https://colab.research.google.com/drive/1upThO_hbxr25ThikrYOAJa-IHNGjq-pD?usp=sharing)

sambitmishra98:
I ran a TGV case with dt-fact = 0.9, 0.95 and compared with base. We see a clear win in terms of how less the physical time step changes.

image

image

pyfr/integrators/base.py
```diff
@@ -62,6 +62,27 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
         # Abort computation
         self.abort = False
 
+        # Smoothly step to target time in the last near_t steps
+        self.dt_fact = self.cfg.getfloat('solver-time-integrator', 'dt-fact', 0.9)
+        self._dt_in = cfg.getfloat('solver-time-integrator', 'dt')
+        self._dt_near = None
+
+    def adjust_step(self, t):
+
```
FreddieWitherden, line 71: Remove newline

pyfr/integrators/base.py
```diff
@@ -62,6 +62,27 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
...
+        t_diff = t - self.tcurr
+        steps_with_dt_far = t_diff/self._dt_in
+        steps_to_t = int(np.ceil(steps_with_dt_far))
```
FreddieWitherden, line 74: Use negated integer division.

pyfr/integrators/base.py
```diff
@@ -62,6 +62,27 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
...
 
+        # Smoothly step to target time in the last near_t steps
+        self.dt_fact = self.cfg.getfloat('solver-time-integrator', 'dt-fact', 0.9)
+        self._dt_in = cfg.getfloat('solver-time-integrator', 'dt')
+        self._dt_near = None
+
+    def adjust_step(self, t):
+
+        t_diff = t - self.tcurr
+        steps_with_dt_far = t_diff/self._dt_in
```
FreddieWitherden, line 73: Spacing.

pyfr/integrators/base.py
```diff
@@ -62,6 +62,27 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
...
+        steps_to_t = int(np.ceil(steps_with_dt_far))
+
+        if steps_with_dt_far < 1:
```
FreddieWitherden, line 76: == 0?

pyfr/integrators/base.py
```diff
@@ -62,6 +62,27 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
...
+            self._dt_near = None
+            self._dt = t_diff
+        elif (steps_with_dt_far-1)/(steps_to_t-1) < self.dt_fact:
```
FreddieWitherden, line 79: Spacing

pyfr/integrators/base.py
```diff
@@ -114,7 +135,7 @@ def get_plugin_data_prefix(name, suffix):
 
     def call_plugin_dt(self, dt):
         ta = self.tlist
-        tb = deque(np.arange(self.tcurr, self.tend, dt).tolist())
+        tb = deque(np.arange(self.tend - dt, self.tcurr, -dt).tolist()[::-1])
```
FreddieWitherden, line 138: arange can generate reversed lists

pyfr/integrators/base.py
```diff
@@ -58,6 +58,29 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
...
+        # Smoothly step to target time in the last near_t steps
+        self.fact = self.cfg.getfloat('solver-time-integrator', 'dt-fact', 0.9)
+        self.fact_max = 1.001
+        self._dt_in = cfg.getfloat('solver-time-integrator', 'dt')
+        self._dt_near = None
+
+    def adjust_step(self, t):
+        t_diff = t - self.tcurr
+        steps_with_dt_far = t_diff / self._dt_in
+        steps_to_t = -((steps_with_dt_far/self.fact_max) // -1)
```
FreddieWitherden, line 70: I think this will simplify a little

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -106,18 +111,18 @@ def __init__(self, *args, **kwargs):
         tplargs['maxf'] = self.cfg.getfloat(sect, 'max-fact', 1.01)
         tplargs['minf'] = self.cfg.getfloat(sect, 'min-fact', 0.98)
         tplargs['saff'] = self.cfg.getfloat(sect, 'safety-fact', 0.8)
-        tplargs['dtau_maxf'] = self.cfg.getfloat(sect, 'pseudo-dt-max-mult',
-                                                 3.0)
+        
+        dtau_minf = self.cfg.getfloat(sect, 'pseudo-dt-min-mult', 1.0)
+        dtau_maxf = self.cfg.getfloat(sect, 'pseudo-dt-max-mult', 3.0)
+
+        self.dtau_min = dtau_minf * self.dtau
```
FreddieWitherden, line 118: No space around *

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -106,18 +111,18 @@ def __init__(self, *args, **kwargs):
...
         if not tplargs['minf'] < 1 <= tplargs['maxf']:
             raise ValueError('Invalid pseudo max-fact, min-fact')
 
-        if tplargs['dtau_maxf'] < 1:
-            raise ValueError('Invalid pseudo-dt-max-mult')
-
-        # Limits for the local pseudo-time-step size
-        tplargs['dtau_min'] = self._dtau
-        tplargs['dtau_max'] = tplargs['dtau_maxf'] * self._dtau
+        if not dtau_minf <= 1 < dtau_maxf:
```
FreddieWitherden, line 124: Rewrite to avoid the not (I think it is possible)

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -140,8 +145,27 @@ def __init__(self, *args, **kwargs):
                     )
                 )
 
+        self.pseudo_step_multiplied(1.0)
+
         self.backend.commit()
 
+    def adjust_pseudo_step(self, dt):
+        old_dtau = self.dtau
+        super().adjust_pseudo_step(dt)
```
WillTrojak, line 154: This is maybe not the most straightforward way of implementing this. It adds spaghetti that will make it hard in the future to refactor. Consider inlining the `super().adjust_pseudo_step()`?

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -140,8 +145,27 @@ def __init__(self, *args, **kwargs):
...
+        self.pseudo_step_multiplied(self.dtau / old_dtau)
+
+    def pseudo_step_multiplied(self, y):
```
WillTrojak, line 157: Maybe make this a function that purely does the binding of dtau_min, dtau_max and dtau_fieldf and move the `dtau_min *= y` and `self.dtau_max *= y`. This should make the calls where you do `pseudo_step_multiplied` more understandable.

pyfr/integrators/base.py
```diff
@@ -58,6 +58,29 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
         # Abort computation
         self.abort = False
 
+        # Smoothly step to target time in the last near_t steps
+        self.fact = self.cfg.getfloat('solver-time-integrator', 'dt-fact', 0.9)
+        self.fact_max = 1.001
+        self._dt_in = cfg.getfloat('solver-time-integrator', 'dt')
+        self._dt_near = None
+
+    def adjust_step(self, t):
```
WillTrojak, line 67: I think this needs to be renamed. Maybe to something like `adjust_dt`.

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -140,8 +145,27 @@ def __init__(self, *args, **kwargs):
                     )
                 )
 
+        self.pseudo_step_multiplied(1.0)
+
         self.backend.commit()
 
+    def adjust_pseudo_step(self, dt):
```
WillTrojak, line 152: Maybe change the name of the name of this, similar to `adjust_step`. I think the name isn't very descriptive of what is actually happening

setup.py
```diff
@@ -1,6 +1,12 @@
 #!/usr/bin/env python
 import re
 from setuptools import setup
+import sys
+
+
+# Check Python version
```
WillTrojak, line 7: This should be a separate PR

WillTrojak (changes requested):
There is some cleanup that is needed around dt and dtau. At the moment, you use `self._dt` and `self.dtau` and also sometimes you pass dtua and dt and other times you use self. I think you need to be more consistent where possible.

WillTrojak:
Also can you rebase off PyFR/develop?

pyfr/integrators/base.py
```diff
@@ -58,6 +58,29 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
...
+        t_diff = t - self.tcurr
+        steps_with_dt_far = t_diff / self._dt_in
+        steps_to_t = -(steps_with_dt_far//-self.fact_max)
+
+        if steps_to_t == 1:
+            self._dt = t_diff
+        elif steps_with_dt_far == 0:
+            self._dt_near = None
```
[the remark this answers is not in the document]
sambitmishra98: The default behaviour is using `self._dt_in` which is set in the config file (or in the future by the adaptive DIRK stepper).
From one `t` to the next, `self._dt_near` is set only once, i.e. when we are close to the target time. As soon as we cross the subsequent target `t`, we set it to `None`.

pyfr/integrators/dual/pseudo/base.py
```diff
@@ -18,8 +18,7 @@ def __init__(self, backend, systemcls, rallocs, mesh,
 
         sect = 'solver-time-integrator'
 
-        self._dtaumin = 1.0e-12
-        self._dtau = cfg.getfloat(sect, 'pseudo-dt')
+        self.dtau = cfg.getfloat(sect, 'pseudo-dt')
```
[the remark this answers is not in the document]
sambitmishra98: This was a result of mixing adaptive physical time stepping with this PR. In my experience, `self._dtaumin` is redundant. I have never found a case where the Δτ-field reached that low.

pyfr/integrators/dual/pseudo/multip.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, backend, systemcls, rallocs, mesh, initsoln, cfg,
         self.npmgcycles = 0
 
         # Multigrid pseudo-time steps
-        dtau = cfg.getfloat(sect, 'pseudo-dt')
+        self.dtau = cfg.getfloat(sect, 'pseudo-dt')
```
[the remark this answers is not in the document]
sambitmishra98: We are not running the `__init__()` from the parent class right? This would have been redundant if we had `super().__init__(*args, **kwargs)`.

pyfr/integrators/base.py
```diff
@@ -58,6 +58,29 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
         # Abort computation
         self.abort = False
 
+        # Smoothly step to target time in the last near_t steps
+        self.fact = self.cfg.getfloat('solver-time-integrator', 'dt-fact', 0.9)
+        self.fact_max = 1.001
+        self.dt_in = cfg.getfloat('solver-time-integrator', 'dt')
+        self.dt_near = None
+
+    def adjust_dt(self, t):
```
WillTrojak, line 67: If I'm not mistaken, can this function not be replaced with
```python
def adjust_dt(self, t):
    if (tdiff := t - self.tcurr) < self.dt*self*fact_max:
        self.dt = tdiff
```
or is there some edge case I've missed?

pyfr/integrators/dual/pseudo/base.py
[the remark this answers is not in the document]
WillTrojak: I've had situations where I've hit dtaumin.

pyfr/integrators/base.py
```diff
@@ -58,6 +58,29 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
         # Abort computation
         self.abort = False
 
+        # Smoothly step to target time in the last near_t steps
+        self.fact = self.cfg.getfloat('solver-time-integrator', 'dt-fact', 0.9)
+        self.fact_max = 1.001
```
WillTrojak, line 63: Shouldn't this be configurable?

WillTrojak (changes requested):

pyfr/integrators/base.py
sambitmishra98: I will make it configurable. 
I only added it since a value slightly higher than the set `dt` took 2 steps instead of 1 slightly larger step. For stiffer problems we will use a conservative `self.fact_max = 1.000001`.

pyfr/integrators/base.py
sambitmishra98: The logic you suggested does not consider a gradual change of `self.dt`. When we are a few time steps away from the target `dt`, the third `if` condition sets a `dt` value we would take as we approach to the target `dt`. Also, I have used a `self.dt_near` variable to decrease the number of times we change `dt`, and set it to `None` when we are done with using the `dt_near` for our target `t`.


Test it here:

https://colab.research.google.com/drive/1upThO_hbxr25ThikrYOAJa-IHNGjq-pD?usp=sharing

sambitmishra98:
Anything else we need to do here? @FreddieWitherden @semi-h @WillTrojak

pyfr/integrators/dual/pseudo/pseudocontrollers.py
```diff
@@ -140,8 +146,30 @@ def __init__(self, *args, **kwargs):
...
+        self.bind_dtau()
+
         self.backend.commit()
 
+    def adjust_dtau(self, dt):
+        old_dtau = self.dtau
+        self.dtau = dt / self._dt_dtau_ratio
+        self.multiply_limits(self.dtau / old_dtau)
+
+    def multiply_limits(self, y):
```
FreddieWitherden, line 158: Why is this its own function?

sambitmishra98: As a result of what @WillTrojak said in 
https://github.com/PyFR/PyFR/pull/335#discussion_r1321524509

FreddieWitherden: At the moment the function is only called once and so adds little to no value.

pyfr/integrators/base.py
```diff
@@ -18,6 +18,7 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
         self.rallocs = rallocs
         self.isrestart = initsoln is not None
         self.cfg = cfg
+        cfgsect = 'solver-time-integrator'
```
FreddieWitherden, line 21: Do we need to declare this?

pyfr/integrators/std/controllers.py
```diff
@@ -175,7 +175,7 @@ def advance_to(self, t):
 
         while self.tcurr < t:
             # Decide on the time step
-            dt = max(min(t - self.tcurr, self._dt, self.dtmax), self.dtmin)
+            dt = max(min(t - self.tcurr, self.dt, self.dtmax), self.dtmin)
```
FreddieWitherden, line 178: I thought the new logic was supposed to take care of this?

pyfr/integrators/base.py
sambitmishra98: Line 64 otherwise exceeds the line limit.

FreddieWitherden: You can just line break it, if it is only one line causing an issue.

pyfr/integrators/std/controllers.py
FreddieWitherden: Just following up here.

sambitmishra98: We will probably need to refactor the code a bit to remove that line. I will do that and test the edge cases by tomorrow.

sambitmishra98: I just realised, we are already using `self.fact` and `self.fact_max` to restrict the allowed change in `dt`. We could remove `self.dtmin` everywhere. Now that we are hitting the target times exactly, is there any reason to keep `self.dtmin`?

FreddieWitherden: How would we prevent dt from going below a certain amount?

sambitmishra98: I added a line to prevent `dt` from going below `self.dtmin`. 

`self.dt` is modified with line 78-80 to make sure that we do not change dt suddenly. The time that we will come across this condition is when `self.dt` is close to `self.dtmin` because the controller requires it. I have not come across this though, so I assumed it is not required.

FreddieWitherden: This literally happens all of the time: https://pyfr.discourse.group/search?expanded=true&q=Minimum%20sized%20time%20step%20rejected

sambitmishra98:
Any reason to keep `dt-max`?

pyfr/integrators/std/controllers.py
sambitmishra98: You're right. I replicated the issue in https://pyfr.discourse.group/t/minimum-sized-time-step-rejected-2d-pipe-flow/720 and could not think of a way to go around setting `self.dtmin`.

sambitmishra98:
What is left in this PR?

FreddieWitherden:
Can someone confirm the status of this PR?

sambitmishra98:
@WillTrojak @semi-h @FreddieWitherden Can we merge this branch? I have been using this functionality for about a year without issues.

FreddieWitherden:
I'll aim to get it merged for 2.1.  Just want to wait for a couple of weeks to be sure that we will not need another 2.0.x point release.

FreddieWitherden:
@semi-h @WillTrojak Can you guys give this a final once over before I merge?

pyfr/integrators/base.py
```diff
@@ -73,6 +73,31 @@ def __init__(self, backend, rallocs, mesh, initsoln, cfg):
...
+            self.dt = t_diff
+        elif fallback_steps == 0:
+            self.dt_near = None
+            self.dt = t_diff
+        elif (fallback_steps - 1) / (steps_to_t - 1) < self.fact_min:
+            self.dt_near = self.dt_near or t_diff / steps_to_t
+            self.dt = self.dt_near
+        else:
+            self.dt = self.dt_fallback
+        self.dt = max(self.dt, self.dtmin)
```
FreddieWitherden, line 99: Move this into the various blocks (and only when needed).

sambitmishra98: I think this is needed only when a controller is enabled. We ensure a large enough step is taken with no controllers enabled. I am moving it into the PI controller.

sambitmishra98:
> At the current stage I think the `dt` is rarely changed, only when we're near a checkpoint, but later on when we introduce adaptive physical time stepping it might get triggered quite often and it would be nice to avoid this if we can.

@semi-h We rework the pseudo-time field every pseudo-time step if we enable the local-PI controller. So irrespective of whether we have a physical time controller or not, the local-PI controller determines modifying pseudo-time field. I see `bind_dtau()` would get triggered every physical step with adaptive physical time stepping, which I assume isn't too expensive. 

> Can you also give a bit more insight about why you've decided to start gradually change dt if there is a checkpoint within next 3*dt? 

I vaguely remember @FreddieWitherden mentioned that the std PI controller struggles a little if we are forced to take a very small time-step very close to a target time, say one of the file writes. This gradually change allows us to take a decently large time step before and after a target time (see https://github.com/PyFR/PyFR/pull/335#issuecomment-1634797022).

sambitmishra98:
Moving this PR to #495

PR #341, opened 2023-03-22
Improved iterative solver for entropy filter.

tarikdzanic:


pyfr/solvers/euler/kernels/entropyfilter.mako
```diff
@@ -144,7 +144,10 @@
                     f3 = (e_high > 0.0) ? f_high : (0.5*f_low*e_high - f_high*e_low)/(0.5*e_high - e_low + ${ill_tol});
 
                     // Compute guess as minima of individual constraints
-                    fnew = fmin(f1, fmin(f2, f3));
+                    fnew = fmin(f1, f2);
+
+                    // Avoid using entropy constraint to guess new bracket if entropy is not well-defined
+                    fnew = (max(e_low, e_high) < ${fpdtype_max + e_tol} - entmin) ? fmin(f3, fnew) : fnew;
```
FreddieWitherden, line 150: I'm not sure this line is well defined.  For double precision adding anything to `fpdtype_max` will give `inf` which likely will not compile (unknown symbol).  For single precision it will give a constant which evaluates to `inf`.  However, all of the backends promise that this will not happen and so the compiler is free to go crazy with our code.

tarikdzanic: `(max(e_low, e_high) < ${0.9*fpdtype_max})` ? Somewhat arbitrary but if the entropy calculation is even remotely close to the max value it's not well-defined

tarikdzanic: Pushed the change. I don't think there's any scenario where the entropy will be between `0.9*fpdtype_max  - fpdtype_max ` and still be well defined

FreddieWitherden: Can `e_low` be greater than `e_high`?  Just wondering if the max is needed or not.

FreddieWitherden: Okay. Finally, stick with either fmin/fmax or min/max.  Not sure if PyFR has a preference (we probably should).

tarikdzanic: Updated

PR #344, opened 2023-04-12
Track time spent in plugins in the stats record

FreddieWitherden:


FreddieWitherden:
The tracking works reasonably well.  However, one caveat is that when running on GPU backends without MPI it is possible for kernel submission to get well ahead of execution.  This results in a large amount of time being accounted to the device-to-host transfers since they block on kernel completion.  This time, in turn, is booked to whatever plugin accesses the solution first.

For real simulations, which use multiple ranks, this run-ahead is not possible to the same degree.  We can think about trying to factor this out.

PR #347, opened 2023-04-19
Fix an issue with benchmarking in the Metal backend

FreddieWitherden:


WillTrojak:
As discussed this, fixed my issue with the adaptive time stepping. I'm happy if you merge

PR #349, opened 2023-04-19
Updated point sampler to use Rtree

pv101:


pyfr/plugins/sampler.py
```diff
@@ -5,51 +5,36 @@
...
-        dmins = [d[a] for d, a in zip(dists, amins)]
-
-        # Find the minimum across all element types
-        yield min(zip(dmins, range(len(epts)), amins))
-
-
-def _closest_pts_kd(epts, pts):
-    from scipy.spatial import cKDTree
+def _closest_pts(epts, pts):
+    from rtree import index
```
FreddieWitherden, line 9: Can import at outer level

pyfr/plugins/sampler.py
```diff
@@ -5,51 +5,36 @@
...
 
-    # Flatten the physical location arrays
     fepts = [e.reshape(-1, e.shape[-1]) for e in epts]
```
FreddieWitherden, line 11: Probably not worth constructing this explicitly if we are going to use a for loop below.  Might as well just do `for e in epts:`.

pyfr/plugins/sampler.py
```diff
@@ -5,51 +5,36 @@
...
+    idx = []
 
-    for p in pts:
-        # Query the distance/index of the closest upt to p
-        dmins, amins = zip(*[t.query(p) for t in trees])
+    for f in fepts:
+        pr = index.Property()
+        pr.dimension = f.shape[1]
+        pr.interleaved = True
+        insl = []
```
FreddieWitherden, line 19: Should be able to simplify this quite a bit.

pyfr/plugins/sampler.py
```diff
@@ -5,51 +5,36 @@
...
-    # For each element type construct a KD-tree of the upt locations
-    trees = [cKDTree(f) for f in fepts]
+    idx = []
 
-    for p in pts:
-        # Query the distance/index of the closest upt to p
-        dmins, amins = zip(*[t.query(p) for t in trees])
+    for f in fepts:
+        pr = index.Property()
+        pr.dimension = f.shape[1]
```
FreddieWitherden, line 17: I think you can pass this as keyword arguments to Property.

pyfr/plugins/sampler.py
FreddieWitherden: `from rtree.index import Index, Property` might be cleaner, too.

pyfr/plugins/sampler.py
FreddieWitherden: Also, may just be able to pass this property directly to `Index` as an argument without assigning it to a variable.

pyfr/plugins/sampler.py
```diff
@@ -5,51 +5,36 @@
...
+            ins = tuple(np.concatenate([f[i,:], f[i,:]]))
+            insl.append((i,ins,None))
+        idx.append(index.Index(insl,properties=pr))
 
-        # Unravel the indices
+    for p in pts:
+        dmins = []
+        amins = []
+        for i, v in enumerate(idx):
+            amin = list(v.nearest(np.concatenate([p, p]), 1))[0]
```
FreddieWitherden, line 29: Why is list needed here?  If it is an iterator `next(...)` should get the first item in the sequence.  Also see if you can one-line amins and do dmins as a separate pass

pyfr/plugins/sampler.py
```diff
@@ -5,51 +5,36 @@
...
+            amins.append(amin)
+            dmins.append(np.linalg.norm(fepts[i][amin,:] - p))
         amins = [np.unravel_index(i, e.shape[:2])
-                 for i, e in zip(amins, epts)]
+                 for i, e in zip(amins, epts)]    
 
-        # Reduce across element types
```
FreddieWitherden, line 40: Comment is probably useful.

pyfr/plugins/sampler.py
```diff
@@ -5,51 +5,36 @@
...
 
-    for p in pts:
-        # Query the distance/index of the closest upt to p
-        dmins, amins = zip(*[t.query(p) for t in trees])
+    for f in fepts:
+        pr = index.Property()
+        pr.dimension = f.shape[1]
+        pr.interleaved = True
+        insl = []
+        for i in range(f.shape[0]):
```
FreddieWitherden, line 20: `for p in f` avoids the need for explicit indexing

pyfr/plugins/sampler.py
```diff
@@ -5,51 +5,36 @@
...
+            ins = tuple(np.concatenate([f[i,:], f[i,:]]))
```
FreddieWitherden, line 21: `(*p, *p)` should do it

pyfr/plugins/sampler.py
```diff
@@ -1,55 +1,33 @@
...
+    ndims = len(pts[0])
+    props = Property(dimension=ndims, interleaved=True)
+    trees = []
 
-    # For each element type construct a KD-tree of the upt locations
-    trees = [cKDTree(f) for f in fepts]
+    for e in epts:
+        # Build list of solution points for each element type
+        ins = [(i, [*p, *p], None) for i, p in enumerate(e.reshape(-1, ndims))]
+        # Buld tree of solution points for each element type
```
FreddieWitherden, line 17: Add newline

pyfr/plugins/sampler.py
```diff
@@ -1,55 +1,33 @@
...
 
     for p in pts:
-        # Query the distance/index of the closest upt to p
-        dmins, amins = zip(*[t.query(p) for t in trees])
-
-        # Unravel the indices
-        amins = [np.unravel_index(i, e.shape[:2])
-                 for i, e in zip(amins, epts)]
+        #Find index of solution point closest to p
+        amins = [np.unravel_index(next(t.nearest([*p, *p], 1)), ept.shape[:2])
```
FreddieWitherden, line 22: Check line length.

pyfr/plugins/sampler.py
```diff
@@ -1,55 +1,33 @@
...
+                 for ept, t in zip(epts, trees)]
+        #Find distance of solution point closest to p
```
FreddieWitherden, line 24: Add newline and space after #

pyfr/plugins/sampler.py
```diff
@@ -1,55 +1,33 @@
...
-        amins = [np.unravel_index(i, e.shape[:2])
-                 for i, e in zip(amins, epts)]
+        #Find index of solution point closest to p
+        amins = [np.unravel_index(next(t.nearest([*p, *p], 1)), ept.shape[:2])
+                 for ept, t in zip(epts, trees)]
+        #Find distance of solution point closest to p     
+        dmins = [np.linalg.norm(e[a] - p) for e, a in zip(epts, amins)]
 
-        # Reduce across element types
+        # Reduce across element types
```
FreddieWitherden, line 27: Remove trailing whitespace.

PR #354, opened 2023-05-10
Added optional lp norm in integrate plugin.

WillTrojak:
Adds an option to the integrate plugin for setting a degree for lp norm allowing the integrate plugin to calculate lp norms.
This is particualrly useful when analysing methods where L1, L2 and Linf norms of error are widely used. Previously you could do L1, and dodge L2 doing the sqrt in post, but there was no way to do Linf.

If the option is not present then the native approach is used where the expressions are just integrated.

### DIFF
--- a/doc/src/user_guide.rst
+++ b/doc/src/user_guide.rst
@@ -1559,17 +1559,23 @@ Integrate quantities over the compuational domain. Parameterised with:
 
     *int*
 
+
 5. ``quad-pts-{etype}`` --- name of quadrature rule (optional):
 
     *string*
 
-6. ``region`` --- region to integrate, specified as either the
+6. ``norm`` --- sets the degree and calculates an :math:`L_p` norm (optional),
+    otherwise standard integration is performed:
+
+    *float* | ``inf`` | ``none``
+
+7. ``region`` --- region to integrate, specified as either the
    entire domain using ``*`` or a combination of the geometric shapes
    specified in :ref:`regions`:
 
     ``*`` | ``shape(args, ...)``
 
-7. ``int``-*name* --- expression to integrate, written as a function of
+8. ``int``-*name* --- expression to integrate, written as a function of
    the primitive variables and gradients thereof, the physical coordinates
    [x, y, [z]] and/or the physical time [t]; multiple expressions,
    each with their own *name*, may be specified:
--- a/pyfr/plugins/integrate.py
+++ b/pyfr/plugins/integrate.py
@@ -31,6 +31,24 @@ class IntegratePlugin(BaseSolnPlugin):
         self.exprs = [self.cfg.getexpr(cfgsect, k, subs=c)
                       for k in self.cfg.items(cfgsect, prefix='int-')]
 
+        # Add optional integral L-p degree
+        if self.cfg.get(cfgsect, 'norm', 'none').lower() != 'none':
+            self.lp = self.cfg.getfloat(cfgsect, 'norm')
+
+            # Modify expressions for L-p norm
+            if self.lp == 1 or self.lp == float('inf'):
+                self.exprs = [f'abs({e})' for e in self.exprs]
+                self._post_func = lambda x: x
+            else:
+                self.exprs = [f'pow(abs({e}), {self.lp})' for e in self.exprs]
+                self._post_func = lambda x: x**(1 / self.lp)
+        else:
+            self.lp = None
+            self._post_func = lambda x: x
+
+        # Set MPI reduction op
+        self.mpi_op = mpi.MAX if self.lp == float('inf') else mpi.SUM
+
         # Integration region pre-processing
         esetmask = self._prepare_esetmask(intg)
 
@@ -173,10 +191,16 @@ class IntegratePlugin(BaseSolnPlugin):
 
             for j, v in enumerate(self.exprs):
                 # Evaluate the expression at each point
-                iex = wts*npeval(v, subs)
+                e = npeval(v, subs)
 
-                # Accumulate
-                intvals[j] += np.sum(iex) - np.sum(iex[emask])
+                # Either max or quadrature and sum
+                if self.lp == float('inf'):
+                    intvals[j] = max(intvals[j], np.amax(e))
+                else:
+                    iex = wts*e
+
+                    # Accumulate
+                    intvals[j] += np.sum(iex) - np.sum(iex[emask])
 
         return intvals
 
@@ -190,9 +214,12 @@ class IntegratePlugin(BaseSolnPlugin):
 
             # Reduce and output if we're the root rank
             if rank != root:
-                comm.Reduce(iintex, None, op=mpi.SUM, root=root)
+                comm.Reduce(iintex, None, op=self.mpi_op, root=root)
             else:
-                comm.Reduce(mpi.IN_PLACE, iintex, op=mpi.SUM, root=root)
+                comm.Reduce(mpi.IN_PLACE, iintex, op=self.mpi_op, root=root)
+
+                # Apply post integration function
+                iintex = (self._post_func(i) for i in iintex)
 
                 # Write
                 print(intg.tcurr, *iintex, sep=',', file=self.outf)


pyfr/plugins/integrate.py
```diff
@@ -173,10 +199,16 @@ def _eval_exprs(self, intg):
...
-                iex = wts*npeval(v, subs)
+                e = npeval(v, subs)
 
-                # Accumulate
-                intvals[j] += np.sum(iex) - np.sum(iex[emask])
+                # Either max or quadrature and sum
+                if self.lp == inf:
+                    intvals[j] = max(intvals[j], np.amax(e[eset]))
+                else:
+                    iex = wts*npeval(v, subs)
```
WillTrojak, line 208: npeval should be e

pyfr/plugins/integrate.py
```diff
@@ -31,6 +32,31 @@ def __init__(self, intg, cfgsect, suffix=None):
         self.exprs = [self.cfg.getexpr(cfgsect, k, subs=c)
                       for k in self.cfg.items(cfgsect, prefix='int-')]
 
+        # Add optional integral L-p degree
+        if self.cfg.hasopt(cfgsect, 'p-deg'):
```
FreddieWitherden, line 36: No need for this, just use `getfloat` with a third argument?

pyfr/plugins/integrate.py
```diff
@@ -31,6 +32,31 @@ def __init__(self, intg, cfgsect, suffix=None):
...
+            self.lp = self.cfg.getfloat(cfgsect, 'p-deg')
+        else:
+            self.lp = 'native'
+
+        # Set post integration function default
+        self._post_func = lambda x : x
```
FreddieWitherden, line 42: Can set this in an else block further down? No space before : for lamdba.

pyfr/plugins/integrate.py
```diff
@@ -31,6 +32,31 @@ def __init__(self, intg, cfgsect, suffix=None):
...
+
+        # Modify expressions for L-p norm
+        if self.lp == 1 or self.lp == inf:
+            self.exprs = [f'(abs({e}))' for e in self.exprs]
+        elif self.lp == 2:
+            self.exprs = [f'({e}*{e})' for e in self.exprs]
```
FreddieWitherden, line 48: **2 to avoid evaluating the expression twice.

pyfr/plugins/integrate.py
```diff
@@ -31,6 +32,31 @@ def __init__(self, intg, cfgsect, suffix=None):
...
+            self._post_func = lambda x : sqrt(x)
```
FreddieWitherden, line 49: **0.5, avoids importing sqrt.

pyfr/plugins/integrate.py
FreddieWitherden, line 48: Use pow here to avoid revaluating the expression.

pyfr/plugins/integrate.py
```diff
@@ -31,6 +32,31 @@ def __init__(self, intg, cfgsect, suffix=None):
...
+        elif self.lp != 'native':
+            self.exprs = [f'(pow(abs({e}), {self.lp}))' for e in self.exprs]
+            self._post_func = lambda x : pow(x, 1/self.lp)
```
FreddieWitherden, line 52: x**1/self.lp to avoid importing pow (which I think we've forgotten to do?)

pyfr/plugins/integrate.py
```diff
@@ -31,6 +32,31 @@ def __init__(self, intg, cfgsect, suffix=None):
...
+
+        # Set MPI reduction op
+        if self.lp == inf:
+            self.mpi_op = mpi.MAX
```
FreddieWitherden, line 56: Single line if is probably simpler here.

pyfr/plugins/integrate.py
```diff
@@ -190,9 +222,12 @@ def __call__(self, intg):
...
             # Reduce and output if we're the root rank
             if rank != root:
-                comm.Reduce(iintex, None, op=mpi.SUM, root=root)
+                comm.Reduce(iintex, None, op=self.mpi_op, root=root)
             else:
-                comm.Reduce(mpi.IN_PLACE, iintex, op=mpi.SUM, root=root)
+                comm.Reduce(mpi.IN_PLACE, iintex, op=self.mpi_op, root=root)
+
+                # Apply post integration function
+                iintex = [self._post_func(i) for i in iintex]
```
FreddieWitherden, line 230: [] => () so it can be a generator.

pyfr/plugins/integrate.py
```diff
@@ -31,6 +32,31 @@ def __init__(self, intg, cfgsect, suffix=None):
...
+        if self.cfg.hasopt(cfgsect, 'p-deg'):
+            self.lp = self.cfg.getfloat(cfgsect, 'p-deg')
+        else:
+            self.lp = 'native'
+
+        # Set post integration function default
+        self._post_func = lambda x : x
+
+        # Modify expressions for L-p norm
+        if self.lp == 1 or self.lp == inf:
```
FreddieWitherden, line 45: No need to import inf.  Just do float('inf').

FreddieWitherden:
An alternative worth considering is a functional approach:
```
int-E = norm(..., 2)
int-vv = norm(..., inf)
```
should be easy enough to parse, although more care is needed in the inf case for MPI in the case of mixed norms.

WillTrojak:
I think for now the current approach is fine and maybe later we can implement this operator.

FreddieWitherden:
> I think for now the current approach is fine and maybe later we can implement this operator.

We probably want to get it right first time to avoid breaking existing code.

FreddieWitherden:
Did we decide what syntax we wanted to go with? @tarikdzanic thoughts?

tarikdzanic:
I don't see any issues with the current implementation, seems like it works for both the usual and general cases.

WillTrojak:
Did you have any more points to add on the @FreddieWitherden? I've been using it in anger recently and it seems good to go.

FreddieWitherden:
Documentation.

FreddieWitherden:
Are you okay to add the relevant documentation?

FreddieWitherden:
@WillTrojak Just following up here.

doc/src/user_guide.rst
```diff
@@ -1559,17 +1559,23 @@ Integrate quantities over the compuational domain. Parameterised with:
 
     *int*
 
+
 5. ``quad-pts-{etype}`` --- name of quadrature rule (optional):
 
     *string*
 
-6. ``region`` --- region to integrate, specified as either the
+6. ``p-deg`` --- sets the degree and calculates an Lp norm (optional),
```
FreddieWitherden, line 1567: $L_p$?

doc/src/user_guide.rst
```diff
@@ -1559,17 +1559,23 @@ Integrate quantities over the compuational domain. Parameterised with:
...
+    otherwise standard integration is performed:
+
+    *float* | ``inf``
```
FreddieWitherden, line 1570: Can we also allow ``none`` here?

doc/src/user_guide.rst
```diff
@@ -1559,17 +1559,23 @@ Integrate quantities over the compuational domain. Parameterised with:
 
     *int*
 
+
```
FreddieWitherden, line 1562: ?

doc/src/user_guide.rst
FreddieWitherden, line 1567: Call it `norm` as the parameter name?

pyfr/plugins/integrate.py
```diff
@@ -31,6 +31,22 @@ def __init__(self, intg, cfgsect, suffix=None):
         self.exprs = [self.cfg.getexpr(cfgsect, k, subs=c)
                       for k in self.cfg.items(cfgsect, prefix='int-')]
 
+        # Add optional integral L-p degree
+        self.lp = self.cfg.getfloat(cfgsect, 'p-deg', -1)
+
+        # Modify expressions for L-p norm
+        if self.lp == 1 or self.lp == float('inf'):
+            self.exprs = [f'(abs({e}))' for e in self.exprs]
```
FreddieWitherden, line 39: Are the () needed?

pyfr/plugins/integrate.py
```diff
@@ -173,10 +190,16 @@ def _eval_exprs(self, intg):
...
             for j, v in enumerate(self.exprs):
                 # Evaluate the expression at each point
-                iex = wts*npeval(v, subs)
+                e = npeval(v, subs)
 
-                # Accumulate
-                intvals[j] += np.sum(iex) - np.sum(iex[emask])
+                # Either max or quadrature and sum
+                if self.lp == float('inf'):
+                    intvals[j] = max(intvals[j], np.amax(e))
```
FreddieWitherden, line 197: e.max() as it is a NumPy array.

doc/src/user_guide.rst
```diff
@@ -1559,17 +1559,23 @@ Integrate quantities over the compuational domain. Parameterised with:
...
 5. ``quad-pts-{etype}`` --- name of quadrature rule (optional):
 
     *string*
 
-6. ``region`` --- region to integrate, specified as either the
+6. ``norm`` --- sets the degree and calculates an $L_p$ norm (optional),
```
FreddieWitherden, line 1567: Not sure this is the right math notation for RST.

pyfr/plugins/integrate.py
```diff
@@ -31,6 +31,23 @@ def __init__(self, intg, cfgsect, suffix=None):
         self.exprs = [self.cfg.getexpr(cfgsect, k, subs=c)
                       for k in self.cfg.items(cfgsect, prefix='int-')]
 
+        # Add optional integral L-p degree
+        if self.cfg.get(cfgsect, 'norm', 'none').lower() != 'none':
+            self.lp = self.cfg.getfloat(cfgsect, 'norm')
+            # Modify expressions for L-p norm
```
FreddieWitherden, line 37: Add newline.

PR #360, opened 2023-08-04
Allow d/dt of solution to be accessed

WillTrojak:


pyfr/integrators/base.py
```diff
@@ -129,6 +125,11 @@ def call_plugin_dt(self, dt):
             if not tlist or t - tlist[-1] > self.dtmin:
                 tlist.append(t)
 
+    def invalidate_caches(self):
```
FreddieWitherden, line 128: Can you make this private with an _?

PR #361, opened 2023-08-07
Add a more robust progress bar architecture.

FreddieWitherden:


pyfr/__main__.py
```diff
@@ -53,8 +56,10 @@ def main():
                               help='input solution files')
     ap_partition.add_argument('outd', help='output directory')
     partitioners = sorted(cls.name for cls in subclasses(BasePartitioner))
-    ap_partition.add_argument('-p', dest='partitioner', choices=partitioners,
+    ap_partition.add_argument('-l', dest='partitioner', choices=partitioners,
```
WillTrojak, line 59: It might be worth adding examples of the `-l` and `-p` flags to the running pyfr examples in the docs. Or to the CLI docs that we now have.

pyfr/__main__.py
```diff
@@ -187,33 +199,41 @@ def process_partition(args):
...
 
-        for etype, emap in sorted(rnum.items()):
-            for k, v in sorted(emap.items()):
-                print(etype, *k, *v, sep=',', file=args.rnumf)
+                pb.advance_by(1)
 
 
 def process_export(args):
+    # Create a suitable progress tracker
+    progress = ProgressSequence() if args.progress else NullProgressSequence()
```
WillTrojak, line 235: Is it worth making `args.progress` an argument to the ProgressSequence class? It could simplify these if statements

pyfr/progress_bar.py
```diff
@@ -1,20 +1,25 @@
...
+import itertools as it
 import shutil
 import sys
 import time
 
 
-def to_hms(delta):
-    hours, remainder = divmod(int(delta), 3600)
-    minutes, seconds = divmod(remainder, 60)
+def format_hms(delta):
```
WillTrojak, line 8: Might it be a bit simpler to use the built-in datetime package?

WillTrojak (changes requested):
Some minor comments, however, when I test it on the cylinder case and restart, I get a new line after the progress bar is first written. Not quite sure where this is coming from.

pyfr/__main__.py
FreddieWitherden: Including `-p` I think is a good call.  `-l` probably just for the docs as otherwise it will confuse our users (as the default tries all available libraries). @pv101 Are you okay to action?

pyfr/progress_bar.py
FreddieWitherden: The equivalent class is `datetime.timedelta` but you are on your own when it comes to formatting.

pyfr/__main__.py
FreddieWitherden: This should now be resolved.

FreddieWitherden:
I've cleaned up the code quite a bit by making the progress argument generic.

WillTrojak:
I haven't been able to work out what this issue I saw was and I haven't been able to reliable reproduce it. So I think it might be a problem on my side somewhere. Happy to merge

pyfr/__main__.py
```diff
@@ -255,11 +258,11 @@ def _process_common(args, mesh, soln, cfg):
 
     # If we are running interactively then create a progress bar
     if args.progress and MPI.COMM_WORLD.rank == 0:
-        pb = ProgressBar(solver.tstart, solver.tcurr, solver.tend)
+        pbar = ProgressBar()
+        pbar.start(solver.tend, start=solver.tstart, curr=solver.tcurr
```
WillTrojak, line 262: type forgot )

PR #363, opened 2023-08-15
Multip dtau factors

sambitmishra98:
Expose pseudo-dt fact across multi-p levels. This will be used in future for optimisation.

### DIFF
--- a/pyfr/integrators/dual/pseudo/multip.py
+++ b/pyfr/integrators/dual/pseudo/multip.py
@@ -2,6 +2,8 @@ from collections import defaultdict
 import itertools as it
 import re
 
+import math
+
 from pyfr.inifile import Inifile
 from pyfr.integrators.dual.pseudo.base import BaseDualPseudoIntegrator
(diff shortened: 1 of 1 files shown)


pyfr/integrators/dual/pseudo/multip.py
```diff
@@ -185,13 +206,13 @@ def mgproject(self, l1, l1reg, l2, l2reg):
         return projk
 
     @memoize
-    def dtauproject(self, l1, l2):
+    def dtauproject(self, l1, l2, dtaufs):
```
FreddieWitherden, line 209: This is likely expensive.

FreddieWitherden:
I still think that the overhead from this approach is too high.  Creating a new matrix multiplication kernel can take well over a minute at p = 6.

I think we instead need a separate kernel which applies the scaling.

sambitmishra98:
I will stop work on this, and look into having a local-PI controller for every level.

PR #365, opened 2023-09-08
Added metadata to ascent PNG

WillTrojak:
This adds the saving of the things to the PNG metadata of images produced by Ascent. Currently, the config, mesh-uuid, and time are saved. There is potential for adding additional syntax to the config to allow user notes, but I'm open to ideas. 

This is reliant on https://github.com/Alpine-DAV/ascent/pull/1199/

### DIFF
--- a/pyfr/plugins/ascent.py
+++ b/pyfr/plugins/ascent.py
@@ -195,6 +195,11 @@ class AscentPlugin(BaseSolnPlugin):
         self._ele_regions_lin.append((d_str, idx, rgn, soln_op))
 
         mesh_n[f'{d_str}/state/domain_id'] = intg.rallocs.prank
+        mesh_n[f'{d_str}/state/config/keyword'] = 'Config'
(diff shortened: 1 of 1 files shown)


WillTrojak:
Also, this metadata can be easily accessed with a package like Pillow, for example:

```python
from PIL import Image

filename = "foo.png"
img = Image.open(filename)
print(img.text)
```
or on the command line via `identify -verbose foo.png` and look under the properties tag.

WillTrojak:
The feature this is dependent on is now merged into Ascent, so unless @FreddieWitherden you had any comments, I think we're good to merge.

pyfr/plugins/ascent.py
```diff
@@ -219,6 +224,9 @@ def _eval_exprs(self, intg):
 
         # Iterate over each element type in our region
         for d_str, idx, rgn, soln_op in self._ele_regions_lin:
+            self.mesh_n[f'{d_str}/state/time/keyword'] = "Time"
```
FreddieWitherden, line 227: Single quotes.

PR #367, opened 2023-09-20
Region cleanups.

FreddieWitherden:
@WillTrojak Let me know if this works for your FWH plugin.  It is now much easier to get the region object.

### DIFF
--- a/pyfr/plugins/ascent.py
+++ b/pyfr/plugins/ascent.py
@@ -1,5 +1,5 @@
 from collections import defaultdict
-from ctypes import c_char_p, c_double, c_int, c_int64, c_void_p, RTLD_GLOBAL
+from ctypes import RTLD_GLOBAL, c_char_p, c_double, c_int, c_int64, c_void_p
 import re
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


WillTrojak:
This doesn't do what I need for the FWH plugin, but I think I should be able to add a `region_surf_data` function. The only thing is that it would need the mesh, so it would make sense for it to be called when the mixin is initialised, but not every plugin needs this data, so its a bit wasteful. What are your thoughts on the best way to handle this?

FreddieWitherden:
> This doesn't do what I need for the FWH plugin, but I think I should be able to add a `region_surf_data` function. The only thing is that it would need the mesh, so it would make sense for it to be called when the mixin is initialised, but not every plugin needs this data, so its a bit wasteful. What are your thoughts on the best way to handle this?

I was expecting you'd use the parse function directly and bypass the mixin (which is mostly intended for those who work with the volume solution with PyFR's I/O functions).

WillTrojak:
We could make a surface mixin? There are sufficient differences, and the functionality could then diverge. For example, a surface mixin might be a good place for the infrastructure behind the vtu manifold stuff we've discussed.

FreddieWitherden:
> We could make a surface mixin? There are sufficient differences, and the functionality could then diverge. For example, a surface mixin might be a good place for the infrastructure behind the vtu manifold stuff we've discussed.

Yes, I think that depends on if the manifold stuff is a plugin or not.  (At the moment the mixins do not work for CLI plugins hence why I started refactoring some of the Ascent region stuff.)

WillTrojak:
Ah, I did wonder why you had made the region function external. I'm happy with what you've done. In the FWH plugin, I'll add a surface mixin and you can see what you think.

PR #368, opened 2023-09-21
Updated alpha-opt tri points to they tet face points

WillTrojak:
The higher-order tet and tri alpha-opt points don't match, meaning shock capturing can't be used.

This fix sets the triangle alpha-opt points to match the face points of the alpha-opt tet points.


FreddieWitherden:
Can we sort/format the points so they are a close as possible to what we have already?  This will break less stuff (existing solutions for instance)

FreddieWitherden:
1 => 1.0 and I think we're good.

PR #369, opened 2023-09-28
Added element mixin for surface quadrature

WillTrojak:
This allows new point sets to be generated for plugins, including surface quadratures.

### DIFF
--- a/doc/src/user_guide.rst
+++ b/doc/src/user_guide.rst
@@ -521,7 +521,7 @@ Parameterises multi-p for dual time-stepping with
 
     where ``order`` in the first and last bracketed pair must be the
     overall polynomial order used for the simulation, ``order`` can
-    only change by one between subsequent bracketed pairs, and 
+    only change by one between subsequent bracketed pairs, and
     ``nsteps`` is a non-negative rational number.
 
 Example::
@@ -1253,12 +1253,22 @@ vectors to a CSV file. Parameterised with
 
     ``(x, y, [z])``
 
+5. ``quad-deg-{etype}`` --- degree of quadrature rule for fluid force
+   integration, optionally this can be specified for different element types:
+
+    *int*
+
+6. ``quad-pts-{etype}`` --- name of quadrature rule (optional):
+
+    *string*
+
 Example::
 
     [soln-plugin-fluidforce-wing]
     nsteps = 10
     file = wing-forces.csv
     header = true
+    quad-deg = 6
     morigin = (0.0, 0.0, 0.5)
 
 [soln-plugin-nancheck]
--- a/pyfr/plugins/base.py
+++ b/pyfr/plugins/base.py
@@ -5,8 +5,11 @@ import shlex
 import numpy as np
 from pytools import prefork
 
+from pyfr.inifile import NoOptionError
 from pyfr.mpiutil import get_comm_rank_root, mpi
+from pyfr.quadrules import get_quadrule
 from pyfr.regions import parse_region_expr
+from pyfr.util import memoize
 
 
 def cli_external(meth):
@@ -167,3 +170,26 @@ class RegionMixin:
 
             if not isinstance(eidxs, slice):
                 self._ele_region_data[f'{etype}_idxs'] = eidxs
+
+
+class SurfaceMixin:
+    @memoize
+    def surf_quad(self, itype, proj):
+        # Obtain quadrature info
+        rname = self.cfg.get(f'solver-interfaces-{itype}', 'flux-pts')
+
+        # Quadrature rule (default to that of the solution points)
+        qrule = self.cfg.get(self.cfgsect, f'quad-pts-{itype}', rname)
+        try:
+            qdeg = self.cfg.getint(self.cfgsect, f'quad-deg-{itype}')
+        except NoOptionError:
+            qdeg = self.cfg.getint(self.cfgsect, 'quad-deg')
+
+        q = get_quadrule(itype, qrule, qdeg=qdeg, flags='s')
+
+        return self._proj_pts(proj, q.pts), q.wts
+
+    @staticmethod
+    def _proj_pts(projector, pts):
+        pts = np.atleast_2d(pts.T)
+        return np.vstack(np.broadcast_arrays(*projector(*pts))).T
--- a/pyfr/plugins/fluidforce.py
+++ b/pyfr/plugins/fluidforce.py
@@ -3,10 +3,10 @@ from collections import defaultdict
 import numpy as np
 
 from pyfr.mpiutil import get_comm_rank_root, mpi
-from pyfr.plugins.base import BaseSolnPlugin, init_csv
+from pyfr.plugins.base import BaseSolnPlugin, SurfaceMixin, init_csv
 
 
-class FluidForcePlugin(BaseSolnPlugin):
+class FluidForcePlugin(SurfaceMixin, BaseSolnPlugin):
     name = 'fluidforce'
     systems = ['ac-euler', 'ac-navier-stokes', 'euler', 'navier-stokes']
     formulations = ['dual', 'std']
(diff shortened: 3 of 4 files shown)


pyfr/plugins/base.py
```diff
@@ -167,3 +169,90 @@ def __init__(self, intg, *args, **kwargs):
...
+
+        # Apply the operator to the mesh elements and reshape
+        e_surf = eles.eles[:, eidx]
+        ploc = plocop @ e_surf.reshape(nspts, -1)
+        ploc = ploc.reshape(len(pts), neles, ndims)
+
+        return ploc
+
+    @staticmethod
+    def jacs_at(eles, eidx, pts):
```
FreddieWitherden, line 220: There looks to be a lot of code duplication here.  It should be possible to eliminate most of this by extending some of the eles `_at_np` methods to support more flexible `name` arguments (see `rcpdjac_at_np` for an example).

pyfr/plugins/base.py
```diff
@@ -167,3 +169,90 @@ def __init__(self, intg, *args, **kwargs):
...
+
+            q = get_quadrule(itype, qrule, qdeg=qdeg, flags='s')
+            ppts, pwts = self._proj_pts(proj, q.pts), q.wts
+            self._surf_quad[itype, proj] = (ppts, pwts)
+
+        return ppts, pwts
+
+    @staticmethod
+    def _proj_pts(projector, pts):
+        pts = np.atleast_2d(pts.T)
```
FreddieWitherden, line 202: Is this check needed in this context?

doc/src/user_guide.rst
```diff
@@ -1253,12 +1253,22 @@ vectors to a CSV file. Parameterised with
 
     ``(x, y, [z])``
 
+5. ``quad-deg-{etype}`` --- degree of quadrature rule for fluidforce
```
FreddieWitherden, line 1256: force

pyfr/plugins/base.py
FreddieWitherden: Just following up here.

pyfr/plugins/fluidforce.py
```diff
@@ -105,17 +113,11 @@ def __init__(self, intg, cfgsect, suffix):
...
-
-                eidxs[etype, fidx].append(eidx)
-                norms[etype, fidx].append(pnorms)
-
                 # Get the flux points position of the given face and element
                 # indices relative to the moment origin
                 if self._mcomp:
-                    fpts_idx = eles.basis.facefpts[fidx]
-                    rfpt = eles.plocfpts[fpts_idx, eidx] - morigin
+                    ploc = eles.plocs_at(ppts)[:,[eidx]]
```
FreddieWitherden, line 119: See if we can go without the [].  Also, spacing,

pyfr/solvers/base/elements.py
```diff
@@ -277,24 +278,28 @@ def qpts(self):
...
-        if np.any(np.sqrt(mag_pnorm_fpts) < 1e-10):
+        if np.any(np.sqrt(mag_pnorm) < 1e-10):
             raise RuntimeError('Zero face normals detected')
 
-        return pnorm_fpts
+        return pnorm
 
     @cached_property
-    def _smats_djacs_mpts(self):
+    def smats_djacs_mpts(self):
```
FreddieWitherden, line 302: Why is this renaming needed?

pyfr/plugins/base.py
```diff
@@ -167,3 +169,35 @@ def __init__(self, intg, *args, **kwargs):
...
                 self._ele_region_data[f'{etype}_idxs'] = eidxs
+
+
+class SurfaceMixin:
+    def __init__(self, intg, *args, **kwargs):
+        super().__init__(intg, *args, **kwargs)
+
+        self._surf_quad = {}
+
+    def surf_quad(self, itype, proj, *args):
```
FreddieWitherden, line 180: Just make this a quasi-static method and avoid the need for a new member variable `_surf_quad`.

FreddieWitherden: Or you can @memoize this (but remove *args).

WillTrojak: I added the `*args` so you could just pass `eles.basis.faces[fidx]` but I think this is a niche use case.

pyfr/solvers/base/elements.py
WillTrojak: It shouldn't be needed anymore.

pyfr/plugins/fluidforce.py
WillTrojak: Sadly not, I tried initially as this is ugly.

pyfr/plugins/base.py
WillTrojak: I think it's a reasonable check in case someone tries to pass 1D quad points.

pyfr/plugins/base.py
FreddieWitherden: Yeah, it is usually just asking for trouble.

pyfr/plugins/base.py
FreddieWitherden: It is not really a check.  It is to add support for 1D projections, which I do not think make sense here (we want to get an exception).

pyfr/plugins/base.py
```diff
@@ -167,3 +170,36 @@ def __init__(self, intg, *args, **kwargs):
 
             if not isinstance(eidxs, slice):
                 self._ele_region_data[f'{etype}_idxs'] = eidxs
+
+
+class SurfaceMixin:
+    def __init__(self, intg, *args, **kwargs):
```
FreddieWitherden, line 176: Do we need this?

pyfr/plugins/base.py
```diff
@@ -167,3 +170,36 @@ def __init__(self, intg, *args, **kwargs):
...
+            # Quadrature rule (default to that of the solution points)
+            qrule = self.cfg.get(self.cfgsect, f'quad-pts-{itype}', rname)
+            try:
+                qdeg = self.cfg.getint(self.cfgsect, f'quad-deg-{itype}')
+            except NoOptionError:
+                qdeg = self.cfg.getint(self.cfgsect, 'quad-deg')
+
+            q = get_quadrule(itype, qrule, qdeg=qdeg, flags='s')
+            ppts, pwts = self._proj_pts(proj, q.pts), q.wts
+            self._surf_quad[itype, proj] = (ppts, pwts)
```
FreddieWitherden, line 198: Do we need this?

pyfr/solvers/base/elements.py
```diff
@@ -277,21 +278,25 @@ def qpts(self):
...
-        smats = self.smat_at_np('fpts').transpose(1, 3, 0, 2)
+        return self.pnorm_at('fpts', self.basis.norm_fpts)
+
+    @memoize
+    def pnorm_at(self, name, norm):
+        smats = self.smat_at_np(name).transpose(1, 3, 0, 2)
 
         # We need to compute |J|*[(J^{-1})^{T}.N] where J is the
-        # Jacobian and N is the normal for each fpt.  Using
+        # Jacobian and N is the normal for each pnt. Using
```
FreddieWitherden, line 288: point?

pyfr/plugins/base.py
```diff
@@ -167,3 +170,26 @@ def __init__(self, intg, *args, **kwargs):
 
             if not isinstance(eidxs, slice):
                 self._ele_region_data[f'{etype}_idxs'] = eidxs
+
+
+class SurfaceMixin:
+    @memoize
+    def surf_quad(self, itype, proj):
```
FreddieWitherden, line 177: Make _ as probably not meant to be called by external users of the plugin?

pyfr/plugins/fluidforce.py
```diff
@@ -86,12 +86,20 @@ def __init__(self, intg, cfgsect, suffix):
 
             for etype, eidx, fidx, flags in mesh[bc].astype('U4,i4,i1,i2'):
                 eles = elemap[etype]
+                itype, proj, norm = eles.basis.faces[fidx]
 
-                if (etype, fidx) not in m0:
-                    facefpts = eles.basis.facefpts[fidx]
+                ppts, pwts = self.surf_quad(itype, proj)
```
FreddieWitherden, line 91: q is probably a better prefix as they're quadrature points.

pyfr/plugins/base.py
```diff
@@ -167,3 +170,26 @@ def __init__(self, intg, *args, **kwargs):
...
+        rname = self.cfg.get(f'solver-interfaces-{itype}', 'flux-pts')
+
+        # Quadrature rule (default to that of the solution points)
+        qrule = self.cfg.get(self.cfgsect, f'quad-pts-{itype}', rname)
+        try:
+            qdeg = self.cfg.getint(self.cfgsect, f'quad-deg-{itype}')
+        except NoOptionError:
+            qdeg = self.cfg.getint(self.cfgsect, 'quad-deg')
+
+        q = get_quadrule(itype, qrule, qdeg=qdeg, flags='s')
```
FreddieWitherden, line 188: Make flags an option argument to surf_quad?

PR #370, opened 2023-09-28
Variable ac-zeta

sambitmishra98:
This PR is the first step to varying ac-zeta across P-multigrid levels. I have compared csv files of `[soln-plugin-pseudostats]` for the develop version and the PR, to check if the PR works. Setting ac-zeta this way did not seem to affect the performance of the inc-cylinder-2d case from PyFR-test-cases. Let me know if we need to test its performance difference with the develop version more.

Also, I was not sure where to initialise ac-zeta, so I shifted it from `[constants]` to `[solver]`.

With this PR, we can now look into the effect of changing ac-zeta across pseudo-iterations, multi-P levels, and stages.

### DIFF
--- a/pyfr/solvers/aceuler/elements.py
+++ b/pyfr/solvers/aceuler/elements.py
@@ -62,12 +62,14 @@ class ACEulerElements(BaseACFluidElements, BaseAdvectionElements):
             self.kernels['tdisf_curved'] = lambda uin: self._be.kernel(
                 'tflux', tplargs=tplargs, dims=[self.nupts, r[c]],
                 u=s(self.scal_upts[uin], c), f=s(self._vect_upts, c),
+                extrns={'ac_zeta': 'scalar fpdtype_t'},
                 smats=self.curved_smat_at('upts')
             )
         elif c in r:
             self.kernels['tdisf_curved'] = lambda: self._be.kernel(
                 'tflux', tplargs=tplargs, dims=[self.nqpts, r[c]],
                 u=s(self._scal_qpts, c), f=s(self._vect_qpts, c),
+                extrns={'ac_zeta': 'scalar fpdtype_t'},
                 smats=self.curved_smat_at('qpts')
             )
 
@@ -75,11 +77,13 @@ class ACEulerElements(BaseACFluidElements, BaseAdvectionElements):
             self.kernels['tdisf_linear'] = lambda uin: self._be.kernel(
                 'tfluxlin', tplargs=tplargs, dims=[self.nupts, r[l]],
                 u=s(self.scal_upts[uin], l), f=s(self._vect_upts, l),
+                extrns={'ac_zeta': 'scalar fpdtype_t'},
                 verts=self.ploc_at('linspts', l), upts=self.upts
             )
         elif l in r:
             self.kernels['tdisf_linear'] = lambda: self._be.kernel(
                 'tfluxlin', tplargs=tplargs, dims=[self.nqpts, r[l]],
                 u=s(self._scal_qpts, l), f=s(self._vect_qpts, l),
+                extrns={'ac_zeta': 'scalar fpdtype_t'},
                 verts=self.ploc_at('linspts', l), upts=self.qpts
             )
--- a/pyfr/solvers/aceuler/inters.py
+++ b/pyfr/solvers/aceuler/inters.py
@@ -13,9 +13,11 @@ class ACEulerIntInters(BaseAdvectionIntInters):
         tplargs = dict(ndims=self.ndims, nvars=self.nvars, rsolver=rsolver,
                        c=self.c)
 
+        self._set_external('ac_zeta', 'scalar fpdtype_t')
         self.kernels['comm_flux'] = lambda: self._be.kernel(
             'intcflux', tplargs=tplargs, dims=[self.ninterfpts],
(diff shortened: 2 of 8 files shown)


pyfr/solvers/aceuler/system.py
```diff
@@ -11,3 +11,25 @@ class ACEulerSystem(BaseAdvectionSystem):
...
+            b.prepare(t)
+
+        for b in binders: 
+            b(t=t, ac_zeta=self._ac_zeta)
+
+    @property
+    def ac_zeta(self):
+        return self._ac_zeta
+
+    @ac_zeta.setter
```
FreddieWitherden, line 33: Why are these needed?

pyfr/solvers/aceuler/system.py
```diff
@@ -11,3 +11,25 @@ class ACEulerSystem(BaseAdvectionSystem):
     intinterscls = ACEulerIntInters
     mpiinterscls = ACEulerMPIInters
     bbcinterscls = ACEulerBaseBCInters
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        
+        self._ac_zeta = self.cfg.getfloat('solver', 'ac-zeta')
+
+    def _prepare_kernels(self, t, uinbank, foutbank):
```
FreddieWitherden, line 20: Too much code duplication here.

pyfr/solvers/acnavstokes/elements.py
```diff
@@ -37,24 +37,28 @@ def set_backend(self, *args, **kwargs):
             self.kernels['tdisf_curved'] = lambda uin: self._be.kernel(
                 'tflux', tplargs=tplargs, dims=[self.nupts, r[c]],
                 u=s(self.scal_upts[uin], c), f=s(self._vect_upts, c),
+                extrns={'ac_zeta': 'scalar fpdtype_t'},
```
FreddieWitherden, line 40: See if we can factor this out.  For example define:
```python
def kernel(*args, **kwargs):
    return lambda: self._be.kernel(*args, **kwargs, tplargs=tplargs,
                                   externs={'ac_zeta': 'scalar fpdtype_t'})
```

pyfr/solvers/aceuler/inters.py
```diff
@@ -45,10 +49,10 @@ def __init__(self, *args, **kwargs):
         tplargs = dict(ndims=self.ndims, nvars=self.nvars, rsolver=rsolver,
                        c=self.c, bctype=self.type)
 
+        self._set_external('ac_zeta', 'scalar fpdtype_t')
         self.kernels['comm_flux'] = lambda: self._be.kernel(
             'bccflux', tplargs=tplargs, dims=[self.ninterfpts],
             extrns=self._external_args, ul=self._scal_lhs, nl=self._pnorm_lhs,
-            **self._external_vals
```
FreddieWitherden, line 51: Why are we removing this?

FreddieWitherden:
What kind of improvement does this have the potential to give for a real-world test case like a TGV or SD7003?

sambitmishra98:
I just tested it for a TGV case. I added a few lines to make `ac-zeta` vary across multi-P levels by a factor, just like `pseudo-dt-fact`.

```
mcfg.set('solver', 'ac-zeta', self.zeta*self.zetaf**(order - l))
```
|ac-zeta-fact|total pseudo-steps|total-time|
| -- | -- | -- |
| 1.0 (develop) | 4062 | 12m 56s |
| 1.0 (PR) | 4062 | 13m 00s |
| 1.5 (PR) | 3982 | 12m 48s |
| 2.0 (PR) | 3996 | 12m 57s |

Last time I tried, I had got speedup by changing with pseudo-iterations. Let me do that and get the speed-up before proceeding any further.

pyfr/solvers/aceuler/elements.py
```diff
@@ -62,24 +62,28 @@ def set_backend(self, *args, **kwargs):
             self.kernels['tdisf_curved'] = lambda uin: self._be.kernel(
                 'tflux', tplargs=tplargs, dims=[self.nupts, r[c]],
                 u=s(self.scal_upts[uin], c), f=s(self._vect_upts, c),
+                extrns={'ac_zeta': 'scalar fpdtype_t'},
```
WillTrojak, line 65: Could you set this as a variable eg `exargs` and then pass that to each kernel? Should reduce the chance of any errors.

pyfr/solvers/acnavstokes/system.py
```diff
@@ -12,3 +12,25 @@ class ACNavierStokesSystem(BaseAdvectionDiffusionSystem):
     intinterscls = ACNavierStokesIntInters
     mpiinterscls = ACNavierStokesMPIInters
     bbcinterscls = ACNavierStokesBaseBCInters
+
```
WillTrojak, line 15: Maybe this could be a mixin? `ExternalSystem` or something. You will probably need to come up with a nice syntax for saying what externals are needed though.

WillTrojak (commented):
Just a few minor changes; the main thing that needs thought is the repetition in the system class.

FreddieWitherden:
Did this turn out to be beneficial in the end?

sambitmishra98:
No, nothing conclusive.

PR #371, opened 2023-10-03
Fusion of gradient and flux kernels

WillTrojak:
Speedup can be gained by fusing the gradient transformation and flux kernel, but only when flux anti-aliasing isn't active.


pyfr/solvers/acnavstokes/kernels/tfluxgradcoru.mako
```diff
@@ -0,0 +1,26 @@
...
+    fpdtype_t ftemp[${ndims}][${nvars}];
+    ${pyfr.expand('inviscid_flux', 'u', 'ftemp')};
+    ${pyfr.expand('viscous_flux_add', 'u', 'gradu', 'ftemp')};
+
+    // Transform the fluxes
+% for i, j in pyfr.ndrange(ndims, nvars):
+    f[${i}][${j}] = ${' + '.join(f'smats[{i}][{k}]*ftemp[{k}][{j}]'
+                                 for k in range(ndims))};
+% endfor
+</%pyfr:kernel>
```
FreddieWitherden, line 26: Missing newline.

FreddieWitherden:
I think we should see if there is a means of eliminating the flux kernel duplication.  Currently, every solver needs four: (linear, curved) x (fused, not fused).  This is usually when we see if there is anything that can be done to simplify.

WillTrojak:
I don't see a way to reduce the number of kernels. The issue is that the inputs are somewhat physics-dependent. For example, nav-stokes has artificial viscosity and ACM is about to get ac-zeta. This means it's hard to move general versions of the kernels to the baseadvecdiff solver. 

One option is to pull out the bits that do the flux calculation and transformation and move them into flux.mako. This would reduce that repetition. But I'm open to other ideas as I agree it is a bit unwieldy.

FreddieWitherden:
> I don't see a way to reduce the number of kernels. The issue is that the inputs are somewhat physics-dependent. For example, nav-stokes has artificial viscosity and ACM is about to get ac-zeta. This means it's hard to move general versions of the kernels to the baseadvecdiff solver.
> 
> One option is to pull out the bits that do the flux calculation and transformation and move them into flux.mako. This would reduce that repetition. But I'm open to other ideas as I agree it is a bit unwieldy.

Could we define a kernel (per system) which takes all of the arguments (linear arrays, curved arrays, gradient) and then has template arguments which tell it what bits to do/generate.  Then there will just be a single kernel which can be generated multiple times.  The argument elimination code in PyFR should do the rest.

pyfr/solvers/navstokes/kernels/tflux.mako
```diff
@@ -1,25 +1,47 @@
...
               u='in fpdtype_t[${str(nvars)}]'
               artvisc='in broadcast-col fpdtype_t'
               f='inout fpdtype_t[${str(ndims)}][${str(nvars)}]'
-              smats='in fpdtype_t[${str(ndims)}][${str(ndims)}]'>
+              gradu='inout fpdtype_t[${str(ndims)}][${str(nvars)}]'
+              smats='in fpdtype_t[${str(ndims)}][${str(ndims)}]'
+              rcpdjac='in fpdtype_t'
+              verts='in broadcast-col fpdtype_t[${str(nverts)}][${str(ndims)}]'
+              upts='in broadcast-row fpdtype_t[${str(ndims)}]'>
+
```
FreddieWitherden, line 22: Remove newline.

FreddieWitherden:
Are you able to test this for Euler, NS, AC-Euler, and AC-NS?

pyfr/solvers/baseadvecdiff/elements.py
```diff
@@ -49,25 +53,31 @@ def set_backend(self, backend, nscalupts, nonce, linoff):
...
+            kerns['gradcoru_u_linear'] = lambda: kernel(
                 'gradcoru', tplargs=tplargs | {'ktype': 'linear'},
                 dims=[self.nupts, regions['linear']],
-                gradu=slicem(self._vect_upts, 'linear'),
+                gradu=slicem(grad, 'linear'),
                 upts=self.upts, verts=self.ploc_at('linspts', 'linear')
             )
 
+        if 'flux' in self.antialias or self.basis.order == 0:
+            if 'linear' in regions:
```
FreddieWitherden, line 73: For loop?

pyfr/solvers/baseadvecdiff/system.py
```diff
@@ -77,10 +77,20 @@ def deps(dk, *names): return self._kdeps(k, dk, *names)
...
+        # Compute the fused transformed flux and corrected gradient
+        for l in k['eles/tdisf_fused_curved'] + k['eles/tdisf_fused_linear']:
+            ldeps = deps(l, 'eles/tgradcoru_upts')
+            g2.add(l, deps=ldeps)
+
         # Interpolate these gradients to the flux points
         for l in k['eles/gradcoru_fpts']:
-            ldeps = deps(l, 'eles/gradcoru_upts_curved',
-                         'eles/gradcoru_upts_linear')
+            if (k['eles/gradcoru_upts_curved'] or
```
FreddieWitherden, line 87: If the check needed here?

pyfr/solvers/navstokes/elements.py
```diff
@@ -36,7 +36,10 @@ def set_backend(self, *args, **kwargs):
             return
 
         # Register our flux kernels
-        self._be.pointwise.register('pyfr.solvers.navstokes.kernels.tflux')
+
```
FreddieWitherden, line 39: Remove newline.

pyfr/solvers/navstokes/elements.py
```diff
@@ -36,7 +36,10 @@ def set_backend(self, *args, **kwargs):
...
+        kernel, kerns = self._be.kernel, self.kernels
```
FreddieWitherden, line 40: Move down.

pyfr/solvers/navstokes/kernels/tflux.mako
```diff
@@ -2,30 +2,41 @@
...
               u='in fpdtype_t[${str(nvars)}]'
               artvisc='in broadcast-col fpdtype_t'
               f='inout fpdtype_t[${str(ndims)}][${str(nvars)}]'
+              gradu='inout fpdtype_t[${str(ndims)}][${str(nvars)}]'
               smats='in fpdtype_t[${str(ndims)}][${str(ndims)}]'
+              rcpdjac='in fpdtype_t'
               verts='in broadcast-col fpdtype_t[${str(nverts)}][${str(ndims)}]'
               upts='in broadcast-row fpdtype_t[${str(ndims)}]'>
-% if 'linear' in ktype:
+% if ktype.startswith('linear'):
```
FreddieWitherden, line 22: Why is this check different to what we do with acnavstokes?

pyfr/solvers/baseadvecdiff/elements.py
```diff
@@ -9,35 +9,39 @@ def _scratch_bufs(self):
...
-        kernel = self._be.kernel
+        kernel, kerns = self._be.kernel, self.kernels
         kprefix = 'pyfr.solvers.baseadvecdiff.kernels'
         slicem = self._slice_mat
 
         # Register our pointwise kernels
         self._be.pointwise.register(f'{kprefix}.gradcoru')
 
+        # Set memory optimisation based on config
+        grad = self._vect_upts if 'flux' in self.antialias else self._grad_upts
```
FreddieWitherden, line 28: Check line length.

pyfr/solvers/baseadvecdiff/system.py
```diff
@@ -112,7 +120,12 @@ def deps(dk, *names): return self._kdeps(k, dk, *names)
 
         # Compute the transformed divergence of the partially corrected flux
         for l in k['eles/tdivtpcorf']:
-            g2.add(l, deps=deps(l, 'eles/tdisf_curved', 'eles/tdisf_linear'))
+            ldeps = deps(l,
```
FreddieWitherden, line 123: Indent this differently:
```
deps(
    l, ...
)
```

pyfr/solvers/baseadvecdiff/elements.py
FreddieWitherden: Also provide an option to disable (it increases memory usage so we may want it off sometimes).

WillTrojak: I added this to `backend` as it felt more like a backend thing than a system thing. But I can move it.

pyfr/solvers/acnavstokes/elements.py
```diff
@@ -16,9 +16,9 @@ def set_backend(self, *args, **kwargs):
             return
 
         # Register our flux kernels
+        kernel, kerns = self._be.kernel, self.kernels
```
FreddieWitherden, line 19: Any chance of better names here?

pyfr/solvers/base/elements.py
```diff
@@ -39,6 +39,10 @@ def __init__(self, basiscls, eles, cfg):
         # If we need quadrature points or not
         haveqpts = 'flux' in self.antialias
 
+        # If we are doing gradient fusion
+        self.grad_fusion = (cfg.getbool('backend', 'grad-fusion', True) and
```
FreddieWitherden, line 43: Should be part of the solver rather than the backend (this code is `pyfr/solvers...` rather than `pyfr/backends`)

pyfr/solvers/base/system.py
```diff
@@ -46,7 +46,9 @@ def __init__(self, backend, rallocs, mesh, initsoln, nregs, cfg):
         # Get all the solution point locations for the elements
         self.ele_ploc_upts = [e.ploc_at_np('upts') for e in eles]
 
-        if hasattr(eles[0], '_vect_upts'):
+        if hasattr(eles[0], '_grad_upts'):
```
FreddieWitherden, line 49: Is this still needed?

pyfr/solvers/acnavstokes/elements.py
WillTrojak: Elsewhere, you use kernels rather than kerns, so let's use that.

PR #373, opened 2023-10-10
Migrate to importlib.resources.

FreddieWitherden:
We also remove the dense pseudo-steppers which do not appear to be in active use.


WillTrojak:
Yep, this seems fine to me.

PR #375, opened 2023-11-10
Lobatto point optimisation

WillTrojak:
When flux points are a subset of solution points, this performs from additional optimisations. Currently, on p6 tets this gives about a 10% reduction in runtime.

### DIFF
--- a/pyfr/nputil.py
+++ b/pyfr/nputil.py
@@ -35,7 +35,8 @@ def clean(origfn=None, tol=1e-10):
 
                 i, ix = 0, amix[0]
                 for j, jx in enumerate(amix[1:], start=1):
-                    if amfl[jx] - amfl[ix] >= tol:
+                    if not np.isclose(amfl[jx], amfl[ix], rtol=tol,
+                                      atol=0.1*tol):
                         if j - i > 1:
                             amfl[amix[i:j]] = np.median(amfl[amix[i:j]])
                         i, ix = j, jx
--- a/pyfr/shapes.py
+++ b/pyfr/shapes.py
@@ -314,6 +314,16 @@ class BaseShape:
     def nmpts(self):
         return len(self.mpts)
 
+    @cached_property
+    def fpts_in_upts(self):
+        return np.all(np.count_nonzero(self.m0, axis=1) == 1)
+
+    @cached_property
+    def fpts_map_upts(self):
+        if not self.fpts_in_upts:
+            raise ValueError('Flux points not subset of solution points')
+        return np.argwhere(np.abs(self.m0 - 1) <= 1e-8)[:, 1]
+
 
 class TensorProdShape:
     @classmethod
--- a/pyfr/solvers/base/elements.py
+++ b/pyfr/solvers/base/elements.py
@@ -1,9 +1,23 @@
-from functools import cached_property
+from functools import cached_property, wraps
 
 import numpy as np
 
 from pyfr.nputil import npeval, fuzzysort
 from pyfr.util import memoize
+
+
+def inters_map(meth):
+    @wraps(meth)
+    def newmeth(self, eidx, fidx):
+        nfp = self.nfacefpts[fidx]
+        cmap = (eidx,)*nfp
+
+        match meth(self, eidx, fidx):
+            case [mid, rmap]:
+                return (mid,)*nfp, rmap, cmap
+            case [mid, rmap, lda]:
+                return (mid,)*nfp, rmap, cmap, (lda,)*nfp
+    return newmeth
 
 
 class BaseElements:
@@ -49,6 +63,13 @@ class BaseElements:
         self.nfpts = basis.nfpts
         self.nfacefpts = basis.nfacefpts
         self.nmpts = basis.nmpts
+
+        if self.basis.fpts_in_upts:
+            self.get_vect_fpts_for_inter = self._get_vect_upts_for_inter
+            self.get_comm_fpts_for_inter = self._get_comm_fpts_for_inter
+        else:
+            self.get_vect_fpts_for_inter = self._get_vect_fpts_for_inter
+            self.get_comm_fpts_for_inter = self._get_vect_fpts_for_inter
 
     @staticmethod
     def validate_formulation(form, intg, cfg):
@@ -188,6 +209,12 @@ class BaseElements:
             self._vect_qpts = valloc('vect_qpts', nqpts)
         if 'vect_fpts' in sbufs:
             self._vect_fpts = valloc('vect_fpts', nfpts)
+
+        # Allocate space if needed for interfaces
+        if 'comm_fpts' in sbufs:
+            self._comm_fpts = salloc('comm_fpts', nfpts)
+        elif 'vect_fpts' in sbufs:
+            self._comm_fpts = self._vect_fpts.slice(0, self.nfpts)
 
         if 'grad_upts' in sbufs and self.grad_fusion:
             self._grad_upts = valloc('grad_upts', nupts)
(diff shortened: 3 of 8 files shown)


pyfr/shapes.py
```diff
@@ -314,6 +314,18 @@ def mpts(self):
...
+    @cached_property
+    def fpts_in_upts(self):
+        mrowsum = np.max(np.abs(np.sum(self.m0, axis=1) - 1.0))
+        return np.min(self.m0) > -1e-8 and mrowsum < 1e-8
+
+    @cached_property
+    def fpts_map_upts(self):
+        if not self.fpts_in_upts:
+            raise ValueError("Flux points not subset of solution points")
+        is_one = lambda r: np.logical_and(1 - 1e-8 <= r, r <= 1 + 1e-8)
```
FreddieWitherden, line 326: relative difference of abs is more typical, no need for an `and`.

pyfr/solvers/base/elements.py
```diff
@@ -383,6 +392,25 @@ def get_vect_fpts_for_inter(self, eidx, fidx):
 
         return (self._vect_fpts.mid,)*nfp, rmap, cmap, rstri
 
+    def _get_vect_upts_for_inter(self, eidx, fidx):
```
FreddieWitherden, line 395: I think we should see if we can template these boiler plate get functions now we have quite a few of them.

pyfr/solvers/baseadvecdiff/elements.py
```diff
@@ -10,6 +10,9 @@ def _scratch_bufs(self):
         if 'flux' in self.antialias:
             bufs |= {'scal_qpts', 'vect_qpts'}
 
+        if self.basis.fpts_in_upts:
```
FreddieWitherden, line 13: Is this always what we want?  Are there ever exceptions based on AA and the such like?

WillTrojak: from my understanding, when we use flux point AA, the flux points 'fpts' are set as the AA quadrature points, and as a result the self.fpts_in_upts will fail. So I think this should be fine.

pyfr/solvers/base/elements.py
```diff
@@ -176,6 +183,8 @@ def set_backend(self, backend, nscalupts, nonce, linoff):
             self._scal_fpts = salloc('scal_fpts', nfpts)
         if 'scal_qpts' in sbufs:
             self._scal_qpts = salloc('scal_qpts', nqpts)
+        if 'comm_fpts' in sbufs:
```
FreddieWitherden, line 186: Also, is it not possible to save some storage in the Lobatto case as we never need to allocate vect_fpts?

pyfr/solvers/baseadvecdiff/elements.py
```diff
@@ -25,9 +28,12 @@ def set_backend(self, backend, nscalupts, nonce, linoff):
         # Mesh regions
         regions = self._mesh_regions
 
+        comm = (self._comm_fpts if self.basis.fpts_in_upts
```
FreddieWitherden, line 31: Any reason not to assign `self._comm_fpts = self._vect_fpts.slice(0, self.nfpts)` when we decide what to allocate to avoid this check?

pyfr/shapes.py
```diff
@@ -314,6 +314,18 @@ def mpts(self):
...
+    @cached_property
+    def fpts_in_upts(self):
+        mrowsum = np.max(np.abs(np.sum(self.m0, axis=1) - 1.0))
+        return np.min(self.m0) > -1e-8 and mrowsum < 1e-8
+
+    @cached_property
+    def fpts_map_upts(self):
+        if not self.fpts_in_upts:
+            raise ValueError("Flux points not subset of solution points")
+        def is_one(r): return np.abs(r - 1) <= 1e-8
```
FreddieWitherden, line 326: I still think this can be cleaned up further.

pyfr/solvers/base/elements.py
```diff
@@ -1,11 +1,22 @@
-from functools import cached_property
+from functools import cached_property, wraps
 
 import numpy as np
 
 from pyfr.nputil import npeval, fuzzysort
 from pyfr.util import memoize
 
 
+def inters(meth):
```
FreddieWitherden, line 9: Rename.

pyfr/solvers/baseadvecdiff/elements.py
```diff
@@ -10,6 +10,10 @@ def _scratch_bufs(self):
         if 'flux' in self.antialias:
             bufs |= {'scal_qpts', 'vect_qpts'}
 
+        if self.basis.fpts_in_upts:
+            bufs |= {'comm_fpts'}
+            bufs.remove('vect_fpts')
```
FreddieWitherden, line 15: Consistency in notation (|= vs add and remove vs -=).

pyfr/shapes.py
```diff
@@ -314,6 +314,17 @@ def mpts(self):
     def nmpts(self):
         return len(self.mpts)
 
+    @cached_property
+    def fpts_in_upts(self):
+        mrowsum = np.max(np.abs(np.sum(self.m0, axis=1) - 1.0))
+        return np.min(self.m0) > -1e-8 and mrowsum < 1e-8
```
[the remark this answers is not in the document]
WillTrojak: I found that at higher orders, the clean wasn't guaranteed to give that behaviour, and this approach seemed more robust.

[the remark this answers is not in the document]
FreddieWitherden: Okay, we probably need to fix that issue with clean as it will be breaking GiMMiK.

pyfr/solvers/baseadvecdiff/inters.py
```diff
@@ -10,6 +10,8 @@ def __init__(self, be, lhs, rhs, elemap, cfg):
         # Generate the additional view matrices
         self._vect_lhs = self._vect_view(lhs, 'get_vect_fpts_for_inter')
         self._vect_rhs = self._vect_view(rhs, 'get_vect_fpts_for_inter')
+        self._comm_lhs = self._scal_view(lhs, 'get_comm_fpts_for_inter')
```
FreddieWitherden, line 13: Can you confirm that this does not result in additional memory allocations?

WillTrojak: Yep, this doesn't result in any more memory as some memory (`vect_fpts`) is removed and some is added (`comm_fpts`).

pyfr/shapes.py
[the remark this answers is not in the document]
WillTrojak: I applied a fix that changes the tolerance calculations to be relative and this seems to fix the problem at least at p7

pyfr/nputil.py
```diff
@@ -35,7 +35,9 @@ def newfn(*args, **kwargs):
 
                 i, ix = 0, amix[0]
                 for j, jx in enumerate(amix[1:], start=1):
-                    if amfl[jx] - amfl[ix] >= tol:
+                    rdiff = ((amfl[jx] - amfl[ix]) /
```
FreddieWitherden, line 38: I think there might be a numpy function for this; see `np.allclose` and its arguments.

pyfr/nputil.py
```diff
@@ -26,7 +25,8 @@ def newfn(*args, **kwargs):
             arr = fn(*args, **kwargs).copy()
 
             # Flush small elements to zero
-            arr[np.abs(arr) < tol] = 0
```
FreddieWitherden, line 29: This seems like it should be absolute?

pyfr/nputil.py
```diff
@@ -35,7 +35,7 @@ def newfn(*args, **kwargs):
 
                 i, ix = 0, amix[0]
                 for j, jx in enumerate(amix[1:], start=1):
-                    if amfl[jx] - amfl[ix] >= tol:
+                    if ~np.isclose(amfl[jx], amfl[ix], rtol=tol, atol=0.1*tol):
```
FreddieWitherden, line 38: `not np.isclose`

pyfr/shapes.py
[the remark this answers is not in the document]
FreddieWitherden: With the fix are we now okay to switch to the more intuitive check?

pyfr/solvers/navstokes/inters.py
```diff
@@ -39,7 +39,7 @@ def __init__(self, *args, **kwargs):
         self.kernels['con_u'] = lambda: self._be.kernel(
             'intconu', tplargs=self._tplargs, dims=[self.ninterfpts],
             ulin=self._scal_lhs, urin=self._scal_rhs,
-            ulout=self._vect_lhs, urout=self._vect_rhs
+            ulout=self._comm_lhs, urout=self._comm_rhs
```
FreddieWitherden, line 42: Why are there no corresponding changes to ac-navier-stokes?

FreddieWitherden:
Can you rebase?

pyfr/solvers/base/elements.py
```diff
@@ -1,11 +1,26 @@
...
+def inters_map(meth):
+    @wraps(meth)
+    def newmeth(self, eidx, fidx):
+        nfp = self.nfacefpts[fidx]
+        cmap = (eidx,)*nfp
+
+        mid, rmap, *lda = meth(self, eidx, fidx)
+
+        if lda:
+            return (mid,)*nfp, rmap, cmap, (lda)*nfp
```
FreddieWitherden, line 18: What is the extra argument?

pyfr/solvers/base/elements.py
```diff
@@ -1,11 +1,26 @@
...
 
+def inters_map(meth):
+    @wraps(meth)
+    def newmeth(self, eidx, fidx):
+        nfp = self.nfacefpts[fidx]
+        cmap = (eidx,)*nfp
+
+        mid, rmap, *lda = meth(self, eidx, fidx)
+
+        if lda:
```
WillTrojak, line 17: @FreddieWitherden you might have some idea of a cleaner way to do this. The reason for needing this is the result gets passed to `np.concatenate` so you can't pass back None

FreddieWitherden: What methods return four things?  If I recall all of the existing interface methods are expected to return 3-tuples.

pyfr/solvers/base/elements.py
WillTrojak: It's the stride for the vector interface terms. Before, I was doing `(LDA,)`, which returned `[]` when it was empty, but when it's not empty the concatenated array becomes 2D. Which in some case leads to huge amounts of memory needing to be allocated

FreddieWitherden: But this code now returns four things?  It does not make sense.

PR #379, opened 2023-11-24
Enable regions to be rotated.

FreddieWitherden:


pyfr/regions.py
```diff
@@ -150,6 +150,27 @@ def interior_eles(self, mesh, rallocs):
...
 
+    def __init__(self, rot=None):
+        rot = np.rad2deg(rot) if rot is not None else rot
+
+        match rot:
+            case None:
+                self.rot = None
+            case [phi, theta, psi]:
+                c, s = np.cos(phi), np.sin(phi)
+                D = np.array([[c, s, 0], [-s, c, 0], [0, 0, 1]])
```
tarikdzanic, line 161: These 3 axis rotation matrices use the form `[c,s; -s,c]` instead of the standard `[c,-s; s,c]` which is used in 2D. I think this makes it rotate the opposite way?

tarikdzanic: This might end up getting canceled out in the matrix multiplication

FreddieWitherden: I hate Euler angles, but went with the x-convention from https://mathworld.wolfram.com/EulerAngles.html

Easy enough to change if we think there is something more logical, though.

tarikdzanic: I believe those equations are wrong due to the negative off-diagonal terms. The picture for D rotation shows it how it should be (mapping x,y to x',y') but the matrix for D actually rotates the wrong way (maps x',y' to x,y). I think https://en.wikipedia.org/wiki/Rotation_matrix#In_three_dimensions shows the correct rotation matrices which also coincide with the 2D rotation matrix.

FreddieWitherden: I switched to the ZYX convention which is the default in MATLAB (http://web.mit.edu/2.05/www/Handout/HO2.PDF)

tarikdzanic: LGTM

PR #381, opened 2023-12-01
Added check for GMSH unique physical IDs.

tarikdzanic:


pyfr/readers/gmsh.py
```diff
@@ -276,9 +277,14 @@ def _read_phys_names(self, mshit):
...
 
             # Ensure we have not seen this name before
-            if name in seen:
+            if name in seen_names:
                 raise ValueError(f'Duplicate physical name: {name}')
 
+            # Ensure physical entitiy IDs are unique
+            if pent in seen_ids:
+                raise ValueError('PyFR only supports GMSH meshes with unique '
+                                 f'physical entity IDs. Duplicate ID: {pent}')
```
FreddieWitherden, line 286: Just raise a one-liner like we do for names.

PR #382, opened 2023-12-01
Added nan dump

WillTrojak:
This adds the ability to dump the solution around detected NaN points. For each partition and each element type, it will find the approximate centroid of any NaN and create a box region around it. It will then dump a single pyfrs file that contains these regions.

Identifying individual contiguous patches of NaN seemed a bit overkill, but that is an option for later if people find this method insufficient.

### DIFF
--- a/doc/src/user_guide.rst
+++ b/doc/src/user_guide.rst
@@ -1116,7 +1116,7 @@ the prismatic elements at multi-p level *order*, with
 1. ``soln-pts`` --- location of the solution points in a prismatic
    element:
 
-    ``alpha-opt~gauss-legendre-lobatto`` | 
+    ``alpha-opt~gauss-legendre-lobatto`` |
     ``williams-shunn~gauss-legendre`` |
     ``williams-shunn~gauss-legendre-lobatto``
 
@@ -1281,10 +1281,32 @@ Periodically checks the solution for NaN values. Parameterised with
 
     *int*
 
+2. ``region-dump`` --- if NaN regions should be dumped once detected:
+
+    *boolean*
+
+3. ``region-len`` --- side length of box region placed around the centriod of
+   NaN points:
+
+    *float*
+
+4. ``basedir`` --- relative path to directory where dump file will be
+   written:
+
+    *string*
+
+5. ``basename`` --- pattern of dump file name:
+
+    *string*
+
 Example::
 
     [soln-plugin-nancheck]
     nsteps = 10
+    region-dump = true
+    region-len = 0.5
+    basedir = data
+    basename = nan_dump
 
 [soln-plugin-residual]
 ^^^^^^^^^^^^^^^^^^^^^^
--- a/pyfr/plugins/base.py
+++ b/pyfr/plugins/base.py
@@ -5,7 +5,7 @@ import shlex
 import numpy as np
 from pytools import prefork
 
-from pyfr.inifile import NoOptionError
+from pyfr.inifile import Inifile, NoOptionError
 from pyfr.mpiutil import get_comm_rank_root, mpi
 from pyfr.quadrules import get_quadrule
 from pyfr.regions import parse_region_expr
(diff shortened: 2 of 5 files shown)


pyfr/solvers/base/system.py
```diff
@@ -130,11 +130,14 @@ def _load_eles(self, rallocs, mesh, initsoln, nregs, nonce):
...
-        int_inters = self.intinterscls(self.backend, lhs, rhs, elemap,
-                                       self.cfg)
-
-        return [int_inters]
+        try:
+            lhs, rhs = mesh[key].astype('U4,i4,i1,i2').tolist()
+            int_inters = [self.intinterscls(self.backend, lhs, rhs, elemap,
+                                            self.cfg)]
+        except KeyError:
+            int_inters = []
```
FreddieWitherden, line 138: Not sure this should be needed for any sane mesh.

doc/src/user_guide.rst
```diff
@@ -1281,10 +1281,32 @@ Periodically checks the solution for NaN values. Parameterised with
...
+
+5. ``basename`` --- pattern of dump file name:
+
+    *string*
+
 Example::
 
     [soln-plugin-nancheck]
     nsteps = 10
+    region-dump = True
```
FreddieWitherden, line 1306: Usually use lower case true.

pyfr/plugins/nancheck.py
```diff
@@ -9,12 +12,99 @@ class NaNCheckPlugin(BaseSolnPlugin):
...
+
+        return ele_regions, ele_region_data
+
     def __call__(self, intg):
         if intg.nacptsteps % self.nsteps == 0:
-            if any(np.isnan(np.sum(s)) for s in intg.soln):
+            comm, rank, root = get_comm_rank_root()
+            isnan = np.array(any(np.isnan(np.sum(s)) for s in intg.soln))
+
+            comm.Allreduce(isnan, isnan, op=mpi.LOR)
```
FreddieWitherden, line 81: Just use regular allreduce.

pyfr/solvers/base/system.py
WillTrojak: I think the reason why I did this was some weirdness that was happening with metis.

pyfr/plugins/nancheck.py
```diff
@@ -9,12 +12,99 @@ class NaNCheckPlugin(BaseSolnPlugin):
...
+        comm, rank, root = get_comm_rank_root()
+
+        rgn_parts = []
+        # For each element type, get nan region centroid and make box
+        for s, etype in zip(intg.soln, self.centroid):
+            i = np.unique(np.argwhere(np.isnan(s))[:,-1])
+
+            if len(i):
+                x = np.mean(self.centroid[etype][i], axis=0)
+                x0, x1 = x - 0.5*self.rgn_len, x + 0.5*self.rgn_len
```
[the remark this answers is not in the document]
WillTrojak: I don't follow, this produces one region per rank and then adds a region to the ini as an easy way to interface with the region code. What were you thinking?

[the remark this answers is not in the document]
FreddieWitherden: Reducing strings is hacky.  Reduce the arguments themselves that would be passed to the region and then have each rank construct the region accordingly.

pyfr/solvers/base/system.py
FreddieWitherden: This can only happen if a rank has no internal connectivity which is not a viable scenario (performance here would be non-existent for example).

pyfr/plugins/base.py
```diff
@@ -196,3 +196,18 @@ def _surf_quad(self, itype, proj, flags=''):
...
+
+class WriterMixin:
+    def __init__(self, intg, *args, **kwargs):
+        super().__init__(intg, *args, **kwargs)
+
+        self._stats = None
+
+    @property
+    def stats(self):
+        if self._stats is None:
```
FreddieWitherden, line 208: cachedprop?

FreddieWitherden:
This functionality should be enabled through a more generic trigger/event-detection approach which can be applied to all plugins.  (So you'd apply the hypothetical event detection or trigger to an ordinary writer plugin where the condition would be "a nan in the solution." This is more flexible as it can also be used for triggering writing when we have high pressures in a location etc.)

PR #385, opened 2023-12-18
Fix the 3rd term in Rot to output 3x3 outer product matrix in plugins/turbulence.py 

amberjia1205:
The 3rd term in Rot in plugins/turbulence.py currently outputs a scalar, which should output an outer product in the form of 3x3 matrix. This PR fixes the equation, and the simulation results are verified locally.


pyfr/plugins/turbulence.py
```diff
@@ -89,7 +89,7 @@ def __init__(self, intg, cfgsect):
         self.shift = shift = np.array(centre)
         self.rot = rot = (np.cos(ran)*np.eye(3) +
                           np.sin(ran)*(np.cross(rax, np.eye(3))) +
-                          (1 - np.cos(ran))*rax @ rax.T)
+                          (1 - np.cos(ran))*np.outer(rax,rax))
```
FreddieWitherden, line 92: Spacing after `,`.

PR #387, opened 2024-01-09
Updated prism alpha-opt points

WillTrojak:


FreddieWitherden:
Keep 1 as 1.0 to avoid spurious changes.

PR #388, opened 2024-01-10
Fix the norm calculation in the residual plugin.

FreddieWitherden:


WillTrojak (approved):
Yep, this looks good and should be more robust.

PR #386, opened 2024-01-04
Rocblas benchmarking and algo picker

WillTrojak:
The heuristics in the rocblas aren't great for some of the gemm kernels used by pyfr. This approach uses a beta function in the rocblas extensions that gets all feasible algorithms, benchmarks them, and returns the best one. For some sizes, the improvement can be quite large.

This has only been tested on rocm 5.7, as I don't have access to 6.0 yet.

PR #389, opened 2024-01-11
cublaslt with algo benchmarking

WillTrojak:
This is similar to #386; it changes the cublas mul provider to cublaslt and utilises their more fine-grained control to benchmark and pick the best gemm algorithm at runtime.


pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,296 @@
...
+                    c_int64, c_size_t, c_uint64, c_void_p, sizeof)
+
+import numpy as np
+
+from pyfr.backends.cuda.driver import CUDADevAlloc
+from pyfr.backends.cuda.provider import CUDAKernel, CUDAKernelProvider
+from pyfr.ctypesutil import LibWrapper
+
+
+# Possible CUBLAS exception types
```
FreddieWitherden, line 12: LT

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,296 @@
...
+    # Constants
+    CUDA_R_32F = 0
+    CUDA_R_64F = 1
+    CUBLASLT_MATMUL_DESC_TRANSA = 3
+    CUBLASLT_MATMUL_DESC_TRANSB = 4
+    CUBLASLT_MATMUL_DESC_TRANSC = 5
+    CUBLASLT_MATMUL_PREF_MAX_WORKSPACE_BYTES = 1
+    CUBLAS_COMPUTE_32F = 68
+    CUBLAS_COMPUTE_64F = 70
+    CUBLAS_OP_N = 0
```
FreddieWitherden, line 60: Remove prefixes (as we know they are local to CUBLASLT).

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,296 @@
...
+        self._create = getattr(lib, f'{self._fn_prefix}Create')
+        self._destory = getattr(lib, f'{self._fn_prefix}Destroy')
+
+    def __del__(self):
+        try:
+            self._destory(self)
+        except AttributeError:
+            pass
+
+    def __int__(self):
```
FreddieWitherden, line 104: I do not think we need this.

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,296 @@
...
+    def __init__(self, cublaslt, handle):
+        self.cublaslt = cublaslt
+        self.h = handle
+
+        lib = self.cublaslt.lib
+        self._create = getattr(lib, f'{self._fn_prefix}Create')
+        self._destory = getattr(lib, f'{self._fn_prefix}Destroy')
+
+    def __del__(self):
+        try:
```
FreddieWitherden, line 99: Just call destroy, no need to check.

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,296 @@
...
+        super().__init__(cublaslt, handle)
+
+        ptr = c_void_p()
+
+        if dtype == np.float64:
+            dtype = cublaslt.lib.CUDA_R_64F
+        else:
+            dtype = cublaslt.lib.CUDA_R_32F
+
+        # PyFR is row-major and cublas is coluum-ajor
```
FreddieWitherden, line 153: Spelling.

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,296 @@
...
+
+    def mul(self, a, b, out, alpha=1.0, beta=0.0):
+        cuda = self.backend.cuda
+        w, h = self.lib, self.handle
+
+        # Ensure the matrices are compatible
+        if a.nrow != out.nrow or a.ncol != b.nrow or b.ncol != out.ncol:
+            raise ValueError('Incompatible matrices for out = a*b')
+
+        # cuBLAS expects inputs to be column-major (or Fortran order in
```
FreddieWitherden, line 192: LT

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,296 @@
...
+        # = (B^T)*(A^T) with a little trickery we can multiply our
+        # row-major matrices directly.
+        m, n, k = b.ncol, a.nrow, a.ncol
+        A, B, C = b, a, out
+
+        # Cache key
+        ckey = (A.dtype, alpha, beta, m, n, k, A.leaddim, B.leaddim, C.leaddim)
+
+        # Size checks
+        if any(sz > 2**31 - 1 for sz in ckey[3:]):
```
FreddieWitherden, line 203: I do not think this is needed (as LT uses 64-bit integers everywhere).

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,296 @@
...
+            desc, dt = self._mul_cache[ckey]
+        except KeyError:
+            # Create matrix layouts
+            a_desc = CUBLASLtMatrixLayout(self, h, A, dtype)
+            b_desc = CUBLASLtMatrixLayout(self, h, B, dtype)
+            c_desc = CUBLASLtMatrixLayout(self, h, C, dtype)
+
+            # Matmul descriptor
+            mm_desc = CUBLASLtMatmulDesc(self, h, ctype, dtype)
+            mm_desc[w.CUBLASLT_MATMUL_DESC_TRANSA] = w.CUBLAS_OP_N
```
FreddieWitherden, line 237: I'd just pass these to the constructor as fixed arguments rather than making it look like a write-only map.

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,296 @@
...
+            mm_desc[w.CUBLASLT_MATMUL_DESC_TRANSB] = w.CUBLAS_OP_N
+            mm_desc[w.CUBLASLT_MATMUL_DESC_TRANSC] = w.CUBLAS_OP_N
+
+            # Heuristic preference descriptor
+            pref = CUBLASLtMatmulPreference(self, h)
+            pref[w.CUBLASLT_MATMUL_PREF_MAX_WORKSPACE_BYTES] = 1024 * 1024 * 4
```
FreddieWitherden, line 243: 32 MiB?

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,296 @@
...
+
+            # Save a copy of the contents of the output matrix
+            out_np = getattr(out, 'parent', out).get()
+
+            # Benchmark the kernel
+            best_kern = None
+            for heur in heurs[:nreturn.value]:
+                # Allocate gemm workspace
+                ws_size = heur.workspaceSize
+                ws = CUDADevAlloc(cuda, ws_size) if ws_size else None
```
FreddieWitherden, line 261: Allocate whatever the max workspace is up front in the constructor and pass this.

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,296 @@
...
+                desc = GEMMDesc(
+                    mm_desc, a_desc, b_desc, c_desc, heur.algo, ws, ws_size
+                )
+
+                dt = self._benchmark(gemm)
+                if best_kern is None or dt < best_kern[-1]:
+                    best_kern = desc, dt
+
+            # Delete the potentially large hueristics array
+            del heurs
```
FreddieWitherden, line 272: This will get cleaned up on its own.

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,296 @@
...
+
+            # Restore the output matrix
+            getattr(out, 'parent', out).set(out_np)
+
+            # Update the cache
+            desc, dt = best_kern
+            self._mul_cache[ckey] = best_kern
+
+        class MulKernel(CUDAKernel):
```
FreddieWitherden, line 281: Need to allocate workspace (the exact amount) around here and then pass it to `MulKernel` to prevent memory leaks.  Current approach has race conditions if two kernels (such as two M0 instances) both execute at the same time (which will happen when we do gradient interpolation).  Given the sizes both will hit in the cache and so they will have the same workspace.

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,296 @@
...
+        ptr = c_void_p()
+
+        if dtype == np.float64:
+            dtype = cublaslt.lib.CUDA_R_64F
+        else:
+            dtype = cublaslt.lib.CUDA_R_32F
+
+        # PyFR is row-major and cublas is coluum-ajor
+        self._create(ptr, dtype, m.ncol, m.nrow, m.leaddim)
+        self._as_parameter_ = ptr.value
```
FreddieWitherden, line 155: Can remove once base class is modified.

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,296 @@
...
+    def __setitem__(self, key, value):
+        value_ct = c_int(value)
+        self._set(self, key, byref(value_ct), sizeof(value_ct))
+
+
+class CUBLASLtMatmulPreference(_CUBLASLtBase):
+    _fn_prefix = 'cublasLtMatmulPreference'
+
+    def __init__(self, cublaslt, handle):
+        super().__init__(cublaslt, handle)
```
FreddieWitherden, line 128: Call after create (will need to add cublaslt as an argument to create).

pyfr/backends/cuda/cublaslt.py
WillTrojak: Yep, they stylise as CUBLASLt, so that's what I've done

FreddieWitherden: Comment says CUBLAS though

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,296 @@
...
+         c_void_p, c_size_t),
+    ]
+
+
+class _CUBLASLtBase:
+    _fn_prefix = None
+
+    def __init__(self, cublaslt, handle):
+        self.cublaslt = cublaslt
+        self.h = handle
```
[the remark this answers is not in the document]
WillTrojak: Actually the handle isn't needed at all

pyfr/backends/cuda/cublaslt.py
WillTrojak: The reccomend 32MiB for H100 and 4MiB for older. I could filter based on compute capability?

FreddieWitherden: Yes, I believe we expose the compute capability via `cuda.compute_capability()`.

pyfr/backends/cuda/cublaslt.py
WillTrojak: The heuristic returns the workspace size needed, but sometimes it's zeros and this handles that.

FreddieWitherden: That's okay, but for benchmarking if you've already pre-allocated the maximum possible size then run with it.  Avoids lots of little memory allocations and deallocations.  Once you know the best kernel you can allocate precisely the right amount of workspace everytime someone requests the kernel.

WillTrojak: Gotcha, will do

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,285 @@
...
+            alpha_ct, beta_ct = c_double(alpha), c_double(beta)
+            ctype = w.COMPUTE_64F
+            dtype = w.R_64F
+        else:
+            alpha_ct, beta_ct = c_float(alpha), c_float(beta)
+            ctype = w.COMPUTE_32F
+            dtype = w.R_32F
+
+        ws_ptr = c_void_p()
+        desc = GEMMDesc(None, None, None, None, None, 0)
```
FreddieWitherden, line 197: Move this down into the key error block

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,285 @@
...
+        self._create(ptr, *args)
+        self._as_parameter_ = ptr.value
+
+        for k, v, t in attr:
+            self._set(self, k, v, t)
+
+    def _create(self, *args):
+        getattr(self.cublaslt.lib, f'{self._fn_prefix}Create')(self, *args)
+
+    def _destory(self):
```
FreddieWitherden, line 107: I think you can inline this

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,285 @@
...
+            out_np = getattr(out, 'parent', out).get()
+
+            # Allocate some temporary worspace for benchmarks
+            ws_ptr = CUDADevAlloc(cuda, w.WORKSPACE_MAX_SIZE)
+
+            # Benchmark the kernel
+            best_kern = None
+            for heur in heurs[:nreturn.value]:
+                # Allocate gemm workspace
+                ws_size = heur.workspaceSize
```
FreddieWitherden, line 246: Not sure this is needed?  Just pass in a larger size and I don't think it will care

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,285 @@
...
+                dt = self._benchmark(gemm)
+                if best_kern is None or dt < best_kern[-1]:
+                    best_kern = desc, dt
+
+            # Restore the output matrix
+            getattr(out, 'parent', out).set(out_np)
+
+            # Update the cache
+            desc, dt = best_kern
+            self._mul_cache[ckey] = best_kern
```
FreddieWitherden, line 261: Can one line these two I think?

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,285 @@
...
+
+        class MulKernel(CUDAKernel):
+            def __init__(self, mats, dt, ws_ptr):
+                super().__init__(mats=mats, dt=dt)
+                self.ws_ptr = ws_ptr
+
+            def add_to_graph(self, graph, deps):
+                stream = cuda.create_stream()
+                ws_ptr = self.ws_ptr
```
FreddieWitherden, line 270: Not sure this will work as you expect (ws_ptr is local and so will not be seen by the GEMM further up).  Either want the obscure `nonlocal` keyword, or just shift the allocation scope.

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,285 @@
...
+            best_kern = None
+            for heur in heurs[:nreturn.value]:
+                # Allocate gemm workspace
+                ws_size = heur.workspaceSize
+
+                desc = GEMMDesc(
+                    mm_desc, a_desc, b_desc, c_desc, heur.algo, ws_size
+                )
+
+                dt = self._benchmark(gemm)
```
FreddieWitherden, line 252: Need a try/except here as NVIDIA says these heuristics can fail for alignment reasons etc.

pyfr/backends/cuda/cublaslt.py
WillTrojak: Where should the allocation go? I really wasn't sure about this bit.

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,285 @@
...
+            getattr(out, 'parent', out).set(out_np)
+
+            # Update the cache
+            desc, dt = best_kern
+            self._mul_cache[ckey] = best_kern
+
+        class MulKernel(CUDAKernel):
+            def __init__(self, mats, dt, ws_ptr):
+                super().__init__(mats=mats, dt=dt)
+                self.ws_ptr = ws_ptr
```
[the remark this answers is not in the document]
WillTrojak: I saw that, but I thought it was ugly

pyfr/backends/cuda/cublaslt.py
FreddieWitherden: Just above the class definition.

pyfr/backends/cuda/cublaslt.py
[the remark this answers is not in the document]
FreddieWitherden: Misc is what libxsmm uses to keep its extra data around. Avoids the need for a constructor/

pyfr/backends/cuda/cublaslt.py
WillTrojak: As we have the number, we might as well give it the right size.

pyfr/backends/cuda/cublaslt.py
WillTrojak: I'm not sure which call you want to catch the error for and what we would do instead. I added the CUBLASLtStatusNotSupported error code, as it seems that that is generally what's returned in that case.

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,271 @@
...
+         c_uint64, c_uint64, c_int64),
+        (c_int, 'cublasLtMatrixLayoutDestroy', c_void_p),
+        (c_int, 'cublasLtMatmulPreferenceCreate', POINTER(c_void_p)),
+        (c_int, 'cublasLtMatmulPreferenceDestroy', c_void_p),
+        (c_int, 'cublasLtMatmulPreferenceSetAttribute', c_void_p, c_int,
+         c_void_p, c_size_t),
+    ]
+
+    # Sets the default max workspace size based on recommendations
+    WORKSPACE_MAX_SIZE = 1024 * 1024 * 32
```
FreddieWitherden, line 89: Probably should not be a library constant as it does not come from the headers.

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,271 @@
...
+        fn(self, k, byref(v_ct), sizeof(v_ct))
+
+    def __del__(self):
+        getattr(self.cublaslt.lib, f'{self._fn_prefix}Destroy')(self)
+
+
+class CUBLASLtMatmulDesc(_CUBLASLtBase):
+    _fn_prefix = 'cublasLtMatmulDesc'
+
+    def __init__(self, cublaslt, ctype, dtype, attr):
```
FreddieWitherden, line 118: What does this do?

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,271 @@
...
+        if a.dtype == np.float64:
+            alpha_ct, beta_ct = c_double(alpha), c_double(beta)
+            ctype = w.COMPUTE_64F
+            dtype = w.R_64F
+        else:
+            alpha_ct, beta_ct = c_float(alpha), c_float(beta)
+            ctype = w.COMPUTE_32F
+            dtype = w.R_32F
+
+        ws_ptr = c_void_p()
```
FreddieWitherden, line 191: Not sure we need to allocate this

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,271 @@
...
+                h, desc.mm, byref(alpha_ct), A, desc.a, B,
+                desc.b, byref(beta_ct), C, desc.c, C, desc.c,
+                byref(desc.algo), ws_ptr, desc.ws_size, stream
+            )
+
+        # Obtain the performance of the kernel
+        try:
+            desc, dt = self._mul_cache[ckey]
+        except KeyError:
+            desc = GEMMDesc(None, None, None, None, None, 0)
```
FreddieWitherden, line 205: What does this do?

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,271 @@
...
+class CUBLASLtStatusNotSupported(CUBLASLtError): pass
+
+
+class CUBLASLtMatmulAlgo(Structure):
+    _fields_ = [('data', c_uint64 * 8)]
+
+
+class CUBLASLtMatmulHeuristicResult(Structure):
+    _fields_ = [('algo', CUBLASLtMatmulAlgo),
+                ('workspaceSize', c_size_t),
```
FreddieWitherden, line 30: workspace_size

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,271 @@
...
+                ('state', c_int),
+                ('wavesCount', c_float),
```
FreddieWitherden, line 32: waves_count

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,271 @@
...
+                    best_kern = desc, dt
+
+            # Restore the output matrix
+            getattr(out, 'parent', out).set(out_np)
+
+            # Update the cache
+            self._mul_cache[ckey] = desc, dt = best_kern
+
+        ws_ptr = CUDADevAlloc(cuda, desc.ws_size)
+        class MulKernel(CUDAKernel):
```
FreddieWitherden, line 256: Add newline

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,271 @@
...
+                gemm(stream)
+                gnode = stream.end_capture()
+
+                # Embed this graph in our main graph
+                return graph.graph.add_graph(gnode, deps)
+
+            def run(self, stream):
+                gemm(stream)
+
+        return MulKernel(mats=[a, b, out], dt=dt)
```
FreddieWitherden, line 271: Need to pass misc=[ws_ptr]

pyfr/backends/cuda/cublaslt.py
FreddieWitherden: Yes, we need to catch the error and continue.  Since some kernels have crazy alignment requirements that our matrices might not meet.  So we need to ignore these kernels and keep going.  Only throw if we can not find anything that works.

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,262 @@
...
+
+
+class CUBLASLtMatmulPreference(_CUBLASLtBase):
+    _fn_prefix = 'cublasLtMatmulPreference'
+
+
+class CUBLASLtMatrixLayout(_CUBLASLtBase):
+    _fn_prefix = 'cublasLtMatrixLayout'
+
+    def __init__(self, cublaslt, m, dtype):
```
FreddieWitherden, line 126: Probably easier not to wrap the constructor.

pyfr/backends/cuda/cublaslt.py
WillTrojak: Ok, I added a try/except to the benchmark call, as that is where the error will get picked up. If the heuristic call can't find any kernels it will throw an error anyway, and---without falling back to cublas---I don't think there is anything else we can do.

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,279 @@
...
+    _fields_ = [('data', c_uint64 * 8)]
+
+
+class CUBLASLtMatmulHeuristicResult(Structure):
+    _fields_ = [('algo', CUBLASLtMatmulAlgo),
+                ('workspace_size', c_size_t),
+                ('state', c_int),
+                ('waves_count', c_float),
+                ('reserved', c_int * 4),
+               ]
```
FreddieWitherden, line 34: Odd indentation.

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,279 @@
...
+        self.lib.cublasLtCreate(self.handle)
+
+        # Maximum number of algorithms to test
+        self.nkerns = backend.cfg.getint('backend-cuda', 'cublas-nkerns', 512)
+
+        # GEMM cache
+        self._mul_cache = {}
+
+        # Sets the default max workspace size based on recommendations
+        WORKSPACE_MAX_SIZE = 1024 * 1024 * 32
```
FreddieWitherden, line 164: Not sure what this does

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,279 @@
...
+                h, desc.mm, byref(alpha_ct), A, desc.a, B,
+                desc.b, byref(beta_ct), C, desc.c, C, desc.c,
+                byref(desc.algo), ws_ptr, desc.ws_size, stream
+            )
+
+        # Obtain the performance of the kernel
+        try:
+            desc, dt = self._mul_cache[ckey]
+        except KeyError:
+
```
FreddieWitherden, line 210: Remove newline

pyfr/backends/cuda/driver.py
```diff
@@ -220,7 +220,8 @@ class CUDADevAlloc(_CUDABase):
 
     def __init__(self, cuda, nbytes):
         ptr = c_void_p()
-        cuda.lib.cuMemAlloc(ptr, nbytes)
+        if nbytes:
```
FreddieWitherden, line 223: This is weird and non-standard.  The low level wrappers should do what CUDA does here if nbytes of 0 is passed.

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,279 @@
...
+    def __init__(self, cublaslt, ptr, attr=()):
+        self.cublaslt = cublaslt
+        self.attr = attr
+        self._as_parameter_ = ptr.value
+
+        for k, v, t in attr:
+            self._set(k, v, t)
+
+    def _set(self, k, v, vtype):
```
FreddieWitherden, line 100: Inline as only ever called above

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,279 @@
...
+         c_void_p, c_size_t),
+    ]
+
+
+class _CUBLASLtBase:
+    _fn_prefix = None
+
+    def __init__(self, cublaslt, ptr, attr=()):
+        self.cublaslt = cublaslt
+        self.attr = attr
```
FreddieWitherden, line 94: No need to set this

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,279 @@
...
+        for k, v, t in attr:
+            self._set(k, v, t)
+
+    def _set(self, k, v, vtype):
+        fn = getattr(self.cublaslt.lib, f'{self._fn_prefix}SetAttribute')
+        v_ct = vtype(v)
+        fn(self, k, byref(v_ct), sizeof(v_ct))
+
+    def __del__(self):
+        getattr(self.cublaslt.lib, f'{self._fn_prefix}Destroy')(self)
```
FreddieWitherden, line 106: Looking again I think you might need to handle an attribute error here.

pyfr/backends/cuda/cublaslt.py
WillTrojak: Forgot the self

pyfr/backends/cuda/driver.py
WillTrojak: Ok, I'll remove and add a check to cublaslt

pyfr/backends/cuda/cublaslt.py
```diff
@@ -0,0 +1,281 @@
...
+        self.lib.cublasLtCreate(self.handle)
+
+        # Maximum number of algorithms to test
+        self.nkerns = backend.cfg.getint('backend-cuda', 'cublas-nkerns', 512)
+
+        # GEMM cache
+        self._mul_cache = {}
+
+        # Sets the default max workspace size based on recommendations
+        self.WORKSPACE_MAX_SIZE = 1024 * 1024 * 32
```
FreddieWitherden, line 164: Make a class constant (line 146).  Also can do 32*1024**2.

PR #398, opened 2024-02-27
Fixed mesh uuid in partitioner

WillTrojak:


pyfr/partitioners/base.py
```diff
@@ -422,7 +422,8 @@ def partition(self, mesh, progress=NullProgressSequence):
                     emap[k] = eleglmap[etype, eidx]
 
         # Generate a new UUID for the mesh
-        newmesh['mesh_uuid'] = newuuid = str(UUID(digest(curruuid, rnum)[:32]))
+        newuuid = np.array(str(UUID(digest(mesh)[:32])), dtype='S')
```
FreddieWitherden, line 425: Not sure this change is needed?

WillTrojak: I thought that it was best if, throughout the code, the UUID was written in the same way, even if we support the edge case of it just being a string.

FreddieWitherden: Is encode() needed?  As we have it in the file writer when writing strings.

FreddieWitherden: Just looking, the str() call is not needed.

PR #401, opened 2024-02-28
Updated Quick Start guides.

pv101:


FreddieWitherden:
Why is the METIS stuff being removed?

pv101:
Because brew and apt etc have 5.1 and we need 5.2.

FreddieWitherden:
> Because brew and apt etc have 5.1 and we need 5.2.

Does SCOTCH work as a replacement?

pv101:
apt has Scotch 6.x which is not new enough. Brew does have 7.x but when I test it I seem to run into this issue that you raised: https://trac.macports.org/ticket/67077 - do you know of a resolution to this? I do export PYFR_SCOTCH_LIBRARY_PATH=/opt/homebrew/lib/libscotch.dylib but then get OSError: dlopen(/opt/homebrew/lib/libscotch.dylib, 0x0006): symbol not found in flat namespace '_SCOTCH_errorPrint'

PR #407, opened 2024-03-15
Update entropy filter inner solver and functional. 

tarikdzanic:
Some tests I've ran on subsonic/supersonic flows indicate that: 1) bisection is significantly more robust than Illinois for the root-bracketing, more than making up for the small efficiency hit; 2) using the numerical entropy instead of the physical entropy (even for Navier-Stokes) seems to be more robust and accurate. This PR adds these changes which also makes the mainline branch more similar to the implementation in the JCP paper. Combined, these changes make the solver more robust around discontinuities (fewer oscillations) and less dissipative around turbulent flow features.


pyfr/solvers/euler/kernels/entropy.mako
```diff
@@ -9,5 +9,9 @@
     p = ${c['gamma'] - 1}*(E - 0.5*rcpd*(${pyfr.dot('u[{i}]', i=(1, ndims + 1))}));
 
     // Compute specific physical entropy
+    % if entropy_func == 'numerical':
+    e = (d > 0 && p > 0) ? d*log(p*pow(rcpd, ${c['gamma']})) : ${fpdtype_max};
```
FreddieWitherden, line 13: This will simplify and avoid the pow.

tarikdzanic: Updated

PR #408, opened 2024-03-27
New File Format

FreddieWitherden:


WillTrojak:
I need to do a proper review, but something that immediately jumps out is we could do with a version flag. Not quite sure how you wold want to handle this without a formal definition of the standard.

PR #410, opened 2024-04-01
fix error: local variable 'std_acc' referenced before assignment

amberjia1205:
When do pyfr tavg merge command, the current latest develop encounters error: local variable 'std_acc' referenced before assignment".  This PR is to fix that error by initializing std_acc


FreddieWitherden:
Thanks for spotting this.  It looks like I did not correctly backport my fix from #408.  It should now be working in develop.

FreddieWitherden:
I believe this is fixed.

PR #412, opened 2024-04-11
Bug fix in pressure/temperature BCs.

tarikdzanic:
Fixes a minor robustness issue for computing `sqrt(udotu)` in pressure/temperature BCs where zero velocity states (such as at startup) can cause roundoff issues due to `pt = p`.


FreddieWitherden:
Looks good.  To make the line breaking cleaner it might be better to have a
```
udotu = fmax(0, udotu);
```
on its own line.

PR #417, opened 2024-05-10
Bug fix in entropy filter functional choice for BCs.

tarikdzanic:


FreddieWitherden:
Can you confirm when the type is equal to physical that this gives binary identical results to what we had in v2?

tarikdzanic:
There was also a change in the EF inner solver between v2 and v2.0.1 so I don't think they'd be exactly the same.

FreddieWitherden:
Can you check with the inner solve changes temporarily reverted?

nnnunnn:
As another data point, I'm using the new entropy filter to try to reproduce this: 

[Transonic 2D flow around Onera OAT15a airfoil at Mach 0.73](https://www.youtube.com/watch?v=yM2rT3teakE)

Using a 3rd order curvilinear quad mesh plus the old [**physical**] + [**Illinois**] entropy options, I can get a good match. However, using the new defaults [**numerical**] + [**bisection**], I see a significant difference in the shock fronts. This is with fp32 + cuda.

But of more interest is a clear difference (on the non-slip wall boundaries) between quad and tri meshes.

As I understand it, the current entropy filter requires GLL flux-pts for lines, GLL soln-pts for quads, and alpha-opt soln-pts for tris. To help explore, I'm running a comparison for a low-speed, low-Reynolds version using quad vs triangle meshes at curvilinear orders {1,2,3}. 

Given the success with quad mesh + GLL nodes, but apparent errors with tri mesh + alpha-opt nodes, could there be some issue when interpolating from tri nodes to alpha-opt [soln-pts]? I get much the same issue for tri meshes using either linear nodes (p1) or curvilinear (p2, p3).

tarikdzanic:
@FreddieWitherden Results appear identical

tarikdzanic:
@nnnunnn This is very helpful! Could you share any images? Are the differences with this bugfix implemented (I force-pushed a change a little bit ago, if your ini file didn't explicitly say e-func = numerical then the default didn't work properly for BCs)? If not, I could imagine the BC bug (which doesn't appear in pre v2.0.1 code) could impact the shock fronts. Did you happen to compare physical vs. numerical with both using bisection? 

On the other hand, the minimum principle on the numerical entropy functional is not necessarily mathematically well-posed for Navier-Stokes (whereas the physical entropy likely is), only for Euler. I still found results for NS to be more well-behaved with the numerical entropy however. I wouldn't think it to be likely that it would cause a drastically different prediction in the shock fronts though. 

The quad/tri issue is more interesting. There's nothing fundamentally different in the flux point choice there, they're both using collocated flux/solution nodes so there's no "real" interpolation happening for either. I can imagine shock capturing on quads being more accurate than tris, but not to the point where they are drastically different flow fields. Could you also share images for this? One possible test to rule out an issue is to try using quads that are very skewed to see if the fact that the tri mesh isn't boundary layer aligned is affecting the flow field predictions.

FreddieWitherden:
I am going to merge the PR as it does appear to resolve the issue.  However, I think some more testing is needed to investigate the secondary issue @nnnunnn raised regarding Illinois vs Bisection.  So I will hold off making a point release until we have a conclusion here.

nnnunnn:
I've run a set of sims for the 2D transonic airfoil case, comparing **rsolver** {exact, hllc}, **e-func** {numerical, physical}, **iter mode** {Illinois, bisection}, **precision** {fp32, fp64}, and **element** {quad, tri}.

For this case, it looks like the issue only appears when I switch the entropy filter inner solver to use bisection mode instead of the original Illinois algorithm. First thought was this glitch might be due to setting the viscosity too low (mu=5e-7) for my single precision RTX 3090 Ti. But then found the same issue with double precision.

Within the next few hours I'll upload to a google folder a gmsh .geo file, the quad mesh, the input file, plus pdf and movie results.

nnnunnn:
Here's a link to a google folder with various files:

https://drive.google.com/drive/folders/1nVH9E6pGhqqi-speZKf6gkfnhFhirtyc?usp=sharing

See Entropy_Check.pdf for a comparison of results at time t=2.
Entropy_Movie.mp4 shows the same results, but includes a movie of a run to t=15.
There's also the input file, a zipped quad mesh and the generating gmsh .geo file.

Over the next few days I'll see if I can spot any likely causes of that issue with the alpha-opt tri nodes. 
Seems to be some anomaly on the triangle edges along the no-slip (wall) wing boundary.

Nigel

tarikdzanic:
@nnnunnn Thank you for this! After running some tests, I think this can be solved by reducing the iteration count of bisection. Can you try with `niters = 2` in the entropy-filter options? That fixed the issues on my end.

To my understanding, what's going on is that the Illinois method has trouble converging the bracket interval (likely due to the fact that the functional isn't really well-behaved so the linearization in the method doesn't work well), which ends up with a more dissipative filter since the bracketing approach chooses the stronger (bounds-preserving) filter strength at the end. Bisection doesn't really have that convergence problem, so with enough iterations it'll find a less dissipative filter strength. Reducing the iteration count increases the dissipation (but only in the elements which are flagged by the entropy criterion) which should increase robustness. 

At the moment, I believe we will change the defaults to a lower iteration count to prioritize general robustness (pending some more accuracy tests on different cases). As a longer term goal, I think we want to try to isolate steady shocks vs. unsteady shocks and treat them slightly differently. For steady shocks, the filter allows a small tolerance in the entropy violation, which generally isn't a problem if the shock stays in an element for a short period of time. However, if the shock is in the same element/group of elements for thousands/millions of time steps, the compounding tolerance basically allows for some non-physical shock behavior which you saw in the results. This can be mitigated by a more dissipative filter (i.e., lower iteration count). For unsteady shocks (and any elements without shocks spuriously flagged), we can afford to use a less dissipative filter (i.e., higher iteration count) for more accuracy. This might be achievable by storing the previous step's filter value per element and apply some sort of exponentially weighted moving average to see the temporal behavior of the filter strength.

nnnunnn:
With niters=2, the issue is gone. Nice work!!
Confirming good result with e-func=numerical, { iter=bisection, niters=2 }
Next I'll take a look at that issue with alpha-opt triangles along the no-slip wall.
PS: incoming soon - movies of a transonic test flow over the 2d OAT15a profile.

PR #420, opened 2024-06-13
Fixed dependency grouping bug.

pv101:


FreddieWitherden:
Another option is to add an argument to `group` along the lines of `ignoredeps` which tells group to explicitly ignore certain dependency violations.

FreddieWitherden:
Here is a rough implementation
```
diff --git a/pyfr/backends/base/types.py b/pyfr/backends/base/types.py
index b97b6f28..de90baf3 100644
--- a/pyfr/backends/base/types.py
+++ b/pyfr/backends/base/types.py
@@ -323,13 +323,17 @@ class Graph:
         self.mpi_reqs = []
         self.mpi_req_deps = []
 
-    def add(self, kern, deps=[]):
+    def add(self, kern, deps=[], pdeps=[]):
         if self.committed:
             raise RuntimeError('Can not add nodes to a committed graph')
 
         if kern in self.knodes:
             raise RuntimeError('Can only add a kernel to a graph once')
 
+        # Handle priority-enforcing (false) dependencies
+        if self.needs_pdeps:
+            deps = [*deps, *pdeps]
+
         # Resolve the dependency list
         rdeps = [self.knodes[d] for d in deps]
 
@@ -337,12 +341,12 @@ class Graph:
         self.knodes[kern] = kern.add_to_graph(self, rdeps)
 
         # Note our dependencies
-        self.kdeps[kern] = deps
+        self.kdeps[kern] = list(deps)
         self.depk.update(deps)
 
-    def add_all(self, kerns, deps=[]):
+    def add_all(self, kerns, deps=[], pdeps=[]):
         for k in kerns:
-            self.add(k, deps)
+            self.add(k, deps, pdeps)
 
     def add_mpi_req(self, req, deps=[]):
         if self.committed:
diff --git a/pyfr/backends/cuda/types.py b/pyfr/backends/cuda/types.py
index 85755af3..f887d0b2 100644
--- a/pyfr/backends/cuda/types.py
+++ b/pyfr/backends/cuda/types.py
@@ -78,6 +78,8 @@ class CUDAXchgMatrix(CUDAMatrix, base.XchgMatrix):
 
 
 class CUDAGraph(base.Graph):
+    needs_pdeps = True
+
     def __init__(self, backend):
         super().__init__(backend)
 
diff --git a/pyfr/backends/hip/types.py b/pyfr/backends/hip/types.py
index 289e7398..138411b4 100644
--- a/pyfr/backends/hip/types.py
+++ b/pyfr/backends/hip/types.py
@@ -79,6 +79,8 @@ class HIPXchgMatrix(HIPMatrix, base.XchgMatrix):
 
 
 class HIPGraph(base.Graph):
+    needs_pdeps = False
+
     def commit(self):
         super().commit()
 
diff --git a/pyfr/backends/metal/types.py b/pyfr/backends/metal/types.py
index 30a11b16..8d52472f 100644
--- a/pyfr/backends/metal/types.py
+++ b/pyfr/backends/metal/types.py
@@ -62,6 +62,8 @@ class MetalXchgMatrix(MetalMatrix, base.XchgMatrix): pass
 
 
 class MetalGraph(base.Graph):
+    needs_pdeps = False
+
     def __init__(self, backend):
         super().__init__(backend)
 
diff --git a/pyfr/backends/opencl/types.py b/pyfr/backends/opencl/types.py
index 66ae40db..dbe81abb 100644
--- a/pyfr/backends/opencl/types.py
+++ b/pyfr/backends/opencl/types.py
@@ -78,6 +78,8 @@ class OpenCLXchgMatrix(OpenCLMatrix, base.XchgMatrix):
 
 
 class OpenCLGraph(base.Graph):
+    needs_pdeps = True
+
     def commit(self):
         super().commit()
 
diff --git a/pyfr/backends/openmp/types.py b/pyfr/backends/openmp/types.py
index c142b7de..64fe0920 100644
--- a/pyfr/backends/openmp/types.py
+++ b/pyfr/backends/openmp/types.py
@@ -57,6 +57,8 @@ class OpenMPView(base.View): pass
 
 
 class OpenMPGraph(base.Graph):
+    needs_pdeps = False
+
     def __init__(self, backend):
         super().__init__(backend)
 
diff --git a/pyfr/solvers/baseadvecdiff/system.py b/pyfr/solvers/baseadvecdiff/system.py
index 478f1b5c..384851a0 100644
--- a/pyfr/solvers/baseadvecdiff/system.py
+++ b/pyfr/solvers/baseadvecdiff/system.py
@@ -97,13 +97,13 @@ class BaseAdvectionDiffusionSystem(BaseAdvectionSystem):
 
         # Compute the common normal flux at our internal/boundary interfaces
         g2.add_all(k['iint/comm_flux'],
-                   deps=ideps + k['mpiint/vect_fpts_pack'])
+                   deps=ideps, pdeps=k['mpiint/vect_fpts_pack'])
         g2.add_all(k['bcint/comm_flux'], deps=ideps)
 
         # Interpolate the gradients to the quadrature points
         for l in k['eles/gradcoru_qpts']:
             ldeps = deps(l, 'eles/gradcoru_upts')
-            g2.add(l, deps=ldeps + k['mpiint/vect_fpts_pack'])
+            g2.add(l, deps=ldeps, pdeps=k['mpiint/vect_fpts_pack'])
 
         # Interpolate the solution to the quadrature points
         g2.add_all(k['eles/qptsu'])
```

pv101:
I'll test that, thanks!

pv101:
@FreddieWitherden your 'rough implementation' seems to work perfectly, and the logic of it makes sense to me. Shall I submit a PR for it?

semi-h:
I think the patch fixes the problem, but makes OpenMP backend completely disregard the false dependencies regardless of the grouping specification. In some grouping configurations OpenMP backend can also take advantage of false dependencies to overlap computation/communication I think.

So how about treating the two dependency lists differently here
https://github.com/PyFR/PyFR/blob/ef369454f7cb690c81ad16802f2821a9753e645f/pyfr/backends/base/types.py#L387
and ignore violations if it is due to a false dependency and `needs_pdeps == False`? (Not sure how the graph will handle this if ignored)

And I would expect `needs_pdeps` to be False only in OpenMP backend and True everywhere else, can't Metal and HIP take advantage of false dependencies?

FreddieWitherden:
> I think the patch fixes the problem, but makes OpenMP backend completely disregard the false dependencies regardless of the grouping specification. In some grouping configurations OpenMP backend can also take advantage of false dependencies to overlap computation/communication I think.
> 
> So how about treating the two dependency lists differently here
> 
> https://github.com/PyFR/PyFR/blob/ef369454f7cb690c81ad16802f2821a9753e645f/pyfr/backends/base/types.py#L387
> 
> and ignore violations if it is due to a false dependency and `needs_pdeps == False`? (Not sure how the graph will handle this if ignored)
> 
> And I would expect `needs_pdeps` to be False only in OpenMP backend and True everywhere else, can't Metal and HIP take advantage of false dependencies?

No, HIP, Metal, and OpenMP all execute kernels in the order they are added to the graph and there will never be two graph kernels running at the same time.  As such false dependencies are not needed.  CUDA and OpenCL can run multiple kernels at once and out of sequence (limited only by dependencies).

pv101:
Shall we move over to: https://github.com/PyFR/PyFR/pull/421 ?

PR #422, opened 2024-06-18
Add capability for entropy filtering with open solution points.

tarikdzanic:
This PR extends EF to open solution node sets (e.g., Gauss-Legendre nodes) by evaluating the bounds and filter at the fpts in addition to the upts.

### DIFF
--- a/pyfr/polys.py
+++ b/pyfr/polys.py
@@ -73,6 +73,9 @@ class BasePolyBasis:
     def vdm(self):
         return self.ortho_basis_at(self.pts)
 
+    def vdm_at(self, pts):
+        return self.ortho_basis_at(pts)
+
     def proj_to(self, tobasis):
         if tobasis.order > self.order:
             return self.nodal_basis_at(tobasis.pts)
--- a/pyfr/polys.py
+++ b/pyfr/polys.py
@@ -1,5 +1,4 @@
 from functools import cached_property
-
 import numpy as np
 
 from pyfr.nputil import clean
@@ -71,7 +70,7 @@ class BasePolyBasis:
 
     @cached_property
     def vdm(self):
-        return self.ortho_basis_at(self.pts)
+        return self.vdm_at(self.pts)
 
     def vdm_at(self, pts):
         return self.ortho_basis_at(pts)
--- a/pyfr/polys.py
+++ b/pyfr/polys.py
@@ -1,4 +1,5 @@
 from functools import cached_property
+
 import numpy as np
 
 from pyfr.nputil import clean
--- a/pyfr/solvers/baseadvec/elements.py
+++ b/pyfr/solvers/baseadvec/elements.py
@@ -1,6 +1,7 @@
 from pyfr.backends.base import NullKernel
 from pyfr.solvers.base import BaseElements
 
+import numpy as np
 
 class BaseAdvectionElements(BaseElements):
     def __init__(self, *kargs, **kwargs):
@@ -148,12 +149,23 @@ class BaseAdvectionElements(BaseElements):
             # Allocate one minimum entropy value per interface
             self.nfaces = len(self.nfacefpts)
             ext = nonce + 'entmin_int'
+            # Set values to -inf for pre-proc filter to enforce positivity
+            entmin_int = np.full((self.nfaces, self.neles),
+                                 -self._be.fpdtype_max)
             self.entmin_int = self._be.matrix((self.nfaces, self.neles),
-                                              tags=tags, extent=ext)
+                                              tags=tags, extent=ext,
+                                              initval=entmin_int)
 
             # Setup nodal/modal operator matrices
-            self.vdm = self._be.const_matrix(self.basis.ubasis.vdm.T)
             self.invvdm = self._be.const_matrix(self.basis.ubasis.invvdm.T)
+            if self.basis.fpts_in_upts:
+                self.vdm_ef = self._be.const_matrix(self.basis.ubasis.vdm.T)
+                self.m0 = None
+            else:
+                vdmu = self.basis.ubasis.vdm.T
+                vdmf = self.basis.ubasis.vdm_at(self.basis.fpts).T
+                self.vdm_ef = self._be.const_matrix(np.vstack((vdmu, vdmf)))
+                self.m0 = self._be.const_matrix(self.basis.m0)
         else:
             self.entmin_int = None
 
--- a/pyfr/solvers/baseadvec/elements.py
+++ b/pyfr/solvers/baseadvec/elements.py
@@ -149,6 +149,7 @@ class BaseAdvectionElements(BaseElements):
             # Allocate one minimum entropy value per interface
             self.nfaces = len(self.nfacefpts)
             ext = nonce + 'entmin_int'
+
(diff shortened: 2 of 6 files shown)


tarikdzanic:
Side note, but I think the method we use of computing `exp(-z*p^2)` actually does`exp(-2*z*p)` instead.

FreddieWitherden:
> Side note, but I think the method we use of computing `exp(-z*p^2)` actually does`exp(-2*z*p)` instead.

Would you be able to investigate if this makes a difference?  It should be easy enough to fix if needed with some extra multiplications.

pyfr/polys.py
```diff
@@ -73,6 +73,9 @@ def jac_nodal_basis_at(self, epts):
     def vdm(self):
         return self.ortho_basis_at(self.pts)
```
FreddieWitherden, line 74: Re implement in terms of vdm_at?

pyfr/solvers/baseadvec/elements.py
```diff
@@ -1,6 +1,7 @@
 from pyfr.backends.base import NullKernel
 from pyfr.solvers.base import BaseElements
 
+import numpy as np
```
FreddieWitherden, line 4: Move import up

FreddieWitherden:
Given this seems to be useful I am inclined to merge.

tarikdzanic:
Let me run some examples and edge-cases to double check before we merge

tarikdzanic:
I think this is ready to merge, I'll update the examples afterwards with L2 projection and open solution points.

pyfr/polys.py
```diff
@@ -1,5 +1,4 @@
 from functools import cached_property
-
```
FreddieWitherden, line 2: Keep newline as functools is part of Python's stdlib but NumPy isn't.

pyfr/solvers/euler/kernels/entropyfilter.mako
```diff
@@ -17,6 +17,19 @@
         ${pyfr.expand('compute_entropy', 'ui', 'd', 'p', 'e')};
         dmin = fmin(dmin, d); pmin = fmin(pmin, p); emin = fmin(emin, e);
     }
+
+    % if not fpts_in_upts:
+    fpdtype_t uf[${nvars}];
+    for (int fidx = 0; fidx < ${nfpts}; fidx++)
+    {
+        for (int vidx = 0; vidx < ${nvars}; vidx++)
```
FreddieWitherden, line 25: I'd unroll this using Mako as nvars is (BGK excluded!) typically small.

FreddieWitherden:
Are there any tricks we can play here by going from modal => fpts in a single step?

tarikdzanic:
Technically yes, but that matrix would be the same size as m0 but typically always dense whereas m0 can be sparse (such as for tensor product elements).

FreddieWitherden:
> Technically yes, but that matrix would be the same size as m0 but typically always dense whereas m0 can be sparse (such as for tensor product elements).

The sparsity doesn't matter though as we don't take advantage of it in the current code.  But it may be advantageous when filtering to include these points (so if one of them does not work out then go back to the next iteration of the filter loop).

tarikdzanic:
This m0 matmul operation is only called once at the beginning (to compute the discrete minima across the upts/fpts). The actual filter loop operates off the VDM form even at the fpts.

pyfr/solvers/euler/kernels/entropyfilter.mako
```diff
@@ -73,8 +86,9 @@
 <%pyfr:kernel name='entropyfilter' ndim='1'
               u='inout fpdtype_t[${str(nupts)}][${str(nvars)}]'
               entmin_int='inout fpdtype_t[${str(nfaces)}]'
-              vdm='in broadcast fpdtype_t[${str(nupts)}][${str(nupts)}]'
-              invvdm='in broadcast fpdtype_t[${str(nupts)}][${str(nupts)}]'>
+              vdm='in broadcast fpdtype_t[${str(nefpts)}][${str(nupts)}]'
```
FreddieWitherden, line 89: Where do we define `nefpts`?

tarikdzanic: Forgot to push that file

pyfr/solvers/euler/kernels/entropyfilter.mako
```diff
@@ -17,6 +17,19 @@
         ${pyfr.expand('compute_entropy', 'ui', 'd', 'p', 'e')};
         dmin = fmin(dmin, d); pmin = fmin(pmin, p); emin = fmin(emin, e);
     }
+
+    % if not fpts_in_upts:
+    fpdtype_t uf[${nvars}];
+    for (int fidx = 0; fidx < ${nfpts}; fidx++)
+    {
+        % for vidx in range(nvars):
+        uf[${vidx}] = ${pyfr.dot('m0[fidx][{k}]', f'u[{{k}}][{vidx}]', k=nupts)};
```
FreddieWitherden, line 26: Depends an implicit definition of m0.  No great solutions here, though.

tarikdzanic: Yeah, we do the same with the vdm matmul as well.

FreddieWitherden: vdm we can fix easily enough with an extern

pyfr/solvers/euler/elements.py
```diff
@@ -96,29 +96,25 @@ def set_backend(self, *args, **kwargs):
             )
 
             # Template arguments
+            fpts_in_upts = self.basis.fpts_in_upts
+            self.nefpts = self.nupts if fpts_in_upts else self.nupts + self.nfpts
             eftplargs = {
                 'ndims': self.ndims,
```
FreddieWitherden, line 102: I think it might be worth putting two entries on each line.

pyfr/solvers/baseadvec/elements.py
```diff
@@ -148,12 +149,23 @@ def modal_filter(uin):
             # Allocate one minimum entropy value per interface
             self.nfaces = len(self.nfacefpts)
             ext = nonce + 'entmin_int'
+            # Set values to -inf for pre-proc filter to enforce positivity
```
FreddieWitherden, line 152: Add newline

pyfr/solvers/baseadvec/system.py
```diff
@@ -105,12 +105,14 @@ def deps(dk, *names): return self._kdeps(k, dk, *names)
         g1 = self.backend.graph()
         g1.add_mpi_reqs(m['ent_fpts_recv'])
 
+        g1.add_all(k['eles/entropy_filter'])
```
FreddieWitherden, line 108: Is this always needed?

tarikdzanic: Sort of. For open upts, then yes to ensure that your solution at the fpts is positivity-preserving to avoid an instant NaN. Same with closed upts if you have L2 projection (which doesn't necessarily ensure that the solution is positive after projection). For closed upts without L2 projection, it technically doesn't require it, but in that case all points are positivity-preserving (assuming that your ICs are positive) and then it just skips the filtering step.

PR #424, opened 2024-07-01
Fix pdeps issue across all backends.

pv101:


pyfr/backends/base/types.py
```diff
@@ -330,12 +330,14 @@ def add(self, kern, deps=[], pdeps=[]):
         if kern in self.knodes:
             raise RuntimeError('Can only add a kernel to a graph once')
 
+        adeps = deps
```
FreddieWitherden, line 333: if/else is probably cleaner than assign + if.

PR #425, opened 2024-07-09
Increased checks in mako params/extern macro definitions.

kaschau:
Checking for a missed comma in params list at macro definition.

If one omits a comma in the params lists, the current checks will not catch them, and the result is a reuse of variable names generated. 

This will catch white space in params list definition, and assert the lengths of a pyfr.expands' param list is equal to the length of a macros param list.

### DIFF
--- a/pyfr/backends/base/makoutil.py
+++ b/pyfr/backends/base/makoutil.py
@@ -98,9 +98,9 @@ def macro(context, name, params, externs=''):
     params = [p.strip() for p in params.split(',')]
     externs = [e.strip() for e in externs.split(',')] if externs else []
 
-    # Ensure no white space in params/extern variables
-    for p in params+externs:
-        if ' ' in p:
+    # Ensure no invalid characters in params/extern variables
+    for p in it.chain(params, externs):
+        if not re.match(r'[A-Za-z_]\w*$', p):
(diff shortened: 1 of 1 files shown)


pyfr/backends/base/makoutil.py
```diff
@@ -98,6 +98,11 @@ def macro(context, name, params, externs=''):
     params = [p.strip() for p in params.split(',')]
     externs = [e.strip() for e in externs.split(',')] if externs else []
 
+    # Ensure no white space in params/extern variables
+    for p in params+externs:
+        if ' ' in p:
```
FreddieWitherden, line 103: I'd be a bit more aggressive and validate it as `if not re.match(r'[A-Za-z_]\w*$', p):` which should catch a whole range of invalid characters for variable names.

pyfr/backends/base/makoutil.py
```diff
@@ -98,6 +98,11 @@ def macro(context, name, params, externs=''):
     params = [p.strip() for p in params.split(',')]
     externs = [e.strip() for e in externs.split(',')] if externs else []
 
+    # Ensure no white space in params/extern variables
+    for p in params+externs:
```
FreddieWitherden, line 102: `it.chain`

PR #428, opened 2024-07-17
Added a framework for implementing SGS models in navstokes solver and added Smagorinsky SGS model option

AlexBogaev:
**Example usage in .ini files:**
```
[constants]
gamma = 1.4
mu = 1.716e-5
Pr = 0.73
rl = 120
pl = 85.7142857143
rr = 1.2
pr = 0.857142857143
Prt = 0.85

[solver]
system = navier-stokes
order = 4
shock-capturing = entropy-filter
sgs-model = smagorinsky

[solver-sgs-model]
Csgs = 0.16
delta = 0.000279
```

**Comments:**
The first approach was to implement the SGS models as a plugin that navstokes could access. However, I needed access to the flux terms to use the pre-computed grad_uin terms and to perform the viscous addition of nu_sgs. I chose to modify navstokes.kernels.flux.mako directly with slight alterations to navstokes.elements.py, navstokes.inters.py, and base.inters.py for the purpose of initialization and passing the necessary model constants to navstokes.kernels.flux.mako.

I did my absolute best to preserve the original structure of the code (both in the elements.py and flux.mako) while implementing. I followed the same modification strategy as that used to implement the sutherland model. I used an if/else statement on the energy flux terms as I was worried about the overhead of computing net-zero flux contributions with the sgs model off. The t_ij flux terms were not put in an if/else statement as the overhead is just an addition of nu_sgs = 0 which should be negligible.

I performed testing on the viscous-shock-tube case at various mesh resolutions, with various Csgs and delta values. It all seems to be numerically stable. The sgs model is compatible with the sutherland model and I have performed testing with both enabled. Please note that the sgs model requires the addition of a turbulent Prandtl number 'Prt' under the [constants] section of the .ini file.

I hope this addition benifits the PyFR community! If approved, I will add vreman as an SGS model. Additionally, I hope to add SGS modeling capabilities to acnavstokes.

Many thanks!


FreddieWitherden:
The framework looks good.  However, we would probably not merge until there is a use case which benefits from the presence of an SGS model.  (So where a simulation is easier/cheaper with an SGS model than the equivalent simulation without perhaps on a finer grid.)

FreddieWitherden:
Is this still being worked on?

AlexBogaev:
I have taken a short break to focus on work. In short, the plan is yes. I would still like to work on and maintain a few SGS models for implementation in ac-navier-stokes and navier-stokes.

Although you mentioned the possibility of using finer grids, the purpose of SGS models is achieve comparable or better results at lower computational cost (assuming the grid scale is much larger than the Kolmogorov scale). This is particularly relevant for large-scale simulations where grid refinement might be prohibitively expensive. 

However, what are your opinions and/or preferences for SGS models in PyFR?  Do you have an objection to the inclusion of SGS models in PyFR? What about wall models (in the wall-stress category, not hybrid LES/RANS or DES methods)?

For SGS, I have considered Smagorinsky, dynamic-smagorinsky, vreman, WALE, WALES, and k_sgs transport. Of those considered, vreman is an excellent balance between accuracy and computational cost... smagorinsky has poor near-wall behavior and anisotropy issues. 

I understand the adversion to smagorinsky, but I would appreciate your feedback before continuing. Is there a better way to implement these in ac-navier-stokes and navier-stokes than my above framework?

Many thanks,
Alex

FreddieWitherden:
The current design of the framework looks reasonable.  However, what we need to check is if there is a benefit to an SGS model for a real simulation.  Generally, we only on board functionality that people use.

AlexBogaev:
Hello Freddie,

I think I am going to abandon this pull request... I am hesitant to continue after reading more papers on implicit LES. I believe that with the numerical dissipation in the current method, well resolved PyFR simulations should be able to mimic natural turbulent dissipation in the unresolved sub grid scales.

Additionally, I just haven't been investing the time into code development to meaningfully contribute clean and robust code to the PyFR community.

Thanks,
Alex

FreddieWitherden:
Okay.

Also, @kaschau has been working on some non-reflecting BCs that you might be interested in testing out.

PR #429, opened 2024-07-18
Allow brackets in ini expressions.

kaschau:
Needed for acces to ploc, etc.


FreddieWitherden:
We disallow [] in expressions as it can allow for arbitrary code execution.  For plocs what we normally do is map them to xyz (https://github.com/PyFR/PyFR/blob/4e5f01b1a8cc4beee20e658c9aeb35465b9420e4/pyfr/plugins/source.py#L19).  This is also what we handle gradients as grad_u_x rather than grad_u[0].

PR #430, opened 2024-07-24
FWH Fixes

rishit2307:


pyfr/plugins/fwh.py
```diff
@@ -67,6 +68,11 @@ def __init__(self, intg, cfgsect, suffix=None, *args, **kwargs):
         self.ele_surface, _ = self._surf_region(intg)
         self._init_surf(ele_map)
 
+        # Get boundary type info
+        sname = self.cfg.get(cfgsect, 'surface')
+        bcsect = f'soln-bcs-{sname}'
```
FreddieWitherden, line 73: Does this all fit on one line?

pyfr/plugins/fwh.py
```diff
@@ -67,6 +68,11 @@ def __init__(self, intg, cfgsect, suffix=None, *args, **kwargs):
...
+        self.bctype = self.cfg.get(bcsect, 'type', None)
```
FreddieWitherden, line 74: Is the None needed here?  I think all boundaries have to have a type?

rishit2307: Surface may not be a boundary?

FreddieWitherden: For this the correct check is `if '(' not in sname:` then you can be sure it is a boundary.

pyfr/plugins/fwh.py
```diff
@@ -75,20 +81,18 @@ def _init_surf(self, ele_map):
...
-            rcpdjac = eles.rcpdjac_at_np(qpts)[..., eidxs]
             pnorm = eles.pnorm_at(qpts, norm)[:, eidxs]
 
             # Get ops and components of the surface
             m0 = eles.basis.ubasis.nodal_basis_at(qpts)
-            norm_mag = np.linalg.norm(pnorm, axis=-1, keepdims=True)
-            n = (pnorm / norm_mag).transpose(2, 0, 1).reshape(self.ndims, -1)
-            area = (qwts[:, None] / rcpdjac).reshape((-1))
+            n = pnorm.transpose(2, 0, 1).reshape(self.ndims, -1)
+            area = np.repeat(qwts, len(eidxs))
```
FreddieWitherden, line 95: This seems wasteful; we should rely on broadcasting when we need the area which is much more efficient.

pyfr/plugins/fwh.py
```diff
@@ -110,13 +114,25 @@ def _distances(self, spts):
...
+        r_tilde_vec = (r_star_tilde_vec - self.qinf['M'])*gamma**2
 
         return r_tilde_vec, r_star_inv, r_star_tilde_vec
 
+    def _apply_wall_bc(self, pris):
+        uvw = pris[self._vidx]
+        rho = self.qinf['rho'] if self.incomp else pris[self._ridx]
+
+        # Calculate velocity magnitude
+        gamma = self.consts['gamma']
```
FreddieWitherden, line 129: How does this work if we are incompressible?

pyfr/plugins/fwh.py
```diff
@@ -75,20 +81,18 @@ def _init_surf(self, ele_map):
...
-            rcpdjac = eles.rcpdjac_at_np(qpts)[..., eidxs]
             pnorm = eles.pnorm_at(qpts, norm)[:, eidxs]
 
             # Get ops and components of the surface
             m0 = eles.basis.ubasis.nodal_basis_at(qpts)
-            norm_mag = np.linalg.norm(pnorm, axis=-1, keepdims=True)
-            n = (pnorm / norm_mag).transpose(2, 0, 1).reshape(self.ndims, -1)
-            area = (qwts[:, None] / rcpdjac).reshape((-1))
+            n = pnorm.transpose(2, 0, 1).reshape(self.ndims, -1)
+            area = (qwts[:, None]*np.ones_like(eidxs)).reshape(-1)
```
FreddieWitherden, line 95: This is still wasteful.  We are just repeating an array dozens of time, wasting memory and memory bandwidth.  Just store the weights and broadcast when they are actually used.

pyfr/plugins/fwh.py
```diff
@@ -67,6 +68,11 @@ def __init__(self, intg, cfgsect, suffix=None, *args, **kwargs):
         self.ele_surface, _ = self._surf_region(intg)
         self._init_surf(ele_map)
 
+        # Get boundary type info
+        self.sname = sname = self.cfg.get(cfgsect, 'surface')
```
FreddieWitherden, line 72: I would not store this

pyfr/plugins/fwh.py
```diff
@@ -67,6 +68,11 @@ def __init__(self, intg, cfgsect, suffix=None, *args, **kwargs):
...
+        if '(' not in sname:
+            self.bctype = self.cfg.get(f'soln-bcs-{sname}', 'type')
```
FreddieWitherden, line 74: Add an else block and set bctype to None.

pyfr/plugins/fwh.py
```diff
@@ -7,7 +7,7 @@
 from pyfr.plugins.base import BaseSolnPlugin, SurfaceMixin, init_csv
 
 
-FWHSurfParams = namedtuple('FWHSurfParams', ('eidxs', 'm0', 'norm', 'area',
+FWHSurfParams = namedtuple('FWHSurfParams', ('eidxs', 'm0', 'nda',
```
FreddieWitherden, line 10: Check if the line breaking can be changed.

FreddieWitherden:
As `self.Minf` is now only used at start-up I do not think that we need to make it a member variable but rather can just pass it as an argument to `_distances` and friends.

pyfr/plugins/fwh.py
```diff
@@ -110,13 +113,24 @@ def _distances(self, spts):
...
 
-        r_star_tilde_vec = r_snorm*r_grad_fac
-        r_tilde_vec = (r_snorm - self.qinf['M'])*gamma**2
+        r_star_tilde_vec = r_snorm[..., None]*r_grad_fac
+        r_tilde_vec = (r_star_tilde_vec - self.qinf['M'])*gamma**2
 
         return r_tilde_vec, r_star_inv, r_star_tilde_vec
 
+    def _apply_wall_bc(self, pris):
+        uvw = pris[self._vidx]
```
FreddieWitherden, line 124: Can just move into the line below.

pyfr/plugins/fwh.py
```diff
@@ -110,13 +113,24 @@ def _distances(self, spts):
...
+        r_snorm = r_star_inv*d
 
-        r_star_tilde_vec = r_snorm*r_grad_fac
-        r_tilde_vec = (r_snorm - self.qinf['M'])*gamma**2
+        r_star_tilde_vec = r_snorm[..., None]*r_grad_fac
+        r_tilde_vec = (r_star_tilde_vec - self.qinf['M'])*gamma**2
 
         return r_tilde_vec, r_star_inv, r_star_tilde_vec
 
+    def _apply_wall_bc(self, pris):
```
FreddieWitherden, line 123: Maybe call enforce no slip bc?

PR #432, opened 2024-08-02
preproc expects a bank not array as inupt

kaschau:
preproc expects a bank not array as input.

This issue was causing the `preproc` method to miss relevant kernels with it's `_get_kernel` call as the input was an array, not a bank. 

The result was most of the preprocessing kernels were not committed and a proper flow field was not being generated when the preprocessor was called, i.e. first calls of boundary conditions, boundary entropy filtering, were being computed with all zero `uin` array on the very first call. It was then corrected with the first pass of the rhs graph.

Test cases appear visually identical after change.

### DIFF
--- a/pyfr/integrators/std/base.py
+++ b/pyfr/integrators/std/base.py
@@ -31,7 +31,7 @@ class BaseStdIntegrator(BaseCommon, BaseIntegrator):
 
         # Pre-process solution if necessary
         self.system.preproc(self.tcurr,
-                            self.system.ele_scal_upts(self._idxcurr))
+                            self._idxcurr)
 
         # Global degree of freedom count
         self._gndofs = self._get_gndofs()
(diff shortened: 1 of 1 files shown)


pyfr/integrators/std/base.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, backend, systemcls, rallocs, mesh, initsoln, cfg):
 
         # Pre-process solution if necessary
         self.system.preproc(self.tcurr,
-                            self.system.ele_scal_upts(self._idxcurr))
```
FreddieWitherden, line 34: See if the second argument will fit on the line above.

pyfr/integrators/std/base.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, backend, systemcls, rallocs, mesh, initsoln, cfg):
 
         # Pre-process solution if necessary
```
FreddieWitherden, line 32: Comment is probably misleading?  If necessary but no if.

FreddieWitherden:
Looks good to me but @tarikdzanic will give a proper review sometime towards the end of next week just to check there are no other issues lingering.

PR #433, opened 2024-08-12
Change entropy filter to second-order.

tarikdzanic:


pyfr/solvers/euler/kernels/entropyfilter.mako
```diff
@@ -56,14 +56,16 @@
...
     // Apply filter to local value
-    fpdtype_t v = 1.0;
+    fpdtype_t v = 1.0, v2;
     for (int pidx = 1; pidx < ${order+1}; pidx++)
     {
-        // Utilize exp(-zeta*p**2) = pow(f, p**2)
+        // Utilize exp(-zeta*p**2) = pow(f, p**2) = pow(pow(f, p), p)
         v *= f;
+        v2 = v;
+        for (int zidx = 1; zidx < pidx; zidx++) v2 *= v;
```
FreddieWitherden, line 65: I think we can avoid this.  Given a = exp(-z*p), b = exp(-z*p**2) and c = exp(-z) then exp(-z*(p + 1)**2) = b*a*a*c.  This should give us the ability to compute v2 given its previous value, v1, and f.

tarikdzanic:
Updated

FreddieWitherden:
Can we do a quick check with print statements to be sure that it is generating the right sequence of exponentials?

tarikdzanic:
Confirmed it does

PR #434, opened 2024-08-13
Minor bugfix in preproc graph.

tarikdzanic:
Now that the preproc graphs are running properly, there seems to be a bug with running p0 with entropy filtering enabled (even though it shouldn't do anything) due to the disu kernel not being called in preproc: `KeyError: <pyfr.backends.cuda.gimmik.CUDAGiMMiKKernels.mul.<locals>.MulKernel object at 0x0000020A8999E230>`

### DIFF
--- a/pyfr/solvers/baseadvec/inters.py
+++ b/pyfr/solvers/baseadvec/inters.py
@@ -4,8 +4,15 @@ import math
 from pyfr.solvers.base import BaseInters
 from pyfr.nputil import npeval
 
+class IntersMixin:
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
 
-class BaseAdvectionIntInters(BaseInters):
+        self._ef_enabled = (self.cfg.get('solver', 'shock-capturing') == 
+                            'entropy-filter' and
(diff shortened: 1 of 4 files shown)


FreddieWitherden:
Is the kernel actually needed?  Or can we fix the problem by just removing a dependency in the graph (I think `disu` is a dependency to a kernel which we might want to check).

FreddieWitherden:
We may want to see if we move the duplicated interface logic out to a mixin.

```python
class X:
    @property
    def _ef_enabled(self):
        return self.cfg.getint(...) and self.cfg.getint(...)
```

tarikdzanic:
Added, let me know if you think the mixin class should be renamed.

pyfr/solvers/baseadvec/inters.py
```diff
@@ -4,8 +4,15 @@
 from pyfr.solvers.base import BaseInters
 from pyfr.nputil import npeval
 
+class IntersMixin:
```
FreddieWitherden, line 7: Add newline

FreddieWitherden, line 7: Maybe rename to BaseAdvectionIntersMixin

PR #435, opened 2024-08-14
Add L2 projection option for ICs.

tarikdzanic:


FreddieWitherden:
Does it work/benefit a test case?

Also, assuming it checks out it'll need documentation.

pyfr/solvers/base/elements.py
```diff
@@ -88,8 +91,26 @@ def set_ics_from_cfg(self):
...
+        upts = self.cfg.get(f'solver-elements-{ename}', 'soln-pts')
+        qdeg = (self.cfg.getint('soln-ics', f'quad-deg-{ename}', 0) or 
+                self.cfg.getint('soln-ics', f'quad-deg', 0))
+        # Default to solution points if quad-pts are not specified
+        qpts = (self.cfg.get('soln-ics', f'quad-pts-{ename}', upts) or 
+                self.cfg.get('soln-ics', f'quad-pts', upts))
+        # Get prolongation/projection operators 
+        if qdeg:
+            qrule = get_quadrule(ename, qpts, qdeg=qdeg)
+            m7 = self.basis.ubasis.nodal_basis_at(qrule.pts)
```
FreddieWitherden, line 107: Not sure this is needed. `ploc_at_np` will take a list of points in lieu of a name.

pyfr/solvers/base/elements.py
```diff
@@ -2,7 +2,10 @@
 
 import numpy as np
 
+from pyfr.inifile import NoOptionError
```
FreddieWitherden, line 5: Is this needed?

pyfr/solvers/base/elements.py
```diff
@@ -2,7 +2,10 @@
...
 from pyfr.nputil import npeval, fuzzysort
+from pyfr.quadrules import get_quadrule
+from pyfr.shapes import _proj_l2
```
FreddieWitherden, line 8: Probably worth renaming the function so that it no longer as an '_'.

pyfr/solvers/base/elements.py
```diff
@@ -88,8 +91,26 @@ def set_ics_from_cfg(self):
...
         # Get the physical location of each solution point
-        coords = self.ploc_at_np('upts').swapaxes(0, 1)
-        vars |= dict(zip('xyz', coords))
+        coords = self.ploc_at_np('upts').swapaxes(0,1)
+
+        # See if performing L2 projection
+        ename = self.basis.name
+        upts = self.cfg.get(f'solver-elements-{ename}', 'soln-pts')
+        qdeg = (self.cfg.getint('soln-ics', f'quad-deg-{ename}', 0) or 
+                self.cfg.getint('soln-ics', f'quad-deg', 0))
```
FreddieWitherden, line 100: Not sure the f is needed.

tarikdzanic:
It appears to help noticeably with cases where a shock is initially within an element.

pyfr/solvers/base/elements.py
```diff
@@ -87,8 +89,23 @@ def set_ics_from_cfg(self):
...
-        # Get the physical location of each solution point
-        coords = self.ploc_at_np('upts').swapaxes(0, 1)
+        # See if performing L2 projection
+        ename = self.basis.name
+        upts = self.cfg.get(f'solver-elements-{ename}', 'soln-pts')
+        qdeg = (self.cfg.getint('soln-ics', f'quad-deg-{ename}', 0) or 
+                self.cfg.getint('soln-ics', f'quad-deg', 0))
+        # Default to solution points if quad-pts are not specified
+        qpts = (self.cfg.get('soln-ics', f'quad-pts-{ename}', upts) or 
+                self.cfg.get('soln-ics', f'quad-pts', upts))
```
FreddieWitherden, line 99: I'd force the element type to always be specified.

pyfr/solvers/base/elements.py
```diff
@@ -87,8 +89,23 @@ def set_ics_from_cfg(self):
...
+        # Get the physical location of each interpolation point
```
FreddieWitherden, line 100: Add newline

pyfr/solvers/base/elements.py
```diff
@@ -87,8 +89,23 @@ def set_ics_from_cfg(self):
...
+        if qdeg:
+            qrule = get_quadrule(ename, qpts, qdeg=qdeg)
+            coords = self.ploc_at_np(qrule.pts).swapaxes(0,1)
```
FreddieWitherden, line 103: Can move swapaxes down into the zip.

pyfr/solvers/base/elements.py
```diff
@@ -87,8 +89,23 @@ def set_ics_from_cfg(self):
...
+            # Compute projection operator
+            m8 = proj_l2(qrule, self.basis.ubasis)
+        else:
+            m8 = None
+            coords = self.ploc_at_np('upts').swapaxes(0,1)
         vars |= dict(zip('xyz', coords))
```
FreddieWitherden, line 109: Add newline

pyfr/solvers/base/elements.py
tarikdzanic: Just for the points or for the degree as well?

FreddieWitherden: Element type for sure, and I'd be inclined to go with degree, too but it might be worth seeing what the integrate plugin does here.

tarikdzanic: Integrate uses a single option for all element quad degs, so I guess I'll go with that for consistency.

FreddieWitherden: Sounds good, we can change both if needed in the future.

FreddieWitherden:
We may also want to update the Euler vortex case to make use of this functionality as it is a good example of one which can benefit from better sampling.

PR #438, opened 2024-08-19
Add linearized constraint/filter kernel for EF.

tarikdzanic:
This PR adds the option to use the linearized formulation for the entropy filter from Ching et al. (10.1016/j.jcp.2024.112881), which linearizes the pressure/entropy constraints and uses a linear Zhang/Shu-type limiter. This avoids the root-finding problem for EF at the expense of a more dissipative limiter (which can also sometimes be beneficial for strong shocks/poor grids).

### DIFF
--- a/doc/src/systems/solver-entropy-filter.rst
+++ b/doc/src/systems/solver-entropy-filter.rst
@@ -24,9 +24,9 @@ Parameterises entropy filter for shock capturing with
 
     *int*
 
-6. ``linearise`` --- linearise constraints and filter kernel:
+6. ``formulation`` --- formulation for constraints and filter kernel:
 
-    *bool*
+    ``nonlinear`` | ``linearised``
 
 Example::
 
@@ -36,7 +36,7 @@ Example::
     e-tol = 1e-6
     e-func = physical
     niters = 2
-    linearise = False
+    formulation = nonlinear
 
 Used in the following Examples:
 
--- a/doc/src/systems/solver-entropy-filter.rst
+++ b/doc/src/systems/solver-entropy-filter.rst
@@ -24,7 +24,7 @@ Parameterises entropy filter for shock capturing with
 
     *int*
 
-6. ``linearize`` --- linearize constraints and filter kernel:
+6. ``linearise`` --- linearise constraints and filter kernel:
 
     *bool*
 
@@ -36,7 +36,7 @@ Example::
     e-tol = 1e-6
     e-func = physical
     niters = 2
-    linearize = False
+    linearise = False
 
 Used in the following Examples:
 
--- a/doc/src/systems/solver-entropy-filter.rst
+++ b/doc/src/systems/solver-entropy-filter.rst
@@ -24,6 +24,10 @@ Parameterises entropy filter for shock capturing with
 
     *int*
 
+6. ``linearize`` --- linearize constraints and filter kernel:
+
+    *bool*
+
 Example::
 
     [solver-entropy-filter]
@@ -32,6 +36,7 @@ Example::
     e-tol = 1e-6
     e-func = physical
     niters = 2
+    linearize = False
 
 Used in the following Examples:
 
--- a/pyfr/solvers/baseadvec/elements.py
+++ b/pyfr/solvers/baseadvec/elements.py
@@ -158,8 +158,7 @@ class BaseAdvectionElements(BaseElements):
                                               initval=entmin_int)
 
             # Setup nodal/modal operator matrices
-            linearize = self.cfg.getbool('solver-entropy-filter', 'linearize',
-                                         False)
+            linearize = self.cfg.getbool('solver-entropy-filter', 'linearize')
             if linearize:
                 self.invvdm = self.vdm_ef = None
             else:
--- a/pyfr/solvers/baseadvec/elements.py
+++ b/pyfr/solvers/baseadvec/elements.py
@@ -158,14 +158,20 @@ class BaseAdvectionElements(BaseElements):
                                               initval=entmin_int)
 
             # Setup nodal/modal operator matrices
-            self.invvdm = self._be.const_matrix(self.basis.ubasis.invvdm.T)
-            if self.basis.fpts_in_upts:
-                self.vdm_ef = self._be.const_matrix(self.basis.ubasis.vdm.T)
-                self.m0 = None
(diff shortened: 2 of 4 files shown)


pyfr/solvers/euler/kernels/entropyfilter.mako
```diff
@@ -102,6 +102,56 @@
...
+        // Compute mean quantities
+        fpdtype_t uavg[${nvars}], davg, pavg, eavg;
+        % for vidx in range(nvars):
+        uavg[${vidx}] = ${' + '.join(f'{jx}*u[{j}][{vidx}]'
+                                     for j, jx in enumerate(meanwts) if jx != 0)};
+        % endfor
+
+        ${pyfr.expand('compute_entropy', 'uavg', 'davg', 'pavg', 'eavg')};
+
+        // Apply density limiting
```
FreddieWitherden, line 115: % for v, t in [('d', d_min), ('p', p_min), ('e', 'e_min')]:
...
% endfor

pyfr/solvers/euler/kernels/entropyfilter.mako
```diff
@@ -102,6 +102,32 @@
...
+        % for vidx in range(nvars):
+        uavg[${vidx}] = ${' + '.join(f'{jx}*u[{j}][{vidx}]'
+                                     for j, jx in enumerate(meanwts) if jx != 0)};
+        % endfor
+
+        ${pyfr.expand('compute_entropy', 'uavg', 'davg', 'pavg', 'eavg')};
+
+        // Apply density, pressure, and entropy limiting sequentially
+        fpdtype_t alpha;
+        % for (fvar, bound) in [('d', f'{d_min}'), ('p', f'{p_min}'), ('e', f'entmin - {e_tol}')]:
```
FreddieWitherden, line 117: Can remove some of the f-strings here.

pyfr/solvers/euler/kernels/entropyfilter.mako
```diff
@@ -102,6 +102,32 @@
...
+        if (${f'{fvar}min < {bound}'})
```
FreddieWitherden, line 118: No need for an f string here; can use native Mako ${}.

pyfr/solvers/baseadvec/elements.py
```diff
@@ -158,14 +158,20 @@ def modal_filter(uin):
                                               initval=entmin_int)
 
             # Setup nodal/modal operator matrices
-            self.invvdm = self._be.const_matrix(self.basis.ubasis.invvdm.T)
-            if self.basis.fpts_in_upts:
-                self.vdm_ef = self._be.const_matrix(self.basis.ubasis.vdm.T)
-                self.m0 = None
+            linearize = self.cfg.getbool('solver-entropy-filter', 'linearize',
```
FreddieWitherden, line 161: Check if the default argument is needed here.

pyfr/solvers/euler/kernels/entropyfilter.mako
tarikdzanic: Do you mean something similar to what I just pushed or different?

pyfr/solvers/baseadvec/elements.py
tarikdzanic: I think by default having it off wouldn't be a bad idea (just for user-friendliness), but I can remove it too.

pyfr/solvers/euler/kernels/entropyfilter.mako
FreddieWitherden: `if (${fvar}min < ${bound})`

pyfr/solvers/baseadvec/elements.py
FreddieWitherden: We have the `getbool('solver-entropy-filter', 'linearize', False)` twice in the PR.  Generally, we try and only use the default argument once and rely on the fact that when a default is used there is an implicit set (so future accesses do not need the default to be specified).

doc/src/systems/solver-entropy-filter.rst
```diff
@@ -24,6 +24,10 @@ Parameterises entropy filter for shock capturing with
 
     *int*
 
+6. ``linearize`` --- linearize constraints and filter kernel:
```
FreddieWitherden, line 27: British spelling!

pyfr/solvers/baseadvec/elements.py
```diff
@@ -158,14 +158,19 @@ def modal_filter(uin):
                                               initval=entmin_int)
 
             # Setup nodal/modal operator matrices
-            self.invvdm = self._be.const_matrix(self.basis.ubasis.invvdm.T)
-            if self.basis.fpts_in_upts:
-                self.vdm_ef = self._be.const_matrix(self.basis.ubasis.vdm.T)
-                self.m0 = None
+            linearize = self.cfg.getbool('solver-entropy-filter', 'linearize')
+            if linearize:
```
FreddieWitherden, line 162: Will this fit on a single line?

pyfr/solvers/euler/elements.py
```diff
@@ -98,12 +98,15 @@ def set_backend(self, *args, **kwargs):
             # Template arguments
             fpts_in_upts = self.basis.fpts_in_upts
             self.nefpts = self.nupts if fpts_in_upts else self.nupts + self.nfpts
+            ub = self.basis.ubasis
+            meanwts = ub.invvdm[:,0]/np.sum(ub.invvdm[:,0])
```
FreddieWitherden, line 102: Spacing around , and /

pyfr/solvers/euler/elements.py
```diff
@@ -98,12 +98,15 @@ def set_backend(self, *args, **kwargs):
             # Template arguments
             fpts_in_upts = self.basis.fpts_in_upts
             self.nefpts = self.nupts if fpts_in_upts else self.nupts + self.nfpts
+            ub = self.basis.ubasis
+            meanwts = ub.invvdm[:, 0] / np.sum(ub.invvdm[:, 0])
```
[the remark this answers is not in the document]
tarikdzanic: What syntax are you thinking? `meanwts` here is the vector of quadrature weights that compute the mean. This could also be replaced with just the quadrature weights (normalized to 1) instead of using invvdm, that might be more obvious.

PR #440, opened 2024-09-19
Added support to detect arch/cpu compiler flag

WillTrojak:
On some hardware the `-march` flag is not supported instead `-mcpu` is used. This autodetects it. The cc_option function can also be used for other flags, for example, I think some compilers don't use `-fopenmp`.

### DIFF
--- a/pyfr/backends/openmp/compiler.py
+++ b/pyfr/backends/openmp/compiler.py
@@ -8,7 +8,7 @@ import tempfile
 import uuid
 
 from platformdirs import user_cache_dir
-from pytools.prefork import call_capture_output
+from pytools.prefork import ExecError, call_capture_output
 
 from pyfr.ctypesutil import platform_libname
 from pyfr.util import digest, mv, rm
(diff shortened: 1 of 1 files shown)


pyfr/backends/openmp/compiler.py
```diff
@@ -85,6 +89,19 @@ def cc_cmd(self, srcname, libname):
         # Append any user-provided arguments and return
         return cmd + self.cflags
 
+    def cc_option(self, opts, helps=[]):
+        helps = [f'--help={h}' for h in helps] + ['--help']
```
FreddieWitherden, line 93: Can't we just run `--help` for me this gives everything including if march is supported?

FreddieWitherden:
In general we need `-fopenmp` since many of our kernels call OpenMP runtime functions so we must have support there.

pyfr/backends/openmp/compiler.py
WillTrojak: No, on the some gcc versions you have to pass an additional are to help flag to get the target info. 
This is the case on `gcc (GCC) 11.4.1 20231218` for example.

FreddieWitherden:
Can you confirm this also works for clang?

FreddieWitherden:
Can you see if `-mcpu=native` works (POWER/x86/AARCH64)?

FreddieWitherden:
@WillTrojak Just following up here.

FreddieWitherden:
Just following up here.

FreddieWitherden:
Fixed by #547

PR #441, opened 2024-09-24
Cleanups in mean quadrature weights.

tarikdzanic:


pyfr/shapes.py
```diff
@@ -162,6 +162,11 @@ def upts(self):
         rname = self.cfg.get(f'solver-elements-{self.name}', 'soln-pts')
         return get_quadrule(self.name, rname, self.nupts).pts
 
+    @cached_property
+    def upts_wts(self):
+        rname = self.cfg.get(f'solver-elements-{self.name}', 'soln-pts')
```
FreddieWitherden, line 167: Can we use `_get_qrule` here to make it a one liner?

tarikdzanic: I think `_get_qrule` would return the qpts weights if `quad-pts` or `quad-deg` are specified in the element section which isn't what we want

FreddieWitherden:
I'm still not 100% here as some legitimate solution point sets do not have weights.  For example, the alpha optimised points.  Am I missing something?

tarikdzanic:
I guess they do not have to be well-defined weights in the sense of they're all strictly positive, but all unique point sets should have a set of "weights" such that `w.u = int(u)` if `u` interpolates a polynomial on those point sets. This would just be the first row of the inverse Vandermonde matrix, which may have entries that are zero.

FreddieWitherden:
> I guess they do not have to be well-defined weights in the sense of they're all strictly positive, but all unique point sets should have a set of "weights" such that `w.u = int(u)` if `u` interpolates a polynomial on those point sets. This would just be the first row of the inverse Vandermonde matrix, which may have entries that are zero.

Right, but the current rules in PyFR often do not have these weights in their tabulations.

tarikdzanic:
Ah I see, in that case I'll close this and we can just use the weights from the inverse VDM matrix like before.

PR #442, opened 2024-09-25
Updated invariant set definition

kaschau:
This pull request clarifies the definition of the invariant set with respect to entropy filtering, as well as the distinction between the entropy on a per volume bases, and the specific entropy on a per mass basis (i.e. $\rho s$ versus $s$). The new invariant set is defined as $G=\{ \rho > 0, p > 0, \chi > 0\}$, where $\chi = \rho(s-s_{0})$, and $s_0$ is the local minimum specific entropy. 

This is different from the current definition, which would define $\chi$ as $\chi=\rho s - (\rho s)_{0}$, which is not necessarily a convex function. The definition of $\chi$ above, however, is convex. (There is further confusion in the distinction between the `physical` and `numerical` definitions in PyFR as one is multiplied by density and the other is not).

This clarifies that the entropy computed in the `entropy.mako` kernel is the specific entropy, $s$.

Further, the use of this invariant set greatly reduces the sensitivity to the definition of $s$. Other valid forms of entropy such as a dimensional entropy, or the numerical entropy plus a constant create dramatically different results. With this formulation, the other valid forms of entropy are visually identical in test cases. 

This formulation is consistent with Ching et al. (https://doi.org/10.1016/j.jcp.2024.112881) and Jiang & Liu (https://doi.org/10.1016/j.jcp.2018.03.004)

### DIFF
--- a/pyfr/solvers/euler/kernels/entropy.mako
+++ b/pyfr/solvers/euler/kernels/entropy.mako
@@ -10,7 +10,7 @@
 
     // Compute numerical or specific physical entropy
     % if e_func == 'numerical':
-    e = (d > 0 && p > 0) ? d*(log(p) - ${c['gamma']}*log(d)) : ${fpdtype_max};
+    e = (d > 0 && p > 0) ? log(p) - ${c['gamma']}*log(d) : ${fpdtype_max};
     % elif e_func == 'physical':
     e = (d > 0 && p > 0) ? p*pow(rcpd, ${c['gamma']}) : ${fpdtype_max};
     % endif
--- a/pyfr/solvers/euler/kernels/entropyfilter.mako
+++ b/pyfr/solvers/euler/kernels/entropyfilter.mako
@@ -2,11 +2,12 @@
 <%namespace module='pyfr.backends.base.makoutil' name='pyfr'/>
 <%include file='pyfr.solvers.euler.kernels.entropy'/>
 
-<%pyfr:macro name='get_minima' params='u, m0, dmin, pmin, emin'>
+<%pyfr:macro name='get_minima' params='u, m0, dmin, pmin, emin, Xmin, entmin'>
     fpdtype_t d, p, e;
     fpdtype_t ui[${nvars}];
 
-    dmin = ${fpdtype_max}; pmin = ${fpdtype_max}; emin = ${fpdtype_max};
+    dmin = ${fpdtype_max}; pmin = ${fpdtype_max};
+    emin = ${fpdtype_max}, Xmin = ${fpdtype_max};
 
     for (int i = 0; i < ${nupts}; i++)
     {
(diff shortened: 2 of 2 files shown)


pyfr/solvers/euler/kernels/entropyfilter.mako
```diff
@@ -90,32 +95,36 @@
               vdm='in broadcast fpdtype_t[${str(nefpts)}][${str(nupts)}]'
               invvdm='in broadcast fpdtype_t[${str(nupts)}][${str(nupts)}]'
               m0='in broadcast fpdtype_t[${str(nfpts)}][${str(nupts)}]'>
-    fpdtype_t dmin, pmin, emin;
+
+    // X = rho*(s-s0)
+    fpdtype_t dmin, pmin, emin, Xmin;
```
tarikdzanic, line 100: Is emin still used anymore?

kaschau: emin is used at the end of the filtering operation to store the new local minimum specific entropy through the get_minima expansion, but not in the interim.

pyfr/solvers/euler/kernels/entropyfilter.mako
```diff
@@ -27,7 +29,8 @@
         % endfor
 
         ${pyfr.expand('compute_entropy', 'uf', 'd', 'p', 'e')};
-        dmin = fmin(dmin, d); pmin = fmin(pmin, p); emin = fmin(emin, e);
+        dmin = fmin(dmin, d); pmin = fmin(pmin, p);
+        emin = fmin(emin, e); Xmin = fmin(Xmin, d*(e-entmin));
```
FreddieWitherden, line 33: e - entmin

pyfr/solvers/euler/kernels/entropyfilter.mako
```diff
@@ -90,32 +95,36 @@
...
 
+        fpdtype_t eavg;
         ${pyfr.expand('compute_entropy', 'uavg', 'davg', 'pavg', 'eavg')};
+        Xavg = davg*(eavg - entmin);
 
         // Apply density, pressure, and entropy limiting sequentially
         fpdtype_t alpha;
-        % for (fvar, bound) in [('d', d_min), ('p', p_min), ('e', f'entmin - {e_tol}')]:
-        if (${fvar}min < ${bound}) 
+        % for (fvar, bound) in [('d', d_min), ('p', p_min), ('X', f'-{e_tol}')]:
```
FreddieWitherden, line 126: -e_tol without the f''?

pyfr/solvers/euler/kernels/entropy.mako
```diff
@@ -8,10 +8,11 @@
     // Compute the pressure
     p = ${c['gamma'] - 1}*(E - 0.5*rcpd*(${pyfr.dot('u[{i}]', i=(1, ndims + 1))}));
 
-    // Compute numerical or specific physical entropy
-    % if e_func == 'numerical':
-    e = (d > 0 && p > 0) ? d*(log(p) - ${c['gamma']}*log(d)) : ${fpdtype_max};
-    % elif e_func == 'physical':
+    // Compute entropy
+    // Testing showed the functional s=exp(log(p*r^-g))
```
FreddieWitherden, line 12: I'd remove the comment; ideally it should end up in a paper rather than here.

PR #449, opened 2024-11-04
Ensure non-blocked exchange matrices across all backends

sambitmishra98:
This PR fixes mismatch in exchange matrix shapes between backends using blocked matrices (currently only OpenMP backend) and non-blocked matrices (currently, all other backends).

### DIFF
--- a/pyfr/backends/base/types.py
+++ b/pyfr/backends/base/types.py
@@ -28,7 +28,8 @@ class MatrixBase:
 
             # Alignment requirement for the leading dimension
             ldmod = csubsz if 'align' in self.tags else 1
-            leaddim = csubsz if backend.blocks else ncol - (ncol % -ldmod)
+            blocked = backend.blocks and 'xchg' not in self.tags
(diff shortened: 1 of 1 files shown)


semi-h (approved):

FreddieWitherden:
Can we test this for a few configurations (MPI + OpenMP) to ensure everything works as expected?

sambitmishra98:
I have tested it for TGV case with different mesh sizes and polynomial orders for HIP+OpenMP and HIP+CUDA+OpenMP backends, and ensured that the PR fixes all issues around matrix shapes. I will keep this commit in the background when working with other meshes in the next few days.

FreddieWitherden:
Okay, if no issues are found over the next day or so I am happy to merge but it does require quite a bit of testing.

semi-h:
I wouldn't expect a mismatch in size between buffer arrays if the element numbers in a mesh is a multiple of 8. So with your tests make sure that the local ranks running with OpenMP backend have element numbers that are not a multiple of 8.

PR #450, opened 2024-11-22
Using a topological sort to order kernels on OpenMP backend

TobyFlynn:
This pull request uses a topological sort to create a run list for the kernels on the OpenMP backend. This fixes an issue where dependencies between kernels might not be respected after setting a cache blocking group.

I've tested this on the AV test case I was using previously and on all cases in the PyFR test case repo with 10 MPI ranks on the OpenMP backend - they all look good. I've not touched any of the other backends but let me know if there are any further testing you'd like me to do for this pull request.

### DIFF
--- a/pyfr/backends/openmp/types.py
+++ b/pyfr/backends/openmp/types.py
@@ -1,6 +1,7 @@
 from collections import defaultdict
 from ctypes import c_int
 from functools import cached_property
+from graphlib import TopologicalSorter
 
 import pyfr.backends.base as base
 from pyfr.backends.openmp.provider import OpenMPBlockKernelArgs, OpenMPKRunArgs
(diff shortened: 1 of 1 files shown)


FreddieWitherden:
Worth checking that all of the MPI requests start at a 'good' time (as in they are not all lumped towards the end) as this can hurt scaling.  Further, if we are doing a topological sort to determine the order then the OpemMP backend may need `pdeps`.

pyfr/backends/openmp/types.py
```diff
@@ -55,6 +56,10 @@ class OpenMPXchgMatrix(OpenMPMatrix, base.XchgMatrix): pass
 class OpenMPXchgView(base.XchgView): pass
 class OpenMPView(base.View): pass
 
+# Need a hashable wrapper for a list of kernels for the topological sort
+class OpenMPGroupNode:
```
FreddieWitherden, line 60: See if we can avoid this by using `id`.

TobyFlynn: Is the latest commit more what you were thinking in terms of using `id`? If it is, I've only tested it quickly - I'll set off some more tests overnight and let you when they've finished before merging

FreddieWitherden:
What would a solution not using pdeps or a topological sort look like?  So just trying to dynamically insert the group but while keeping as close as possible to the original ordering?

PR #451, opened 2024-12-11
Add impermeable ghost state calculation for viscous walls

kaschau:
This PR addresses viscous wall boundary conditions and their impermeability.

**Adiabatic No Slip Walls - Temperature Gradient**

The viscous flux is adiabatic when the normal flux calculation for the energy equation is zero. For heat conduction, this is achieved by setting the normal temperature gradient to be zero (shear stress energy is zero from zero velocity at the wall). Currently:

1. `ldg-state` (prescribing `ur` from `ul`) sensibly sets the wall solution momentum to zero, and matches density and internal energy from the fluid side. 
2. `ldg-grad-state` (prescribing `gradur` from `gradul`,`ul`) we copy all gradients from the left to the right, and then modify the total energy gradient s.t. when computing the temperature gradient in the viscous flux calculation, the normal temperature gradient is zero. This transformation, however, us done using the values of `gradul` and `ul` in the LDG Grad state routine. Critically, the temperature gradient calculation requires solution variables AND gradient variables. 
3. When the viscous flux at the wall is calculated, only the _right_ state is considered, i.e. $f_{wall} = g(ur, gradur)$. But, `gradur` was modified to yield zero normal temperature gradient based on `ul` solution values. Thus when the temperature gradient is computed in the viscous flux add routine, it does not yield zero normal temperature flux, and the impermeability of the wall BC is compromised.

**LDG Penalty term**

The next issue is the LDG penalty term $\tau(ul-ur)$. Since `ul`$\neq$`ur`, this term is non-zero, and directly adds normal flux to all variables (except density, which is the same on the left and right). This, again, compromises the impermeability of the wall boundary condition.

This PR creates a new ghost routine called `ghost-imperm` for the `no-slp-adia-wall` and `no-slp-isot-wall` boundary condition in the navier-stokes system. The difference being that `ghost-imperm` computes the temperature gradient in `ldg_grad_state` based on the previously defined `ur` and `gradul` (see L13 of `ghost-imperm`) so the values are correct in `viscous_flux_add`. Also the LDG penalty term is removed. Behavior of inlets and outlets remains the same.

I have tested the PyFR-Test-Cases that use walls and there is no significant change. The only noticeable change is in the viscous shock tube (see results below, current develop on top, new boundary condition on bottom). The issue is more apparent when the temperature gradient near the wall is very large, the velocity near the wall is non-zero, the wall normal is not aligned with a cartesian vector. 

Screenshot 2024-12-11 at 5 06 01 PM

### DIFF
--- a/pyfr/solvers/navstokes/kernels/bcs/ghost-imperm.mako
+++ b/pyfr/solvers/navstokes/kernels/bcs/ghost-imperm.mako
@@ -4,8 +4,6 @@
 <%include file='pyfr.solvers.euler.kernels.rsolvers.${rsolver}'/>
 <%include file='pyfr.solvers.navstokes.kernels.flux'/>
 
-<% tau = c['ldg-tau'] %>
-
 <%pyfr:macro name='bc_common_flux_state' params='ul, gradul, artviscl, nl, magnl'>
     // Viscous states
     fpdtype_t ur[${nvars}], gradur[${ndims}][${nvars}];
--- a/pyfr/solvers/navstokes/kernels/bcs/ghost-imperm.mako
+++ b/pyfr/solvers/navstokes/kernels/bcs/ghost-imperm.mako
@@ -1,0 +1,30 @@
+<%namespace module='pyfr.backends.base.makoutil' name='pyfr'/>
+
+<%include file='pyfr.solvers.baseadvecdiff.kernels.artvisc'/>
+<%include file='pyfr.solvers.euler.kernels.rsolvers.${rsolver}'/>
+<%include file='pyfr.solvers.navstokes.kernels.flux'/>
+
+<% tau = c['ldg-tau'] %>
+
+<%pyfr:macro name='bc_common_flux_state' params='ul, gradul, artviscl, nl, magnl'>
+    // Viscous states
+    fpdtype_t ur[${nvars}], gradur[${ndims}][${nvars}];
+    ${pyfr.expand('bc_ldg_state', 'ul', 'nl', 'ur')};
+    ${pyfr.expand('bc_ldg_grad_state', 'ur', 'nl', 'gradul', 'gradur')};
+
+    fpdtype_t fvr[${ndims}][${nvars}] = {{0}};
+    ${pyfr.expand('viscous_flux_add', 'ur', 'gradur', 'fvr')};
+    ${pyfr.expand('artificial_viscosity_add', 'gradur', 'fvr', 'artviscl')};
+
+    // Inviscid (Riemann solve) state
+    ${pyfr.expand('bc_rsolve_state', 'ul', 'nl', 'ur')};
+
+    // Perform the Riemann solve
+    fpdtype_t ficomm[${nvars}], fvcomm;
+    ${pyfr.expand('rsolve', 'ul', 'ur', 'nl', 'ficomm')};
+
+% for i in range(nvars):
+    fvcomm = ${' + '.join(f'nl[{j}]*fvr[{j}][{i}]' for j in range(ndims))};
+    ul[${i}] = magnl*(ficomm[${i}] + fvcomm);
+% endfor
+</%pyfr:macro>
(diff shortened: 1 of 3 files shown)


pyfr/solvers/navstokes/kernels/bcs/ghost-imperm.mako
```diff
@@ -0,0 +1,30 @@
+<%namespace module='pyfr.backends.base.makoutil' name='pyfr'/>
+
+<%include file='pyfr.solvers.baseadvecdiff.kernels.artvisc'/>
+<%include file='pyfr.solvers.euler.kernels.rsolvers.${rsolver}'/>
+<%include file='pyfr.solvers.navstokes.kernels.flux'/>
+
+<% tau = c['ldg-tau'] %>
```
FreddieWitherden, line 7: Not sure if we need this?

FreddieWitherden:
Thanks for this.  Can you do a check and see if the incompressible solver also needs changing?

kaschau:
`acnavier-stokes` does not compute heat flux as far as I can tell, and uses `bc_common_grad_copy` to set the grad state for no slip walls, so no this change does not apply.

FreddieWitherden:
@WillTrojak Do you recall what you did in your code for the no slip walls?

WillTrojak:
I need a bit more time to think about this. But, is there another, simpler, test case that can be made that demonstrates the issue and the difference compared to your new approach? Something in 2D should be enough

kaschau:
Here is what I am using to validate. I have a square domain that is comprised of 2x2 hex elements with physical extents of 1x1 length unit, see image. I am setting an initial condition that (with perfect interpolation, this uses p=1 so it will be off) yields a temperature gradient of 300K in `x` and 300K in `y`:

```
[soln-ics]
p = 101325.0
u = 0.0
v = 0.0
;; T = 300 + 300*y + 300*x
rho = 101325/(337.75*(300+300*y+300*x))
``` 

Screenshot 2024-12-13 at 1 33 22 PM

Then simply taking a single forward euler step in time, printing out the values of `T_x` and `T_y` in the `no-slp-adia-wall` `ldg_grad_state` and `viscous_flux_add` kernels. Since the bottom wall is adiabatic, the value of `T_y` must be zero in the `viscous_flux_add` routine (as this is the wall normal direction). 

In the current develop branch, the results are actually correct with these initial conditions:
Screenshot 2024-12-13 at 1 20 25 PM

But this is correct only because the velocity ICs are zero, which matches the state set to `ur` in `ldg_state`. If we impose a non zero initial velocity:
```
[soln-ics]
p = 101325.0
u = 5.0
v = 5.0
;; T = 300 + 300*y + 300*x
rho = 101325/(337.75*(300+300*y+300*x))
```
The value of `T_y` is no longer zero in the `viscous_flux_add` kernel at the wall:
Screenshot 2024-12-13 at 1 25 13 PM


With the new approach, and non zero velocity, we are back to zero normal temperature gradient at the wall:
Screenshot 2024-12-13 at 1 27 49 PM


The only thing I do not have a verification for is the LDG penalty term $\tau(ul-ur)$. This just seems conceptually wrong to go through the trouble of specifying things like zero temperature gradient at the wall, just to dump flux based on solution differences. For example, as is currently in the develop branch, one would have to justify:

> Why do we impose a total energy flux through an adiabatic wall that is equal to $\tau*KE_{left}$ (the kinetic energy on the left (fluid) side, as that is how it evaluates now due to the values we impose on the right solution state).

kaschau:
As an addendum to the above, I have also rotated the domain 45 degrees counter clockwise, with the same initial condition as above. In this configuration, a normal temperature gradient equal to zero requires that `T_x` and `T_y` are identical (s.t. the gradient at the wall is parallel to the sloped lower wall).

I have validated that $\frac{\partial T}{\partial y}$ and $\frac{\partial T}{\partial x}$ are equal with the proposed approach, and are different in magnitude and sometimes sign currently.

PR #452, opened 2024-12-12
Variable not necessary

kaschau:
I don't think it is necessary to set this to none and is causing incorrect behavior downstream when entropy filtering is not being used and checks using `hasattr` are made, see: https://github.com/PyFR/PyFR/blob/52b34313755cc4160b686057229804c880339d36/pyfr/solvers/base/system.py#L65

I cannot find any logic where entorpy filtering is not being used, but the `entmin_int` variable is required to be `None`.

### DIFF
--- a/pyfr/solvers/baseadvec/elements.py
+++ b/pyfr/solvers/baseadvec/elements.py
@@ -175,8 +175,6 @@ class BaseAdvectionElements(BaseElements):
                 self.m0 = None
             else:
                 self.m0 = self._be.const_matrix(self.basis.m0)
-        else:
-            self.entmin_int = None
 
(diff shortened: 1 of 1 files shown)


tarikdzanic:
Yeah it looks like it isn't necessary. Does it end up changing the solver behavior when EF is off? It seems like it just doesn't get used even though the conditional is wrong.

kaschau:
> Yeah it looks like it isn't necessary. Does it end up changing the solver behavior when EF is off? It seems like it just doesn't get used even though the conditional is wrong.

No other than creating an `eles_entmin_int` attribute in the system (which ends up as just a list of Nones with EF off) I cant find anywhere  where it changes any behavior. Just creates an unused variable in the system.

PR #453, opened 2024-12-12
Respecting dependencies when inserting a cache blocking group into the run order for the OpenMP backend

TobyFlynn:
This pull request inserts a cache blocking group into a graph's run order while respecting dependencies. Previously, a group replaced its last running kernel in a graph's run order without checking for dependencies. Now, the position of each kernel within a group is considered, with the earliest position that doesn't violate any dependencies being selected.

I've also added a quick sanity check afterwards to check that this positioning does not violate other dependencies.

### DIFF
--- a/pyfr/backends/openmp/types.py
+++ b/pyfr/backends/openmp/types.py
@@ -185,12 +185,13 @@ class OpenMPGraph(base.Graph):
         gdeps = []
         for k in kerns:
             gdeps = gdeps + [dep for dep in self.kdeps[k] if dep not in kerns]
-        gid = None
+ 
         if gdeps:
             lk = max(self.knodes[dep] for dep in gdeps)
             gid = min(self.knodes[k] for k in kerns if self.knodes[k] > lk) - 1
         else:
             gid = min(self.knodes[k] for k in kerns) - 1
+
         self.kins[gid] = groups
         
         # Sanity check that other dependencies haven't been violated
--- a/pyfr/backends/openmp/types.py
+++ b/pyfr/backends/openmp/types.py
@@ -182,12 +182,22 @@ class OpenMPGraph(base.Graph):
             groups.append(rargs)
 
         # Arrange for the groupings to be inserted into the final run list
-        group_deps = [dep for k in kerns for dep in self.kdeps[k]]
+        group_deps = [dep for k in kerns for dep in self.kdeps[k] if dep not in kerns]
+        group_kins_ind = None
         if group_deps:
             max_dep_ind = max(self.knodes[dep] for dep in group_deps)
-            self.kins[min(self.knodes[k] for k in kerns if self.knodes[k] > max_dep_ind) - 1] = groups
+            group_kins_ind = min(self.knodes[k] for k in kerns if self.knodes[k] > max_dep_ind) - 1
+            self.kins[group_kins_ind] = groups
         else:
-            self.kins[min(self.knodes[k] for k in kerns) - 1] = groups
+            group_kins_ind = min(self.knodes[k] for k in kerns) - 1
+            self.kins[group_kins_ind] = groups
+        
+        # Sanity check that other dependencies haven't been violated
+        for k in self.kdeps:
+            if k not in kerns:
+                for dep in self.kdeps[k]:
+                    if dep in kerns and self.knodes[k] < group_kins_ind:
+                        raise RuntimeError('Graph dependencies not met after grouping')
 
         # Finally, prevent grouped being added to the final run list
         for k in kerns:
--- a/pyfr/backends/openmp/types.py
+++ b/pyfr/backends/openmp/types.py
@@ -182,18 +182,12 @@ class OpenMPGraph(base.Graph):
             groups.append(rargs)
 
         # Arrange for the groupings to be inserted into the final run list
-        gdeps = []
-        for k in kerns:
-            gdeps = gdeps + [dep for dep in self.kdeps[k] if dep not in kerns]
- 
-        if gdeps:
-            lk = max(self.knodes[dep] for dep in gdeps)
-            gid = min(self.knodes[k] for k in kerns if self.knodes[k] > lk) - 1
-        else:
-            gid = min(self.knodes[k] for k in kerns) - 1
-
+        gdeps = [dep for k in kerns \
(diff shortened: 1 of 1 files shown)


pyfr/backends/openmp/types.py
```diff
@@ -182,7 +182,22 @@ def group(self, kerns, subs=[]):
             groups.append(rargs)
 
         # Arrange for the groupings to be inserted into the final run list
-        self.kins[max(self.knodes[k] for k in kerns) - 1] = groups
+        group_deps = [dep for k in kerns for dep in self.kdeps[k] if dep not in kerns]
+        group_kins_ind = None
+        if group_deps:
+            max_dep_ind = max(self.knodes[dep] for dep in group_deps)
+            group_kins_ind = min(self.knodes[k] for k in kerns if self.knodes[k] > max_dep_ind) - 1
```
FreddieWitherden, line 189: Line length; maybe use shorter variable names.

pyfr/backends/openmp/types.py
```diff
@@ -182,7 +182,22 @@ def group(self, kerns, subs=[]):
...
+            self.kins[group_kins_ind] = groups
+        else:
+            group_kins_ind = min(self.knodes[k] for k in kerns) - 1
+            self.kins[group_kins_ind] = groups
```
FreddieWitherden, line 193: Can move this outside the block

pyfr/backends/openmp/types.py
```diff
@@ -182,7 +182,23 @@ def group(self, kerns, subs=[]):
             groups.append(rargs)
 
         # Arrange for the groupings to be inserted into the final run list
-        self.kins[max(self.knodes[k] for k in kerns) - 1] = groups
+        gdeps = []
+        for k in kerns:
+            gdeps = gdeps + [dep for dep in self.kdeps[k] if dep not in kerns]
+        gid = None
```
FreddieWitherden, line 188: Why is this needed?  gid is assigned in both branches of the loop.

pyfr/backends/openmp/types.py
```diff
@@ -182,7 +182,23 @@ def group(self, kerns, subs=[]):
             groups.append(rargs)
 
         # Arrange for the groupings to be inserted into the final run list
-        self.kins[max(self.knodes[k] for k in kerns) - 1] = groups
+        gdeps = []
```
FreddieWitherden, line 185: Use a list comprehension here.

pyfr/backends/openmp/types.py
```diff
@@ -182,7 +182,23 @@ def group(self, kerns, subs=[]):
...
+        for k in kerns:
+            gdeps = gdeps + [dep for dep in self.kdeps[k] if dep not in kerns]
+        gid = None
+        if gdeps:
+            lk = max(self.knodes[dep] for dep in gdeps)
+            gid = min(self.knodes[k] for k in kerns if self.knodes[k] > lk) - 1
+        else:
+            gid = min(self.knodes[k] for k in kerns) - 1
+        self.kins[gid] = groups
```
FreddieWitherden, line 194: Add newline.

pyfr/backends/openmp/types.py
TobyFlynn: That was what I was doing previously with `[dep for k in kerns for dep in self.kdeps[k] if dep not in kerns]`

FreddieWitherden: Should just need a line break or two.

pyfr/backends/openmp/types.py
TobyFlynn: Oops I've very used to pre-declaring variables from other languages sorry

FreddieWitherden:
May be able to unify everything by doing
```python
lk = max((self.knodes[dep] for dep in gdeps), default=-1)
```
which avoids the if completely.

TobyFlynn:
> May be able to unify everything by doing
> 
> ```python
> lk = max((self.knodes[dep] for dep in gdeps), default=-1)
> ```
> 
> which avoids the if completely.

Nice!

pyfr/backends/openmp/types.py
FreddieWitherden: Avoid `\`, they should almost never be used.  Newlines are allowed inside [] and ().

FreddieWitherden: Also, we might want to move down the `self.kins[gid] = groups` bit until after the sanity check.

pyfr/backends/openmp/types.py
```diff
@@ -182,7 +182,18 @@ def group(self, kerns, subs=[]):
...
+        lk = max((self.knodes[dep] for dep in gdeps), default=-1)
+        gid = min(self.knodes[k] for k in kerns if self.knodes[k] > lk) - 1
+        self.kins[gid] = groups
+
+        # Sanity check that other dependencies haven't been violated
+        for k in self.kdeps:
+            if k not in kerns:
+                for dep in self.kdeps[k]:
+                    if dep in kerns and self.knodes[k] < gid:
+                        raise RuntimeError('Graph grouping dependency error')
```
FreddieWitherden, line 196: Line length.

pyfr/backends/openmp/types.py
TobyFlynn: I think I've now fixed this

pyfr/backends/openmp/types.py
```diff
@@ -182,7 +182,19 @@ def group(self, kerns, subs=[]):
             groups.append(rargs)
 
         # Arrange for the groupings to be inserted into the final run list
-        self.kins[max(self.knodes[k] for k in kerns) - 1] = groups
+        gdeps = [dep for k in kerns
+                        for dep in self.kdeps[k] if dep not in kerns]
```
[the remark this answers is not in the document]
TobyFlynn: No indent at all like this?
```
gdeps = [dep for k in kerns 
for dep in self.kdeps[k] if dep not in kerns]
```
or so it lines up after the assignment?
```
gdeps = [dep for k in kerns 
         for dep in self.kdeps[k] if dep not in kerns]
```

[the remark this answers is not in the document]
FreddieWitherden: The second form

PR #454, opened 2024-12-16
Allow point sampler cmd to process multiple solutions

FrankKTH:
Allow point sampler cmd to process multiple solutions

### DIFF
--- a/pyfr/plugins/sampler.py
+++ b/pyfr/plugins/sampler.py
@@ -351,6 +351,9 @@ class SamplerCLIPlugin(BaseCLIPlugin):
             dims = 'xyz'[:mesh.ndims]
             fields = soln['stats'].get('data', 'fields').split(',')
 
+            # Current time
+            tcurr = float(soln["stats"].items("solver-time-integrator")["tcurr"])
+
             # Read the sample points from a CSV file
             if fidx == 0:
                 # Init the output file
@@ -372,7 +375,6 @@ class SamplerCLIPlugin(BaseCLIPlugin):
             # Handle conversion from conservative to primitive variables
             if args.format == 'primitive':
                 from pyfr.solvers.base import BaseSystem
-
                 if soln['stats'].get('data', 'prefix') != 'soln':
                     raise ValueError('Primitive output only supported for '
                                     'solution files')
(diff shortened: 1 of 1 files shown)


FreddieWitherden:
Usually we only add bulk interfaces when there is a meaningful performance improvement compared with just using a shell script.  Are there any big benefits here?

FrankKTH:
> Usually we only add bulk interfaces when there is a meaningful performance improvement compared with just using a shell script.  Are there any big benefits here?

Not necessarily in terms of performance (in any case for sure there isn't a loss) but in terms of usability it is better since you don't dump the header everytime and makes gathering and analyzing data easier

FreddieWitherden:
So the header at the moment is easy enough to strip with `pyfr sampler sample ... | tail -n + 2`.  The issue with the new PR is there is no way of knowing *what* samples came from what file.  It is even possible for the header to be wrong if some solutions were done with a different system.  This seems like a lot of additional edge cases for very few benefits.

FrankKTH:
> The issue with the new PR is there is no way of knowing what samples came from what file.

To address the uniqueness of the file the time can be dumped as well as it is done in the runtime plugin. 
This would allow time series processing in a similar way if using the sampling in runtime or in the post simulation phase

> It is even possible for the header to be wrong if some solutions were done with a different system

If I understood correctly you might refer to the case where one of the solution in the file series passed belong to a different system. An additional logical check could be added if this is the case

The old output for two sample pts for two solution files is (dumped to the stream):
`x       y       rho     u       v       p
1.0     0.7     0.7288963478924152      0.8992335404875036      -0.28461602591129626    2.8673729084625985
1.0     0.8     0.744845096330979       0.9939963197009544      -0.2424982404050138     2.955590215111508
x       y       rho     u       v       p
1.0     0.7     0.9824071479600371      0.00830508875868363     0.9822433186245745      4.434546040405308
1.0     0.8     0.9834200919037982      0.0073885783094391365   0.9856347451371101      4.441654869882605`

New output for two sample pts for two solution files (dumped in a csv):
`t,x,y,rho,u,v,p
0.0,1.0,0.7,0.7288963478924152,0.8992335404875036,-0.28461602591129626,2.8673729084625985
0.0,1.0,0.8,0.744845096330979,0.9939963197009544,-0.2424982404050138,2.955590215111508
99.99999999999724,1.0,0.7,0.9824071479600371,0.00830508875868363,0.9822433186245745,4.434546040405308
99.99999999999724,1.0,0.8,0.9834200919037982,0.0073885783094391365,0.9856347451371101,4.441654869882605
`

pv101:
Thanks for the PR, but my preference here is to use shell scripts were at all possible for such tasks, as Freddie suggests, rather than adding any extra surface area to the code base on the PyFR side. Thanks.

PR #460, opened 2025-02-14
Make partition weight specification more flexible.

FreddieWitherden:


sambitmishra98:
I see no issues here, tested it for the last three days on a few meshes.

PR #463, opened 2025-03-18
Outputting tavg files at expected times when restarting

TobyFlynn:
This PR is in relation to this forum post: https://pyfr.discourse.group/t/tavg-merge-and-tavg-plugin-using-adaptive-time-step/1229/5

There are 2 parts to this PR, the first part is setting `tout_last` in the tavg plugin when restarting so that the output files are written at the expected times. For example, when restarting from t=0.01 and setting dt-out=0.01 you would expect the next tavg file to be written at 0.02 but it was instead being output at 0.02 + dt. This PR fixes it (tested on the example in the forum post).

The second part is writing tavg files at the expected time when setting tstart to a number which is not a multiple of `dt-out` (or a number that doesn't fit restart_time + n * `dt-out`). E.g. setting tstart = 0.015 and dt-out=0.01. I've edited `call_plugin_dt` to now account for this by taking tstart as an input. I've not had first hand experience of the dual time stepping but from I understand the same restriction of dt-out being a multiple of dt should apply to the user specified tstart time - let me know if I have misunderstood though.

### DIFF
--- a/pyfr/plugins/tavg.py
+++ b/pyfr/plugins/tavg.py
@@ -116,9 +116,8 @@ class TavgPlugin(PostactionMixin, RegionMixin, TavgMixin, BaseSolnPlugin):
         # Check if we are restarting and not before when tavg begins
         if intg.isrestart and intg.tcurr >= self.tstart:
             self.tout_last = intg.tcurr
-            self.init_tout_last = False
         else:
-            self.init_tout_last = True
+            self.tout_last = None
 
     def _prepare_exprs(self):
         cfg, cfgsect = self.cfg, self.cfgsect
@@ -161,10 +160,9 @@ class TavgPlugin(PostactionMixin, RegionMixin, TavgMixin, BaseSolnPlugin):
     def _init_accumex(self, intg):
         self.tstart_acc = self.prevt = intg.tcurr
 
-        if self.init_tout_last:
+        # Don't change tout_last if we are restarting past tstart
+        if self._started or self.tout_last is None:
             self.tout_last = intg.tcurr
-        else:
-            self.init_tout_last = True
 
         self.prevex = self._eval_acc_exprs(intg)
         self.accex = [np.zeros_like(p, dtype=np.float64) for p in self.prevex]
--- a/pyfr/plugins/tavg.py
+++ b/pyfr/plugins/tavg.py
@@ -160,10 +160,12 @@ class TavgPlugin(PostactionMixin, RegionMixin, TavgMixin, BaseSolnPlugin):
 
     def _init_accumex(self, intg):
         self.tstart_acc = self.prevt = intg.tcurr
+
         if self.init_tout_last:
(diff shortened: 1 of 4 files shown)


FreddieWitherden:
I think this looks reasonable but this code has, historically, been extremely fiddly.  Does everything else (i.e., writer related stuff) still work as expected in terms of how many files we get and the such like?

TobyFlynn:
This is what I have run so far (using the case from the forum as the base):
- From t=0.0 to 0.02, Writer: `dt-out = 0.01`, TAvg: `dt-out = 0.01` and `tstart = 0.0`
  - Writer files at t = 0.0, 0.01, 0.02
  - TAvg files at t = 0.01, 0.02
- Restarting from t=0.01 to 0.03, Writer: `dt-out = 0.01`, TAvg: `dt-out = 0.01` and `tstart = 0.0`
  - Writer files at t = 0.02, 0.03
  - TAvg files at t = 0.02, 0.03
- Restarting from t=0.01 to 0.026, Writer: `dt-out = 0.01`, TAvg: `dt-out = 0.01` and `tstart = 0.015`
  - Writer files at t = 0.02
  - TAvg files at t = 0.025
- Restarting from t=0.01 to 0.04, Writer: `dt-out = 0.0025`, TAvg: `dt-out = 0.005` and `tstart = 0.015`
  - Writer files at t = 0.0125, 0.015, 0.0175, 0.02, 0.0225, 0.025, 0.0275, 0.03, 0.0325, 0.035, 0.0375. 0.04
  - TAvg files at t = 0.02, 0.025, 0.03, 0.035, 0.04

I'm happy to run any additional tests if there is anything else you'd like me to check?

FreddieWitherden:
Additional things to check: no files are being overwritten by mistake and that the tavg merge command works where we expect it to.

pyfr/plugins/tavg.py
```diff
@@ -152,7 +159,11 @@ def _init_gradients(self):
                            for pname in gradpnames]
 
     def _init_accumex(self, intg):
-        self.tstart_acc = self.prevt = self.tout_last = intg.tcurr
+        self.tstart_acc = self.prevt = intg.tcurr
+        if self.init_tout_last:
+            self.tout_last = intg.tcurr
+        else:
+            self.init_tout_last = True
         self.prevex = self._eval_acc_exprs(intg)
```
FreddieWitherden, line 167: Add newline(s)

TobyFlynn:
I've just checked and the restart file (or anything else) is not being overwritten and the tavg merge works as expected (including with tavg files from original and restarted runs, e.g. avg-0.01.pyfrs from the original run and avg-0.02.pyfrs from a run restarted from a checkpoint at t = 0.01)

TobyFlynn:
I now use `_start` and if `tout_last` is/isn't `None` to remove the unnecessary auxiliary variable.

PR #466, opened 2025-04-08
Cleanups.

FreddieWitherden:


pyfr/plugins/turbulence.py
```diff
@@ -179,7 +179,7 @@ def get_vort_buf(self):
 
     def get_vort_structs(self, intg):
         ls, avgu = self.ls, self.avgu
-        props = Property(dimension=3, interleaved=True)
+        props = Property(dimension=3)
```
FreddieWitherden, line 182: @pv101 Can you check this does not break the plugin?

PR #478, opened 2025-06-18
Improve reproducibility of kernel selection.

FreddieWitherden:
This requires each kernel to be appreciably faster than the current best in order to usurp it.  This should make the code easier to profile and also improve reproducibility.

The factor of 0.95 seems about right, but we may need to make it 0.9 (so a new kernel has to perform 10% better than the current best).

### DIFF
--- a/pyfr/backends/base/backend.py
+++ b/pyfr/backends/base/backend.py
@@ -198,7 +198,7 @@ class BaseBackend:
                     continue
 
                 # Evaluate this kernel compared to the best seen so far
-                if best_kern is None or kern.dt < best_kern.dt:
+                if best_kern is None or kern.dt < 0.95*best_kern.dt:
                     best_kern = kern
 
                     # If there is no benchmark data then short circut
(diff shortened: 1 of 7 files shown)


pyfr/backends/hip/rocblas.py
TobyFlynn, line 1: self.ifac isn't used on line 148 when checking if a kernel is better than the current best

pyfr/backends/base/backend.py
TobyFlynn, line 1: I don't know if you'd want this to also be configurable from the ini file or if you are happy with it being hardcoded?

TobyFlynn:
I've done a (very) quick test of this on a CUDA backend and it looks good - performance is maintained between this and the develop branch for a simple TGV case.

pyfr/backends/base/backend.py
FreddieWitherden: I thought about it, although if possible I'd like these to be hard coded just to keep things simple.

FreddieWitherden:
Hopefully this should ensure we generally get the same result for the same test case.

PR #480, opened 2025-07-09
Mass flow controller BC

TobyFlynn:
This is the mass flow controller I've been using updated for the latest version of PyFR. Its currently a draft pull request as I'll make it more general (i.e. specify a target mass flow rate in the ini file rather than the current approach which is to specify a Mach number + inflow boundary in the ini file and then calculate a mass flow rate at startup) and then write up an example of how to use it to target a specific Mach number.

### DIFF
--- a/pyfr/solvers/euler/inters.py
+++ b/pyfr/solvers/euler/inters.py
@@ -220,8 +220,10 @@ class MassFlowBCMixin:
 
                 # Current p
                 p0 = self.interp_m * t + self.interp_c
+
                 # Next target p
                 p1 = p0 + dt * self.eta * (1 - self.target_mfr / self.mf_avg)
+
                 # Update interpolation
                 self.interp_m = (p1 - p0) / dt
                 self.interp_c = p0 - self.interp_m * t


FreddieWitherden:
Few more notes:
 - You can probably eliminate a lot of boilerplate by making the extern a broadcast array with one element rather than a scalar.
 - I am not sure what the ODE is? The method for $dp/dt$ is independent of $p$ so it reduces down to an integral (where RK4 is just reproducing Simpson's rule) and the integral itself is trivial to evaluate analytically as there is no dependence on $t$.
 - Is there any reason this is not being made available to the Euler system, too?

FreddieWitherden:
Externs should work (see how the turbulent inflow plugin works which defines a macro with externs).

TobyFlynn:
> Another note: look into using an exponentially weighted moving average. It is simpler and has nicer theoretical properties.

This looks good and some quick tests showed it performed very nicely - I've switched over to it now. Thanks for the advice!

FreddieWitherden:
@FrankFrank9 Once this PR has been updated would you be able to take a look and confirm if it is suitable for your purposes?

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,120 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg):
 
 class EulerSlpAdiaWallBCInters(EulerBaseBCInters):
     type = 'slp-adia-wall'
+
+
+class MassFlowIntegrator(SurfaceIntegrator):
```
FreddieWitherden: Not sure this is needed (just instantiate the parent with the right arguments)

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,120 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg):
...
+        self.mf_avg = 0.0
+        self.mf_alpha = 0.4
+        # Parameter to control the strength of the controller
+        self.eta = self.cfg.getfloat(cfgsect, 'eta')
+        # Frequency that mf.csv should be updated
+        self.nsteps = self.cfg.getint(cfgsect, 'nsteps', 100)
+        self.nflush = self.cfg.getint(cfgsect, 'nflush', 10)
+
+        self._set_external('var_p', 'scalar fpdtype_t')
+        self.tprev = -1.0
```
FreddieWitherden: Use None

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,120 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg):
...
+        self.nflush_counter = 0
+        
+        self.mf_int = MassFlowIntegrator(cfg, cfgsect, elemap, lhs)
+
+        if self.bccomm.rank == 0:
+            self.outf = init_csv(self.cfg, self.cfgsect, 't,mf,pbc')
+
+    def calculate_mass_flow(self, solns):
+        ndims, nvars = self.ndims, self.nvars
+        fm = np.zeros((ndims))
```
FreddieWitherden: Can remove ()

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,120 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg):
...
+            qwts = self.mf_int.qwts[etype, fidx]
+            norms = self.mf_int.norms[etype, fidx]
+
+            # Do the quadrature for each dimension
+            # RhoU = ufpts[1], RhoV = ufpts[2], RhoW = ufpts[2]
+            for i in range(0, ndims):
+                rhoVel = ufpts[1 + i]
+                fm[i] += np.einsum('i...,ij,ji', qwts, rhoVel, norms[:,:,i])
+        self.bccomm.Allreduce(mpi.IN_PLACE, fm, op=mpi.SUM)
+        return sum(fm)
```
FreddieWitherden: fm.sum() and maybe cast to float

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,120 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg):
...
+    
+    def update_mf(self, solns):
```
FreddieWitherden: Can inline many of these functions.

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,120 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg):
...
+        self.p += dt * self.eta * (1.0 - self.target_mfr / self.mf_avg)
+    
+    def bind_p(self, kerns):
+        for k in kerns:
+            kerns[k].bind(var_p=self.p)
+
+    def prepare(self, system, ubank, t, kerns):
+        # Check if past tstart
+        if t < self.tstart:
+            self.bind_p(kerns)
```
FreddieWitherden: See if you can refactor the logic so we don't have bind_p called in three different places.

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,120 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg):
...
+
+            # Output mass flow and pressure at outflow
+            if self.bccomm.rank == 0:
+                print(f'{t},{self.mf_avg},{self.p}', 
+                      file=self.outf)
+            self.nflush_counter = self.nflush_counter + 1
+        self.bind_p(kerns)
+
+        # Flush to file
+        if self.nflush_counter % self.nflush == 0:
```
FreddieWitherden: and

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,120 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg):
...
+            ufpts = ufpts.reshape(nfpts, nvars, -1)
+            ufpts = ufpts.swapaxes(0, 1)
+
+            # Get the quadrature weights and normal vectors
+            qwts = self.mf_int.qwts[etype, fidx]
+            norms = self.mf_int.norms[etype, fidx]
+
+            # Do the quadrature for each dimension
+            # RhoU = ufpts[1], RhoV = ufpts[2], RhoW = ufpts[2]
+            for i in range(0, ndims):
```
FreddieWitherden: zip

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,120 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg):
...
+    def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
+        super().__init__(be, lhs, elemap, cfgsect, cfg, bccomm)
+
+        self.c |= self._exp_opts(
+            ['rho', 'u', 'v', 'w'][:self.ndims + 1], lhs
+        )
+
+        self.target_mfr = self.cfg.getfloat(cfgsect, 'mass-flow-rate')
+        # When to start the mass flow controller
+        self.tstart = self.cfg.getfloat(cfgsect, 'tstart', 0.0)
```
FreddieWitherden: A default shouldn't be needed here?

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,120 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg):
...
+        # Start p value
+        self.p = self.cfg.getfloat(cfgsect, 'p')
+        self.bcname = cfgsect.removeprefix('soln-bcs-')
+        # Mass flow average
+        self.mf_avg = 0.0
+        self.mf_alpha = 0.4
```
FreddieWitherden: Should this be configurable?

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,96 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+            self.tprev = t
+
+            # Output mass flow and pressure at BC
+            if self.bccomm.rank == 0:
+                print(f'{t},{self.mf_avg},{self.p}', file=self.outf)
+
+            self.nflush_counter = self.nflush_counter + 1
+        
+        # Bind p to kernels
+        for k in kerns:
```
FreddieWitherden: kerns.values()

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,96 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+            kerns[k].bind(var_p=self.p)
+
+        # Flush to file
+        if self.nflush_counter % self.nflush == 0 and self.bccomm.rank == 0:
+            self.outf.flush()
+        self.nstep_counter = self.nstep_counter + 1
```
FreddieWitherden: +=

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,96 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+
+            dt = (t - self.tprev)
+            self.p += dt * self.eta * (1.0 - self.target_mfr / self.mf_avg)
+            self.tprev = t
+
+            # Output mass flow and pressure at BC
+            if self.bccomm.rank == 0:
+                print(f'{t},{self.mf_avg},{self.p}', file=self.outf)
+
+            self.nflush_counter = self.nflush_counter + 1
```
FreddieWitherden: Want to += 1

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,96 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+            # Get the quadrature weights and normal vectors
+            qwts = self.mf_int.qwts[etype, fidx]
+            norms = self.mf_int.norms[etype, fidx]
+
+            # Do the quadrature for each dimension
+            for i, _ufpts in enumerate(ufpts[1:ndims+1]):
+                fm[i] += np.einsum('i...,ij,ji', qwts, _ufpts, norms[:,:,i])
+
+        self.bccomm.Allreduce(mpi.IN_PLACE, fm, op=mpi.SUM)
+        return fm.sum().astype(float)
```
FreddieWitherden: float(fm.sum()) to remove the NumPy crud

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,96 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+            ufpts = m0 @ uupts.reshape(nupts, -1)
+            ufpts = ufpts.reshape(nfpts, nvars, -1)
+            ufpts = ufpts.swapaxes(0, 1)
+
+            # Get the quadrature weights and normal vectors
+            qwts = self.mf_int.qwts[etype, fidx]
+            norms = self.mf_int.norms[etype, fidx]
+
+            # Do the quadrature for each dimension
+            for i, _ufpts in enumerate(ufpts[1:ndims+1]):
```
FreddieWitherden: zip over norms (may want to rollaxis when you fetch it)

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,96 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+        for etype, fidx in self.mf_int.m0:
+            # Get the interpolation operator
+            m0 = self.mf_int.m0[etype, fidx]
+            nfpts, nupts = m0.shape
+
+            # Extract the relevant elements from the solution
+            uupts = solns[etype][..., self.mf_int.eidxs[etype, fidx]]
+
+            # Interpolate to the face
+            ufpts = m0 @ uupts.reshape(nupts, -1)
```
FreddieWitherden, line 188: Don't bother interpolating quantities you don't need (first and last).

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,96 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+
+class MassFlowBCMixin:
+    def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
+        super().__init__(be, lhs, elemap, cfgsect, cfg, bccomm)
+
+        self.c |= self._exp_opts(
+            ['rho', 'u', 'v', 'w'][:self.ndims + 1], lhs
+        )
+
+        self.target_mfr = self.cfg.getfloat(cfgsect, 'mass-flow-rate')
```
FreddieWitherden: No need for self.

pyfr/solvers/euler/inters.py
TobyFlynn: `nflush_counter` and `nstep_counter` are separate (writing to the CSV vs how often to update p).

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,96 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+            # First update to begin history
+            solns = dict(zip(system.ele_types, system.ele_scal_upts(ubank)))
+            self.mf_avg = self.calculate_mass_flow(solns)
+            self.tprev = t
+        elif t >= self.tstart and self.nstep_counter % self.nsteps == 0:
+            solns = dict(zip(system.ele_types, system.ele_scal_upts(ubank)))
+            mf = self.calculate_mass_flow(solns)
+            self.mf_avg = self.alpha * mf + (1.0 - self.alpha) * self.mf_avg
+
+            dt = (t - self.tprev)
```
FreddieWitherden: Remember this can be negative or 0.

TobyFlynn: Is a simple check for these cases sufficient? i.e. if dt > 0 then update p

FreddieWitherden: It isn't ideal.  The condition really should be differentially consistent.  It should also explicitly try to handle rejected steps to avoid hysteresis.

Approaches we've used in the past to help here are having target values and then interpolating towards them (so pass two pressures to the kernel and have the kernel interpolate with respect to time between them).

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,96 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+        self.tprev = None
+        self.nstep_counter = 0
+        self.nflush_counter = 0
+        self.mf_avg = 0.0
+        
+        surf_list = [(etype, fidx, eidxs) for etype, eidxs, fidx in lhs]
+        self.mf_int = SurfaceIntegrator(cfg, cfgsect, elemap, surf_list)
+
+        if self.bccomm.rank == 0:
+            self.outf = init_csv(self.cfg, self.cfgsect, 't,mf,pbc')
```
FreddieWitherden: The file stuff should be optional.  If the condition is well designed we shouldn't need this.

pyfr/solvers/euler/inters.py
```diff
@@ -1,7 +1,12 @@
+from pyfr.mpiutil import mpi
+from pyfr.plugins.base import init_csv
```
FreddieWitherden: Awkward dependency (solvers shouldn't depend on plugins which are higher level).

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,100 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+        # Get the sizes for the area calculation
+        for etype, fidx in self.mf_int.m0:
+            # Get the interpolation operator
+            m0 = self.mf_int.m0[etype, fidx]
+            nfpts, nupts = m0.shape
+
+            # Extract the relevant elements from the solution
+            uupts = solns[etype][..., self.mf_int.eidxs[etype, fidx]]
+            # Remove unused variables
+            uupts = uupts[:,1:-1,:]
```
FreddieWitherden: Can't we just `.sum(axis=...)` here? No need for final :

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,100 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+
+            # Get the quadrature weights and normal vectors
+            qwts = self.mf_int.qwts[etype, fidx]
+            norms = np.rollaxis(self.mf_int.norms[etype, fidx], 2)
+
+            # Do the quadrature for each dimension
+            for _ufpts, _norms in zip(ufpts, norms):
+                mf += np.einsum('i...,ij,ji', qwts, _ufpts, _norms)
+
+        self.bccomm.allreduce(mf, op=mpi.SUM)
```
FreddieWitherden: scal_coll

FreddieWitherden:
Also, do your best to handle restarts.  Practically you'll want to do `self.cfg.set(...)` when you update the pressure value.  This way on restart the BC will pick up the correct value.

pyfr/solvers/euler/inters.py
TobyFlynn: I could move `init_csv` (maybe also `open_hdf5_a` and `DatasetAppender`?) to writers.base? Though they are different kinds of 'writers' than what that directory is aimed at

TobyFlynn:
> Do you have the underlying mathematical expression for what we've defined as the mass flow rate?

This is what I'm using, so integrating across the boundary the mass flow normal to it:
image

pyfr/solvers/euler/inters.py
TobyFlynn: I'm not sure I follow about `.sum(axis=...)`

pyfr/solvers/euler/inters.py
TobyFlynn: I've added an interpolation between pressures but I need to think more about handling rejected steps

pyfr/solvers/euler/inters.py
FreddieWitherden: Based on my reading of the code you can sum the three momentum components there-and-then (pre interpolation, which is linear, followed by quadrature, which is linear).

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,113 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+            # Extract the relevant elements from the solution
+            uupts = solns[etype][..., self.mf_int.eidxs[etype, fidx]]
+            # Remove unused variables
+            uupts = uupts[:,1:-1]
+
+            # Interpolate to the face
+            ufpts = m0 @ uupts.reshape(nupts, -1)
+            ufpts = ufpts.reshape(nfpts, nvars-2, -1)
+            ufpts = ufpts.swapaxes(0, 1)
+
```
FreddieWitherden, line 190: I'd double check this expression and see if you can't get einum to do the loop with an extra index

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,113 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+            # Get the quadrature weights and normal vectors
+            qwts = self.mf_int.qwts[etype, fidx]
```
FreddieWitherden, line 192: scal_coll

pyfr/solvers/euler/inters.py
FreddieWitherden: Nevermind, I think I understand everything now. Disregard this.

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,113 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+        ndims, nvars = self.ndims, self.nvars
+        mf = 0.0
+
+        # Get the sizes for the area calculation
+        for etype, fidx in self.mf_int.m0:
+            # Get the interpolation operator
+            m0 = self.mf_int.m0[etype, fidx]
+            nfpts, nupts = m0.shape
+
+            # Extract the relevant elements from the solution
```
FreddieWitherden: ndims

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,113 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+        if self.writecsv and bccomm.rank == 0:
+            self.outf = init_csv(cfg, cfgsect, 't,mf,pbc')
+
+    def calculate_mass_flow(self, solns):
+        ndims, nvars = self.ndims, self.nvars
+        mf = 0.0
+
+        # Get the sizes for the area calculation
+        for etype, fidx in self.mf_int.m0:
+            # Get the interpolation operator
```
FreddieWitherden, line 180: Can you move above?

pyfr/solvers/euler/inters.py
TobyFlynn: I've moved `init_csv`, `open_hdf5_a` and `DatasetAppender` to writers.timeseries. The other option would be to have a mass flow stats plugin that would handle writing out a csv, let me know if that would be better instead.

FreddieWitherden:
Just to keep things progressing quickly all of the changes outlined above are infrastructure so can go into their own PR.

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,128 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+            uupts = solns[etype][..., self.mf_int.eidxs[etype, fidx]][:,1:-1]
+
+            # Interpolate to the face
+            ufpts = m0 @ uupts.reshape(nupts, -1)
+            ufpts = ufpts.reshape(nfpts, ndims, -1)
+            ufpts = ufpts.swapaxes(0, 1)
+
+            # Get the quadrature weights and normal vectors
+            qwts = self.mf_int.qwts[etype, fidx]
+            norms = np.rollaxis(self.mf_int.norms[etype, fidx], 2)
```
FreddieWitherden: Can probably avoid the rollaxis just by being smarter with the einsum stuff

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,128 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+
+            # Do the quadrature
+            mf += np.einsum('i,hij,hji', qwts, ufpts, norms)
+
+        mf = scal_coll(self.bccomm.Allreduce, mf, op=mpi.SUM)
```
FreddieWitherden: Straight return here?

pyfr/solvers/euler/kernels/bcs/char-riem-inv-mass-flow.mako
```diff
@@ -0,0 +1,41 @@
+<%namespace module='pyfr.backends.base.makoutil' name='pyfr'/>
+
+<% gmo = c['gamma'] - 1.0 %>
+<% gamma = c['gamma'] %>
+
+<%pyfr:macro name='bc_rsolve_state' params='ul, nl, ur' externs='ploc, t, t0, t1, p0, p1'>
+    fpdtype_t p = p0 + ((t - t0) / (t1 - t0)) * (p1 - p0);
```
FreddieWitherden: Should only need to pass in two things (straight line => two coefficients).

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,124 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+            self.t1 = 1.0
+            self.mf_avg = 0.0
+            self.tprev = None
+
+        self.nstep_counter = 0
+        
+        surf_list = [(etype, fidx, eidxs) for etype, eidxs, fidx in lhs]
+        self.mf_int = SurfaceIntegrator(cfg, cfgsect, elemap, surf_list)
+
+        self.csv = None
```
FreddieWitherden: Move into else branch

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,124 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+        ndims, nvars = self.ndims, self.nvars
+        mf = 0.0
+
+        for etype, fidx in self.mf_int.m0:
+            # Get the interpolation operator
+            m0 = self.mf_int.m0[etype, fidx]
+            nfpts, nupts = m0.shape
+
+            # Extract the relevant elements and variables from the solution
+            uupts = solns[etype][..., self.mf_int.eidxs[etype, fidx]][:,1:-1]
```
FreddieWitherden: Should be able to merge [][] => []

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,124 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+
+            # Interpolate to the face
+            ufpts = m0 @ uupts.reshape(nupts, -1)
+            ufpts = ufpts.reshape(nfpts, ndims, -1)
+            ufpts = ufpts.swapaxes(0, 1)
```
FreddieWitherden: Not sure the swapaxes is needed (einsum it)

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,124 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+            norms = self.mf_int.norms[etype, fidx]
+
+            # Do the quadrature
+            mf += np.einsum('i,hij,jih', qwts, ufpts, norms)
+
+        return scal_coll(self.bccomm.Allreduce, mf, op=mpi.SUM)
+    
+    @classmethod
+    def preparefn(cls, bciface, mesh, elemap):
+        return lambda s, u, t, k : bciface.prepare(s, u, t, k)
```
FreddieWitherden: Need to handle case of bciface == None and return None

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,124 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+            self.cfg.set(self.cfgsect, 't0', self.t0)
+            self.cfg.set(self.cfgsect, 't1', self.t1)
+            self.cfg.set(self.cfgsect, 'mf_avg', self.mf_avg)
+            self.cfg.set(self.cfgsect, 'tprev', self.tprev)
+
+            # Output mass flow and pressure at BC
+            if self.csv:
+                self.csv(t, self.mf_avg, self.p1)
+        
+        # Bind p to kernels
```
FreddieWitherden: Update comment

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,124 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+        self._set_external('t0', 'scalar fpdtype_t')
+        self._set_external('t1', 'scalar fpdtype_t')
+
+        # Check if using values from restart
+        if cfg.hasopt(cfgsect, 't0'):
+            self.p0 = cfg.getfloat(cfgsect, 'p0')
+            self.p1 = cfg.getfloat(cfgsect, 'p1')
+            self.t0 = cfg.getfloat(cfgsect, 't0')
+            self.t1 = cfg.getfloat(cfgsect, 't1')
+            self.mf_avg = cfg.getfloat(cfgsect, 'mf_avg')
```
FreddieWitherden: Use -

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,124 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+class MassFlowBCMixin:
+    def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
+        super().__init__(be, lhs, elemap, cfgsect, cfg, bccomm)
+
+        self.c |= self._exp_opts(
+            ['rho', 'u', 'v', 'w'][:self.ndims + 1], lhs
+        )
+
+        self.target_mfr = cfg.getfloat(cfgsect, 'mass-flow-rate')
+        self.tstart = cfg.getfloat(cfgsect, 'tstart')
```
FreddieWitherden: Can this be defaulted?

pyfr/solvers/euler/inters.py
FreddieWitherden: Also in the primary branch can return `bciface.prepare`.

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,124 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+            self.p0 = cfg.getfloat(cfgsect, 'p')
+            self.p1 = self.p0
+            self.t0 = 0.0
+            self.t1 = 1.0
+            self.mf_avg = 0.0
+            self.tprev = None
+
+        self.nstep_counter = 0
+        
+        surf_list = [(etype, fidx, eidxs) for etype, eidxs, fidx in lhs]
```
FreddieWitherden: Don't think eidxs should be plural.

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,119 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+        else:
+            return None
+
+    def prepare(self, system, ubank, t, kerns):
+        if t >= self.tstart and not self.tprev:
+            # First update to begin history
+            solns = dict(zip(system.ele_types, system.ele_scal_upts(ubank)))
+            self.mf_avg = self.calculate_mass_flow(solns)
+            self.tprev = t
+        elif t >= self.tstart and self.nstep_counter % self.nsteps == 0:
```
FreddieWitherden: First two lines of both if/elif blocks look identical

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,119 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+            else:
+                self.mf_avg = self.alpha * mf + (1 - self.alpha) * self.mf_avg
+                dt = (t - self.tprev)
+                self.tprev = t
+
+                # Current p
+                p0 = self.interp_m * t + self.interp_c
+                # Next target p
+                p1 = p0 + dt * self.eta * (1 - self.target_mfr / self.mf_avg)
+                # Update interpolation
```
FreddieWitherden, line 227: Newline before comment

pyfr/solvers/euler/inters.py
```diff
@@ -124,3 +129,119 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+                self.mf_avg = mf
+                self.tprev = t
+            else:
+                self.mf_avg = self.alpha * mf + (1 - self.alpha) * self.mf_avg
+                dt = (t - self.tprev)
+                self.tprev = t
+
+                # Current p
+                p0 = self.interp_m * t + self.interp_c
+                # Next target p
```
FreddieWitherden, line 224: Newline before comment

PR #481, opened 2025-07-11
Time averaging bug fix

TobyFlynn:
The time average plugin was getting the output data precision from its own section in the ini file rather than the backend section. This could cause a mismatch in precision when passing data to the writer (e.g. simulation run in double precision but tavg done in single precision) which would cause half the domain to be zeroed out and the other half to have invalid data.

### DIFF
--- a/pyfr/plugins/tavg.py
+++ b/pyfr/plugins/tavg.py
@@ -82,8 +82,8 @@ class TavgPlugin(PostactionMixin, RegionMixin, TavgMixin, BaseSolnPlugin):
         ershapes = {etype: (nfields, emap[etype].nupts) for etype in erdata}
 
         # Construct the file writer
-        self._writer = NativeWriter.from_integrator(intg, basedir, 
-                                                    basename, 'tavg', 
+        self._writer = NativeWriter.from_integrator(intg, basedir, basename,
+                                                    'tavg', 
                                                     fpdtype=self.fpdtype)
         self._writer.set_shapes_eidxs(ershapes, erdata)
 
--- a/pyfr/plugins/tavg.py
+++ b/pyfr/plugins/tavg.py
@@ -63,7 +63,7 @@ class TavgPlugin(PostactionMixin, RegionMixin, TavgMixin, BaseSolnPlugin):
         nfields = self._prepare_exprs()
 
         # Output data type
-        fpdtype = self.cfg.get('backend', 'precision', 'single')
+        fpdtype = self.cfg.get(cfgsect, 'precision', 'single')
         if fpdtype == 'single':
             self.fpdtype = np.float32
         elif fpdtype == 'double':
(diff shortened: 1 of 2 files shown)


FreddieWitherden:
The capability to write time averages in lower precision is deliberate (as you never restart from them it is often beneficial and reduces write times).  The correct fix should probably be to properly handle this in the writer.

pyfr/plugins/tavg.py
```diff
@@ -82,8 +82,10 @@ def __init__(self, intg, cfgsect, suffix=None):
         ershapes = {etype: (nfields, emap[etype].nupts) for etype in erdata}
 
         # Construct the file writer
-        self._writer = NativeWriter.from_integrator(intg, basedir, basename,
```
FreddieWitherden, line 85: Probably best to extend `from_integrator` to accept a precision option.

pyfr/plugins/tavg.py
```diff
@@ -83,7 +83,7 @@ def __init__(self, intg, cfgsect, suffix=None):
 
         # Construct the file writer
         self._writer = NativeWriter.from_integrator(intg, basedir, basename,
-                                                    'tavg')
+                                                'tavg', fpdtype=self.fpdtype)
```
FreddieWitherden: Indentation.

pyfr/writers/native.py
```diff
@@ -74,8 +74,10 @@ def __init__(self, mesh, cfg, fpdtype, basedir, basename, prefix, *,
         self._awriter = None
 
     @staticmethod
-    def from_integrator(intg, basedir, basename, prefix, *, extn='.pyfrs'):
-        return NativeWriter(intg.system.mesh, intg.cfg, intg.backend.fpdtype,
+    def from_integrator(intg, basedir, basename, prefix, *, extn='.pyfrs',
+                        fpdtype=None):
+        _ftype = intg.backend.fpdtype if fpdtype is None else fpdtype
```
FreddieWitherden: I think or will work here.

pyfr/plugins/tavg.py
TobyFlynn: Fixed now

pyfr/writers/native.py
TobyFlynn: I've now changed to or

pyfr/plugins/tavg.py
```diff
@@ -82,8 +82,9 @@ def __init__(self, intg, cfgsect, suffix=None):
         ershapes = {etype: (nfields, emap[etype].nupts) for etype in erdata}
 
         # Construct the file writer
-        self._writer = NativeWriter.from_integrator(intg, basedir, basename,
-                                                    'tavg')
+        self._writer = NativeWriter.from_integrator(intg, basedir, 
+                                                    basename, 'tavg',
```
FreddieWitherden: Not sure why basename has moved down?

pyfr/writers/native.py
```diff
@@ -74,8 +74,10 @@ def __init__(self, mesh, cfg, fpdtype, basedir, basename, prefix, *,
         self._awriter = None
 
     @staticmethod
-    def from_integrator(intg, basedir, basename, prefix, *, extn='.pyfrs'):
-        return NativeWriter(intg.system.mesh, intg.cfg, intg.backend.fpdtype,
+    def from_integrator(intg, basedir, basename, prefix, *, extn='.pyfrs',
+                        fpdtype=None):
+        _ftype = fpdtype or intg.backend.fpdtype
+        return NativeWriter(intg.system.mesh, intg.cfg, _ftype,
```
FreddieWitherden: Reindent.

PR #482, opened 2025-07-11
Infrastructure for the mass flow BC

TobyFlynn:
Allows BCs to be created with a shared MPI controller and also to set externs that vary at runtime.


pyfr/solvers/base/system.py
```diff
@@ -152,16 +154,29 @@ def _load_bc_inters(self, mesh, elemap):
...
             bcclass = bcmap[self.cfg.get(cfgsect, 'type')]
-            bciface = bcclass(self.backend, interarr, elemap, cfgsect,
-                                self.cfg)
-            bc_inters.append(bciface)
+
+            # Check if requires an MPI communicator
+            bccomm = None
+            if bcclass.req_mpi_comm:
+                comm, rank, root = get_comm_rank_root()
+                bccomm = comm.Split(1 if bname in mesh.bcon else mpi.UNDEFINED)
```
FreddieWitherden: Memory leak (see `autofree`).

pyfr/solvers/base/system.py
```diff
@@ -152,16 +154,29 @@ def _load_bc_inters(self, mesh, elemap):
         bccls = self.bbcinterscls
         bcmap = {b.type: b for b in subclasses(bccls, just_leaf=True)}
 
+        # Iterate over all BCs (including ones not on this rank)
         bc_inters = []
-        for bname, interarr in mesh.bcon.items():
+        bcnames = [s[9:] for s in self.cfg.sections() if s[:9] == 'soln-bcs-']
```
FreddieWitherden: Don't rely on the config file for the master list of BCs across all ranks.  Use the mesh codec.

TobyFlynn:
I've made these changes and removed the extra externs bits as I now use a similar method to the turbulence plugin as you suggested by defining the extern as a broadcast (a 2D 1x1 matrix as it was saying the broadcasts must be 2D).

pyfr/solvers/aceuler/inters.py
```diff
@@ -73,11 +73,14 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg):
 class ACEulerSlpWallBCInters(ACEulerBaseBCInters):
     type = 'slp-wall'
 
+    def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
```
FreddieWitherden: Should not need to define this

pyfr/solvers/aceuler/inters.py
```diff
@@ -73,11 +73,14 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg):
...
+        super().__init__(be, lhs, elemap, cfgsect, cfg)
+
 
 class ACEulerCharRiemInvBCInters(ACEulerBaseBCInters):
     type = 'ac-char-riem-inv'
 
-    def __init__(self, be, lhs, elemap, cfgsect, cfg):
+    def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
         super().__init__(be, lhs, elemap, cfgsect, cfg)
```
FreddieWitherden: Always pass up (redefine parent if needed).

pyfr/solvers/base/system.py
```diff
@@ -9,6 +9,7 @@
 from pyfr.cache import memoize
 from pyfr.shapes import BaseShape
 from pyfr.util import subclasses
+from pyfr.mpiutil import autofree, get_comm_rank_root, mpi
```
FreddieWitherden: Alphabetical order imports.

pyfr/solvers/base/system.py
```diff
@@ -152,16 +153,26 @@ def _load_bc_inters(self, mesh, elemap):
         bccls = self.bbcinterscls
         bcmap = {b.type: b for b in subclasses(bccls, just_leaf=True)}
 
+        # Iterate over all BCs (including ones not on this rank)
         bc_inters = []
-        for bname, interarr in mesh.bcon.items():
+        bcnames = [s[3:] for s in mesh.codec if s[:3] == 'bc/']
```
FreddieWitherden: `.startswith` and `.removeprefix`

FreddieWitherden: May be cleaner to iterate in the for loop below as:

```
for c in mesh.codec:
    if not c.startswith('bc/'):
        continue

    bname = c.removeprefix('bc/')
```

pyfr/solvers/base/system.py
```diff
@@ -152,16 +153,26 @@ def _load_bc_inters(self, mesh, elemap):
...
             # Determine the config file section
             cfgsect = f'soln-bcs-{bname}'
 
-            # Instantiate
-            bcclass = bcmap[self.cfg.get(cfgsect, 'type')]
-            bciface = bcclass(self.backend, interarr, elemap, cfgsect,
-                                self.cfg)
-            bc_inters.append(bciface)
+            # Construct MPI communicator for this BC
+            comm, rank, root = get_comm_rank_root()
```
FreddieWitherden: Move up to parent scope (no need to keep calling `get_comm_rank_root`).

TobyFlynn:
> > I've made these changes and removed the extra externs bits as I now use a similar method to the turbulence plugin as you suggested by defining the extern as a broadcast (a 2D 1x1 matrix as it was saying the broadcasts must be 2D).
> 
> If you do want to use a scalar argument your best bet is to ensure the kernel itself is passed to prepare. Then one can call the bind method to bind the scalar. The problem with your previous approach (passing everything up to the system and having it bind) is that it will break if two plugins choose the same name for a scalar. Giving the plugin access in prepare to the kernel object avoids this (as does using a broadcasted array).

I've been having a look and I don't see an elegant way of passing the kernel back in prepare or matching kernels to the BC that produced them. Yeah I now see the problem with my previous approach, thanks for steering me in the right direction. There is probably a little overhead for updating the extern matrix on GPU backends but given that the BC only updates the matrix ~30 or more time steps it should be minimal so I am happy to stick with using a broadcast array if you are.

FreddieWitherden:
> > > I've made these changes and removed the extra externs bits as I now use a similar method to the turbulence plugin as you suggested by defining the extern as a broadcast (a 2D 1x1 matrix as it was saying the broadcasts must be 2D).
> > 
> > 
> > If you do want to use a scalar argument your best bet is to ensure the kernel itself is passed to prepare. Then one can call the bind method to bind the scalar. The problem with your previous approach (passing everything up to the system and having it bind) is that it will break if two plugins choose the same name for a scalar. Giving the plugin access in prepare to the kernel object avoids this (as does using a broadcasted array).
> 
> I've been having a look and I don't see an elegant way of passing the kernel back in prepare or matching kernels to the BC that produced them. Yeah I now see the problem with my previous approach, thanks for steering me in the right direction. There is probably a little overhead for updating the extern matrix on GPU backends but given that the BC only updates the matrix ~30 or more time steps it should be minimal so I am happy to stick with using a broadcast array if you are.

Yes, some extra plumbing would be needed to make this work.  There are a few ways of doing it if you're curious.  Probably the most direct is to add extra metadata into the system so it knows more about where kernels came from so it can then pass them back later on.  Another would be to have BC's track what kernels they generate for the system and retain a weak reference to their bind functions.

Still, as things stand, arrays are the recommended solution

pyfr/solvers/base/system.py
```diff
@@ -152,16 +153,30 @@ def _load_bc_inters(self, mesh, elemap):
...
+        # Iterate over all BCs (including ones not on this rank)
         bc_inters = []
-        for bname, interarr in mesh.bcon.items():
+        for c in mesh.codec:
+            if not c.startswith('bc/'):
+                continue
+
             # Determine the config file section
+            bname = c.removeprefix('bc/')
             cfgsect = f'soln-bcs-{bname}'
```
FreddieWitherden: Move into localbc block as not needed otherwise.

pyfr/solvers/euler/inters.py
```diff
@@ -98,8 +98,8 @@ def __init__(self, *args, **kwargs):
 class EulerSupInflowBCInters(EulerBaseBCInters):
     type = 'sup-in-fa'
 
-    def __init__(self, be, lhs, elemap, cfgsect, cfg):
-        super().__init__(be, lhs, elemap, cfgsect, cfg)
+    def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
```
FreddieWitherden, line 101: Should not be needed.

TobyFlynn:
> > > > I've made these changes and removed the extra externs bits as I now use a similar method to the turbulence plugin as you suggested by defining the extern as a broadcast (a 2D 1x1 matrix as it was saying the broadcasts must be 2D).
> > > 
> > > 
> > > If you do want to use a scalar argument your best bet is to ensure the kernel itself is passed to prepare. Then one can call the bind method to bind the scalar. The problem with your previous approach (passing everything up to the system and having it bind) is that it will break if two plugins choose the same name for a scalar. Giving the plugin access in prepare to the kernel object avoids this (as does using a broadcasted array).
> > 
> > 
> > I've been having a look and I don't see an elegant way of passing the kernel back in prepare or matching kernels to the BC that produced them. Yeah I now see the problem with my previous approach, thanks for steering me in the right direction. There is probably a little overhead for updating the extern matrix on GPU backends but given that the BC only updates the matrix ~30 or more time steps it should be minimal so I am happy to stick with using a broadcast array if you are.
> 
> Yes, some extra plumbing would be needed to make this work. There are a few ways of doing it if you're curious. Probably the most direct is to add extra metadata into the system so it knows more about where kernels came from so it can then pass them back later on. Another would be to have BC's track what kernels they generate for the system and retain a weak reference to their bind functions.
> 
> Still, as things stand, arrays are the recommended solution

I'm happy to stick with arrays

pyfr/solvers/base/system.py
```diff
@@ -248,7 +261,7 @@ def _prepare_kernels(self, t, uinbank, foutbank):
         _, binders = self._get_kernels(uinbank, foutbank)
 
         for b in self._bc_inters:
-            b.prepare(t)
+            b.prepare(t, self, uinbank)
```
FreddieWitherden: How is the bank used?  The system normally tries to hide these details

FreddieWitherden:
Also on the infrastructure side we probably want to do something about the surface quadrature.  It should be factored from the plugin side so BCs can use it, too.

pyfr/solvers/base/system.py
TobyFlynn: I was using `system.ele_scal_upts(uinbank)` to get the current solution to do the mass flow integration on. I could just use a fixed bank in the BC like `system.ele_scal_upts(0)` as the difference between the solution at a different RK stage/previous time step would have negligible effect on the mass flow controller (which checks out on a simple test case). Or I could pass the solution to prepare, i.e. `b.prepare(self, t, kerns, self.ele_scal_upts(uinbank))`?

FreddieWitherden: I think the current approach is fine if a bit inefficient (as if multiple BCs all want the same thing we'll end up copying it multiple times).  With that said the long term solution is to migrate the surface quadrature to the backend so it is fine to keep things as they are for the moment (this being a much better solution as it avoids ever needing to copy the solution).

pyfr/plugins/base.py
```diff
@@ -236,6 +236,9 @@ def __init__(self, intg, *args, **kwargs):
 
 
 class SurfaceMixin:
+    def __init__(self, *args, **kwargs):
```
FreddieWitherden: Is this needed?  I think it should be the default

pyfr/solvers/euler/inters.py
```diff
@@ -1,7 +1,48 @@
...
 from pyfr.solvers.baseadvec import (BaseAdvectionIntInters,
                                     BaseAdvectionMPIInters,
                                     BaseAdvectionBCInters)
 
+class BCIntersSurfaceMixin(SurfaceMixin):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    # Setup integrating over boundary
+    def _init_surface_integration(self, system, elemap, bcname):
```
FreddieWitherden: This still seems to have quite a lot of code duplication with the plugins.  Is there anything more we can do here?

pyfr/plugins/base.py
TobyFlynn: Yeah, you're right - I'll remove

pyfr/solvers/euler/inters.py
TobyFlynn: I'll see if I can move some of the common setup/init code into the mixin

TobyFlynn:
> > For the BC surface quadrature, I am importing the SurfaceMixin from plugins.base and then extending that class with BCIntersSurfaceMixin which also implements some setup code (copied with some adaptions from plugins.fluidforce).
> > Not in this PR but for the actual mass flow controller I then extended BCIntersSurfaceMixin to get a MassFlowBCMixin class that both the Euler and NS boundary conditions inherit from (with the Euler and NS classes having almost no code in, its all in MassFlowBCMixin).
> 
> I might be tempted to move the mixin to quadrules (which is arguably a better place for it if multiple things are now going to depend on it).

If I move the mixin in its current state to quadrules I end up with circular imports due to region import needed for `_surf_region`. This function is only used in the fwh plugin, so if we want to put `SurfaceMixin` in quadrules, I could remove `_surf_region` from the mixin. Then I could extend `SurfaceMixin` in the base plugin file to add `_surf_region` back in to get `SurfaceRegionMixin`?

pyfr/solvers/euler/inters.py
TobyFlynn: This is now in `SurfaceMixin._surf_init()` which is the setup code that is used by both the mass flow controller and fluidforce plugin

pyfr/quadrules/__init__.py
```diff
@@ -102,3 +106,85 @@ class StoredQuadRule(BaseStoredQuadRule):
...
+                if morigin:
+                    ploc = eles.ploc_at_np(ppts)[..., eidx]
+                    rfpt = ploc - morigin
+                    rfpts[etype, fidx].append(rfpt)
+
+            self._eidxs = {k: np.array(v) for k, v in eidxs.items()}
+            self._norms = {k: np.array(v) for k, v in norms.items()}
+            self._rfpts = {k: np.array(v) for k, v in rfpts.items()}
+            self._rcpjact = {k: rcpjact[k[0]][..., v]
+                                    for k, v in self._eidxs.items()}
```
FreddieWitherden: Indentation

pyfr/quadrules/__init__.py
```diff
@@ -102,3 +106,85 @@ class StoredQuadRule(BaseStoredQuadRule):
...
+            qdeg = self.cfg.getint(self.cfgsect, f'quad-deg-{itype}')
+        except NoOptionError:
+            qdeg = self.cfg.getint(self.cfgsect, 'quad-deg')
+
+        # Get the quadrature rule
+        q = get_quadrule(itype, qrule, qdeg=qdeg, flags=flags)
+
+        # Project its points onto the provided surface
+        pts = np.atleast_2d(q.pts.T)
+        return np.vstack(np.broadcast_arrays(*proj(*pts))).T, q.wts
```
FreddieWitherden: Add newline

pyfr/quadrules/__init__.py
```diff
@@ -102,3 +106,85 @@ class StoredQuadRule(BaseStoredQuadRule):
...
+
+                    # Get |J|^-1 at the solution points
+                    rcpdjac = eles.rcpdjac_at_np('upts')
+
+                    # Product to give J^-T at the solution points
+                    rcpjact[etype] = smat*rcpdjac
+
+                # Get the flux points position of the given face and element
+                # indices relative to the moment origin
+                if morigin:
```
FreddieWitherden: This probably should not be part of a mixin as it is highly specific to fluid force

FreddieWitherden:
> > We also do not need to use a mixin. A freestanding class may be more convenient (so it takes a list of faces to integrate over and handles the rest). Fluid force can then subclass this to handle moments and then create an instance of the new class. Not sure what is cleaner but we probably need to do something. Also, see the FWH plugin for another case where we have similar code duplication.
> 
> Yeah the FWH plugin does have similarities at points but its setup and handling of the norms/weights diverges from the the fluidforce/mass flow controller. Would the idea be to rewrite it to handle things in a similar way to fluidforce?
> 
> And would the new class handle the actual integration itself or keep to just the setup/initialisation?

Only setup I would have thought.

FreddieWitherden:
Also looping in @WillTrojak as he might have some opinions here.

TobyFlynn:
> Looking through I am no longer convinced that quadrules is the right place given all of the implied dependencies (system, elements classes, operator matrices). Maybe it is better off in solvers/base?

Moving it to solvers/base ended up creating a circular import: integrators.base wants get_plugin from base.plugins -> base.plugins wants SurfaceMixin from solvers.base -> solvers wants get_integrator from integrators. I don't know if there is somewhere else it would belong?

pyfr/plugins/fluidforce.py
```diff
@@ -63,65 +64,45 @@ def __init__(self, intg, cfgsect, suffix):
                 case _:
                     raise ValueError('Invalid file format')
 
-        # Interpolation matrices and quadrature weights
-        self._m0 = m0 = {}
-        self._qwts = qwts = defaultdict(list)
-
+        # Set interpolation matrices and quadrature weights
+        surf_list = []
+        if suffix in mesh.bcon:
```
FreddieWitherden: Try `mesh.bcon.get(suffix, [])`

pyfr/quadrules/__init__.py
```diff
@@ -102,3 +106,58 @@ class StoredQuadRule(BaseStoredQuadRule):
             ndim = ndims[eletype]
 
         return StoredQuadRule(rule, npts, qdeg, flags)
+
+
+class SurfaceMixin:
```
FreddieWitherden: Consider making this its own class rather than a mixin.

pyfr/plugins/fluidforce.py
```diff
@@ -63,65 +64,45 @@ def __init__(self, intg, cfgsect, suffix):
...
+                        ploc = self._locs[etype, fidx][i]
+                        rfpt = ploc - morigin
+                        rfpts[etype, fidx].append(rfpt)
+        
             self._rfpts = {k: np.array(v) for k, v in rfpts.items()}
-
             if self._viscous:
-                self._rcpjact = {k: rcpjact[k[0]][..., v]
-                                 for k, v in self._eidxs.items()}
+                self._rcpjact = {k: rcpjact[k[0]][..., v]
```
FreddieWitherden: Remove trailing space.

FreddieWitherden:
Your current approach looks pretty good now, although maybe we should move the class out from __init__.py to something like surface.py inside pyfr.quadrules. Seems to be the best compromise.

TobyFlynn:
Is this what you were thinking for a freestanding class rather than a mixin? What I have is a little clunky for fwh but works well for the fluidforce plugin

pyfr/plugins/fwh.py
```diff
@@ -59,15 +113,16 @@ def __init__(self, intg, cfgsect, suffix=None, *args, **kwargs):
...
-            raise ValueError('FWH farfield Mach number greater than 1')
 
         # Initialise surface data
         ele_map = intg.system.ele_map
         self.emap = {k: i for i, k in enumerate(ele_map)}
         self.ele_surface, _ = self._surf_region(intg)
-        self._init_surf(ele_map, Minf)
+
+        surf_list = [(s[1], s[3], s[2]) for s in self.ele_surface]
+        self.fwh_int = FWHIntegrator(self.cfg, self.cfgsect, self.ndims,
```
FreddieWitherden: I think quite a few of these instance variables can be made local (or already are like cfgsect).

pyfr/plugins/fwh.py
```diff
@@ -59,15 +113,16 @@ def __init__(self, intg, cfgsect, suffix=None, *args, **kwargs):
...
-        if Minf >= 1:
-            raise ValueError('FWH farfield Mach number greater than 1')
 
         # Initialise surface data
         ele_map = intg.system.ele_map
         self.emap = {k: i for i, k in enumerate(ele_map)}
         self.ele_surface, _ = self._surf_region(intg)
-        self._init_surf(ele_map, Minf)
+
+        surf_list = [(s[1], s[3], s[2]) for s in self.ele_surface]
```
FreddieWitherden: Unpack if you can

pyfr/plugins/fwh.py
```diff
@@ -14,7 +15,60 @@
 )
 
 
-class FWHPlugin(SurfaceMixin, BaseSolnPlugin):
+class FWHIntegrator(SurfaceIntegrator):
+    def __init__(self, cfg, cfgsect, ndims, obsv_pts, qinf, elemap, surf_list):
+        self.cfg = cfg
```
FreddieWitherden: Are these assignments needed?

pyfr/plugins/fluidforce.py
```diff
@@ -3,11 +3,56 @@
...
+                    # Get the flux points position of the given face and 
+                    # element indices relative to the moment origin
+                    if morigin is not None:
+                        ploc = self._locs[etype, fidx][i]
+                        rfpt = ploc - morigin
+                        rfpts[etype, fidx].append(rfpt)
+        
+            self._rfpts = {k: np.array(v) for k, v in rfpts.items()}
+            if viscous:
+                self._rcpjact = {k: rcpjact[k[0]][..., v]
```
FreddieWitherden: Probably should be _ if they are going to be accessed outside the class.

pyfr/plugins/fluidforce.py
```diff
@@ -3,11 +3,56 @@
...
+
+        self._surf_init(system.ele_map, surf_list, flags='s')
+
+        if viscous:
+            self._m4 = m4 = {}
+            rcpjact = {}
+
+        if suffix in mesh.bcon:
+            rfpts = defaultdict(list)
+            for etype, fidx in self._eidxs.keys():
```
FreddieWitherden: No need for .keys()

pyfr/plugins/fluidforce.py
```diff
@@ -3,11 +3,56 @@
...
                                init_csv, open_hdf5_a)
+from pyfr.quadrules.surface import SurfaceIntegrator
 
 
-class FluidForcePlugin(SurfaceMixin, BaseSolnPlugin):
+class FluidForceIntegrator(SurfaceIntegrator):
+    def __init__(self, cfg, cfgsect, system, suffix, viscous, morigin):
+        self.cfg = cfg
+        self.cfgsect = cfgsect
+        mesh = system.mesh
```
FreddieWitherden: Check which of these assignments are needed and what can be moved inline

pyfr/plugins/fwh.py
TobyFlynn: cfg and cfgsect are used to get the quadrature rule in SurfaceIntegrator._surf_quad. ndims, obsv_pts and qinf are used by _distances() - I could instead pass them as arguments to the function instead?

FreddieWitherden: Is the a nice solution where we can call `SurfaceIntegrator`'s constructor?  This is good hygiene and avoids an implicit dependency.

TobyFlynn: Yeah that makes a lot more sense, sorry

pyfr/quadrules/surface.py
```diff
@@ -0,0 +1,64 @@
...
+        except NoOptionError:
+            qdeg = self.cfg.getint(self.cfgsect, 'quad-deg')
+
+        # Get the quadrature rule
+        q = get_quadrule(itype, qrule, qdeg=qdeg, flags=flags)
+
+        # Project its points onto the provided surface
+        pts = np.atleast_2d(q.pts.T)
+
+        return np.vstack(np.broadcast_arrays(*proj(*pts))).T, q.wts
```
FreddieWitherden: Missing newline

pyfr/plugins/fwh.py
```diff
@@ -29,9 +79,9 @@ def __init__(self, intg, cfgsect, suffix=None, *args, **kwargs):
         self.tstart = self.cfg.getfloat(cfgsect, 'tstart', 0.0)
         self.t_last = -np.inf
         self.dt = self.cfg.getfloat(cfgsect, 'dt')
-        self.obsv_pts = np.array(self.cfg.getliteral(self.cfgsect,
-                                                     'observer-pts'))
-        self.nobvs = len(self.obsv_pts)
+        self.obsv_pts = obsv_pts = np.array(self.cfg.getliteral(self.cfgsect,
```
FreddieWitherden: Try to avoid double storing things (if the integrator keeps it as an attribute then probably not worth duplicating here)

pyfr/plugins/fwh.py
```diff
@@ -45,29 +95,28 @@ def __init__(self, intg, cfgsect, suffix=None, *args, **kwargs):
...
+        qinf['M'] = np.array([qinf[k] / qinf['c'] for k in 'uvw'[:self.ndims]])
 
         # Initialise surface data
         ele_map = intg.system.ele_map
         self.emap = {k: i for i, k in enumerate(ele_map)}
-        self.ele_surface, _ = self._surf_region(intg)
-        self._init_surf(ele_map, Minf)
+        ele_surf, _ = self._surf_region(intg)
+
+        surf_list = [(etype, eidx, fidx) for _, etype, fidx, eidx in ele_surf]
```
FreddieWitherden: Odd ordering (element type, list of indices, face number).  eidx should be plural and the ordering from ele_surf is more natural.

TobyFlynn: I was following the ordering I was getting from `system.mesh.bcon.get(suffix, [])` but I've now switched to (element type, list of indices, face number). I'm reordering in `FluidForceIntegrator` as system is already being passed to the constructor, is this alright or would reordering in the `FluidForcePlugin` constructor before initialising the integrator be better?

FreddieWitherden: `mesh.bcon` is in a slightly different format as if I recall it is a list of scalars (etype, eidx, fidx) whereas the surface stuff effectively does a group-by operation on (etype, fidx): eidx**s**.  The latter is more appropriate for what we're doing.

pyfr/plugins/fluidforce.py
```diff
@@ -3,11 +3,54 @@
...
 from pyfr.mpiutil import get_comm_rank_root, mpi
-from pyfr.plugins.base import (BaseSolnPlugin, DatasetAppender, SurfaceMixin,
+from pyfr.plugins.base import (BaseSolnPlugin, DatasetAppender,
                                init_csv, open_hdf5_a)
+from pyfr.quadrules.surface import SurfaceIntegrator
 
 
-class FluidForcePlugin(SurfaceMixin, BaseSolnPlugin):
+class FluidForceIntegrator(SurfaceIntegrator):
+    def __init__(self, cfg, cfgsect, system, suffix, viscous, morigin):
```
FreddieWitherden: Rename suffix

pyfr/plugins/fluidforce.py
```diff
@@ -3,11 +3,54 @@
...
+        surf_list = system.mesh.bcon.get(suffix, [])
+        surf_list = [(etype, fidx, eidxs) for etype, eidxs, fidx in surf_list]
+        
+        super().__init__(cfg, cfgsect, system.ele_map, surf_list, flags='s')
+
+        if viscous:
+            self.m4 = m4 = {}
+            rcpjact = {}
+
+        if suffix in system.mesh.bcon:
```
FreddieWitherden: if surf_list

PR #484, opened 2025-07-22
Unused entmin

FrankFrank9:
It seems to be unused throughout the code.


FreddieWitherden:
@tarikdzanic Do you recall why these accessors were added?

PR #488, opened 2025-07-31
Avoid using NumPy's tofile method as it can cause issues with Lustre.

FreddieWitherden:


tarikdzanic (approved):
Tested and it fixed the writing issues on Lustre.

PR #490, opened 2025-08-04
Refactoring CSV writing into a class

TobyFlynn:
This adds a `CSVWrapper` class (happy to rename if there is a better suggestion) to pyfr.writers.csv that handles some of what is in `init_csv` and also flushing to file. `init_csv` in pyfr.plugins.base now creates and returns a `CSVWrapper` class.

Let me know if there was a better way to do this or if I should also pull the `DatasetAppender` class outside of pyfr.plugins.base (I don't have any plans to use it in the mass flow controller BC, I just wasn't sure if it should live in the same place as `CSVWrapper`).

### DIFF
--- a/pyfr/plugins/base.py
+++ b/pyfr/plugins/base.py
@@ -10,7 +10,7 @@ from pytools import prefork
 
 from pyfr.mpiutil import get_comm_rank_root, mpi
 from pyfr.regions import parse_region_expr
-from pyfr.writers.csv import CSVWriter
+from pyfr.writers.csv import CSVStream
 
 
 def cli_external(meth):
@@ -21,14 +21,15 @@ def cli_external(meth):
     return classmethod(newmeth)
 
 
-def init_csv(cfg, cfgsect, header, *, filekey='file', headerkey='header'):
+def init_csv(cfg, cfgsect, header, *, filekey='file', headerkey='header', 
+             nflush=None):
     # Determine the file path
     fname = cfg.get(cfgsect, filekey)
 
     header = header if cfg.getbool(cfgsect, headerkey, True) else None
-    nflush = cfg.getint(cfgsect, 'flushsteps', 10)
+    nflush = cfg.getint(cfgsect, 'flushsteps', nflush or 10)
 
-    return CSVWriter(fname, header=header, nflush=nflush)
+    return CSVStream(fname, header=header, nflush=nflush)
 
 
 def open_hdf5_a(path):
--- a/pyfr/plugins/base.py
+++ b/pyfr/plugins/base.py
@@ -10,6 +10,7 @@ from pytools import prefork
 
 from pyfr.mpiutil import get_comm_rank_root, mpi
 from pyfr.regions import parse_region_expr
+from pyfr.writers.csv import CSVWrapper
 
 
 def cli_external(meth):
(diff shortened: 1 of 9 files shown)


pyfr/plugins/base.py
```diff
@@ -20,23 +21,14 @@ def newmeth(cls, args):
...
-def init_csv(cfg, cfgsect, header, *, filekey='file', headerkey='header'):
+def init_csv(cfg, cfgsect, header, *, filekey='file', headerkey='header', 
+             nflush=100):
     # Determine the file path
     fname = cfg.get(cfgsect, filekey)
 
-    # Append the '.csv' extension
-    if not fname.endswith('.csv'):
-        fname += '.csv'
+    _header = header if cfg.getbool(cfgsect, headerkey, True) else None
```
FreddieWitherden: No need for _

pyfr/plugins/dtstats.py
```diff
@@ -22,9 +20,11 @@ def __init__(self, intg, cfgsect, prefix):
 
         # The root rank needs to open the output file
         if rank == root:
-            self.outf = init_csv(self.cfg, cfgsect, 'n,t,dt,action,error')
+            nflush = self.cfg.getint(self.cfgsect, 'flushsteps', 500)
```
FreddieWitherden: Add this to `init_csv`

pyfr/writers/csv.py
```diff
@@ -0,0 +1,25 @@
+
+
+class CSVWrapper:
```
FreddieWitherden: CSVWriter

pyfr/writers/csv.py
```diff
@@ -0,0 +1,25 @@
...
+        self.outf = open(fname, 'a')
+
+        # Output a header if required
+        if self.outf.tell() == 0 and header:
+            print(header, file=self.outf)
+        
+        self.nprint = 0
+        self.nflush = nflush
+    
+    def print(self, *args, **kwargs):
```
FreddieWitherden: Rename to write and have it just accept a list and then have it pass sep=',' (as we're writing a CSV)

pyfr/writers/csv.py
FreddieWitherden: Another option is CSVStream since it is specifically for streaming (rather than one-shot writing).

pyfr/writers/csv.py
```diff
@@ -0,0 +1,25 @@
+
+
+class CSVWrapper:
+    def __init__(self, fname, header=None, nflush=100):
```
FreddieWitherden: fname, *,

pyfr/writers/csv.py
FreddieWitherden: Actually, ``__call__`` might be cleaner.

pyfr/plugins/base.py
```diff
@@ -21,14 +21,14 @@ def newmeth(cls, args):
     return classmethod(newmeth)
 
 
-def init_csv(cfg, cfgsect, header, *, filekey='file', headerkey='header', 
-             nflush=100):
+def init_csv(cfg, cfgsect, header, *, filekey='file', headerkey='header'):
```
FreddieWitherden: Have nflush as an optional argument so it can be overridden by the caller.

PR #493, opened 2025-08-11
Adding documentation for the mass flow controller

TobyFlynn:


doc/src/boundary-initial-conditions/soln-bcs.rst
```diff
@@ -89,6 +89,73 @@ dependent, boundary condition labelled *name* in the .pyfrm file with
...
+           *float* | *string*
+
+        - ``w`` --- z-velocity
+
+           *float* | *string*
+
+        - ``p`` --- initial static pressure, the controller will vary this to
+          target a mass flow rate.
+
+           *float* | *string*
```
FreddieWitherden, line 115: Can this be a string?

TobyFlynn: Sorry, I thought cfg.getfloat automatically evaluated constants/expressions in the ini file - I now see that is handled differently, I'll correct this.

PR #494, opened 2025-08-14
Pre-compute sqrt(rhos) in hll and hllc

kaschau:
Pre-compute sqrt(rho) in hll and hllc.

### DIFF
--- a/pyfr/solvers/euler/kernels/rsolvers/hll.mako
+++ b/pyfr/solvers/euler/kernels/rsolvers/hll.mako
@@ -8,6 +8,8 @@
     fpdtype_t pl, pr;
     fpdtype_t va[${ndims}];
     fpdtype_t nf_sub, nf_fl, nf_fr;
+    fpdtype_t sqrtrl = sqrt(ul[0]);
+    fpdtype_t sqrtrr = sqrt(ur[0]);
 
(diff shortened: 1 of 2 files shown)


FreddieWitherden:
Does this make a difference?  Compilers in my experience are very good at factoring out sub-expressions.

kaschau:
not that I can detect reliably, was assuming it helped as in rhoinv pre-calculation. Compiler may just be picking it up.

FreddieWitherden:
One thing we can do is move `inv_rar` up and use it to eliminate a division.

pyfr/solvers/euler/kernels/rsolvers/hll.mako
```diff
@@ -20,17 +22,17 @@
     fpdtype_t ar = sqrt(${c['gamma']}*pr/ur[0]);
 
     // Compute the Roe-averaged velocity
-    fpdtype_t nv = (sqrt(ul[0])*nvl + sqrt(ur[0])*nvr)
-                 / (sqrt(ul[0]) + sqrt(ur[0]));
+    fpdtype_t nv = (sqrtrl*nvl + sqrtrr*nvr)
+                 / (sqrtrl + sqrtrr);
```
FreddieWitherden, line 26: Can use inv_rar here if we move it up.  Not sure if the other changes are needed.

PR #495, opened 2025-08-26
Support for variable time stepping

sambitmishra98:
This PR adds support for variable physical time stepping for both std and dual controllers. This is a continuation of PR #335 (refer to it for more details) with cleaner commits and few bug fixes. I have used this feature has been used for ~2 years for all my tests, see no issues in it at this point.


FreddieWitherden:
Aspects of this are in #542

PR #497, opened 2025-09-02
Use corrected gradients in fluidforce

FrankFrank9:
This PR follows from https://pyfr.discourse.group/t/fluidforce-with-intg-grad-soln/1292/4

### DIFF
--- a/pyfr/plugins/fluidforce.py
+++ b/pyfr/plugins/fluidforce.py
@@ -14,10 +14,7 @@ class FluidForceIntegrator(SurfaceIntegrator):
         super().__init__(cfg, cfgsect, system.ele_map, surf_list, flags='s')
 
         if surf_list and morigin is not None:
-            self.rfpts = {
-                (etype, fidx): self.locs[etype, fidx] - morigin
-                for etype, fidx in self.locs
-            }
+            self.rfpts = {k: loc - morigin for k, loc in self.locs.items()}
 
 class FluidForcePlugin(BaseSolnPlugin):
     name = 'fluidforce'


FreddieWitherden:
I think you need to move the `intg.grad_soln` bit outside the for loop otherwise the code is liable to deadlock if a rank in the simulation does not contain part of the boundary.

You should also be able to remove a lot of the `.m4` related stuff as it should no longer be needed.

FreddieWitherden:
> > I think you need to move the `intg.grad_soln` bit outside the for loop otherwise the code is liable to deadlock if a rank in the simulation does not contain part of the boundary.
> 
> In that case `self.ff_int.m0` is empty dict hence it should not step into the code

Can you confirm this with a simple problem?

FreddieWitherden:
Just following up here

pyfr/plugins/fluidforce.py
```diff
@@ -162,6 +137,7 @@ def __call__(self, intg):
 
         # Solution matrices indexed by element type
         solns = dict(zip(intg.system.ele_types, intg.soln))
+        cgrads = dict(zip(intg.system.ele_types, intg.grad_soln))
```
FreddieWitherden: I think we want this in an if statement otherwise the plugin will no longer work for Euler.

pyfr/plugins/fluidforce.py
```diff
@@ -9,45 +9,24 @@
...
-                        # Get |J|^-1 at the solution points
-                        rcpdjac = eles.rcpdjac_at_np('upts')
-
-                        # Product to give J^-T at the solution points
-                        rcpjact[etype] = smat*rcpdjac
-
-                    # Get the flux points position of the given face and 
+                    # Get the flux points position of the given face and
                     # element indices relative to the moment origin
                     if morigin is not None:
```
FreddieWitherden: I think we can move this if block outside of the loops (so `if surf_list and morigin is not None` or something in that direction)

pyfr/plugins/fluidforce.py
```diff
@@ -162,6 +137,8 @@ def __call__(self, intg):
 
         # Solution matrices indexed by element type
         solns = dict(zip(intg.system.ele_types, intg.soln))
+        if self._viscous: # Corrected solution gradients indexed by element type
```
FreddieWitherden: Move comment up/remove.

pyfr/plugins/fluidforce.py
FrankFrank9: Please check by running the for loops, they should work but I am still troubleshooting

pyfr/plugins/fluidforce.py
```diff
@@ -9,46 +9,20 @@
...
-            if viscous:
-                self.rcpjact = {k: rcpjact[k[0]][..., v]
-                                for k, v in self.eidxs.items()}
+        super().__init__(cfg, cfgsect, system.ele_map, surf_list, flags='s')
 
+        if surf_list and morigin is not None:
+            self.rfpts = {
+                (etype, fidx): np.array([
+                    self.locs[etype, fidx][i] - morigin
+                    for i, _ in enumerate(self.eidxs[etype, fidx])
```
FreddieWitherden: I think this loop can be simplified if we loop over locs instead and then avoid looping over eidx's entirely.

PyFR-Test-Cases
```diff
@@ -0,0 +1 @@
+Subproject commit e741571456e0f7bfba88250d0025710364778a68
```
FreddieWitherden: Should probably remove this.

pyfr/plugins/fluidforce.py
```diff
@@ -108,8 +79,7 @@ def __init__(self, intg, cfgsect, suffix):
                     raise ValueError('Invalid file format')
 
         # Set interpolation matrices and quadrature weights
-        self.ff_int = FluidForceIntegrator(self.cfg, cfgsect, intg.system,
-                                           suffix, self._viscous, morigin)
+        self.ff_int = FluidForceIntegrator(self.cfg, cfgsect, intg.system, suffix, morigin)
```
FreddieWitherden: Line length.

PR #498, opened 2025-09-18
Improve the stability of rocblas auto tuning.

z-yuan131:
This PR deals with the possible invalid value error raised by the rocblas during the auto tuning process. The changes add exceptions while benchmarking and ignore invalid solutions.

### DIFF
--- a/pyfr/backends/hip/rocblas.py
+++ b/pyfr/backends/hip/rocblas.py
@@ -151,12 +151,12 @@ class HIPRocBLASKernels(HIPKernelProvider):
                 except RocBLASInvalidValue:
                     pass
             
+            # Restore the output matrix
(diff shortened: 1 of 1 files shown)


pyfr/backends/hip/rocblas.py
```diff
@@ -143,9 +143,17 @@ def get_solutions(sidx):
...
+                try:
+                    dt = self._benchmark(gemm)
+                    if best_kern is None or dt < ifac*best_kern[-1]:
+                        best_kern = algo, dt
+                # In the case of invalid values raised by rocblas
+                except RocBLASInvalidValue:
+                    pass
+            
+            # If all tests fail
+            if best_kern is None:
```
FreddieWitherden, line 158: Move this down so it is after the restore output matrix block

PR #500, opened 2025-09-30
Dynamic macros

kaschau:
This PR proposes a new style of macro creation I am calling dynamic macros, `dmacros`. This allows creation of macros that utilize python data in their definition that may change from expansion to expansion. An example may be a mat vec mult in which mat is a numpy array, and we create a standard macro to use the data passed to the macro at the point of macro expansion. To do this there are a few mako quirks. Most notably, we cannot capture the output of the macro when we define the macro, as this is where all the python code is executed. Therefore we have to simply store the macro body for later capture. Second, the `@support_caller` decorator allows for `args` specification in the dmacro definition, this allows for arguments to be passed to the macro body callable. These arguments are the python variables. 

This allows for more general macros using python data, as one does not need to have consistent global names (i.e. invvdm) available to all macros, and allows for intra-kernel calculations to take place, and be passed to a macro. Here is an example `dmacro` definition:

```
<%namespace module='pyfr.backends.base.makoutil' name='pyfr'/>

<%pyfr:dmacro name='simple' params='y, x, off' args='A_matrix, B_matrix'>
    // Matrix-vector multiply: y = A_matrix @ x
    <%
    print("A_matrix is accessible:", A_matrix, flush=True)
    print("B_matrix is accessible:", B_matrix, flush=True)
    %>
    fpdtype_t test = off;
    // Test comment: A_matrix shape is ${A_matrix.shape}
    // Test comment: B_matrix shape is ${B_matrix.shape}
    y[0] = x[0];  // Dummy text
</%pyfr:dmacro>
```

And `dexpand` usage:

```
<%inherit file='base'/>
<%namespace module='pyfr.backends.base.makoutil' name='pyfr'/>
<%include file='ilp_gemm_data_simple'/>

<%
se0 = math.log10(c['s0']/order**4)
%>

<%doc>
Optimized shocksensor kernel using matrix multiplication macros.
Each backend provides its own optimized mm_gemv implementation that:
- Automatically detects and optimizes for sparsity in the inverse Vandermonde matrix
- Uses backend-specific parallelization strategies
For OpenMP: Uses ndim='1-ilp' to enable nested parallelism within elements.
For CUDA/HIP: Uses warp/wavefront cooperation for inner-loop parallelism.
</%doc>

<%pyfr:kernel name='test' ndim='1'
              blah='in fpdtype_t[${str(nupts)}][${str(nvars)}]'
              blahblah='out fpdtype_t'>
    <%
    import numpy as np
    test = np.ones((3,3))
    %>

    ## NOTE: invvdm is a tplarg passed to the kernel, test is created within the kernel
    ${pyfr.dexpand('mm_gemv_simple', 'modes', 'u_col', invvdm, test, off=1)}

</%pyfr:kernel>
```

The usage of kwargs follows the standard macro/expand where the kw is intended to be a string substituted param.

### DIFF
--- a/pyfr/backends/base/kernels.py
+++ b/pyfr/backends/base/kernels.py
@@ -74,6 +74,9 @@ class BasePointwiseKernelProvider(BaseKernelProvider):
         # Macro definitions
         tplargs['_macros'] = {}
 
+        # Dynamic macro definitions
+        tplargs['_dmacros'] = {}
+
         # External kernel arguments dictionary
         tplargs['_extrns'] = extrns
 
--- a/pyfr/backends/base/makoutil.py
+++ b/pyfr/backends/base/makoutil.py
@@ -1,4 +1,5 @@
 from collections.abc import Iterable
+import inspect
 import itertools as it
 import re
 
@@ -179,6 +180,109 @@ def kernel(context, name, ndim, **kwargs):
 
     # Render and return the complete kernel
     return kern.render()
+
+
+@supports_caller
+def dmacro(context, name, params, externs=''):
+    """
+    Define a dynamic macro with Python data parameters.
+
+    Unlike regular macros where Python executes at definition time, dmacros
+    allow Python data to be passed at expansion time. This enables reusable
+    constructs that accept runtime Python data (e.g., numpy arrays) while
+    supporting C variable string substitution.
+
+    Syntax: <%pyfr:dmacro name='...' params='c_vars' args='py_vars'>
+
+    Args:
+        name: Macro name
+        params: Comma-separated C variable names (string substitution)
+        externs: Comma-separated external variable names
+    """
+    if name in context['_dmacros']:
+        raise RuntimeError(f'Attempt to redefine dmacro "{name}"')
+
+    # Parse C parameters and external variables
+    params = [p.strip() for p in params.split(',')]
+    externs = [e.strip() for e in externs.split(',')] if externs else []
+
+    # Extract Python data parameters from caller signature
+    sig = inspect.signature(context['caller'].body).parameters
+    dparams = [p.name for p in sig.values()]
+
+    # Validate all parameter names
+    for p in it.chain(params, externs, dparams):
+        if not re.match(r'[A-Za-z_]\w*$', p):
+            raise ValueError(f'Invalid param "{p}" in dmacro "{name}"')
+
+    # Store for later expansion
+    context['_dmacros'][name] = {
+        'params': params,
+        'externs': externs,
+        'dparams': dparams,
+        'template_callable': context['caller'].body
+    }
+
+    return ''
+
+
+def dexpand(context, name, /, *args, **kwargs):
+    """
+    Expand a dynamic macro with Python data and C variable substitution.
+
+    Retrieves the dmacro, calls its template callable with Python data,
+    then performs string substitution for C variables.
+
+    Args:
+        name: Dmacro name
+        *args: C param values followed by Python data values (positional)
+        **kwargs: C param and Python data values (named)
+    """
+    # Retrieve the dmacro definition
+    if '_dmacros' not in context.keys() or name not in context['_dmacros']:
+        raise ValueError(f'Dmacro "{name}" not defined')
+
+    dmacro_info = context['_dmacros'][name]
+    params = dmacro_info['params']
+    externs = dmacro_info['externs']
+    dparams = dmacro_info['dparams']
+    template_callable = dmacro_info['template_callable']
+
+    # Parse arguments (C params first, then Python data params)
+    all_params = params + dparams
+    all_args = dict(zip(all_params, args))
+    for k, v in kwargs.items():
+        if k in all_args:
+            raise ValueError(f'Duplicate parameter {k} in dmacro {name}')
+        all_args[k] = v
+
+    # Validate all parameters provided
+    if sorted(all_params) != sorted(all_args.keys()):
+        raise ValueError(f'Inconsistent parameter list in dmacro {name}')
+
+    # Separate C variable substitutions from Python data
+    c_subs = {k: all_args[k] for k in params}
+    py_data = {k: all_args[k] for k in dparams}
+
+    # Call template callable with Python data and capture output
+    body = capture(context, template_callable, **py_data)
+
+    # Suffix local variable declarations to avoid conflicts
+    lvars = _locals(body)
+    if lvars:
+        body = re.sub(r'\b({0})\b'.format('|'.join(lvars)), r'\1_', body)
+
+    # Validate external variables are available
+    for extrn in externs:
+        if (extrn not in context['_extrns'].keys() and
+            re.search(rf'\b{extrn}\b', body)):
+            raise ValueError(f'Missing external {extrn} in {name}')
+
+    # Perform string substitution for C variables
+    for pname, pval in c_subs.items():
+        body = re.sub(rf'\b{pname}\b', str(pval), body)
+
+    return f'{{\n{body}\n}}'
 
 
 def alias(context, name, func):
--- a/pyfr/backends/base/makoutil.py
+++ b/pyfr/backends/base/makoutil.py
@@ -243,46 +243,50 @@ def dexpand(context, name, /, *args, **kwargs):
         raise ValueError(f'Dmacro "{name}" not defined')
 
     dmacro_info = context['_dmacros'][name]
-    params = dmacro_info['params']
-    externs = dmacro_info['externs']
-    dparams = dmacro_info['dparams']
+    mparams = dmacro_info['params']
+    mexterns = dmacro_info['externs']
+    mdparams = dmacro_info['dparams']
     template_callable = dmacro_info['template_callable']
 
-    # Parse arguments (C params first, then Python data params)
-    all_params = params + dparams
-    all_args = dict(zip(all_params, args))
-    for k, v in kwargs.items():
-        if k in all_args:
-            raise ValueError(f'Duplicate parameter {k} in dmacro {name}')
-        all_args[k] = v
+    # Validate argument count
+    # C params can come from args or kwargs, Python data must be positional
+    num_c_provided = len(kwargs)  # kwargs go to C params
+    num_py_provided = 0
 
-    # Validate all parameters provided
-    if sorted(all_params) != sorted(all_args.keys()):
-        raise ValueError(f'Inconsistent parameter list in dmacro {name}')
+    # Count how many positional args are for C params vs Python data
+    # Python data params come after we've satisfied C params
+    if len(args) <= len(mparams) - num_c_provided:
+        # All positional args are C params
+        num_c_provided += len(args)
+    else:
+        # Some positional args are C params, rest are Python data
+        num_c_provided += len(mparams) - len(kwargs)
+        num_py_provided = len(args) - (len(mparams) - len(kwargs))
 
-    # Separate C variable substitutions from Python data
-    c_subs = {k: all_args[k] for k in params}
-    py_data = {k: all_args[k] for k in dparams}
+    if num_c_provided != len(mparams):
+        raise ValueError(f'Expected {len(mparams)} C parameters in dmacro {name}')
+    if num_py_provided != len(mdparams):
+        raise ValueError(f'Expected {len(mdparams)} data parameters in dmacro {name}')
+
+    # Split positional args between C params and Python data
+    c_positional_count = len(mparams) - len(kwargs)
+    params = args[:c_positional_count]
+    dparams = args[c_positional_count:]
+
+    # Build Python data dict for template callable
+    py_data = dict(zip(mdparams, dparams))
 
     # Call template callable with Python data and capture output
     body = capture(context, template_callable, **py_data)
 
-    # Suffix local variable declarations to avoid conflicts
-    lvars = _locals(body)
-    if lvars:
-        body = re.sub(r'\b({0})\b'.format('|'.join(lvars)), r'\1_', body)
+    # Create temporary macro and delegate to expand() for processing
+    # Pass both positional C params and any kwargs (e.g., off=1)
+    temp_name = f'_dmacro_temp_{id(body)}'
+    context['_macros'][temp_name] = (mparams, mexterns, body)
+    result = expand(context, temp_name, *params, **kwargs)
+    del context['_macros'][temp_name]
 
-    # Validate external variables are available
-    for extrn in externs:
-        if (extrn not in context['_extrns'].keys() and
-            re.search(rf'\b{extrn}\b', body)):
-            raise ValueError(f'Missing external {extrn} in {name}')
-
-    # Perform string substitution for C variables
-    for pname, pval in c_subs.items():
-        body = re.sub(rf'\b{pname}\b', str(pval), body)
-
-    return f'{{\n{body}\n}}'
+    return result
 
 
 def alias(context, name, func):


kaschau:
> Is there a reason not to try and make all such macros dynamic in this sense? As in can we make dmacros replace all regular macros?

Not that I can think of, except we would then allow macros to be defined multiple times, as in every time a macro is expanded it will be re-captured. Though I suppose we could detect dynamic variables and define/capture once.

FreddieWitherden:
> > Is there a reason not to try and make all such macros dynamic in this sense? As in can we make dmacros replace all regular macros?
> 
> Not that I can think of, except we would then allow macros to be defined multiple times, as in every time a macro is expanded it will be re-captured. Though I suppose we could detect dynamic variables and define/capture once.

I think we'd probably want to sort out the logic to avoid the internal temporary redefinition by pulling some functions out and refactoring as appropriate.

pyfr/backends/base/kernels.py
```diff
@@ -74,6 +74,9 @@ def _render_kernel(self, name, mod, extrns, tplargs):
         # Macro definitions
         tplargs['_macros'] = {}
 
+        # Dynamic macro definitions
+        tplargs['_dmacros'] = {}
```
FreddieWitherden: Is this still needed?

pyfr/backends/base/makoutil.py
```diff
@@ -1,8 +1,9 @@
 from collections.abc import Iterable
+import inspect
 import itertools as it
 import re
 
-from mako.runtime import supports_caller, capture
+from mako.runtime import supports_caller, capture, Undefined
```
FreddieWitherden: Capital first.

pyfr/backends/base/makoutil.py
```diff
@@ -98,47 +99,73 @@ def macro(context, name, params, externs=''):
...
+    dparams = [p.name for p in sig.values()]
+
+    # Validate all parameter names
     for p in it.chain(params, externs):
         if not re.match(r'[A-Za-z_]\w*$', p):
-            raise ValueError(f'Invalid param "{p}" in macro "{name}"')
-
-    # Capture the function body
-    body = capture(context, context['caller'].body)
+            raise ValueError(f'Invalid param "{p}" in dmacro "{name}"')
```
FreddieWitherden: Should it be macro rather than dmacro?

pyfr/backends/base/makoutil.py
```diff
@@ -98,47 +99,73 @@ def macro(context, name, params, externs=''):
...
-    lvars = _locals(body)
-
-    # Suffix these variables by a '_'
-    if lvars:
-        body = re.sub(r'\b({0})\b'.format('|'.join(lvars)), r'\1_', body)
-
-    # Save
-    context['_macros'][name] = (params, externs, body)
+    # Store for later capture
+    context['_macros'][name] = (
```
FreddieWitherden: Use a namedtuple here.

pyfr/backends/base/makoutil.py
```diff
@@ -98,47 +99,73 @@ def macro(context, name, params, externs=''):
...
+
+    # Validate argument count
+    # Params can use args or kwargs; dparams must be positional
+    nparams_kw = len(kwargs)
+    nparams_pos = len(mparams) - nparams_kw
+    ndparams_pos = len(args) - nparams_pos
+
+    # Ensure correct counts
+    if nparams_pos + nparams_kw != len(mparams):
+        emps = f'Expected {len(mparams)} parameters in {name}'
```
FreddieWitherden: Move into the call (but break with a suitable newline; see some other examples).

pyfr/backends/base/makoutil.py
```diff
@@ -98,47 +99,73 @@ def macro(context, name, params, externs=''):
...
     for k, v in kwargs.items():
         if k in params:
             raise ValueError(f'Duplicate macro parameter {k} in {name}')
 
         params[k] = v
 
+    # Check all python data is defined
+    dparams = dict(zip(mdparams, args[nparams_pos:]))
+    for k, v in dparams.items():
+        if type(v) == Undefined:
```
FreddieWitherden: Check if we can do a direct comparison, otherwise prefer isinstance over type.

pyfr/backends/base/makoutil.py
```diff
@@ -98,47 +99,73 @@ def macro(context, name, params, externs=''):
...
+            raise ValueError(f'Undefined data parameter {k} passed to {name}')
```
FreddieWitherden: Quote "" for consistency with other exceptions.

pyfr/backends/base/makoutil.py
kaschau: check now

pyfr/backends/base/makoutil.py
kaschau: cant do direct due to numpy arrays true value error

pyfr/backends/base/makoutil.py
```diff
@@ -1,8 +1,10 @@
 from collections.abc import Iterable
+import inspect
 import itertools as it
 import re
+from collections import namedtuple
```
FreddieWitherden: Move up so alphabetical.

pyfr/backends/base/makoutil.py
```diff
@@ -1,8 +1,10 @@
...
 
-from mako.runtime import supports_caller, capture
+from mako.runtime import Undefined, supports_caller, capture
```
FreddieWitherden: Sort alphabetically (think this is a pre-existing issue).

pyfr/backends/base/makoutil.py
```diff
@@ -98,52 +100,80 @@ def macro(context, name, params, externs=''):
...
-    # Identify any local variable declarations
-    lvars = _locals(body)
-
-    # Suffix these variables by a '_'
-    if lvars:
-        body = re.sub(r'\b({0})\b'.format('|'.join(lvars)), r'\1_', body)
-
-    # Save
-    context['_macros'][name] = (params, externs, body)
+    Macro = namedtuple('Macro', ['params', 'externs', 'dparams', 'caller'])
```
FreddieWitherden: Move Macro up to global scope.

pyfr/backends/base/makoutil.py
```diff
@@ -98,52 +100,80 @@ def macro(context, name, params, externs=''):
...
+    # Store for later capture
+    context['_macros'][name] = Macro(
```
FreddieWitherden: Should be able to reduce the number of newlines here.

pyfr/backends/base/makoutil.py
```diff
@@ -98,52 +100,80 @@ def macro(context, name, params, externs=''):
...
-    # Get the macro parameter list and the body
-    mparams, mexterns, body = context['_macros'][name]
 
-    # Ensure an appropriate number of arguments have been passed
-    if len(mparams) != len(args) + len(kwargs):
-        raise ValueError(f'Inconsistent macro parameter list in {name}')
-
-    # Parse the parameter list
-    params = dict(zip(mparams, args))
+    mparams = context['_macros'][name].params
```
FreddieWitherden: Might be easier to assign `macro = context['_macros'][name]` and use this rather than extra variables.

kaschau: ok to redefine macro (def macro exists) or find tweaked name?

FreddieWitherden: Probably use a different name.

FreddieWitherden:
It would be nice if we had uniform syntax support (so Python arguments can be passed by name).  Also, we might want to rethink the syntax for the definition a little since args and params is a bit confusing.  Maybe an argument is Python if prefixed with `py:`.  Would be backwards compatible.

kaschau:
> It would be nice if we had uniform syntax support (so Python arguments can be passed by name). Also, we might want to rethink the syntax for the definition a little since args and params is a bit confusing. Maybe an argument is Python if prefixed with `py:`. Would be backwards compatible.

Are you looking to unify the macro definition to a single list of "args" (merge current params and args distinction?). `args` is a Mako thing, it is defined by the @supports_caller decorator I believe, so things defined as `args` in the macro must be passed into the capture upon expansion. That is why we need the `inspect.signature` deal to get their names, they are not explicitly passed to the `macro` function.

kaschau:
The `args` distinction is necessary as far as I can tell. It serves to tell the macro callable to expect arguments when it is run (now in expand). If you tried to pass the py:d variable to the caller it would error with "unexpected argument passed". It defines the callable's signature.

FreddieWitherden:
Right, but we should be able to construct the args string ourself and pass it to Mako (or otherwise obtain the relevant callable).

kaschau:
So this would be something like, get the params list and separate on the backend, then get a string of the mako file, remove the python arguments from the params list in the macro def string, then inject the python args into an `args = ...` segment of the macro file string, create a new Template from the modified string, then save THAT template's callable to the kernel's context for expansion later. Is that what you are thinking?

pyfr/backends/base/makoutil.py
```diff
@@ -88,62 +91,122 @@ def _locals(body):
...
-    # Ensure no invalid characters in params/extern variables
     for p in it.chain(params, externs):
         if not re.match(r'[A-Za-z_]\w*$', p):
-            raise ValueError(f'Invalid param "{p}" in macro "{name}"')
+            raise ValueError(f'Invalid parameter name "{p}"')
 
-    # Capture the function body
-    body = capture(context, context['caller'].body)
+    # Extract pyparams from callable signature
+    pyparams = list(signature(context['caller'].body).parameters.keys())
```
FreddieWitherden: See if the signature.bind method can help.

pyfr/backends/base/kernels.py
```diff
@@ -71,8 +71,9 @@ def _render_kernel(self, name, mod, extrns, tplargs):
         # Backend-specfic generator classes
         tplargs['_kernel_generator'] = self.kernel_generator_cls
 
-        # Macro definitions
+        # Macro definitions and IDs
         tplargs['_macros'] = {}
+        tplargs['_macro_ids'] = {}
```
FreddieWitherden: Can we not store the id in the macros entry?

pyfr/backends/base/makoutil.py
```diff
@@ -1,8 +1,11 @@
+from collections import namedtuple
 from collections.abc import Iterable
 import itertools as it
 import re
 
-from mako.runtime import supports_caller, capture
+from inspect import signature
```
FreddieWitherden: Move inspect up as it is part of Python.

pyfr/backends/base/makoutil.py
```diff
@@ -88,62 +91,105 @@ def _locals(body):
...
+        elif k in mparams:
+            paramskw[k] = v
+        else:
+            raise ValueError(f'Unknown parameter "{k}" in macro "{name}"')
+
+    # Build params dict from positional args and kwargs
+    # Positional args fill in params first, then any remaining go to pyparams
+    nparamspos = len(mparams) - len(paramskw)
+    params = dict(zip(mparams, args[:nparamspos]))
+    params.update(paramskw)
```
FreddieWitherden: dict(paramskw, zip(...))

pyfr/backends/base/makoutil.py
```diff
@@ -88,62 +91,105 @@ def _locals(body):
...
+
+    # Check we got all params
+    if len(params) != len(mparams):
+        raise ValueError(f'Incomplete or duplicate parameters in "{name}"')
+
+    # Parse pyparams
+    pyparamspos = args[nparamspos:]
```
FreddieWitherden: Can we inline this?

pyfr/template.py
```diff
@@ -28,9 +30,66 @@ def get_template(self, name):
...
+        src = self._format_macro_def(src)
+
         # Subclass Template to support implicit arguments
         class DefaultTemplate(Template):
             def render(iself, *args, **kwargs):
                 return super().render(*args, **self.dfltargs, **kwargs)
 
         return DefaultTemplate(src, lookup=self)
+
+    def _inject_macro_id_registration(self, source):
```
FreddieWitherden: Avoid having this code in Template if possible.  Rather add a more generic way of adding in a filter to template and then have the init function in backend register a suitable function.  Keeps the Template code more generic.

pyfr/template.py
```diff
@@ -28,9 +30,66 @@ def get_template(self, name):
...
+            macro = match.group(1)
+            name = match.group(2)
+            macroid = hashlib.md5(macro.strip().encode()).hexdigest()
+            ids[name] = macroid
+
+        return ids
+
+    def _format_macro_def(self, source):
+        # Transform py:params into args attribute for <%pyfr:macro> tags
+        pattern = r'(<%pyfr:macro\s+name=[\'"][^\'"]+[\'"]\s+params=[\'"])([^\'"]+)([\'"](?:\s+externs=[\'"][^\'"]*[\'"])?(?:\s+\w+=[\'"][^\'"]*[\'"])*\s*>)'
```
FreddieWitherden: Break this up.  Have one line to match the macro definition then another which searches that to extract the rest (say looking for attributes with an re.findall).  This will let attributes be defined in any order.

pyfr/backends/base/makoutil.py
```diff
@@ -1,8 +1,11 @@
+from collections import namedtuple
 from collections.abc import Iterable
 import itertools as it
 import re
 
-from mako.runtime import supports_caller, capture
+from inspect import signature
+from mako.runtime import Undefined, capture, supports_caller
+from mako.template import Template
```
FreddieWitherden: Do we still need this?

pyfr/template.py
```diff
@@ -28,9 +30,66 @@ def get_template(self, name):
...
+            prefix = match.group(1)
+            params = [p.strip() for p in match.group(2).split(',')]
+            suffix = match.group(3)
+
+            # Partition params into regular and py: prefixed
+            pyparams = [p[3:] for p in params if p.startswith('py:')]
+            params = [p for p in params if not p.startswith('py:')]
+
+            if not pyparams:
+                return match.group(0)
```
FreddieWitherden: Can use [0] rather than group.

pyfr/backends/base/kernels.py
kaschau: yes, currently it stores the callable by name in the dict, do you want to store as a tuple like (callable, id)?

kaschau: in Macro now and no longer injecting python code but computing as part of 'filtering' operation and sending to the macro caller as an attribute. much cleaner

pyfr/template.py
kaschau: let me know what you think of structure

pyfr/backends/base/makoutil.py
```diff
@@ -88,62 +91,149 @@ def _locals(body):
...
+
+        # Extract and update attributes from the opening tag
+        attrs = dict(re.findall(apattern, match[1]))
+        attrs['id'] = macid
+
+        # Reconstruct the opening tag and append the body
+        attrstr = ' '.join(f'{k}="{v}"' for k, v in attrs.items())
+        return f'<%pyfr:macro {attrstr}>{match[2]}'
+
+    mpattern = r'(<%pyfr:macro\s+[^>]+>)(.*?</%pyfr:macro>)'
```
FreddieWitherden, line 120: Try using lookbehind and lookahead here.

pyfr/backends/base/makoutil.py
```diff
@@ -88,62 +91,149 @@ def _locals(body):
...
 
+Macro = namedtuple('Macro', ['params', 'externs', 'argsig', 'caller', 'id'])
+
+
+def mfiltid(source):
+    apattern = r'(\w+)=[\'"]([^\'"]*)[\'"]'
+
+    def add_id(match):
+        # Compute hash of the entire macro definition
+        macid = hashlib.sha256(match[0].encode('utf-8')).hexdigest()
```
FreddieWitherden: I think we already have a hash function in utils which does a lot of this?

pyfr/backends/base/makoutil.py
```diff
@@ -88,62 +91,149 @@ def _locals(body):
...
+                             f'passed to "{name}"')
+
+    return params, pyparams
 
 
 def expand(context, name, /, *args, **kwargs):
-    # Get the macro parameter list and the body
-    mparams, mexterns, body = context['_macros'][name]
+    macrodef = context['_macros'][name]
+    mparams = macrodef.params
```
FreddieWitherden: Can probably inline quite a few of these variables.

pyfr/template.py
```diff
@@ -28,6 +32,13 @@ def get_template(self, name):
         if not src:
             raise RuntimeError(f'Template "{name}" not found')
 
+        # Decode bytes to string
+        src = src.decode('utf-8')
```
FreddieWitherden: Just .decode() will work here as 'utf-8' is default.

FreddieWitherden:
Can you also update: https://github.com/PyFR/PyFR/blob/develop/doc/src/developer_guide.rst#pyfr-mako-macros

pyfr/backends/base/makoutil.py
FreddieWitherden: This should clean up the code a little.

pyfr/backends/base/makoutil.py
```diff
@@ -103,47 +143,73 @@ def macro(context, name, params, externs=''):
...
+        unknown = unknown.pop()
+        raise ValueError(f'Unknown parameter "{unknown}" in macro "{name}"')
+
+    paramskw = {k: v for k, v in kwargs.items() if k in mparams}
+    pyparamskw = {k: v for k, v in kwargs.items() if k in margs}
+
+    # Build params dict from positional args and kwargs
+    # Positional args fill in params first, then any remaining go to pyparams
+    nparamspos = len(mparams) - len(paramskw)
+    params = dict(paramskw, **dict(zip(mparams, args[:nparamspos])))
```
FreddieWitherden: dict(zip(...), **paramskw)

pyfr/backends/base/makoutil.py
```diff
@@ -103,47 +143,73 @@ def macro(context, name, params, externs=''):
...
+        raise ValueError(f'Incomplete or duplicate parameters in "{name}"')
+
+    # Parse pyparams
+    try:
+        bound = margsig.bind(*args[nparamspos:], **pyparamskw)
+        pyparams = dict(bound.arguments)
+    except TypeError as e:
+        raise ValueError(f'Invalid Python data parameters in "{name}": {e}')
+
+    # Check all python data is defined
```
FreddieWitherden: Python

pyfr/backends/base/makoutil.py
```diff
@@ -88,13 +90,51 @@ def _locals(body):
...
+
+        # Add the ID attribute
+        attrs['id'] = macid
+
+        # Reconstruct the opening tag and append the body
+        attrstr = ' '.join(f'{k}="{v}"' for k, v in attrs.items())
+        return f'<%pyfr:macro {attrstr}>{match[2]}'
+
+    mpattern = r'(<%pyfr:macro\s+[^>]+>)(.*?</%pyfr:macro>)'
+    return re.sub(mpattern, process_tag, source, flags=re.DOTALL)
```
FreddieWitherden: re.S

pyfr/backends/base/makoutil.py
```diff
@@ -88,13 +90,51 @@ def _locals(body):
...
 
+Macro = namedtuple('Macro', ['params', 'externs', 'argsig', 'caller', 'id'])
+
+
+def mfilttag(source):
+    apattern = r'(\w+)=[\'"]([^\'"]*)[\'"]'
+
+    def process_tag(match):
+        # Compute hash of the entire macro definition
+        macid = util.digest(match[0])
```
FreddieWitherden: Can we inline this?

pyfr/template.py
```diff
@@ -8,6 +8,10 @@ class DottedTemplateLookup(TemplateLookup):
     def __init__(self, pkg, dfltargs):
         self.dfltpkg = pkg
         self.dfltargs = dfltargs
+        self._filters = []
```
FreddieWitherden: Can just make this a non _ variable and then remove the method

PR #502, opened 2025-10-06
[WIP] Use a different partition for plugins. 

sambitmishra98:
This PR adds an opt-in path to redistribute element-indexed data from the solver’s mesh partitioning to a separate plugin partitioning. Set partition = <name> in a plugin’s (e.g., writer or tavg) config block to load that partition for the plugin; the runtime will build an interconnector from the compute mesh to the plugin mesh and relocate array data across ranks before outputting to disk. If partition is omitted, behavior is unchanged and data is written using the compute partition. For now, ensure all writer/tavg plugins reference the same partition name.


FreddieWitherden:
It can still be in the ini file, just handled by BasePlugin.  Key thing is all plugins should work with zero changes needed.

PR #503, opened 2025-10-07
Add better kernel rendering error capture.

kaschau:
This PR saves approximately 1000 hours of future development time by providing better errors within kernel renders. It adds the `strict_undefined=True` argument to the Mako Template so undefined Python variable access immediately raises a NameError and provides the variable name.

Further, because macros can are dynamic now, one macro expansion can render successfully, and another not. So this provides a "call path" from kernel through nested macro expansions to the final offending macro.

### DIFF
--- a/pyfr/backends/base/makoutil.py
+++ b/pyfr/backends/base/makoutil.py
@@ -199,11 +199,6 @@ def _parse_expand_args(name, mparams, margsig, args, kwargs):
     except TypeError as e:
         raise MacroError(name, f'Invalid Python data parameters: {e}')
 
-    # Check all Python data is defined
-    for k, v in pyparams.items():
-        if isinstance(v, Undefined):
-            raise MacroError(name, f'Undefined Python data parameter "{k}"')
-
     return params, pyparams
 
 
--- a/pyfr/backends/base/makoutil.py
+++ b/pyfr/backends/base/makoutil.py
@@ -11,6 +11,29 @@ import pyfr.nputil as nputil
 import pyfr.util as util
 
 
+class MacroError(Exception):
+    def __init__(self, mname, error):
+        if isinstance(error, MacroError):
+            self.path = [mname, *error.path]
+            self.error = error.error
+        else:
+            self.path = [mname]
+            self.error = error
+        super().__init__(mname)
+
+
+class KernelError(Exception):
+    def __init__(self, kname, error):
+        if isinstance(error, MacroError):
+            path = ' -> '.join(it.chain([kname], error.path))
+            msg = str(error.error)
+        else:
+            path = kname
+            msg = str(error)
+
+        super().__init__(f'\n  {path}: {msg}')
+
+
 def ndrange(context, *args):
     return util.ndrange(*args)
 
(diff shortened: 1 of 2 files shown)


pyfr/backends/base/makoutil.py
```diff
@@ -11,6 +11,29 @@
...
 
+class MacroError(Exception):
+    def __init__(self, mname, error):
+        if isinstance(error, MacroError):
+            self.path = [mname, *error.path]
+            self.error = error.error
+        else:
+            self.path = [mname]
+            self.error = error
+        super().__init__(mname)
```
FreddieWitherden: Add newline

pyfr/backends/base/makoutil.py
```diff
@@ -11,6 +11,29 @@
...
+class KernelError(Exception):
+    def __init__(self, kname, error):
+        if isinstance(error, MacroError):
+            path = ' -> '.join(it.chain([kname], error.path))
+            msg = str(error.error)
+        else:
+            path = kname
+            msg = str(error)
+
+        super().__init__(f'\n  {path}: {msg}')
```
FreddieWitherden: Newline is a bit unusual here.

kaschau: The custom exception displays the full path of the exception, then we display the call path and error message after that it gets cluttered and a very long line. 


`pyfr.backends.base.makoutil.KernelError: shocksensor -> test_macro -> test_helper: 'undefined_var' is not defined`

kaschau: I couldnt figure out how to display just `KernelError`

FreddieWitherden:
I thought the exception group class could represent that?  (Or an arbitrary hierarchy.)

FreddieWitherden:
```python
>>> raise ExceptionGroup("Foo", [ExceptionGroup("bar", [ValueError("baz")])])
  + Exception Group Traceback (most recent call last):
  |   File "<python-input-2>", line 1, in <module>
  |     raise ExceptionGroup("Foo", [ExceptionGroup("bar", [ValueError("baz")])])
  | ExceptionGroup: Foo (1 sub-exception)
  +-+---------------- 1 ----------------
    | ExceptionGroup: bar (1 sub-exception)
    +-+---------------- 1 ----------------
      | ValueError: baz
      +------------------------------------
```

pyfr/backends/base/makoutil.py
```diff
@@ -154,8 +154,8 @@ def _parse_expand_args(name, mparams, margsig, args, kwargs):
 
     # Separate kwargs into params and Python data params
     if unknown := set(kwargs) - set(mparams) - set(margs):
-        unknown = unknown.pop()
-        raise ValueError(f'Unknown parameter "{unknown}" in macro "{name}"')
+        raise ExceptionGroup(f'In macro: {name}',
+                           [ValueError(f'Unknown param "{unknown.pop()}"')])
```
FreddieWitherden: Might as well use a list comprehension here to nab all of the unknown params and raise a value error for each one rather than just popping.

pyfr/backends/base/makoutil.py
```diff
@@ -167,20 +167,15 @@ def _parse_expand_args(name, mparams, margsig, args, kwargs):
 
     # Check we got all params
     if len(params) != len(mparams):
-        raise ValueError(f'Incomplete or duplicate parameters in "{name}"')
+        raise ExceptionGroup(f'In macro: {name}',
+                           [ValueError('Incomplete or duplicate parameters')])
```
FreddieWitherden: Indentation.

PR #504, opened 2025-10-07
Using mako preprocessor for filter stack

kaschau:
Here and implementation using the preprocessor argument instead of manually calling filter on the src string.

### DIFF
--- a/pyfr/template.py
+++ b/pyfr/template.py
@@ -35,13 +35,10 @@ class DottedTemplateLookup(TemplateLookup):
         if not src:
             raise RuntimeError(f'Template "{name}" not found')
 
-        # Decode bytes to string for preprocessing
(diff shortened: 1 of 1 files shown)


pyfr/template.py
```diff
@@ -29,16 +35,13 @@ def get_template(self, name):
...
-        # Apply all filters to source
-        for filter in self.filters:
-            src = filter(src)
-
         # Subclass Template to support implicit arguments
         class DefaultTemplate(Template):
             def render(iself, *args, **kwargs):
                 return super().render(*args, **self.dfltargs, **kwargs)
 
-        return DefaultTemplate(src, lookup=self)
```
FreddieWitherden, line 44: Think we can do src.decode() here when we pass

PR #505, opened 2025-10-17
Making sure that dt_type is always defined for reduction mako kernel

TobyFlynn:
Currently, `dt_type` is only defined if `method == 'resid'` which trips the `strict_undefined=True` check in Mako (even if `dt_type` would never be used in that particular rendering of the template). This makes sure it always set to something (i.e. `None` when it is not relevant). Otherwise you get an error when running the 3d-triangular-aerofoil test case.


pyfr/backends/cuda/blasext.py
```diff
@@ -78,6 +78,7 @@ def reduction(self, *rs, method, norm, dt_mat=None):
 
         tplargs = dict(norm=norm, method=method)
 
+        tplargs['dt_type'] = None
```
FreddieWitherden: Can we move this up in the dict(...) call?

Issue #508, opened 2025-11-15
Python 3.14 Metal backend crash


SpeedyTurtle599:
Dear PyFR developers,

I've found a segmentation fault when using the Metal backend on macOS 26.1. My system specs are:

- PyFR: 2.1
- Python Version: 3.14.0
- Platform: macOS 26.1 (Build 25B78)
- Architecture: Apple Silicon (arm64)
- GPU: Apple M3 Max
- Metal Framework: pyobjc-framework-metal 12.1

I was able to reproduce this issue across multiple test cases. For example, for the Euler vortex test case, I tweaked the existing case to use single precision and then ran with:

```bash
uv run pyfr --progress --verbose run -b metal euler-vortex.pyfrm euler-vortex.ini
```

The results, without metal validation:
```
[BGC-MBP14:65473] *** Process received signal ***
[BGC-MBP14:65473] Signal: Segmentation fault: 11 (11)
[BGC-MBP14:65473] Signal code: Invalid permissions (2)
[BGC-MBP14:65473] Failing at address: 0x3a
...
[BGC-MBP14:65473] [ 1] 0   AGXMetalG15X_M1    ...  _ZN3AGX14ComputeContextINS_3G158EncodersE...setPipelineCommonEPNS1_15ComputePipelineE + 52
[BGC-MBP14:65473] [ 2] 0   AGXMetalG15X_M1    ...  -[AGXG15XFamilyComputeContext setComputePipelineState:] + 68

Exit code: 139 (SIGSEGV)
```

With metal validation enabled (running with `MTL_DEBUG_LAYER=1 MTL_SHADER_VALIDATION=1`):

```
2025-11-15 14:04:41.752 Python[67799:11603386] Metal API Validation Enabled
2025-11-15 14:04:41.752 Python[67799:11603386] Metal GPU Validation Enabled
-[MTLDebugComputeCommandEncoder setComputePipelineState:]:345: failed assertion `computePipelineState is not a MTLComputePipelineState.'
```

Happy to provide any further details and test as necessary.

Thank you!

FreddieWitherden:
This is likely a pyobjc-framework-metal issue.  You can try the following snippet:
```python
from Metal import (MTLCreateSystemDefaultDevice, MTLCompileOptions,
                   MTLComputePipelineDescriptor)


src = """kernel void test(device float* v, uint i [[thread_position_in_grid]])
{ if (i < 32) v[i] = i; }
"""

def main():
[...]

PR #509, opened 2025-11-15
Resolve Python 3.14 Metal backend segfault #508

SpeedyTurtle599:
Addresses #508.

The Metal backend crashes with a segmentation fault when attempting to run simulations on macOS. The crash occurs during compute pipeline state initialization with the error:

```
-[MTLDebugComputeCommandEncoder setComputePipelineState:]:345: failed assertion `computePipelineState is not a MTLComputePipelineState.'
```

After a bit of research, I believe PyObjC follows a convention where Objective-C methods with `error:` output parameters (typed as `NSError**`) return a tuple `(result, error)` in Python, rather than just the result. 

The method `newComputePipelineStateWithDescriptor:error:` has the signature:
```objc
- (id<MTLComputePipelineState>)newComputePipelineStateWithDescriptor:(MTLComputePipelineDescriptor*)descriptor
                                                                error:(NSError**)error;
```

In PyObjC, this returns `(pipeline_state, error)` as a tuple. The current version assigns the entire tuple to `cpsf`:
```python
# Obtain the corresponding compute pipeline
cpsf = call_(self.backend.dev, 'newComputePipelineStateWith',
             descriptor=desc, error=None)
if cpsf is None:
    raise RuntimeError('Unable to create compute pipeline state')
```

When this tuple is later passed to `setComputePipelineState_`, Metal validation correctly rejects it because it's not a `MTLComputePipelineState` object. The fix is to properly unpack the tuple return value from `newComputePipelineStateWithDescriptor_error_`:
```python
# Obtain the corresponding compute pipeline
cpsf, err = call_(self.backend.dev, 'newComputePipelineStateWith',
                  descriptor=desc, error=None)
if cpsf is None:
    raise RuntimeError('Unable to create compute pipeline state')
```

Now `cpsf` correctly receives only the pipeline state object, and `err` receives the error (which will be `None` on success or `NSError_object` on failure).

I asked GitHub Copilot to walk me through the function signatures to verify this fix. Its output is paraphrased here in case anyone would like to see:

### PyObjC Error Handling Convention

PyObjC automatically transforms Objective-C methods with output error parameters:

```python
# Method signature metadata shows:
# arg3: b'o^@' (OUTPUT parameter, null_accepted=True)
# The 'o' prefix indicates an output parameter

# All methods ending in _error_ that have NSError** parameters:
result, error = obj.someMethod_error_(arg, None)
```

This is documented in the [PyObjC documentation](https://pyobjc.readthedocs.io/en/latest/api/module-objc.html#objective-c-output-parameters). Without Metal validation enabled (`MTL_DEBUG_LAYER=1`), the crash manifests as a generic segmentation fault. The tuple being passed has the correct pipeline state as its first element, so shallow checks might not catch it before runtime.

Type Encoding Reference:
@   = object (id/NSObject*)
:   = selector (SEL)
^@  = pointer to object
o^@ = OUTPUT pointer to object (error parameter)
i   = int
f   = float
d   = double
v   = void

### DIFF
--- a/pyfr/backends/metal/provider.py
+++ b/pyfr/backends/metal/provider.py
@@ -84,9 +84,9 @@ class MetalKernelProvider(BaseKernelProvider):
 
         # Obtain the corresponding compute pipeline
         cpsf, err = call_(self.backend.dev, 'newComputePipelineStateWith',
-                     descriptor=desc, error=None)
-        if cpsf is None:
-            raise RuntimeError('Unable to create compute pipeline state')
+                          descriptor=desc, error=None)
+        if err is not None:
+            raise ValueError(f'Pipeline creation error: {err}')
 
         # Classify the arguments as either pointers or scalars
         pargs, sargs = [], []
--- a/setup.py
+++ b/setup.py
@@ -121,7 +121,7 @@ install_requires = [
 
 # Soft dependencies
 extras_require = {
-    'metal': ['pyobjc-framework-Metal >= 9.0']
+    'metal': ['pyobjc-framework-Metal >= 12.0']
 }
 
 # Scripts


pyfr/backends/metal/provider.py
```diff
@@ -83,7 +83,7 @@ def _build_kernel(self, name, src, argtypes, argn=[]):
         desc.setThreadGroupSizeIsMultipleOfThreadExecutionWidth_(True)
 
         # Obtain the corresponding compute pipeline
-        cpsf = call_(self.backend.dev, 'newComputePipelineStateWith',
+        cpsf, err = call_(self.backend.dev, 'newComputePipelineStateWith',
                      descriptor=desc, error=None)
         if cpsf is None:
```
SpeedyTurtle599: Could take advantage of the err being available to parse, and borrow some code from earlier in the file:
```
if err is not None:
            raise ValueError(f'Pipeline creation error: {err}')
```

pyfr/backends/metal/provider.py
```diff
@@ -83,7 +83,7 @@ def _build_kernel(self, name, src, argtypes, argn=[]):
         desc.setThreadGroupSizeIsMultipleOfThreadExecutionWidth_(True)
 
         # Obtain the corresponding compute pipeline
-        cpsf = call_(self.backend.dev, 'newComputePipelineStateWith',
+        cpsf, err = call_(self.backend.dev, 'newComputePipelineStateWith',
                      descriptor=desc, error=None)
```
FreddieWitherden: Indentation.

FreddieWitherden:
Can you confirm all of the test cases run (in single precision).

SpeedyTurtle599:
Per your request, I ran each test case with `-b metal` and `precision=single`. I've listed them below, indicating which ones I ran to completion and which ones I aborted a few steps in for my laptop battery's sake:
- 2d-couette-flow (tested to completion, walltime 2 mins 2 seconds)
- 2d-double-mach-reflection (waited for 4 outputs, estimated total walltime 13 mins 56 seconds)
- 2d-euler-vortex (completed, walltime 11 seconds)
- 2d-inc-cylinder (completed, walltime 4 mins 34 seconds)
- 2d-viscous-shock-tube (waited for 4 outputs, est total walltime 9 mins 28 seconds)
- 3d-taylor-green (waited for 4 outputs, est total walltime 22 mins 49 seconds)
- 3d-triangular-airfoil (waited for 4 outputs, est total walltime 1 hour 20 mins)

If you have any existing data for me to compare against, I'm happy to provide outputs or compare. That said, I hope this shouldn't impact correctness at all, since this seems to be an architectural issue rather than an actual math issue.

PR #527, opened 2026-02-06
A fix to the possible point sampler fail

z-yuan131:
I found a issue when I tried to use sampler, there are a few points cannot be founded.
I think this related to the bounding box where the expansion is too aggressive. The expansion should be related to the element size rather than the element location. The error disappeared after this fix.

### DIFF
--- a/pyfr/points.py
+++ b/pyfr/points.py
@@ -125,7 +125,7 @@ class PointLocator:
         smin, smax = spts.min(axis=0), spts.max(axis=0)
 
         # Expand by the scale factor to better account for strong curvature
-        expand = (0.5*(scale - 1))*(smax - smin)
+        expand = (scale - 1)*(smax - smin)
         smin -= expand
         smax += expand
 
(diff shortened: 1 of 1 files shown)


FreddieWitherden:
Can you provide a case which fails with the current code?  Expanding the box too far should never cause a failure (we get more points returned, but they'll be quickly tested and discarded).

z-yuan131:
[mesh_sample.zip](https://github.com/user-attachments/files/25130747/mesh_sample.zip)
Here is an example and the point list where the program could fail. 
In the code expand = factor*(smax + smin)/2 where expand itself can be a negative number depends on the mesh coordinate. Another possible fix is to have expand = factor*abs(smax + smin)/2

FreddieWitherden:
Confirmed.  I think the 0.5 should still be there though?

z-yuan131:
expand = (scale - 1)*(smax - smin) gives 5% of the element size as the expansion to each side. I am not sure if it is too tight with factor 0.5. A wider box is safer.

PR #529, opened 2026-02-07
Soln plugin stop

hahnpv:
Added a stop plugin. Issuing a 'touch stop' in the working folder will gracefully stop the simulation.

### DIFF
--- a/pyfr/plugins/stop.py
+++ b/pyfr/plugins/stop.py
@@ -1,0 +1,40 @@
+import os
+from pyfr.plugins.base import BaseSolnPlugin
+
+class StopPlugin(BaseSolnPlugin):
+    name = 'stop'
+    systems = ['*']
+    formulations = ['dual', 'std']
+    dimensions = [2, 3]
+
+    def __init__(self, intg, cfgsect, suffix=None):
+        super().__init__(intg, cfgsect, suffix)
+        # Check for this file to trigger a graceful stop
+        self.stop_file = self.cfg.get(cfgsect, 'file', 'STOP')
+
+    def __call__(self, intg):
+        if os.path.exists(self.stop_file):
+            # The root rank handles the notification and save triggering
+            from pyfr.mpiutil import get_comm_rank_root
+            comm, rank, root = get_comm_rank_root()
+            
+            if rank == root:
+                print(f"Graceful stop requested via {self.stop_file}")
+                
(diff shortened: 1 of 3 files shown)


pyfr/plugins/stop.py
```diff
@@ -0,0 +1,40 @@
...
+    formulations = ['dual', 'std']
+    dimensions = [2, 3]
+
+    def __init__(self, intg, cfgsect, suffix=None):
+        super().__init__(intg, cfgsect, suffix)
+        # Check for this file to trigger a graceful stop
+        self.stop_file = self.cfg.get(cfgsect, 'file', 'STOP')
+
+    def __call__(self, intg):
+        if os.path.exists(self.stop_file):
```
FreddieWitherden, line 16: This likely isn't safe in parallel.  Possible one rank will see a file before another does and cause a deadlock.

pyfr/plugins/stop.py
```diff
@@ -0,0 +1,40 @@
...
+            comm, rank, root = get_comm_rank_root()
+            
+            if rank == root:
+                print(f"Graceful stop requested via {self.stop_file}")
+                
+            # Try to trigger a final save if a writer exists
+            for p in intg.plugins:
+                if getattr(p, 'name', None) == 'writer':
+                    # Force the writer to output now
+                    p.tout_last = intg.tcurr - p.dt_out - 1e-12
```
FreddieWitherden, line 28: Hacking other plugins is fragile; probably want a more generic mechanism here which enables plugins to decide themselves what they want to do on abort rather than someone else deciding for them.

pyfr/plugins/stop.py
```diff
@@ -0,0 +1,40 @@
...
+class StopPlugin(BaseSolnPlugin):
+    name = 'stop'
+    systems = ['*']
+    formulations = ['dual', 'std']
+    dimensions = [2, 3]
+
+    def __init__(self, intg, cfgsect, suffix=None):
+        super().__init__(intg, cfgsect, suffix)
+        # Check for this file to trigger a graceful stop
+        self.stop_file = self.cfg.get(cfgsect, 'file', 'STOP')
```
FreddieWitherden, line 13: A signal may be cleaner.

FreddieWitherden:
Just following up here.

PR #530, opened 2026-02-07
plugin extern binding callback plumbing

kaschau:
This PR adds callback infrastructure so plugins can add extrns to kernels, and provides an API for plugins to register and update bound extrn values in the `__call__`

### DIFF
--- a/pyfr/backends/base/kernels.py
+++ b/pyfr/backends/base/kernels.py
@@ -188,11 +188,8 @@ class BasePointwiseKernelProvider(BaseKernelProvider):
             # Process the argument list
             argb, argm, argv = self._build_arglst(dims, argn, argt, kwargs)
 
-            # Instantiate and annotate with runtime-bindable arg names
-            kern = self._instantiate_kernel(dims, fun, argb, argm, argv)
-            kern.rtnames = frozenset(k for k in argb if isinstance(k, str))
-
-            return kern
+            # Return a Kernel subclass instance
+            return self._instantiate_kernel(dims, fun, argb, argm, argv)
 
         # Attach the module to the method as an attribute
         kernel_meth._mod = mod
--- a/pyfr/backends/base/kernels.py
+++ b/pyfr/backends/base/kernels.py
@@ -7,7 +7,7 @@ from pyfr.cache import memoize
 
 class Kernel:
     compound = False
-    rtnames = frozenset()
+    rtnames = ()
 
     def __init__(self, mats=[], views=[], misc=[], dt=float('nan')):
         self.mats = mats
--- a/pyfr/backends/base/kernels.py
+++ b/pyfr/backends/base/kernels.py
@@ -7,6 +7,7 @@ from pyfr.cache import memoize
 
 class Kernel:
     compound = False
+    rtnames = frozenset()
 
     def __init__(self, mats=[], views=[], misc=[], dt=float('nan')):
         self.mats = mats
@@ -187,8 +188,11 @@ class BasePointwiseKernelProvider(BaseKernelProvider):
             # Process the argument list
             argb, argm, argv = self._build_arglst(dims, argn, argt, kwargs)
 
-            # Return a Kernel subclass instance
-            return self._instantiate_kernel(dims, fun, argb, argm, argv)
+            # Instantiate and annotate with runtime-bindable arg names
+            kern = self._instantiate_kernel(dims, fun, argb, argm, argv)
+            kern.rtnames = frozenset(k for k in argb if isinstance(k, str))
+
+            return kern
 
         # Attach the module to the method as an attribute
         kernel_meth._mod = mod
--- a/pyfr/plugins/base.py
+++ b/pyfr/plugins/base.py
@@ -148,6 +148,9 @@ class BasePlugin:
     def setup(self, sdata, serialiser):
         pass
 
+    def bind_externs(self):
+        pass
+
     def get_serialiser_prefix(self):
         if self.suffix:
             return f'plugins/{self.name}-{self.suffix}'
@@ -181,7 +184,7 @@ class BaseSolverPlugin(BasePlugin):
     def _extern_callback(self, kern):
         self._extern_binders.append(kern.bind)
 
-    def _bind_externs(self):
+    def bind_externs(self):
         for b in self._extern_binders:
             b(**self._extern_values)
 
--- a/pyfr/plugins/base.py
+++ b/pyfr/plugins/base.py
@@ -179,9 +179,11 @@ class BaseSolverPlugin(BasePlugin):
         intg.system.register_kernel_callback(names, self._extern_callback)
 
     def _extern_callback(self, kern):
-        if kern.bind not in self._extern_binders:
-            self._extern_binders.append(kern.bind)
-            kern.bind(**self._extern_values)
+        self._extern_binders.append(kern.bind)
+
+    def _bind_externs(self):
+        for b in self._extern_binders:
+            b(**self._extern_values)
 
 
 class BaseCLIPlugin:
--- a/pyfr/plugins/base.py
+++ b/pyfr/plugins/base.py
@@ -148,9 +148,6 @@ class BasePlugin:
     def setup(self, sdata, serialiser):
         pass
 
-    def bind_externs(self):
-        pass
-
     def get_serialiser_prefix(self):
         if self.suffix:
             return f'plugins/{self.name}-{self.suffix}'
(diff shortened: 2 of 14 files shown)


pyfr/plugins/base.py
```diff
@@ -162,6 +162,36 @@ class BaseSolnPlugin(BasePlugin):
...
+    def _extern_callback(self, kern):
+        if bind := getattr(kern, 'bind', None):
+            if bind not in self._extern_binders:
+                self._extern_binders.append(bind)
+                bind(**self._extern_values)
+
+    def _update_extern_values(self):
+        pass
+
+    def _bind_externs(self):
```
FreddieWitherden: Who calls this?

pyfr/solvers/base/system.py
```diff
@@ -240,6 +243,11 @@ def tag_kern(pname, prov, kern):
 
                         tag_kern(pn, p, kern)
 
+        for kerns in kernels.values():
```
FreddieWitherden: Do we want to go through every bindable kernel here?  Probably not a huge number, but with some minor changes we might be able to find out what kernels are bindable and to what argument names.

pyfr/plugins/base.py
kaschau: the plugin during its `__call__` when it is ready to update its extrns (via its defined _update_extern_values method it defines)

pyfr/solvers/base/system.py
kaschau: see if this looks right

pyfr/backends/base/kernels.py
```diff
@@ -187,8 +188,11 @@ def kernel_meth(self, tplargs, dims, extrns={}, **kwargs):
             # Process the argument list
             argb, argm, argv = self._build_arglst(dims, argn, argt, kwargs)
 
-            # Return a Kernel subclass instance
-            return self._instantiate_kernel(dims, fun, argb, argm, argv)
+            # Instantiate and annotate with runtime-bindable arg names
+            kern = self._instantiate_kernel(dims, fun, argb, argm, argv)
+            kern.rtnames = frozenset(k for k in argb if isinstance(k, str))
```
FreddieWitherden: We might try to do it at a lower level so that it can potentially work for non pointwise kernels (but this would be per backend).

pyfr/solvers/base/system.py
```diff
@@ -261,8 +269,8 @@ def _get_kernels(self, uinbank, foutbank):
         binders, bckerns = [], defaultdict(dict)
         for kn, kerns in kernels.items():
             for k in kerns:
-                if bind := getattr(k, 'bind', None):
-                    binders.append(bind)
+                if k.rtnames:
```
FreddieWitherden, line 281: So we should be able to do slightly better; we know what externs each plugin is providing so can match against the rtarg names and only give it the relevant kernels.  This way we only call bind on kernels which have the argument.  (And can probably catch name clashes in the process.)

kaschau: check now

pyfr/backends/base/kernels.py
```diff
@@ -7,6 +7,7 @@
 
 class Kernel:
     compound = False
+    rtnames = frozenset()
```
FreddieWitherden: I'd just use a tuple as the number of args is small.

pyfr/plugins/base.py
```diff
@@ -162,6 +162,42 @@ class BaseSolnPlugin(BasePlugin):
...
+        self._extern_values = {}
+        self._extern_binders = []
+
+    def _register_externs(self, intg, names, spec='scalar fpdtype_t'):
+        self._update_extern_values()
+
+        names = frozenset(names)
+
+        # Check for extern name clashes with other plugins
+        for cb_names, _ in intg.system._kernel_callbacks:
```
FreddieWitherden: Try to avoid _ attribute access. We should either make it public or refactor.

pyfr/plugins/base.py
```diff
@@ -162,6 +162,42 @@ class BaseSolnPlugin(BasePlugin):
...
+            if clash := names & cb_names:
+                raise ValueError(f'Extern name clash: {clash}')
+
+        for eles in intg.system.ele_map.values():
+            for name in names:
+                eles._set_external(name, spec)
```
FreddieWitherden: Same with set external (this has been a long running issue I've been meaning to fix and is my fault).

kaschau: do you want me to change the usage of set_external in this PR? arent too many places.

FreddieWitherden: Yeah, I think that's sensible.

pyfr/plugins/base.py
FreddieWitherden: Do we get much value here?  Or is it easier to just let the plugins themselves loop over their binders?

FreddieWitherden:
Other than the above comments it looks good.  Does it work for your internal requirements?

pyfr/plugins/base.py
kaschau: at this point probably not, intention was to just give hints as to what is needed on the plugin developers end, it used to be more substantial but now its pretty weak.

pyfr/plugins/base.py
```diff
@@ -162,6 +162,27 @@ class BaseSolnPlugin(BasePlugin):
...
+        for eles in intg.system.ele_map.values():
+            for name in names:
+                eles.set_external(name, spec)
+
+        intg.system.register_kernel_callback(names, self._extern_callback)
+
+    def _extern_callback(self, kern):
+        if kern.bind not in self._extern_binders:
+            self._extern_binders.append(kern.bind)
+            kern.bind(**self._extern_values)
```
FreddieWitherden: Should we do this?  I think the plugins just need a list of things they need to bind (_extern_binders) and they're good.

kaschau: would you rather the system own binders and reference the plugin state (its _extern_values dict)? and the plugin just references a callable to loop over its binders?

```
# In register_kernel_callback — return a closure
def register_kernel_callback(self, names, values):
    binders = []
    self._kernel_callbacks.append((tuple(names), binders, values))
    return lambda: [b(**values) for b in binders]
```

Then the plugin just has:
```
# In _register_externs
self._rebind = intg.system.register_kernel_callback(names, self._extern_values)

# In __call__
self._update_extern_values()
self._rebind()
```

kaschau: Also do you want to support mixed `spec` values? right now when the plugin registers its exerns it assumes uniformity and defaults to `fpdtype scalar`

FreddieWitherden: It is probably easier for the plugin to handle calling bind itself on its kernels.  At the moment the only thing that can be bound is fpdtype scalar.  However, this will change to also allow ixdtype in the future.  But that should be it (only limited scalars, arrays you just need to allocate your own device matrix for and then update that matrix as you please).

kaschau: @FreddieWitherden I remember why I had this, the values need to be bound for the first RHS call, and they cant do it in plugins' `__init__` because kernels don't exist yet. so it needs to happen after commit. otherwise they dont get anything until after the first RHS with the plugins first `__call__`.

FreddieWitherden: Makes sense. Can you see if the turbulent inflow plugin needs updating?  Or should be updated.

kaschau: no turbulence is bound at kernel creation time via matrices (like the nirf_R is now), so its fine, and I dont see anything that is run time variable that is a global scalar that would use this.

FreddieWitherden: Could it be simpler to require that the callback function (when it is called) be expect to do an initial bind?

kaschau:
> Other than the above comments it looks good. Does it work for your internal requirements?

Yes this is working well for the nirf cases.

FreddieWitherden:
What do we need to do here?

kaschau:
This is working well for the NIRF plugin, I think its good to go.

pyfr/plugins/base.py
```diff
@@ -162,6 +162,30 @@ class BaseSolnPlugin(BasePlugin):
...
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Runtime extern binding
+        self._extern_values = {}
+        self._extern_binders = []
+
+    def _register_externs(self, intg, names, spec='scalar fpdtype_t'):
+        self._update_extern_values()
```
FreddieWitherden, line 176: Who defines this?

kaschau: should have a no-op added to the base that actual plugins override... added now.

PR #532, opened 2026-02-11
[WIP] NIRF Plugin

kaschau:
This WIP PR adds a Non-Inertial Reference Frame plugin for simulating prescribed/reacting motion of the mesh frame.


pyfr/solvers/euler/inters.py
```diff
@@ -244,3 +246,114 @@ def serialisefn(cls, bciface, prefix, srl):
...
+            for p in params
+        }
+        self._tplargs |= nirf_origin_tplargs(self.cfg, sect, self.ndims)
+
+        if motion == 'free':
+            for p in params:
+                self.set_external(_to_extern(p), 'scalar fpdtype_t')
+
+            # Rotation matrix externs: R(-θ) for lab-to-body transform
+            for i in range(3):
```
FreddieWitherden: ndrange will help here, but it might be easier to use a matrix and then just call .set() on it.

pyfr/plugins/nirf.py
```diff
@@ -0,0 +1,433 @@
...
+                              moment[0], moment[1], moment[2])
+
+    def setup(self, sdata, serialiser):
+        if self._motion != 'free':
+            return
+
+        if sdata is not None:
+            sdata = np.asarray(sdata, dtype=float)
+            ndims = self.ndims
+            self._fomega = sdata[0:3].copy()
```
FreddieWitherden: np.void and a structured data type will help a lot here (maps to a struct in HDF5 so you can have named arrays etc).

PR #534, opened 2026-02-19
add general soln copy request

kaschau:
Add ability for downstream entities to request a solution copy, not just in case of source term presence.


FreddieWitherden:
May try to unify with the existing property (renaming as appropriate).

kaschau:
actually hold up on this PR for a moment, it was for the nscbcs but I think there is a better solution that makes this unnecessary.

FreddieWitherden:
Is this still needed?

PR #536, opened 2026-02-25
Add anisotropicturbulence plugins

rtrtcom93:


FreddieWitherden:
This seems to introduce a huge amount of code duplication with the existing plugin.  Is there a reason it can't be extended to support anisotropic turbulence?

rtrtcom93:
> This seems to introduce a huge amount of code duplication with the existing plugin. Is there a reason it can't be extended to support anisotropic turbulence?

I just wanted to keep them separate for the moment to maintain the integrity of the original plugin while I test these minor changes. Once the anisotropic support is stable and the logic is set, I’ll definitely merge them to eliminate the duplication you mentioned.

WillTrojak:
The CGNS IC reader should be split into a separate PR, as it’s a separate feature and we can discuss it there.

FreddieWitherden:
Just following up here.

PR #537, opened 2026-02-26
provide sdata to bcs/plugins/intg with previous cfg for stale data reuse on restart

kaschau:
This PR removes the assumption of stale serialised data on restart and passes all sdata to bcs, plugins, and intg. It also passes in previous cfg (if present) to allow bcs and plugins to make decisions about what to do with this data on restart.

### DIFF
--- a/pyfr/inifile.py
+++ b/pyfr/inifile.py
@@ -121,6 +121,35 @@ class Inifile:
     def remove_option(self, section, option):
         self._cp.remove_option(section, option)
 
+    def sect_eq(self, other, section):
+        if other is None:
+            return False
+
+        try:
+            sitems = dict(self._cp.items(section))
+            oitems = dict(other._cp.items(section))
+            return sitems == oitems
+        except NoSectionError:
+            return False
+
+    def sect_diff(self, other, section):
+        # No previous config, everything is new
+        if other is None:
+            return {k: (v, None)
+                    for k, v in self._cp.items(section)}
+
+        sitems = dict(self._cp.items(section))
+
+        # Section missing from other, everything is new
+        try:
+            oitems = dict(other._cp.items(section))
+        except NoSectionError:
+            return {k: (v, None) for k, v in sitems.items()}
+
+        # Return {key: (self_val, other_val)} for keys that differ
+        return {k: (sitems.get(k), oitems.get(k))
+                for k in sitems.items() ^ oitems.items()}
+
     def sections(self):
         return self._cp.sections()
 
--- a/pyfr/inifile.py
+++ b/pyfr/inifile.py
@@ -138,7 +138,7 @@ class Inifile:
         except NoSectionError:
             return {k: (v, None) for k, v in sitems.items()}
 
-        # Return {key: (self_val, other_val)} for keys that differ
+        # Return {key: (self_val, other_val)} for options that differ
         return {k: (sitems.get(k), oitems.get(k))
                 for k in sitems.items() ^ oitems.items()}
 
--- a/pyfr/inifile.py
+++ b/pyfr/inifile.py
@@ -122,9 +122,6 @@ class Inifile:
         self._cp.remove_option(section, option)
 
     def sect_eq(self, other, section):
-        if other is None:
-            return False
-
         try:
             sitems = dict(self._cp.items(section))
             oitems = dict(other._cp.items(section))
@@ -133,11 +130,6 @@ class Inifile:
             return False
 
     def sect_diff(self, other, section):
-        # No previous config, everything is new
-        if other is None:
-            return {k: (v, None)
-                    for k, v in self._cp.items(section)}
-
         sitems = dict(self._cp.items(section))
 
         # Section missing from other, everything is new
--- a/pyfr/inifile.py
+++ b/pyfr/inifile.py
@@ -140,7 +140,7 @@ class Inifile:
 
         # Return {key: (self_val, other_val)} for options that differ
         return {k: (sitems.get(k), oitems.get(k))
-                for k in sitems.items() ^ oitems.items()}
+                for k, _ in sitems.items() ^ oitems.items()}
 
     def sections(self):
         return self._cp.sections()
--- a/pyfr/integrators/base.py
+++ b/pyfr/integrators/base.py
@@ -127,6 +127,7 @@ class BaseIntegrator(metaclass=RegisterMeta):
             self.triggers.restore(initsoln.state)
 
         plugins = []
+        prevcfg = initsoln.config if initsoln else None
 
         for s in self.cfg.sections():
             if (m := re.match('(soln|solver)-plugin-(.+?)(?:-(.+))?$', s)):
@@ -144,7 +145,7 @@ class BaseIntegrator(metaclass=RegisterMeta):
                 plugin = get_plugin(*args)
                 sprefix = plugin.sprefix
                 sdata = initsoln.state.get(sprefix) if initsoln else None
-                plugin.setup(sdata, self.serialiser)
+                plugin.setup(sdata, prevcfg, self.serialiser)
                 plugins.append(plugin)
 
         # Validate plugin trigger references
--- a/pyfr/integrators/explicit/controllers.py
+++ b/pyfr/integrators/explicit/controllers.py
@@ -78,9 +78,8 @@ class ExplicitPIController(PIControllerMixin, BaseExplicitController):
         # Initialise dt/errprev from restart or defaults
         if initsoln and (sd := initsoln.state.get('intg/ctrl')):
             self.dt, self._errprev = sd
-            diff = self.cfg.sect_diff(initsoln.config,
-                                      'solver-time-integrator')
-            if any(k in {'rtol'} or k.startswith('atol') for k in diff):
+            diff = self.cfg.sect_diff(initsoln.config, 'solver-time-integrator')
+            if any(k.startswith(('atol', 'rtol')) for k in diff):
                 self._errprev = 1.0
         else:
             self._errprev = 1.0
--- a/pyfr/integrators/explicit/controllers.py
+++ b/pyfr/integrators/explicit/controllers.py
@@ -78,6 +78,10 @@ class ExplicitPIController(PIControllerMixin, BaseExplicitController):
         # Initialise dt/errprev from restart or defaults
         if initsoln and (sd := initsoln.state.get('intg/ctrl')):
             self.dt, self._errprev = sd
+            diff = self.cfg.sect_diff(initsoln.config,
+                                      'solver-time-integrator')
+            if any(k in {'rtol'} or k.startswith('atol') for k in diff):
+                self._errprev = 1.0
         else:
             self._errprev = 1.0
 
--- a/pyfr/integrators/implicit/controllers.py
+++ b/pyfr/integrators/implicit/controllers.py
@@ -238,6 +238,10 @@ class ImplicitPIController(ThroughputLimitMixin, PIControllerMixin,
         # Initialise dt/errprev from restart or defaults
         if initsoln and (sd := initsoln.state.get('intg/ctrl')):
             self.dt, self._errprev = sd[:2]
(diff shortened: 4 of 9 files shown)


TobyFlynn:
This looks good and handles restarting with a different config file better than what I had previously. It worked nicely when restarting on a small mass flow controller test case

FreddieWitherden:
As an alternative do we always want to provide the sdata (even on restart with different code) and then let the plugins or whatever decide what to do?  The usually have access to the config file stack so can, if they want, check if things are compatible (or just always throw everything away).

TobyFlynn:
Yeah that would work too, I think the only thing that currently needs to discard the sdata if we restart with a new config file is the PI controller for adaptive time stepping as the previous error won't reflect any changed tolerances in the config file. Then the `_init_dt_err` function would be something along the lines of:

```
def _init_dt_err(self, initsoln):
        sdata = initsoln.get('intg/pi') if initsoln else None

        if sdata is not None:
            self._dt, self._errprev = sdata

            prev_cfg = initsoln.get('config')
            sect = 'solver-time-integrator'
            prev_atol = prev_cfg.getfloat(sect, 'atol')
            prev_rtol = prev_cfg.getfloat(sect, 'rtol')

            if prev_atol != self._atol or prev_rtol != self._rtol:
                self._errprev = 1.0
        else:
            self._errprev = 1.0

        self.serialiser.register('intg/pi', lambda: [self._dt, self._errprev])
```

kaschau:
since bcs and plugins are much more simple in their response to cfg, we could only remove the intg sdata, and leave bcs and plugins to do as they will. just remove the bc/plugin strings from the stale data check.

TobyFlynn:
Yeah that might be the simplest solution

FreddieWitherden:
> Yeah that would work too, I think the only thing that currently needs to discard the sdata if we restart with a new config file is the PI controller for adaptive time stepping as the previous error won't reflect any changed tolerances in the config file. Then the `_init_dt_err` function would be something along the lines of:
> 
> ```
> def _init_dt_err(self, initsoln):
>         sdata = initsoln.get('intg/pi') if initsoln else None
> 
>         if sdata is not None:
>             self._dt, self._errprev = sdata
> 
>             prev_cfg = initsoln.get('config')
>             sect = 'solver-time-integrator'
>             prev_atol = prev_cfg.getfloat(sect, 'atol')
>             prev_rtol = prev_cfg.getfloat(sect, 'rtol')
> 
>             if prev_atol != self._atol or prev_rtol != self._rtol:
>                 self._errprev = 1.0
>         else:
>             self._errprev = 1.0
> 
>         self.serialiser.register('intg/pi', lambda: [self._dt, self._errprev])
> ```

I'd also argue that this is the least important one.  The nice thing about adaptive time stepping is that it readily recovers from bad error predictions quite quickly.

kaschau:
@FreddieWitherden do you have an opinion on treating all 3 (bc,plugin,intg) the same versus keeping intg as is and just letting bcs and plugins manage themselves?

FreddieWitherden:
I'd give all of the data to all of the plugins, integrators, and boundary conditions all of the time.  Then, I might add some helper methods to the ini file to help them figure out if stuff has changed in their section so we get a bit less code duplication around them deciding what to do.

kaschau:
> I'd give all of the data to all of the plugins, integrators, and boundary conditions all of the time. Then, I might add some helper methods to the ini file to help them figure out if stuff has changed in their section so we get a bit less code duplication around them deciding what to do.

If this looks more like what you had in mind, I can add some helper functions to the ini class to allow for comparison of sections, etc.

FreddieWitherden:
The approach looks good to me.

FreddieWitherden:
Do we need to do anything else here?

kaschau:
I don't think so, this now just provides all data for re-use and the individual plugins need to decide what to do, and we provide the previous cfg file for comparison. I think when a use case comes up we can add some utility for detecting changes, but I don't think there are any that need it, currently.

FreddieWitherden:
I think we can merge when the existing serialises are updated (even if it is just a case of `if prevcfg is not None: return` to emulate their current behaviour.

kaschau:
do you want to exactly emulate current behavior or utilize previous data now that it's available like Toby's suggestion above for the PI controller?

FreddieWitherden:
Ideally the latter although that does require more work/testing.

kaschau:
I think its just the PI controller that can do anything at the moment, @TobyFlynn correct me if Im wrong but the MassFlowBCs can stay the same as they already check for the sdata (and I don't think care about changes to the cfg anyway?)

FreddieWitherden:
I guess if the config has changed but the BC arguments are the same it is probably okay to reuse the serialised BC data?

kaschau:
I added some helper methods in the ini class to determine if a cfg sect has changed (one that is a bool and another returns a dict of different values). So now the MassFluxBC ensures the cfg sect is the same for reuse, and the pi controller resets the errest if either `a-tol` or `r-tol` have changed. Running TGV tests now on the PI controller and its working.

pyfr/integrators/std/controllers.py
```diff
@@ -201,6 +201,12 @@ def _init_dt_err(self, initsoln):
 
         if sdata is not None:
             self._dt, self._errprev = sdata
+            prevcfg = initsoln.get('config')
+            sect = 'solver-time-integrator'
```
FreddieWitherden: Can we inline sect assuming the line fits

pyfr/integrators/base.py
```diff
@@ -99,7 +99,8 @@ def _get_plugins(self, initsoln):
                 plugin = get_plugin(*args)
                 sprefix = plugin.get_serialiser_prefix()
                 sdata = initsoln.get(sprefix) if initsoln else None
-                plugin.setup(sdata, self.serialiser)
+                prevcfg = initsoln.get('config') if initsoln else None
```
FreddieWitherden: I think we can factor this out.

pyfr/solvers/base/system.py
```diff
@@ -174,12 +174,13 @@ def _load_bc_inters(self, mesh, elemap, initsoln, serialiser):
 
             # Check if there is serialised data for this boundary in initsoln
             sdata = initsoln.get(f'bcs/{bname}') if initsoln else None
+            prevcfg = initsoln.get('config') if initsoln else None
```
FreddieWitherden: I think we can factor this out

pyfr/solvers/euler/inters.py
```diff
@@ -157,10 +157,12 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
 
-    def setup(self, sdata):
-        if sdata is not None and sdata[4] != 0:
-            (self.interp_c, self.interp_m, 
-            self.mf_avg, self.tprev, self.nstep_counter) = sdata
+    def setup(self, sdata, prevcfg):
+        sect_eq = self.cfg.sect_eq(prevcfg, self.cfgsect)
+
+        if sdata is not None and sdata[4] != 0 and sect_eq:
+            (self.interp_c, self.interp_m,
```
FreddieWitherden: See if we can avoid line breaks on the LHS.

pyfr/inifile.py
```diff
@@ -123,6 +123,35 @@ def items_as(self, section, type, prefix=''):
     def remove_option(self, section, option):
         self._cp.remove_option(section, option)
 
+    def sect_eq(self, other, section):
```
FreddieWitherden, line 124: Allowing None for other seems a bit weird API wise.

FreddieWitherden:
Also target should be develop.

pyfr/inifile.py
```diff
@@ -123,6 +123,27 @@ def items_as(self, section, type, prefix=''):
...
+        sitems = dict(self._cp.items(section))
+
+        # Section missing from other, everything is new
+        try:
+            oitems = dict(other._cp.items(section))
+        except NoSectionError:
+            return {k: (v, None) for k, v in sitems.items()}
+
+        # Return {key: (self_val, other_val)} for keys that differ
+        return {k: (sitems.get(k), oitems.get(k))
```
FreddieWitherden, line 142: What happens for same keys but possibly different values?

pyfr/solvers/euler/inters.py
```diff
@@ -157,10 +157,14 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
-    def setup(self, sdata):
-        if sdata is not None and sdata[4] != 0:
-            (self.interp_c, self.interp_m, 
-            self.mf_avg, self.tprev, self.nstep_counter) = sdata
+    def setup(self, sdata, prevcfg):
+        sect_eq = (prevcfg is not None
+                   and self.cfg.sect_eq(prevcfg, self.cfgsect))
+
+        if sdata is not None and sdata[4] != 0 and sect_eq:
+            self.interp_c, self.interp_m = sdata[0], sdata[1]
```
FreddieWitherden: Can slice on rhs

pyfr/inifile.py
kaschau: that comment is misleading... it should be options that differ or something. the symmetric difference ^ picks up keys that have one value in the first cfg and a different value in the other cfg.

pyfr/solvers/euler/inters.py
```diff
@@ -157,10 +157,14 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
         else:
             self.csv = None
 
-    def setup(self, sdata):
-        if sdata is not None and sdata[4] != 0:
-            (self.interp_c, self.interp_m, 
-            self.mf_avg, self.tprev, self.nstep_counter) = sdata
+    def setup(self, sdata, prevcfg):
+        sect_eq = (prevcfg is not None
```
FreddieWitherden: I think we can clean up with an early exit if sdata is None

kaschau: I think we still have to set the class attrs to something even if sdata is None.

FreddieWitherden: Ah, yes.

pyfr/integrators/explicit/controllers.py
```diff
@@ -78,6 +78,10 @@ def __init__(self, backend, systemcls, mesh, initsoln, cfg):
         # Initialise dt/errprev from restart or defaults
         if initsoln and (sd := initsoln.state.get('intg/ctrl')):
             self.dt, self._errprev = sd
+            diff = self.cfg.sect_diff(initsoln.config,
+                                      'solver-time-integrator')
```
FreddieWitherden: Will this string fit above?

pyfr/solvers/euler/inters.py
```diff
@@ -130,10 +130,14 @@ def __init__(self, be, lhs, elemap, cfgsect, cfg, bccomm):
...
+                   and self.cfg.sect_eq(prevcfg, self.cfgsect))
```
FreddieWitherden: Move and up.

pyfr/integrators/implicit/controllers.py
```diff
@@ -238,6 +238,10 @@ def __init__(self, backend, systemcls, mesh, initsoln, cfg):
         # Initialise dt/errprev from restart or defaults
         if initsoln and (sd := initsoln.state.get('intg/ctrl')):
             self.dt, self._errprev = sd[:2]
+            diff = self.cfg.sect_diff(initsoln.config,
+                                      'solver-time-integrator')
+            if any(k in {'rtol'} or k.startswith('atol') for k in diff):
```
FreddieWitherden: In scalar seems weird.

pyfr/integrators/explicit/controllers.py
```diff
@@ -78,6 +78,10 @@ def __init__(self, backend, systemcls, mesh, initsoln, cfg):
...
+            if any(k in {'rtol'} or k.startswith('atol') for k in diff):
```
FreddieWitherden: k.startswith(('atol', 'rtol'))

PR #538, opened 2026-03-04
Fix for visualising gradients with ascent

TobyFlynn:
This fixes the issue when trying to use gradients in the ascent plugin as seen in this forum post: https://pyfr.discourse.group/t/compute-q-criterion-in-ascent-and-slice-multiple-times/1337/2


robertsawko:
Thanks @TobyFlynn for this update. I will give it a shot and let you know.

PR #539, opened 2026-03-10
Add plane region

kaschau:
this PR adds a plane region which will select all elements straddling a plane defined by a point and normal. The plane can be clipped using other regions and set math: plane - (plane - sphere).

### DIFF
--- a/pyfr/regions.py
+++ b/pyfr/regions.py
@@ -319,6 +319,28 @@ class SphereRegion(EllipsoidRegion):
         super().__init__(x0, r, r, r, **kwargs)
 
 
+class PlaneRegion(BaseGeometricRegion):
+    name = 'plane'
+
+    def __init__(self, x0, n, **kwargs):
+        super().__init__(**kwargs)
+
+        self.x0 = np.array(x0)
+        self.n = np.array(n, dtype=float)
+        self.n /= np.linalg.norm(self.n)
+
+    def _pts_in_region(self, pts):
+        dist = (pts - self.x0) @ self.n
+
+        # Shape points: detect straddle points
+        if dist.ndim == 2:
+            straddle = dist.min(axis=0) * dist.max(axis=0) <= 0
+            return np.broadcast_to(straddle, dist.shape).copy()
+
+        # Centroids: return false
+        return np.zeros(dist.shape, dtype=bool)
+
+
 class STLRegion(BaseGeometricRegion):
     name = 'stl'
 
--- a/pyfr/regions.py
+++ b/pyfr/regions.py
@@ -337,6 +337,28 @@ class SphereRegion(EllipsoidRegion):
         super().__init__(x0, r, r, r, **kwargs)
 
 
+class PlaneRegion(BaseGeometricRegion):
+    name = 'plane'
+
+    def __init__(self, x0, n, **kwargs):
+        super().__init__(**kwargs)
+
+        self.x0 = np.array(x0)
+        self.n = np.array(n, dtype=float)
+        self.n /= np.linalg.norm(self.n)
+
+    def _pts_in_region(self, pts):
+        dist = (pts - self.x0) @ self.n
+
+        # Shape points: detect straddle points
+        if dist.ndim == 2:
+            straddle = dist.min(axis=0) * dist.max(axis=0) <= 0
+            return np.broadcast_to(straddle, dist.shape).copy()
+
+        # Centroids: return false
+        return np.zeros(dist.shape, dtype=bool)
+
+
 class STLRegion(BaseGeometricRegion):
     name = 'stl'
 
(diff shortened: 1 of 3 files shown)


pyfr/regions.py
```diff
@@ -268,6 +268,28 @@ def __init__(self, x0, r, **kwargs):
...
+
+        self.x0 = np.array(x0)
+        self.n = np.array(n, dtype=float)
+        self.n /= np.linalg.norm(self.n)
+
+    def _pts_in_region(self, pts):
+        dist = (pts - self.x0) @ self.n
+
+        # Shape points: detect straddle points
+        if dist.ndim == 2:
```
FreddieWitherden, line 336: When does the ndim check trigger?

kaschau: > When does the ndim check trigger?

since this is a 2d region, and we kinda have to hack pts_in_region to allow the CSG math to work, we need to all false when centroids are ping'd and only do the straddle math when spts are checked.

FreddieWitherden: That makes sense. It would be nice if we could come up with a more elegant solution here (so we don't give different answers depending on the shape of the points array).  Not quite sure what it should be, though.

kaschau: yeah I agree its not the best check... ill try and think of something

kaschau: so just to be clear, the `pts_in_region` methods return different results based on input shape for all regions (they get called with an spts array then a centroid array), this one just has to consider them differently due to 2d nature where 3d regions do not. another option is to change the CSG logic to do element-set math, where elements are bool'd instead of points. I think the only thing that changes is slight difference in how elements on a boundary would be treated with subtractions, but I'd have to think more.

FreddieWitherden: So ideally the methods should not care about the shape; as in `pts_in_region(x.ravel()).reshape(x.shape) == pts_in_region(x)` which is consistent with most basic NumPy methods (like `np.cos` etc).

FreddieWitherden:
Are we looking to factor the CSG routines to enable this?

kaschau:
> Are we looking to factor the CSG routines to enable this?

yes, I have the implementation sorted Ive just been waiting for things to settle down so I can rebase. its not a huge change but it'll require some restructuring

pyfr/regions.py
```diff
@@ -9,6 +10,15 @@
 from pyfr.util import match_paired_paren, subclass_where
 
 
+def rotated(method):
+    @wraps(method)
+    def wrapper(self, pts):
+        if self.rot is not None:
+            pts = pts @ self.rot
```
FreddieWitherden: Think we might have @= in recent Python.

pyfr/regions.py
```diff
@@ -235,22 +253,18 @@ def __init__(self, rdata=None, rot=None):
...
+        inside = self.test_pts(centroids)
+        if not inside.all():
+            inside[~inside] = self.test_pts(spts[:, ~inside]).any(axis=0)
+        inside &= self.test_eles(spts)
+        return inside
 
-    def pts_in_region(self, pts):
-        if self.rot is not None:
-            pts = np.einsum('ij,klj->kli', self.rot.T, pts)
+    def test_pts(self, pts):
```
FreddieWitherden: Are these sentinel methods needed?

kaschau: 2 choices. Currently, this is here to run a point wise test and an element wise test for each region transparently, regardless of their particular method. If the region only does point wise (all previous regions), its element test just `True`s all so nothing happens. and vice versa for an element wise region test (plane).

The alternative is to either return `None`, or class attr them as `None` and have `if None:` tests to skip over them, but that breaks the ufunc nature of the point wise tests. 

Though I would argue that even the previous `pts_in_region` methods were not ufunc, as they relied on collapsing the `ndims` dimension at the end, so one couldn't just throw an array at them. Element wise tests are even worse as they rely on a specific shape to construct element bounds as well. 

If we relax the ufunc goal for these tests it simplifies things quite a bit.

FreddieWitherden: Wonder if there are any larger scale refactorings we can undertake here, as with the current architecture all roads seem clunky.

kaschau: so if we allow for both centroids + spts to be passed to the masking methods, it cleans a lot up. each thing can decide what to do with it (pointwise v element wise decisions). see current state

pyfr/regions.py
```diff
@@ -74,15 +74,23 @@ def to_dict(self):
 
 
 class BaseRegion:
-    def interior_eles(self, mesh):
+    def get_region_eles(self, mesh):
```
FreddieWitherden: Probably want a naming convention which gives symmetry with `surface_faces`.  No strong preference as to what should be named what so long as our convention is consistent.

PR #541, opened 2026-03-20
Post Processing Plugins

kaschau:
This PR reworks the export writer classes a little to prepare data and pass to a new plugin type, `postproc` plugins. The post proc plugins are split between surface (meant for boundary exports, Cf, y+, etc.) and flow (meant for volume exports, Mach #, vorticity, Qcrit, etc.), with further ideas for things like properties (thermo/transport). 

These post processors sit between the data loading and writing on the export command. 

The idea is to alleviate computing things in ParaView for many common quantities of interest, as ParaView scripting is tedious and annoying on the best of days. 

To request these quantities, one can now pass a comma separated list of desired postproc plugins to be run, i.e. `--postproc cf,yplus` would compute skin friction coeff and y+ and add them to the exported .vtu file. As these examples are surface type quantities, they are only valid for a boundary export (which limits them to 3d). Other quantities like Cp care valid for both boundary and surface, and are filtered as such. 

For non-self contained quantities that require user input, the post proc plugins follow the same structure in the ini file as other plugins, and user input values can be set there:

```
[postproc-plugin-cf]
rho-inf = 1.2
u-inf = 1.0
; mu is pulled from constants
```

If this section is not in the solution cfg, a user can pass in a new cfg with the appropriate data upon export with `--cfg file_with_pp_sect.ini` that will be additively loaded to the solver cfg. Otherwise, the plugins assume the simulation was run with the postproc sections present and stored in the pyfrs files. 

The largest change to enable easy extension is the creation of a postproc plugin data adapter, what exposes these plugins to as much data as possible to allow for complex calculations, including element classes, primitive, ploc, face info, sdata, etc. These properties are cached and lazily computed so multiple postproc variables can re-use the same data. This way we can easily add general quantities to the data adapter and simply access them in the post processor's compute methods.

This should provide an extensible platform for adding interesting data for visualization.


pyfr/plugins/surface.py
```diff
@@ -0,0 +1,75 @@
...
+        u_inf = self.cfg.getfloat(self.cfgsect, 'u-inf')
+        mu = self.cfg.getfloat('constants', 'mu')
+
+        q_inf = 0.5 * rho_inf * u_inf**2
+
+        ndims = self.ndims
+        normals = data.normals
+
+        # Velocity gradient tensor: grad_vel[i][d] = du_i/dx_d
+        grad_vel = np.stack([data.grad_pris[1 + i] for i in range(ndims)])
```
FreddieWitherden: Can we slice here? grad_pris[1:ndims + 1]?

pyfr/plugins/surface.py
```diff
@@ -0,0 +1,75 @@
...
+
+        # Stress tensor: tau_ij = mu * (du_i/dx_j + du_j/dx_i)
+        # Wall shear: tau_w = tau . n - (n . tau . n) * n
+        tau_n = mu * np.einsum('ijkl,jkl->ikl',
+                               grad_vel + grad_vel.swapaxes(0, 1), normals)
+
+        # Remove normal component
+        nn = np.einsum('ikl,ikl->kl', tau_n, normals)
+        tau_wall = tau_n - nn[np.newaxis] * normals
```
FreddieWitherden: Prefer None rather then newaxis.

pyfr/plugins/surface.py
```diff
@@ -0,0 +1,75 @@
...
+        return {'yplus': ['y+']}
+
+    def compute(self, data):
+        mu = self.cfg.getfloat('constants', 'mu')
+        ndims = self.ndims
+        normals = data.normals
+        rho_wall = data.pris[0]
+
+        # Wall shear stress
+        grad_vel = np.stack([data.grad_pris[1 + i] for i in range(ndims)])
```
FreddieWitherden: See if we can share some code with Cf.

pyfr/writers/vtk/base.py
```diff
@@ -9,6 +10,131 @@
...
+
+        # Project each upt onto face: move along ref normal to face plane
+        t = (face_pt - upts) @ norm_arr / np.dot(norm_arr, norm_arr)
+        upts_on_face = upts + t[:, np.newaxis] * norm_arr
+
+        # Physical positions at solution points and their face projections
+        upt_op = self._shape.sbasis.nodal_basis_at(upts)
+        face_op = self._shape.sbasis.nodal_basis_at(upts_on_face)
+
+        x_upt = interpolate_pts(upt_op, self._spts)    # (nupts, neles, ndims)
```
FreddieWitherden: Try to avoid inline comments

pyfr/writers/vtk/base.py
```diff
@@ -9,6 +10,131 @@
...
+        return self.pnorm / mag[np.newaxis]
+
+    @cached_property
+    def wall_dist(self):
+        # Project each solution point onto the face along the ref normal
+        # Face is a hyperplane in ref space; project by replacing the
+        # constrained coordinate(s)
+        itype, proj, norm = self._shape.faces[self._fidx]
+        norm_arr = np.array(norm, dtype=float)
+        upts = np.array(self._shape.upts)
```
FreddieWitherden: Is the np.array needed here?

pyfr/writers/vtk/base.py
```diff
@@ -391,6 +518,23 @@ def __init__(self, meshf, pname=None, *, prec='single', order=None,
...
+        extras = []
+
+        for pp in self.pp_plugins:
+            if pp.needs_gradients and not self._gradients:
+                raise RuntimeError(f'Postproc {pp.name} requires '
+                                   f'gradient data in the solution')
+
+            for components in pp.compute(adapter).values():
+                for arr in components:
+                    extras.append(arr[:, np.newaxis, :])
```
FreddieWitherden: Prefer None

pyfr/writers/vtk/base.py
```diff
@@ -533,6 +677,26 @@ def _load_soln(self, *args, **kwargs):
             self._vtk_vars = {k: [k] for k in self._soln_fields}
             self.tcurr = None
 
+        # Instantiate postproc plugins
+        from pyfr.plugins import get_plugin
```
FreddieWitherden: I presume we have to import locally due to circular imports?

pyfr/writers/vtk/base.py
```diff
@@ -533,6 +677,26 @@ def _load_soln(self, *args, **kwargs):
...
+
+        self.pp_plugins = []
+        for name in self._pp_plugin_names:
+            pp = get_plugin('postproc', name, self.ndims, self.cfg)
+
+            if not ('*' in pp.export_types or self.type in pp.export_types):
```
FreddieWitherden: We seem to have some validation code here and some in init.  Best to try and unify if possible.

pyfr/__main__.py
```diff
@@ -418,8 +428,15 @@ def process_export(args):
 
     # Common arguments
     kargs = [args.eargs] if 'eargs' in args else []
+
+    # Flatten comma-separated postproc names
+    postprocs = [p for arg in (args.postprocs or []) for p in arg.split(',')]
```
FreddieWitherden: Is the or needed if we have a default?

pyfr/writers/vtk/stl.py
```diff
@@ -65,7 +66,15 @@ def _load_soln(self, *args, **kwargs):
             self.einfo = []
 
     def _prepare_pts(self, etype):
-        return *self._stl_info, None, None
+        vpts, vsoln = self._stl_info
+
+        if self.pp_plugins:
+            adapter = _STLPostProcAdapter(
+                self, vpts, vsoln, None, None
```
FreddieWitherden: May fit on two lines.

pyfr/writers/vtk/volume.py
```diff
@@ -61,4 +62,12 @@ def _prepare_pts(self, etype):
         # Interpolate the solution to the vis points
         vsoln = interpolate_pts(soln_vtu_op, soln)
 
+        # Run postproc plugins
+        if self.pp_plugins:
+            shape = subclass_where(BaseShape, name=etype)(len(spts), self.cfg)
+            adapter = _VolumePostProcAdapter(
+                self, vpts, vsoln, shape, spts
```
FreddieWitherden: May fit on two lines.

FreddieWitherden:
Does it make sense to move post processing plugins to either pyfr.plugins.post or pyfr.postproc or something?  Am open to ideas, as plugins.base is getting pretty long.

pyfr/writers/vtk/base.py
kaschau: correct

kaschau:
> Does it make sense to move post processing plugins to either pyfr.plugins.post or pyfr.postproc or something? Am open to ideas, as plugins.base is getting pretty long.

agreed, ill try to think of something clean

FreddieWitherden:
For things like y+ and Cf we might see if we can fuse some of the einsums.  It is usually possible with a bit of care.

FreddieWitherden:
Also, is the architecture compatible with Ascent?

pyfr/plugins/__init__.py
```diff
@@ -5,6 +5,8 @@
 from pyfr.plugins.fluidforce import FluidForcePlugin
 from pyfr.plugins.fwh import FWHPlugin
 from pyfr.plugins.integrate import IntegratePlugin
+from pyfr.plugins.flow import CpPostProc, MachPostProc, VorticityPostProc
```
FreddieWitherden: Ordering.

pyfr/plugins/flow.py
```diff
@@ -0,0 +1,74 @@
...
+    systems = ['euler', 'navier-stokes']
+    dimensions = [2, 3]
+    export_types = ['*']
+
+    def fields(self):
+        return {'mach': ['Ma']}
+
+    def compute(self, data):
+        rho, p = data.pris[0], data.pris[-1]
+        vs = data.pris[1:-1]
```
FreddieWitherden: Can either inline this into the sum or move up, `rho, *vs, p = data.pris`

pyfr/writers/vtk/base.py
```diff
@@ -9,6 +10,131 @@
...
+        self._face_norm = face_norm
+
+    @cached_property
+    def _eles(self):
+        return self._elementscls(type(self._shape), self._spts, self._cfg)
+
+    @cached_property
+    def pnorm(self):
+        nsvpts = len(self._svpts)
+        norm_tiled = np.tile(np.array(self._face_norm, dtype=float),
```
FreddieWitherden: See if we can get rid of some the np.array calls.

pyfr/writers/vtk/base.py
```diff
@@ -9,6 +10,131 @@
...
+
+        # pnorm_at returns (nsvpts, neles, ndims)
+        pn = self._eles.pnorm_at(self._svpts, norm_tiled)
+
+        # Transpose to (ndims, nsvpts, neles)
+        return pn.transpose(2, 0, 1)
+
+    @cached_property
+    def normals(self):
+        mag = np.sqrt(np.einsum('ijk,ijk->jk', self.pnorm, self.pnorm))
```
FreddieWitherden: Aim to be consistent between **0.5 and np.sqrt.

pyfr/writers/vtk/boundary.py
```diff
@@ -97,23 +102,39 @@ def _get_surface_info(self, etype, eoffs, fidxs):
 
             idxs[f].append(e)
 
-        return [(*info[f], idxs[f]) for f in info]
+        return [(info[f][0],
```
FreddieWitherden: Maybe should be fused into the loop above.

pyfr/writers/vtk/base.py
```diff
@@ -9,6 +10,131 @@
...
+    @cached_property
+    def ploc(self):
+        return self._vpts.transpose(2, 0, 1)
+
+
+class _VolumePostProcAdapter(_BasePostProcAdapter):
+    pass
+
+
+class _STLPostProcAdapter(_BasePostProcAdapter):
```
FreddieWitherden: Do we want _ here?

pyfr/writers/vtk/boundary.py
kaschau: The namedtuple makes it hard to fuse since idxs needs to be built up.

kaschau:
> Also, is the architecture compatible with Ascent?

They are very close, the sticking point is boundary export, I dont think ascent does boundary face extraction, so that would stay isolated to vtk. I think if I move the pris building into the writer it will make the signatures for the adapter identical between the 2 so they could share that. Ill see what that looks like.

FreddieWitherden:
Yeah, Ascent currently doesn't handle surfaces.  I think we probably could do something but maybe not in this PR.  Key is to ensure that it could be done in the future.  Volume stuff should work though and may be better than Ascent.

kaschau:
Yeah they will work, (note to self) in a future PR to get ascent implemented we just need a few tweaks to get things to line up:

The VTK writer has:
writer.cfg
writer.soln
writer.elementscls
writer._gradients

Ascent's _CLIAdapter has:
adapter.scfg (not .cfg)
soln available but not as an attribute
adapter.elementscls
no _gradients flag

kaschau:
this new structure is still a bit of an import mess... I have to include the pyfr.postproc imports in the main plugin import so that subclass_where will pick things up

FreddieWitherden:
One possibility is to break out the class hierarchy a bit more so we can avoid some cross imports.

pyfr/plugins/cli/ascent.py
```diff
@@ -0,0 +1,51 @@
+from argparse import FileType
+
+from pyfr.inifile import Inifile
+from pyfr.mpiutil import init_mpi
+from pyfr.plugins.cli.base import BaseCLIPlugin
+from pyfr.plugins.cli.base import cli_external
```
FreddieWitherden: Combine with above

pyfr/plugins/postproc/adapters.py
```diff
@@ -0,0 +1,93 @@
...
+        pn = self._eles.pnorm_at(self._svpts, norm_tiled)
+
+        return pn.transpose(2, 0, 1)
+
+    @cached_property
+    def normals(self):
+        return self.pnorm / np.linalg.norm(self.pnorm, axis=0)
+
+    @cached_property
+    def tau_wall(self):
```
FreddieWitherden: Seems a bit specific to flow to be in a Base

pyfr/plugins/postproc/adapters.py
```diff
@@ -0,0 +1,93 @@
...
+        grad_vel = np.stack(self.grad_pris[1:self.ndims + 1])
+        sij = grad_vel + grad_vel.swapaxes(0, 1)
+        tau_n = mu * np.einsum('ijkl,jkl->ikl', sij, normals)
+
+        nn = (tau_n * normals).sum(axis=0)
+        tau_tang = tau_n - nn * normals
+        return np.linalg.norm(tau_tang, axis=0)
+
+    @cached_property
+    def wall_dist(self):
```
FreddieWitherden: Be careful here! Name may be misleading as true wall distance is a PITA to get.

pyfr/plugins/postproc/base.py
```diff
@@ -0,0 +1,24 @@
...
+    export_types = None
+    needs_gradients = False
+    needs_normals = False
+
+    def __init__(self, ndims, cfg, export_type=None):
+        cfgsect = f'postproc-plugin-{self.name}'
+        super().__init__(cfg, cfgsect, ndims)
+
+        if export_type is not None:
+            if not ('*' in self.export_types
```
FreddieWitherden: May want to change export_types and friends to be a regular expression, may cleanup the code. (Lists with a special case meaning for ['*'] were likely not a great idea in retrospect.

pyfr/plugins/soln/sampler.py
```diff
@@ -0,0 +1,188 @@
...
+                g = g.reshape(len(g), self.ndims*self.nvars, -1)
+                soln[i] = np.hstack([soln[i], g])
+
+        # Perform the sampling
+        samps = self.psampler.sample(soln, process=self._process)
+
+        # If we're the root rank then output
+        if samps is not None:
+            self._write(intg.tcurr, samps)
+
```
FreddieWitherden: Extra newline.

pyfr/writers/vtk/stl.py
```diff
@@ -4,6 +4,7 @@
 from pyfr.points import PointLocator, PointSampler
 from pyfr.polys import TriPolyBasis
 from pyfr.shapes import TriShape
+from pyfr.plugins.postproc.adapters import STLPostProcAdapter
```
FreddieWitherden: Move up.

pyfr/plugins/postproc/adapters.py
kaschau: This is in a BoundaryAdapter which inherits from base, were you thinking split it out farther?

pyfr/writers/vtk/boundary.py
FreddieWitherden: Can we do a info.items() to clean up a bit?

pyfr/writers/vtk/volume.py
```diff
@@ -61,4 +62,14 @@ def _prepare_pts(self, etype):
         # Interpolate the solution to the vis points
         vsoln = interpolate_pts(soln_vtu_op, soln)
 
+        # Run postproc plugins
+        if self.pp_plugins:
+            shape = subclass_where(BaseShape, name=etype)(len(spts), self.cfg)
+            pris, grad_pris = self._extract_pris(vsoln)
+            ploc = vpts.transpose(2, 0, 1)
+            adapter = VolumePostProcAdapter(
```
FreddieWitherden: This may fit on two lines with a different line breaking.

pyfr/plugins/postproc/adapters.py
FreddieWitherden: Yes, as the BoundaryAdapter should be system agnostic and contain things that are useful to all conceivable systems.

kaschau:
See if you like the current method of mixing in viscous boundary calculation to the data adapter... it gets a bit tricky without having a data adapter per system, and now relies on the presence of gradients to determine if it exposes the viscous stuff (currently just tau_wall)

pyfr/plugins/postproc/adapters.py
```diff
@@ -0,0 +1,98 @@
...
+        mu = self.cfg.getfloat('constants', 'mu')
+        normals = self.normals
+
+        grad_vel = np.stack(self.grad_pris[1:self.ndims + 1])
+        sij = grad_vel + grad_vel.swapaxes(0, 1)
+        tau_n = mu * np.einsum('ijkl,jkl->ikl', sij, normals)
+
+        nn = (tau_n * normals).sum(axis=0)
+        tau_tang = tau_n - nn * normals
+        return np.linalg.norm(tau_tang, axis=0)
```
FreddieWitherden: Missing newline.

pyfr/writers/vtk/boundary.py
```diff
@@ -86,34 +92,59 @@ def _itype_opmats(self, etype, fidx, cfg):
...
+            face_vsoln = interpolate_pts(fi.soln_op, soln)
+
+            # Run postproc plugins per-batch with full shape context
+            if self.pp_plugins:
+                pris, grad_pris = self._extract_pris(face_vsoln)
+                ploc = face_vpts.transpose(2, 0, 1)
+                shape = self._get_shape(fi.etype, self.cfg)
+                bases = (BoundaryPostProcAdapter,)
+                if grad_pris is not None:
+                    bases = (ViscousBoundaryAdapterMixin, *bases)
```
FreddieWitherden: Is there a reason the classes themselves (like Cf) can't inherit from this directly?

pyfr/plugins/postproc/adapters.py
```diff
@@ -0,0 +1,98 @@
...
+    def normals(self):
+        return self.pnorm / np.linalg.norm(self.pnorm, axis=0)
+
+    @cached_property
+    def _min_upt_wall_dist_approx(self):
+        _, proj, norm = self._shape.faces[self._fidx]
+        upts = self._shape.upts
+
+        face_pt = proj(*([0]*(self._shape.ndims - 1)))
+        t = (face_pt - upts) @ norm / np.dot(norm, norm)
```
FreddieWitherden: np.lingalg.norm may help here for normalising the normals.

pyfr/plugins/postproc/adapters.py
```diff
@@ -0,0 +1,98 @@
...
+            r = op @ spts.reshape(op.shape[1], -1)
+            return r.reshape(op.shape[0], *spts.shape[1:])
+
+        x_upt = interp(sbasis.nodal_basis_at(upts))
+        x_face = interp(sbasis.nodal_basis_at(upts_on_face))
+
+        dist = np.linalg.norm(x_upt - x_face, axis=2)
+        dist[dist == 0] = np.inf
+
+        return dist.min(axis=0)
```
FreddieWitherden: Might be easier to filter dist[dist > 0].min(...)

pyfr/__main__.py
```diff
@@ -418,8 +428,15 @@ def process_export(args):
...
+
+    # Flatten comma-separated postproc names
+    postprocs = [p for arg in args.postprocs for p in arg.split(',')]
+
+    # Load postproc config if provided (merged into solution config)
+    ppcfg = Inifile.load(args.ppcfg) if args.ppcfg else None
+
     kwargs = {'fields': args.fields, 'prec': args.precision,
-              'pname': args.pname}
+              'pname': args.pname, 'pp_plugins': postprocs, 'ppcfg': ppcfg}
```
FreddieWitherden: See if we can have a consistent convention between pp_ and pp.

pyfr/writers/vtk/base.py
```diff
@@ -533,6 +565,20 @@ def _load_soln(self, *args, **kwargs):
             self._vtk_vars = {k: [k] for k in self._soln_fields}
             self.tcurr = None
 
+        # Instantiate postproc plugins
+        from pyfr.plugins import get_plugin
```
FreddieWitherden: Can we have a function like get_postp_plugin and does this help us with imports at all?

pyfr/writers/vtk/boundary.py
kaschau: You mean the ViscousBondaryAdapterMixin? Or the entire data adapter machinery?

pyfr/writers/vtk/base.py
kaschau: yes.. finally

pyfr/writers/vtk/boundary.py
kaschau: The data the Adapters need is in the writer objects (and the render objects in the future version where ascent can use postproc), so we could send those objects to the postproc classes, then the data stays local to the class and isnt shared between plugins that use the same data.

pyfr/writers/vtk/boundary.py
```diff
@@ -1,12 +1,18 @@
...
 import numpy as np
 
 from pyfr.cache import memoize
+from pyfr.plugins.postproc.adapters import (BoundaryPostProcAdapter,
+                                            ViscousBoundaryAdapterMixin)
 from pyfr.shapes import BaseShape
 from pyfr.util import subclass_where
 from pyfr.writers.vtk.base import BaseVTKWriter, interpolate_pts
 
+FaceInfo = namedtuple('FaceInfo',
```
FreddieWitherden: Add newline

pyfr/writers/vtk/boundary.py
FreddieWitherden: I think the ideal situation is where we have a mixin for things like tau which plugins who need it can inherit from.

pyfr/plugins/postproc/adapters.py
```diff
@@ -0,0 +1,99 @@
+from functools import cached_property
+
+import numpy as np
+
+
+
```
FreddieWitherden, line 8: Remove newline.

pyfr/writers/vtk/boundary.py
```diff
@@ -86,34 +93,58 @@ def _itype_opmats(self, etype, fidx, cfg):
...
-            part.append(self.soln[f'{etype}-parts'][idxs])
+            face_vpts = interpolate_pts(fi.mesh_op, spts)
+            face_vsoln = interpolate_pts(fi.soln_op, soln)
+
+            # Run postproc plugins per-batch with full shape context
+            if self.pp_plugins:
+                pris, grad_pris = self._extract_pris(face_vsoln)
+                ploc = face_vpts.transpose(2, 0, 1)
+                shape = self._get_shape(fi.etype, self.cfg)
+                AdapterCls = (GradBoundaryPostProcAdapter
```
FreddieWitherden: Is there anything we can do to clean up the adapter stuff here?

kaschau: moved into function and consolidated signature a bit

pyfr/plugins/postproc/adapters.py
```diff
@@ -0,0 +1,107 @@
...
+        # Primitive variables and gradients (views into vsoln)
+        nvars = len(self.elementscls.privars(self.ndims, self.cfg))
+        self.pris = [vsoln[:, i, :] for i in range(nvars)]
+
+        self.grad_pris = None
+        if has_grads:
+            nd = self.ndims
+            self.grad_pris = [
+                vsoln[:, nvars + i*nd:nvars + (i + 1)*nd, :].transpose(1, 0, 2)
+                for i in range(nvars)
```
FreddieWitherden: I think range with a step can help here

pyfr/plugins/postproc/adapters.py
```diff
@@ -0,0 +1,108 @@
...
+
+        # Primitive variables and gradients (views into vsoln)
+        nvars = len(self.elementscls.privars(self.ndims, self.cfg))
+        self.pris = [vsoln[:, i, :] for i in range(nvars)]
+
+        self.grad_pris = None
+        if has_grads:
+            nd = self.ndims
+            self.grad_pris = [
+                vsoln[:, j:j + nd, :].transpose(1, 0, 2)
```
FreddieWitherden: Can we be less aggressive with line breaking?

pyfr/plugins/soln/fluidforce.py
```diff
@@ -15,13 +16,13 @@ def __init__(self, cfg, cfgsect, system, bcname, morigin):
 
         if self.locs and morigin is not None:
             self.rfpts = {k: loc - morigin[:, None, None]
-                          for k, loc in self.locs.items()}
+                         for k, loc in self.locs.items()}
```
FreddieWitherden: Indentation.

pyfr/plugins/__init__.py
```diff
@@ -1,18 +1,7 @@
...
-from pyfr.plugins.residual import ResidualPlugin
-from pyfr.plugins.sampler import SamplerCLIPlugin, SamplerPlugin
-from pyfr.plugins.source import SourcePlugin
-from pyfr.plugins.tavg import TavgCLIPlugin, TavgPlugin
-from pyfr.plugins.turbulence import TurbulencePlugin
-from pyfr.plugins.writer import WriterPlugin
+from pyfr.plugins.base import BasePlugin, BaseCLIPlugin
+from pyfr.plugins.soln.base import BaseSolnPlugin
+from pyfr.plugins.solver.base import BaseSolverPlugin
+from pyfr.plugins import postproc, soln, solver, cli
```
FreddieWitherden: Is this not circular as we are the init for `pyfr.plugins`.

pyfr/writers/vtk/boundary.py
```diff
@@ -104,6 +109,13 @@ def _get_surface_info(self, etype, eoffs, fidxs):
 
         return [(*info[f], idxs[f]) for f in info]
 
+    def _make_adapter(self, face_vsoln, face_vpts, spts, finfo):
+        AdapterCls = (GradBoundaryPostProcAdapter if self._gradients
```
FreddieWitherden: Avoid line breaks in single line if.

pyfr/writers/vtk/boundary.py
```diff
@@ -133,16 +145,28 @@ def _prepare_pts(self, itype):
         cellf, pointf = defaultdict(list), defaultdict(list)
 
         pshapes = self._extra_point_shapes(itype)
-        for etype, mesh_op, soln_op, lin_op, idxs in self._surface_info[itype]:
+        for etype, mesh_op, soln_op, lin_op, fidx, svpts, norm, idxs \
```
FreddieWitherden: Too much unpacking.

pyfr/writers/vtk/boundary.py
```diff
@@ -133,16 +145,28 @@ def _prepare_pts(self, itype):
...
 
-            vspts.append(interpolate_pts(mesh_op, spts))
-            vsoln.append(interpolate_pts(soln_op, soln))
+            face_vpts = interpolate_pts(mesh_op, spts)
+            face_vsoln = interpolate_pts(soln_op, soln)
+
+            # Run postproc plugins
+            if self.pp_plugins:
+                finfo = FaceInfo(etype, fidx, svpts, norm)
+                adapter = self._make_adapter(face_vsoln, face_vpts,
```
FreddieWitherden: Check if spts fits above.

pyfr/writers/vtk/volume.py
```diff
@@ -73,6 +74,13 @@ def _prepare_pts(self, etype):
         # Interpolate the solution to the vis points
         vsoln = interpolate_pts(soln_vtu_op, soln)
 
+        # Run postproc plugins
+        if self.pp_plugins:
+            adapter = VolumePostProcAdapter(self, vsoln, vpts, etype, spts,
+                                           has_grads=self._gradients)
```
FreddieWitherden: Indentation of has_grads.

pyfr/__main__.py
```diff
@@ -166,6 +166,14 @@ def main():
             '--eopt', dest='eopts', action='append', default=[],
             metavar='key:value', help='exporter-specific option'
         )
+        ap_export_type.add_argument(
+            '--postproc', dest='pp_plugins', action='append', default=[],
+            metavar='PLUGIN', help='postprocessing plugin; may be repeated'
+        )
+        ap_export_type.add_argument(
+            '--cfg', dest='pp_cfg', type=FileType('r'),
```
FreddieWitherden: Avoid FileType for new code as it is deprecated. We are working to get rid of existing uses.

pyfr/plugins/__init__.py
kaschau: right

pyfr/writers/vtk/boundary.py
```diff
@@ -1,14 +1,19 @@
...
 from pyfr.cache import memoize
+from pyfr.plugins.postproc.adapters import (BoundaryPostProcAdapter,
+                                            GradBoundaryPostProcAdapter)
 from pyfr.polys import get_polybasis
 from pyfr.shapes import BaseShape
 from pyfr.util import first, subclass_where
 from pyfr.writers.vtk.base import BaseVTKWriter, interpolate_pts
 
 
+FaceInfo = namedtuple('FaceInfo', ['etype', 'fidx', 'svpts', 'norm'])
```
FreddieWitherden: I think we are moving to strings rather than lists for named tuples `namedtuple('X', 'a b c')`

pyfr/writers/vtk/boundary.py
```diff
@@ -104,9 +111,18 @@ def _get_surface_info(self, etype, eoffs, fidxs):
...
         return [(*info[f], idxs[f]) for f in info]
 
+    def _make_adapter(self, face_vsoln, face_vpts, spts, finfo):
+        if self._gradients:
+            cls = GradBoundaryPostProcAdapter
+        else:
+            cls = BoundaryPostProcAdapter
+
+        return cls(self, face_vsoln, face_vpts, spts, finfo,
+                   has_grads=self._gradients)
```
FreddieWitherden: Is has_grads needed? Should be known based on cls.

pyfr/__main__.py
```diff
@@ -166,6 +166,14 @@ def main():
             '--eopt', dest='eopts', action='append', default=[],
             metavar='key:value', help='exporter-specific option'
         )
+        ap_export_type.add_argument(
+            '--postproc', dest='pp_plugins', action='append', default=[],
+            metavar='PLUGIN', help='postprocessing plugin; may be repeated'
+        )
+        ap_export_type.add_argument(
+            '--cfg', dest='pp_cfg',
```
FreddieWitherden: This might fit on two lines with different breaking.

FreddieWitherden:
STLPostProcAdapter doesn't seem to be used?

pyfr/writers/vtk/base.py
```diff
@@ -390,6 +394,29 @@ def __init__(self, meshf, pname=None, *, prec='single', order=None,
...
+                                   f'gradient data in the solution')
+            pp.process(adapter)
+
+        # Inject into soln.aux with (neles, nupts[, ncomps]) convention
+        aux = self.soln.aux.setdefault(etype, {})
+        for fname, arrs in adapter.fields.items():
+            if len(arrs) == 1:
+                aux[fname] = arrs[0].T.astype(self.dtype)
+            else:
+                aux[fname] = np.stack(arrs, axis=-1).transpose(1, 0, 2) \
```
FreddieWitherden: Avoid \.

pyfr/writers/vtk/base.py
```diff
@@ -390,6 +394,29 @@ def __init__(self, meshf, pname=None, *, prec='single', order=None,
...
+        for pp in plugins:
+            if pp.needs_grads and not self._gradients:
+                raise RuntimeError(f'Postproc {pp.name} requires '
+                                   f'gradient data in the solution')
+            pp.process(adapter)
+
+        # Inject into soln.aux with (neles, nupts[, ncomps]) convention
+        aux = self.soln.aux.setdefault(etype, {})
+        for fname, arrs in adapter.fields.items():
+            if len(arrs) == 1:
```
FreddieWitherden: See if we can get a solution which works for len 1 and not equal to one.

pyfr/writers/vtk/base.py
```diff
@@ -558,6 +585,29 @@ def _load_soln(self, *args, **kwargs):
...
+            cfg = self._pp_cfg or self.cfg
+            self.pp_plugins = [
+                get_plugin('postproc', name, self.ndims, cfg, self.type)
+                for name in self._pp_plugin_names
+            ]
+
+            # Run volume-capable pp plugins at upts and inject into aux
+            volume_pp = [p for p in self.pp_plugins
+                         if re.fullmatch(p.export_types, 'volume')]
+            if volume_pp:
```
FreddieWitherden: See if we can avoid the if, better to have things just work for empty inputs.

pyfr/writers/vtk/base.py
```diff
@@ -558,6 +585,29 @@ def _load_soln(self, *args, **kwargs):
...
+                for etype in self.mesh.eidxs:
+                    self._run_volume_postprocs(etype, volume_pp)
+
+            # Rebuild _extra_fields after aux augmentation
+            self._extra_fields = list(
```
FreddieWitherden: [] is more common here.

pyfr/writers/vtk/base.py
```diff
@@ -730,27 +780,29 @@ def _extra_point_shapes(self, etype):
...
 
+    def _field_kind(self, name, etype):
+        shape = self.soln.aux[etype][name].shape[1:]
+        pshapes = self._extra_point_shapes(etype)
+        if shape in pshapes:
+            return 'point', 1
+        elif shape[:-1] in pshapes:
+            return 'point', shape[-1]
+        else:
+            return 'cell', int(np.prod(shape))
```
FreddieWitherden: Check if cast is needed (it can be fiddly with np.prod).

pyfr/writers/vtk/base.py
```diff
@@ -730,27 +780,29 @@ def _extra_point_shapes(self, etype):
...
-        aux = self.soln.aux[etype]
-        shape = aux[name].shape[1:]
-
-        # Point fields have one component; cell fields flatten extra dims
-        pshapes = self._extra_point_shapes(etype)
-        ncomps = 1 if shape in pshapes else int(np.prod(shape))
+        _, ncomps = self._field_kind(name, etype)
 
-        adtype = aux[name].dtype
+        adtype = np.dtype(self.soln.aux[etype][name].dtype)
```
FreddieWitherden: Check if the np.dtype call is needed.

pyfr/writers/vtk/boundary.py
```diff
@@ -59,10 +77,32 @@ def _load_soln(self, *args, **kwargs):
...
-                self._surface_info[stype].append((etype, *info))
+                self._surface_info[stype].append(info)
 
         self.einfo = list(ecount.items())
 
+    def _run_postprocs(self, adapter):
+        for pp in self._boundary_pp:
+            if pp.needs_grads and not self._gradients:
+                raise RuntimeError(f'Postproc {pp.name} requires '
+                                   f'gradient data in the solution')
```
FreddieWitherden: No need for f.

pyfr/writers/vtk/boundary.py
```diff
@@ -59,10 +77,32 @@ def _load_soln(self, *args, **kwargs):
...
+            pp.process(adapter)
+
+        return {fname: np.dstack(arrs).astype(adapter.dtype)
+                for fname, arrs in adapter.fields.items()}
+
+    def _field_kind(self, name, etype):
+        if name in self._boundary_pp_fields:
+            return 'point', len(self._boundary_pp_fields[name])
+        return super()._field_kind(name, etype)
```
FreddieWitherden: if/lse

pyfr/writers/vtk/boundary.py
```diff
@@ -59,10 +77,32 @@ def _load_soln(self, *args, **kwargs):
...
+
+    def _field_info(self, name, etype):
+        if name in self._boundary_pp_fields:
+            _, ncomps = self._field_kind(name, etype)
+            adtype = np.dtype(self.dtype)
+            return self._vtk_dtype(adtype), adtype.itemsize*ncomps, ncomps
+        return super()._field_info(name, etype)
```
FreddieWitherden: if/else and check dtype cast.

pyfr/writers/vtk/stl.py
```diff
@@ -136,29 +136,42 @@ def _load_soln(self, *args, **kwargs):
...
-            arr = np.empty((nupts, nvars + len(anames), neles))
+            arr = np.empty((nupts, nvars + naux, neles))
             arr[:, :nvars] = soln.data[etype]
-            for i, name in enumerate(anames, start=nvars):
-                arr[:, i] = soln.aux[etype][name].T
+
+            off = nvars
+            for name, n in aux_info:
+                a = soln.aux[etype][name]
+                if n == 1:
```
FreddieWitherden: See if we can unify.

pyfr/writers/vtk/base.py
kaschau: yeah now that things like ploc are lazy this is no longer needed, it was mean as an early exit so we dont waste time when not postprocessing

pyfr/writers/vtk/boundary.py
kaschau: cast is needed for itemsize to be a value

FreddieWitherden: See if it is worth calling when self sets dtype (I think this is what backends do, but the dtype is a confusing NumPy mess).

pyfr/writers/vtk/boundary.py
```diff
@@ -133,32 +181,46 @@ def _prepare_pts(self, itype):
...
+                    pshape = shape
+                elif shape[:-1] in pshapes:
+                    pshape = shape[:-1]
                 else:
                     cellf[fname].append(data)
+                    continue
+
+                op = soln_op if pshape == (nupts,) else lin_op
+                pointf[fname].append(
+                    interpolate_pts(op, np.moveaxis(data, 0, 1))
```
FreddieWitherden, line 183: Does this fit on one line?

pyfr/plugins/postproc/adapters.py
```diff
@@ -0,0 +1,104 @@
...
+from pyfr.shapes import BaseShape
+from pyfr.util import subclass_where
+
+
+class BasePostProcAdapter:
+    pass
+
+
+class PostProcData(BasePostProcAdapter):
+    # pris: (nvars*(1+ndims), *pts_shape)
```
FreddieWitherden: ?

pyfr/plugins/postproc/adapters.py
```diff
@@ -0,0 +1,104 @@
...
+        self.soln = soln
+        self.nvars = len(soln.fields)
+        self.ndims = ploc.shape[0]
+        self._pris = pris
+        self.ploc = ploc
+        self.fields = {}
+
+    @cached_property
+    def pris(self):
+        return [self._pris[i] for i in range(self.nvars)]
```
FreddieWitherden: `list(self._pris[:self.nvars])`

pyfr/plugins/postproc/adapters.py
```diff
@@ -0,0 +1,104 @@
...
+
+    @cached_property
+    def grad_pris(self):
+        if len(self._pris) <= self.nvars:
+            return None
```
FreddieWitherden: Should we raise?

pyfr/plugins/postproc/adapters.py
```diff
@@ -0,0 +1,104 @@
...
+    def normals(self):
+        return self.pnorm / np.linalg.norm(self.pnorm, axis=0)
+
+    @cached_property
+    def min_upt_wall_dist_approx(self):
+        shape = self._shape
+        _, proj, norm = shape.faces[self._finfo.fidx]
+        upts = shape.upts
+
+        norm = norm / np.linalg.norm(norm)
```
FreddieWitherden: /=

pyfr/plugins/postproc/adapters.py
```diff
@@ -0,0 +1,104 @@
...
+        face_pt = proj(*([0]*(shape.ndims - 1)))
```
FreddieWitherden, line 69: Is this always correct for quads and tris?

pyfr/writers/vtk/base.py
```diff
@@ -512,7 +526,19 @@ def _load_soln(self, *args, **kwargs):
         etype = next(k for k in self.soln.data if k in self.mesh.eidxs)
         self._extra_etype = etype
 
-        self._extra_fields = list(self.soln.aux.get(etype, {}))
+        # Extra fields: classify aux by shape, pp appended below
```
FreddieWitherden: pp appended below?

pyfr/writers/vtk/base.py
```diff
@@ -512,7 +526,19 @@ def _load_soln(self, *args, **kwargs):
...
+        self._extra_fields = {}
+        pshapes = self._extra_point_shapes(etype)
+        for name, arr in self.soln.aux.get(etype, {}).items():
+            shape = arr.shape[1:]
+            if shape in pshapes:
+                self._extra_fields[name] = FieldMeta('point', 1, arr.dtype)
+            elif shape[:-1] in pshapes:
+                kind, ncomps = 'point', shape[-1]
```
FreddieWitherden: Inline kind?

FreddieWitherden: We should have point, cell, and full fat field.

pyfr/writers/vtk/base.py
```diff
@@ -558,6 +584,23 @@ def _load_soln(self, *args, **kwargs):
...
+        self.pp_plugins = [get_plugin('postproc', name, self.ndims, cfg,
+                                      self.type)
+                           for name in self._pp_plugin_names]
+
+        # Append pp fields to extra fields with their metadata
+        for pp in self.pp_plugins:
+            for name, varnames in pp.fields().items():
+                if name in self._extra_fields:
+                    raise ValueError(f'Postproc field {name!r} collides with '
+                                     'an existing field.')
```
FreddieWitherden: Usually no .

pyfr/plugins/postproc/adapters.py
```diff
@@ -0,0 +1,100 @@
...
+        return dist[t != 0].min(axis=0)
+
+    @cached_property
+    def tau_wall(self):
+        mu = self.cfg.getfloat('constants', 'mu')
+        normals = self.normals
+
+        grad_vel = np.stack(self.grad_pris[1:self.ndims + 1])
+        sij = grad_vel + grad_vel.swapaxes(0, 1)
+        tau_n = mu * np.einsum('ijkl,jkl->ikl', sij, normals)
```
FreddieWitherden: Minor, but this neglects Sutherlands law.

pyfr/writers/vtk/boundary.py
```diff
@@ -27,10 +32,21 @@ def __init__(self, meshf, boundaries, **kwargs):
             raise RuntimeError('Boundary export only supported for 3D grids')
 
     def _load_soln(self, *args, **kwargs):
+        # Initialise empty surface_info so super()'s aux classification can
+        # call our _extra_point_shapes override (it falls through to etype)
+        self._surface_info = defaultdict(list)
```
FreddieWitherden: Seems a bit fragile.

FreddieWitherden:
Check that the code complains in the expected way if given a tavg file.

pyfr/writers/vtk/base.py
```diff
@@ -558,6 +584,23 @@ def _load_soln(self, *args, **kwargs):
             self._vtk_vars = {k: [k] for k in self._soln_fields}
             self.tcurr = None
 
+        # Instantiate postproc plugins
+        from pyfr.plugins import get_plugin
+
+        cfg = self._pp_cfg or self.cfg
+        self.pp_plugins = [get_plugin('postproc', name, self.ndims, cfg,
```
FreddieWitherden: Any reason not to fuse with the loop below.

pyfr/plugins/cli/sampler.py
```diff
@@ -450,13 +282,41 @@ def sample_cmd(self, args):
...
 
         # Sample the solution
         samps = sampler.sample(sdata, process=process)
 
+        # Run postproc plugins at the sampled points (primitive only)
+        if rank == root and pp_plugins:
+            samps_t = samps.T.astype(np.float64)
+            ploc = pts.T
+            adapter = PostProcData(soln.config, soln, samps_t, ploc)
+            for pp in pp_plugins:
```
FreddieWitherden: May be able to factor a bit of this logic out as it appears in a couple of places (same with getting the plugins themselves).

FreddieWitherden (changes requested):

pyfr/plugins/postproc/adapters.py
kaschau: So the post proc adapters now require data in a specified shape, and the callers are responsible for the reshapes.

pyfr/plugins/postproc/adapters.py
kaschau: There is already a check upstream for any plugin that needs gradients but the soln does not have one and explicitly tells the user which plugin is failing, so really we dont even need this return None as we should never error at this point.

pyfr/writers/vtk/base.py
kaschau: what is a full fat field?

FreddieWitherden: Degree p.

kaschau: so all point fields collapse here and get inspected by dimension at write for things like interpolation op: `op = soln_op if pshape == (nupts,) else lin_op`. do you want to increase the FieldMeta to capture this at load time?

kaschau:
> Check that the code complains in the expected way if given a tavg file.

addressed

pyfr/writers/vtk/base.py
FreddieWitherden: I think this should be fine, so long as we can correctly handle these aux fields in the future.

pyfr/plugins/postproc/adapters.py
kaschau: yeah, maybe theres a better way to do this, but currently we just approximate a min height above the wall per element (not per face vis point) as a proxy, since y+ is such a fudgy number anyway and we just need an approximation. So we just need ANY point on the reference face, and we are dotting to take normals from there.

pyfr/plugins/postproc/adapters.py
FreddieWitherden: More wondering about the comment

pyfr/plugins/postproc/adapters.py
```diff
@@ -0,0 +1,115 @@
...
+    def grad_pris(self):
+        return np.split(self._pris[self.nvars:], self.nvars)
+
+    @cached_property
+    def mu(self):
+        cfg = self.cfg
+        mu_ref = cfg.getfloat('constants', 'mu')
+
+        if cfg.get('solver', 'viscosity-correction', 'none') != 'sutherland':
+            return mu_ref
```
FreddieWitherden: if/else

pyfr/plugins/cli/sampler.py
```diff
@@ -450,13 +282,34 @@ def sample_cmd(self, args):
...
+            adapter = PostProcData(soln.config, soln, samps.T, pts.T)
+            pp_field_map = {}
+            for pp in pp_plugins:
+                pp.run(adapter)
+                pp_field_map.update(pp.fields())
+
+            extra_cols = []
+            for name, arr in adapter.fields.items():
+                fields.extend(pp_field_map[name])
+                extra_cols.append(arr if arr.ndim == 2 else arr[:, None])
```
FreddieWitherden: atleast2d

pyfr/plugins/postproc/base.py
```diff
@@ -0,0 +1,54 @@
...
+        if self.needs_grads and not data.has_grads:
+            raise RuntimeError(f'Postproc {self.name} requires gradient '
+                               'data in the solution')
+        self._process(data)
+
+    def _process(self, data):
+        pass
+
+
+def resolve_pp_plugins(names, ndims, cfg, export_type):
```
FreddieWitherden: get_pp_plugins?

pyfr/plugins/postproc/adapters.py
```diff
@@ -0,0 +1,102 @@
...
+from pyfr.shapes import BaseShape
+from pyfr.util import subclass_where
+
+
+class PostProcData:
+    def __init__(self, cfg, soln, pris, ploc):
+        self.cfg = cfg
+        self.soln = soln
+        self.nvars = len(soln.fields)
+        self.ndims = ploc.shape[0]
```
FreddieWitherden: len(ploc)

pyfr/plugins/postproc/adapters.py
```diff
@@ -0,0 +1,102 @@
...
+    @cached_property
+    def pris(self):
+        return list(self._pris[:self.nvars])
+
+    @cached_property
+    def grad_pris(self):
+        return np.split(self._pris[self.nvars:], self.nvars)
+
+    @cached_property
+    def mu(self):
```
FreddieWitherden: This is annoyingly NS specific in the base adapter.

pyfr/plugins/postproc/base.py
```diff
@@ -0,0 +1,52 @@
...
+            raise RuntimeError(f'Postproc {self.name} requires gradient '
+                               'data in the solution')
+        self._process(data)
+
+    def _process(self, data):
+        pass
+
+
+def get_pp_plugins(names, ndims, cfg, export_type):
+    available = {c.name: c for c in BasePostProcPlugin.__subclasses__()}
```
FreddieWitherden: I think the PyFR method is better here as it handles inheritance

pv101:
Hi @kaschau - I got

tau_wall.py", line 19, in _process
    mu = data.fields['_mu']
         ~~~~~~~~~~~^^^^^^^
KeyError: '_mu'

when running the Cf plugin - are you able to reproduce?

pv101:
Also, for the vorticity and Mach plugins, are the .ini file section headings required even though no parameters are required?

kaschau:
> Also, for the vorticity and Mach plugins, are the .ini file section headings required even though no parameters are required?

cfg access is all local to a particular plugin, so no, if the plugin does not require data it will never ping the cfg. so no section required.

FreddieWitherden:
Thoughts about using keys in the `[constants]` block for this?  We do this already for gamma, Pr, so keys for free-stream are not unreasonable and may already exist.

kaschau:
> Hi @kaschau - I got
> 
> tau_wall.py", line 19, in _process mu = data.fields['_mu'] ~~~~~~~~~~~^^^^^^^ KeyError: '_mu'
> 
> when running the Cf plugin - are you able to reproduce?

@pv101 its not coming up on my tests, but I think I can see why it may happen for a particular case. Please try your case again with this most recent push.

kaschau:
> Thoughts about using keys in the `[constants]` block for this? We do this already for gamma, Pr, so keys for free-stream are not unreasonable and may already exist.

Only concern is consistency, some plugins need bools or other one-offs that don't naturally fit into the `[constants]` section heuristically, though, this is a bit canonical in pyfr already (pulling some things from a cfgsect and some from constants. For example, NIRF requires a `True` if you want to include translation in your re-orientation postproc. So not sure that makes sense in the constants section.

FreddieWitherden:
Yes, certainly some will need their own sections as they are a bit more involved and I think that is okay for the most part.

PR #542, opened 2026-03-27
Implicit

FreddieWitherden:
This is a major PR which:
 - Removes dual time stepping and artificial compressibility
 - Adds proper implicit time stepping with GMRES and optional block Jacobi preconditioning (beta)
 - Adds a baseline partitioner which is pure Python
 - Improves a lot of the backend infrastructure, including the use of atomic operations for reductions
 - Updates the file format to support auxiliary fields and a new `pyfr upgrade` command to update old files
 - Greatly improves artificial viscosity (can now be C0, but shock sensor still needs work)
 - Adds a `pyfr mesh` command for analysing mesh quality
 - Accelerates most plugins via the backend
 - Adds a trigger mechanism which can be used for trigger plugins when certain criteria are met (see documentation)
 - Adds a CFL time step controller which can run at a fixed CFL with the wavespeed being computed from the flow field
 - Improves STL export through spherigon subdivision
 - Improves the solution resampling with WENO-Z and TENO reconstructions
 - Project modernisation with a proper `pyproject.toml`


pyfr/quadrules/surface.py
```diff
@@ -8,40 +6,29 @@
...
+            ploc = eles.ploc_at_np(ppts)[..., eidxs]
+            pnorm = eles.pnorm_at(ppts, [norm]*nppts)[:, eidxs]
 
-        self.eidxs = {k: np.array(v) for k, v in eidxs.items()}
-        self.norms = {k: np.array(v) for k, v in norms.items()}
-        self.locs = {k: np.array(v) for k, v in locs.items()}
+            self.m0[etype, fidx] = eles.basis.ubasis.nodal_basis_at(ppts)
+            self.qwts[etype, fidx] = pwts
+            self.eidxs[etype, fidx] = eidxs
+            self.norms[etype, fidx] = pnorm.transpose(1, 0, 2)
```
kaschau: Im getting different layouts for norms (neles, nfpts, ndims) and locs (ndims, nfpts, neles)

I think we need both  to be (ndims, nfpts, neles) for things like the fluidforce plugin to broadcast correctly (loc - morigin)

FreddieWitherden: Will fix

pyfr/writers/upgrade.py
```diff
@@ -0,0 +1,135 @@
...
+
+    # Copy everything except eles and version
+    for k in src:
+        if k not in ('eles', 'version'):
+            src.copy(k, dst)
+
+    dst['version'] = 2
+
+    # Read element datasets
+    eles = {etype: src[f'eles/{etype}'][:] for etype in src['eles']}
```
kaschau: the `'colour'` key is causing issues here as it doesnt exist in the `src`

pyfr/solvers/baseadvec/entfilter.py
```diff
@@ -0,0 +1,215 @@
...
+            'linearise': form == 'linearised',
+            'ubdegs': [int(max(dd)) for dd in ub.degrees],
+        }
+
+    def _setup_interfaces(self, system, int_inters, mpi_inters, bc_inters):
+        be = self._be
+
+        # Create interface views for entmin using system API
+        iint_v, mpi_v, bc_v = system.make_field_views(
+            self._entmin, layout='face', bc_layout='face-expand'
```
kaschau, line 124: need a guard somewhere for p=0, entmin is empty and make field key errors

kaschau: likely just want to None self._ef if p=0

pyfr/solvers/baseadvec/entfilter.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+class EntropyFilter:
+    def __init__(self, backend, cfg, system, int_inters, mpi_inters,
+                 bc_inters):
+        self._be = backend
+
+        # Register pointwise kernel templates
+        backend.pointwise.register(
+            'pyfr.solvers.euler.kernels.entropylocal'
```
kaschau: would be nice if we split this out a bit so as not to hard code downstream kernels, as these vary by system (and even EOS)

kaschau: perhaps in BaseFluidElement derive an entropy_kernel_prefix attr and use that

PR #543, opened 2026-04-13
Add annotation support to the backend API.

FreddieWitherden:
Currently only CUDA and HIP are supported.

### DIFF
--- a/pyfr/backends/base/backend.py
+++ b/pyfr/backends/base/backend.py
@@ -1,4 +1,5 @@
 from collections import namedtuple
+from contextlib import contextmanager
 from functools import cached_property, wraps
 from itertools import count
 import math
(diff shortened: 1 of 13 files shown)


WillTrojak:
I think a note should be added to the documentation on usage and potentially what regions are supported. 

At the moment the regions end once the host returns, which may not be the same time as when the device returns. It might be worth adding some explicit stream syncs when annotating. This will reduce the performance of runs with annotation a little, but can be dropped from production runs.

FreddieWitherden:
Let me know if this works for you:
```python
    @contextmanager
    def region(self, name):
        if self._annotator:
            self._annotator.push(name)

        try:
            yield
        finally:
            if self._annotator:
                self.wait()
                self._annotator.pop(name)
```

WillTrojak:
Yep looks good, I think there are just some indents that are tabs instead of spaces, but ruff should pick that up.

PR #549, opened 2026-04-30
Correct pp plugin ndims and cfg clarification.

kaschau:
Move vtk point padding to after postproc runs so true ndims are computed and plugins behave intuitively w.r.t ndims. Also clear up confusing multiple cfg definitions between plugins and data adapters.

Now access to the simulation's cfg is through the adapter's soln.cfg and the (possibly extra) CLI cfg is accessed through the plugin's cfg attr.


FreddieWitherden:
While we are here I think it is worth making fields either a class member (simple case) or a @property in the complex (ndim dependent) case.

PR #551, opened 2026-05-11
Reset surface_info for batch processing.

kaschau:
Reset surface_info between each batch file.

### DIFF
--- a/pyfr/writers/vtk/base.py
+++ b/pyfr/writers/vtk/base.py
@@ -3,6 +3,7 @@ from pathlib import Path
 
 import numpy as np
 
+from pyfr.cache import memoize
 from pyfr.mpiutil import get_comm_rank_root, mpi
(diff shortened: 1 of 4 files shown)


FreddieWitherden:
Do we need to reset it before the `super()` or can we do it further down?

kaschau:
has to be before as `_build_extra_fields` uses it via `_extra_point_shapes` -> `_resolve_etype`

kaschau:
Thats why it got bumped into __init__ the first time around

FreddieWitherden:
Okay, might be worth seeing if there are any nice refactorings we can do here to clean this up a little.

FreddieWitherden:
Just looking through it seems as if both the volume and the boundary export want to add in the linear (P = 1) shapes.  Wonder if it is worth doing in the base class (but need to check nothing else breaks and that STL is still happy).

PR #552, opened 2026-05-11
Merge aux_field across ranks so alltoall agrees on size

kaschau:
Merge aux_fields across ranks so the per-etype dtype is identical everywhere — otherwise the persistent Alltoallv buffers don't agree and we get MPI_ERR_TRUNCATE at wait time.

### DIFF
--- a/pyfr/writers/native.py
+++ b/pyfr/writers/native.py
@@ -139,6 +139,13 @@ class NativeWriter:
                          aux_fields=None, *, ndims=0):
         comm, _, _ = get_comm_rank_root()
 
+        # Merge aux_fields across ranks
+        if aux_fields is not None:
+            merged = {}
+            for r_aux in comm.allgather(aux_fields):
+                merged.update(r_aux)
+            aux_fields = merged
+
         # Prepare the element information
         self._einfo = {}
         self._futures = {}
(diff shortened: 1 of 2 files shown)


pyfr/writers/native.py
```diff
@@ -139,6 +139,13 @@ def set_shapes_eidxs(self, shapes, eidxs, field_groups,
                          aux_fields=None, *, ndims=0):
         comm, _, _ = get_comm_rank_root()
 
+        # Merge aux_fields across ranks
+        if aux_fields is not None:
```
FreddieWitherden: Can we ever have one rank with `aux_fields = None` and another without?

pyfr/writers/native.py
```diff
@@ -139,6 +139,13 @@ def set_shapes_eidxs(self, shapes, eidxs, field_groups,
...
+            merged = {}
+            for r_aux in comm.allgather(aux_fields):
+                merged.update(r_aux)
+            aux_fields = merged
```
FreddieWitherden: Think this can be one lined with a dict comp `{k: v for aux in comm.allgather(aux_fields) for k, v in aux.items()}` or something like this.

PR #554, opened 2026-05-13
Add spanwise averaging and support for clean-to-grid.

FreddieWitherden:


TobyFlynn (approved):
I've had another look through and everything looks good. I've tested extrude, sampling, serial and MPI for the spanwise export. I took a closer look at the canonical functions and they look good to me. I've also checked that the node maps weren't messed up when they were moved around.

PR #555, opened 2026-05-13
Improve the performance of implicit preconditioner construction.

FreddieWitherden:


pyfr/backends/cuda/kernels/base.mako
```diff
@@ -29,12 +29,70 @@ __device__ void atomic_${op}_fpdtype(fpdtype_t* addr, fpdtype_t val)
 }
 % endfor
 
+struct half {
```
FreddieWitherden, line 32: @WillTrojak Disgusting!

WillTrojak: Why do you need to handle this explicitly? I think the compiler should be able to do this without the need to call down to PTX: https://godbolt.org/z/7vExGxq1h

FreddieWitherden: The need for an include is a big issue with NVRTC as the SDK can basically be anywhere.

PR #556, opened 2026-05-15
Support for GiMMiK PTX Provider

WillTrojak:
This adds support for the PTX GiMMiK backend of GiMMiK [PR 18](https://github.com/PyFR/GiMMiK/pull/18).

Besides adding calls to the new provider, PTX GiMMiK adds new dense kernels. Some of these kernels use 2D TMA copy instructions which require a device side tensor descriptor which has to be initialised on the host by the driver. 

Additionally, an optimisation has been added that zeros the lower 32-bits from the GEMM constants which is best handled on the PyFR side rather than in GiMMiK. In testing this causes a tiny change in precision, but can lead to a large performance boost at higher orders.


WillTrojak:
The CUDA GiMMiK provider is still called, but this can probably be made conditional on PTX support. The GiMMiK PTX backend should work on Ampere, Hopper, and Blackwell. But if you want to continue supporting legacy GPUs, you may want to maintain the CUDA provider. Additionally this may be a good fallback for GeForce or Tegra products which have varying support for the features used.

FreddieWitherden:
Might be worth looking into having a single GiMMiK provider which uses __new__ to decide which instance to yield (CUDA or PTX).  This way we can keep compatibility with older GiMMiK (good for dependencies and deployments) and not use PTX where it may cause issues.

pyfr/backends/cuda/driver.py
```diff
@@ -571,3 +602,39 @@ def create_event(self, timing=False):
 
     def create_graph(self):
         return CUDAGraph(self)
+
+    def set_tensormap(self, tm_addr, a, tile, interleave=None, swizzle=None,
+                      l2_promotion=None, oob_fill=None):
+        lib = self.lib
+        a_ptr = a.data
+
+        dtype = 0
```
FreddieWitherden: Not needed?

pyfr/backends/cuda/driver.py
```diff
@@ -571,3 +602,39 @@ def create_event(self, timing=False):
...
+        ld = c_ulonglong(a.leaddim * a.itemsize)
+        tile_dim = (c_uint * ndims)(*tile)
+        tile_stride = (c_uint * ndims)(*([1] * ndims))
+
+        interleave = interleave or lib.TENSOR_MAP_INTERLEAVE_NONE
+        swizzle = swizzle or lib.TENSOR_MAP_SWIZZLE_NONE
+        l2_promotion = l2_promotion or lib.TENSOR_MAP_L2_PROMOTION_NONE
+        oob_fill = oob_fill or lib.TENSOR_MAP_FLOAT_OOB_FILL_NONE
+
+        align = 64
```
FreddieWitherden: Try allocating with NumPy like the OpenMP backend does (see its aligned allocation stuff).  Might also simplify the hasattr bit, too.

pyfr/backends/cuda/gimmik.py
```diff
@@ -119,3 +120,163 @@ def run(self, stream):
...
+            if best_kern is None:
+                raise NotSuitableError('GiMMiK produced no viable kernels')
+
+            kern, grid, block, dt, sharedb = best_kern
+            ws_info = best_ws_info
+            self._mul_kerns[ckey] = (kern, grid, block, dt, kname, ws_info,
+                                     sharedb)
+            finalize(a, lambda: self._mul_kerns.pop(ckey))
+
+        # Re-bind tensor maps in case build- and run-time pointers differ
```
FreddieWitherden: Is this safe?  As the same kernel can arise more than one (M0 is a good example) and so need to be careful with anything global that might not be constant throughout the life of the kernel.

pyfr/backends/cuda/driver.py
```diff
@@ -316,6 +340,12 @@ def __init__(self, cuda, cucode):
     def get_function(self, name, argspec):
         return CUDAFunction(self.cuda, self, name, argspec)
 
+    def get_global(self, name):
+        ptr = c_void_p()
```
FreddieWitherden: Can one line these assignments

pyfr/backends/cuda/driver.py
```diff
@@ -571,3 +602,39 @@ def create_event(self, timing=False):
...
+        lib = self.lib
+        a_ptr = a.data
+
+        dtype = 0
+        if a.dtype == np.float64:
+            dtype = lib.TENSOR_MAP_DATA_TYPE_FLOAT64
+        elif a.dtype == np.float32:
+            dtype = lib.TENSOR_MAP_DATA_TYPE_FLOAT32
+        else:
+            raise AttributeError(f"Type {a.dtype} tensor map not supported")
```
FreddieWitherden: Probably valueerror error

pyfr/backends/cuda/driver.py
```diff
@@ -571,3 +602,39 @@ def create_event(self, timing=False):
...
+
+        ndims = 2
+        dims = (c_ulonglong * ndims)(*(a.ncol, a.nrow))
```
FreddieWitherden: Can just pass `(c_ulonglong*ndims)(a.ncol, a.nrow)`

pyfr/backends/cuda/driver.py
```diff
@@ -571,3 +602,39 @@ def create_event(self, timing=False):
 
     def create_graph(self):
         return CUDAGraph(self)
+
+    def set_tensormap(self, tm_addr, a, tile, interleave=None, swizzle=None,
```
FreddieWitherden: Usually the CUDA driver wrappers are lower level and so do not assume what a CUDA matrix looks like (a.ncol, a.nrow, etc).  Probably best to refactor.

pyfr/backends/cuda/gimmik.py
```diff
@@ -1,6 +1,7 @@
+import os
```
FreddieWitherden: Is this needed?

pyfr/backends/cuda/driver.py
```diff
@@ -173,6 +191,12 @@ class CUDAWrappers(LibWrapper):
          POINTER(c_int), POINTER(c_void_p)),
         (c_int, 'cuModuleUnload', c_void_p),
         (c_int, 'cuModuleGetFunction', POINTER(c_void_p), c_void_p, c_char_p),
+        (c_int, 'cuModuleGetGlobal_v2', POINTER(c_void_p), POINTER(c_size_t),
+         c_void_p, c_char_p),
+        (c_int, 'cuTensorMapEncodeTiled',
```
FreddieWitherden: Does this impact our minimum supported version of CUDA?

pyfr/backends/cuda/gimmik.py
```diff
@@ -119,3 +120,163 @@ def run(self, stream):
...
+        # Trim the lower 32 bits of fp64 A constants
+        self.f64trim = backend.cfg.getbool('backend-cuda', 'gimmik-trim-fp64',
+                                           False)
+
+        # Kernel cache
+        self._mul_kerns = {}
+
+    @staticmethod
+    def _trim_f64_low32(x):
+        a = np.ascontiguousarray(np.asarray(x, dtype=np.float64)).copy()
```
FreddieWitherden: Can simplify.

pyfr/backends/cuda/driver.py
```diff
@@ -571,3 +605,30 @@ def create_event(self, timing=False):
 
     def create_graph(self):
         return CUDAGraph(self)
+
+    def set_tensormap(self, tm, a, dims, ld, tile, tile_stride=None,
+                      interleave=None, swizzle=None, l2_promotion=None,
+                      oob_fill=None):
+        a_ptr = a.data
```
FreddieWitherden: Assume a is a CUDA allocation

pyfr/backends/cuda/driver.py
```diff
@@ -571,3 +605,30 @@ def create_event(self, timing=False):
...
+
+        if a.dtype == np.float64:
+            dtype = self.lib.TENSOR_MAP_DATA_TYPE_FLOAT64
+        elif a.dtype == np.float32:
+            dtype = self.lib.TENSOR_MAP_DATA_TYPE_FLOAT32
+        else:
+            raise ValueError(f"Type {a.dtype} tensor map not supported")
```
FreddieWitherden: Single quotes

pyfr/backends/cuda/driver.py
FreddieWitherden: Make this symbol weak

pyfr/backends/cuda/gimmik.py
```diff
@@ -119,3 +86,151 @@ def run(self, stream):
...
+
+
+                # Set the parameters
+                params = kern.make_params(meta['grid'], meta['block'],
+                                          sharedb)
+
+                # Setup input args using tensor maps if required
+                tm = {'b_tile': meta.get('ws_b_tile'),
+                      'out_tile': meta.get('ws_out_tile')}
+                b_arg = (b.tensormap(tm['b_tile'])
```
FreddieWitherden: Avoid breaking single line if's as it can get really hard to read

pyfr/backends/cuda/gimmik.py
```diff
@@ -119,3 +86,151 @@ def run(self, stream):
...
+                }
+        except StopIteration:
+            pass
+
+        # Restore the output matrix
+        getattr(out, 'parent', out).set(out_np)
+
+        return best_kern
+
+    def _ptx_mul(self, arr, b, out, beta):
```
FreddieWitherden: Try and share code with cuda_mul as there is a lot of boiler plate at the top and bottom

pyfr/backends/cuda/gimmik.py
```diff
@@ -119,3 +86,151 @@ def run(self, stream):
...
+            pass
+
+        # Restore the output matrix
+        getattr(out, 'parent', out).set(out_np)
+
+        return best_kern
+
+    @staticmethod
+    def _trim_f64_low32(x):
+        a = np.ascontiguousarray(np.asarray(x, dtype=np.float64)).copy()
```
FreddieWitherden: This can be cleaned up

pyfr/backends/cuda/driver.py
```diff
@@ -571,3 +605,30 @@ def create_event(self, timing=False):
...
+
+        ndims = len(dims)
+        dims = (c_ulonglong*ndims)(*dims)
+        ld = (c_ulonglong*(ndims - 1))(*ld)
+        tile = (c_uint * ndims)(*tile)
```
FreddieWitherden: Be consistent with spacing around *

pyfr/backends/cuda/driver.py
```diff
@@ -529,10 +560,13 @@ def mem_info(self):
     def mem_alloc(self, nbytes):
         return CUDADevAlloc(self, nbytes)
 
+    def pagelocked(self, nbytes):
```
FreddieWitherden: Not sure this is needed? NumPy arrays are the best way of carrying allocations

pyfr/backends/cuda/driver.py
```diff
@@ -316,6 +340,12 @@ def __init__(self, cuda, cucode):
     def get_function(self, name, argspec):
         return CUDAFunction(self.cuda, self, name, argspec)
 
+    def get_global(self, name):
```
FreddieWitherden: Is this still needed?

pyfr/backends/cuda/driver.py
```diff
@@ -138,6 +142,20 @@ class CUDAWrappers(LibWrapper):
     FUNC_ATTR_PREFERRED_SHARED_MEMORY_CARVEOUT = 9
     MEMORYTYPE_UNIFIED = 4
 
+    TENSOR_MAP_DATA_TYPE_FLOAT32 = 7
```
FreddieWitherden, line 150: Are all of these constants needed?

pyfr/backends/cuda/gimmik.py
```diff
@@ -29,87 +33,50 @@ def mul(self, a, b, out, alpha=1.0, beta=0.0):
...
-            getattr(out, 'parent', out).set(out_np)
-
-            # Update the cache
-            self._mul_kerns[ckey] = kern, grid, block, dt = best_kern
+            # Fetch the matrix, premultiply, and optionally trim
+            arr = alpha*a.get()
+            if a.dtype == np.float64 and self.f64trim:
+                arr = self._trim_f64_low32(arr)
+
+            cc = self.backend.cuda.compute_capability()
```
FreddieWitherden: Do this in init

pyfr/backends/cuda/driver.py
WillTrojak: No, not at the moment. I have reduced to FP32, FP64 and the various NONEs.

pyfr/backends/cuda/types.py
```diff
@@ -3,13 +3,26 @@
...
     def _as_parameter_(self):
         return self.data
 
+    @memoize
+    def tensormap(self, tile, interleave=None, swizzle=None,
+                  l2_promotion=None, oob_fill=None):
+        dims = (self.ncol, self.nrow)
+        ld = (self.leaddim * self.itemsize,)
+        tile_stride = (1, 1)
+        args = (dims, ld, tile, tile_stride, interleave, swizzle, l2_promotion,
```
FreddieWitherden: Can't we just move these to the call site?

pyfr/backends/cuda/gimmik.py
```diff
@@ -119,3 +104,59 @@ def run(self, stream):
...
+
+                kdata = {
+                    'runtime': dt,
+                    'registers': kern.nreg,
+                    'local_mem': kern.local_mem,
+                    'shared_mem': kern.shared_mem
+                }
+        except StopIteration:
+            pass
+        return best_kern
```
FreddieWitherden, line 160: Add newline.

pyfr/backends/cuda/driver.py
```diff
@@ -56,7 +58,17 @@ def __init__(self, func, grid, block, sharedb):
...
     def set_arg(self, i, v):
-        self._args[i].value = getattr(v, '_as_parameter_', v)
+        # Handle implicit tensormap inputs
+        if isinstance(v, tuple):
+            a, tile, *other = v
+            if tile is not None:
+                kwargs = other[0] if other else {}
+                tm = a.tensormap(tile, **kwargs)
+                self._args[i].value = tm.ctypes.data
+            else:
```
FreddieWitherden: How can this case trigger?

FreddieWitherden:
Also, can you ensure PyFR still works with the old GiMMiK version if at all possible?  Avoids us having to immediately mind a new GiMMiK release.

pyfr/backends/cuda/driver.py
```diff
@@ -173,6 +189,8 @@ class CUDAWrappers(LibWrapper):
          POINTER(c_int), POINTER(c_void_p)),
         (c_int, 'cuModuleUnload', c_void_p),
         (c_int, 'cuModuleGetFunction', POINTER(c_void_p), c_void_p, c_char_p),
+        (c_int, 'cuModuleGetGlobal_v2', POINTER(c_void_p), POINTER(c_size_t),
```
FreddieWitherden: Do we still need this? I dislike globals.

pyfr/backends/cuda/gimmik.py
```diff
@@ -29,87 +37,67 @@ def mul(self, a, b, out, alpha=1.0, beta=0.0):
...
+            kern, grid, block, dshared, tm, dt = self._mul_kerns[ckey]
         except KeyError:
-            ifac = self.backend.autotune_ifac
-            kname = f'gimmik_mm_{arr.shape[0]}x{arr.shape[1]}'
-            kdata = None
-            best_kern = None
+            # Fetch the matrix, premultiply, and optionally trim
+            arr = alpha*a.get()
+            if a.dtype == np.float64 and self.f64trim:
+                arr = self._trim_f64_low32(arr)
```
FreddieWitherden: Can we not do astype(np.float32).astype(np.float64) or just leave it as float32

pyfr/backends/cuda/gimmik.py
```diff
@@ -29,87 +37,67 @@ def mul(self, a, b, out, alpha=1.0, beta=0.0):
...
-            self._mul_kerns[ckey] = kern, grid, block, dt = best_kern
-            finalize(a, lambda: self._mul_kerns.pop(ckey))
-
         # Set the parameters
-        params = kern.make_params(grid, block)
-        params.set_args(b, out)
+        params = kern.make_params(grid, block, dshared)
+
+        # Set the input args using tensormaps if needed
+        b_args = CUDAGiMMiKKernels._arg_pointer(b, tm.get('b_tile'))
```
FreddieWitherden: Check that b_args and out_args are not GC'ed.

pyfr/backends/cuda/gimmik.py
```diff
@@ -29,87 +37,67 @@ def mul(self, a, b, out, alpha=1.0, beta=0.0):
...
-        ldb, ldc = b.leaddim, out.leaddim
-
         # Alignment
         if 'align' in b.tags and 'align' in out.tags:
             aligne = self.backend.alignb // b.itemsize
         else:
             aligne = None
 
+        # Dimensions
+        ldb, ldc = b.leaddim, out.leaddim
```
FreddieWitherden: Not sure why this moved?

pyfr/backends/cuda/gimmik.py
```diff
@@ -29,87 +37,67 @@ def mul(self, a, b, out, alpha=1.0, beta=0.0):
...
+                arr = self._trim_f64_low32(arr)
+
+            # Alignment
+            if 'align' in b.tags and 'align' in out.tags:
+                aligne = self.backend.alignb // b.itemsize
+            else:
+                aligne = None
+
+            # Pick MM generator
+            if PTXMatMul.is_suitable(arr, self.cc):
```
FreddieWitherden: How do we handle older GiMMiK?

pyfr/backends/cuda/types.py
```diff
@@ -3,13 +3,25 @@
...
     @cached_property
     def _as_parameter_(self):
         return self.data
 
+    @memoize
+    def tensormap(self, tile, interleave=None, swizzle=None,
+                  l2_promotion=None, oob_fill=None):
+        dims = (self.ncol, self.nrow)
+        ld = (self.leaddim * self.itemsize,)
+        tm = self.backend.cuda.pagelocked_empty((128,), np.uint8)
```
FreddieWitherden, line 19: Is 128 arbitrary?

WillTrojak: No, the tensor map required 128 bytes.

pyfr/backends/cuda/gimmik.py
WillTrojak: GC'ed?

pyfr/backends/cuda/gimmik.py
WillTrojak: This is an optimisation for FP64 so not relevant to Fp32

pyfr/backends/cuda/driver.py
WillTrojak: No, this is no longer needed.

pyfr/backends/cuda/gimmik.py
FreddieWitherden: When we call set_arg it'll pull the pointer from b_args and then since we don't hold a reference to b_args it will go out and scope and get garbage collected.  So we want to be sure that whatever pointer set_arg got is still valid even after b_args goes out of scope.

pyfr/backends/cuda/gimmik.py
WillTrojak: It moved through the splitting of cuda and ptx and then a recombining. I'll move back.

pyfr/backends/cuda/gimmik.py
WillTrojak: This would require conditionally importing PTXMatmul from gimmik if it's detected. Seems like it would be messy

FreddieWitherden:
Once we have backwards compatibility I think this is basically good to merge.

pyfr/backends/cuda/gimmik.py
FreddieWitherden: We can have a _get_matmul method which does something like:
```python
def _get_matmul(self, arr):
    try:
        from gimmik import PTXMatMul
        ...
    except ImportError:
        pass

    # CUDA Path
```

I think we can also just import GiMMiK and then do a getattr based approach which means the entire thing can be list driven with the list having provider class names we want to try.

pyfr/backends/cuda/gimmik.py
WillTrojak: `b_args`/`out_args` is set to either b/out or a tensor descriptor for b/out. Then in `set_args` the kernel arguments will be set to a pointer to b/out or the tensormaps. The tensormap is cached in the cudamatrix class, so I don't see when either would go out of scope.

pyfr/backends/cuda/driver.py
FreddieWitherden: Just following up here.

pyfr/backends/cuda/driver.py
FreddieWitherden: Just following up here.

pyfr/backends/cuda/gimmik.py
```diff
@@ -12,14 +12,22 @@ def __init__(self, backend):
...
 
         # Maximum number of kernels to consider
-        self.nkerns = backend.cfg.getint('backend-cuda', 'gimmik-nkerns', 8)
+        self.nkerns = backend.cfg.getint('backend-cuda', 'gimmik-nkerns', 10)
 
         # Number of benchmarking runs
         self.nbench = backend.cfg.getint('backend-cuda', 'gimmik-nbench', 5)
 
+        # Trim the lower 32 bits of fp64 A constants
+        self.f64trim = backend.cfg.getbool('backend-cuda', 'gimmik-trim-fp64',
```
FreddieWitherden: Usually use double for fp64.

pyfr/backends/cuda/gimmik.py
FreddieWitherden: Can use self here I think. Shorter.

FreddieWitherden:
Are you okay to address the merge conflicts?

pyfr/backends/cuda/gimmik.py
FreddieWitherden: Just following up here.

pyfr/backends/cuda/gimmik.py
FreddieWitherden: Can we not call this through self?

pyfr/backends/cuda/gimmik.py
```diff
@@ -119,3 +111,67 @@ def run(self, stream):
...
+    def _arg_pointer(a, tile=None, **kwargs):
+        if tile is not None:
+            tm = a.tensormap(tile, **kwargs)
+            return tm.ctypes.data
+        else:
+            return a
+
+    @staticmethod
+    def _trim_f64_low32(x):
+        if x.dtype != np.float64:
```
FreddieWitherden: Overly defensive. View should throw.

pyfr/backends/cuda/driver.py
FreddieWitherden: Can we assume this comes from the driver rather than being a CUDAMatrix (which is a higher level object).

pyfr/backends/cuda/driver.py
```diff
@@ -357,8 +379,9 @@ def _set_attr(self, attr, val):
         attr = getattr(self.cuda.lib, f'FUNC_ATTR_{attr.upper()}')
         self.cuda.lib.cuFuncSetAttribute(self, attr, val)
 
-    def set_shared_size(self, *, dynm_shared=0, carveout=None):
-        self._set_attr('max_dynamic_shared_size_bytes', dynm_shared)
+    def set_shared_size(self, *, dynm_shared=None, carveout=None):
```
FreddieWitherden: Are we still doing the dynamic shared thing?

WillTrojak: No, I'll remove.

pyfr/backends/cuda/driver.py
WillTrojak: Where this is used it currently takes a numpy array. I can change it to expect `a` to be a pointer

PR #558, opened 2026-05-22
Fixing OpenCL bugs

TobyFlynn:
The first issue was the `self.depk` (set of kernels that are depended on by something) was only being updated for kernels, not MPI dependencies. This means the OpenCL backend didn't generate events for kernels that are only depended on by a MPI request, so when the MPI request called `wait_for_events` we got an error. Only the OpenCL backend uses `self.depk`, so the other backends don't trigger this bug.

The second issue is that when the OpenCL backend was getting the events that kernels depend on, it was including the pdeps as well as the hard dependencies which lead to errors due to the way the event indices were being generated generated. Right now with this the OpenCL backend does not try to enforce the pdeps where possible like OpenMP and Metal, do we want to do that on this backend or only on CUDA/HIP which have the vendor graphs?

### DIFF
--- a/pyfr/backends/opencl/types.py
+++ b/pyfr/backends/opencl/types.py
@@ -87,7 +87,7 @@ class OpenCLGraph(base.Graph):
             evtidxs[k] = i
 
             # Resolve the event indices of kernels we depend on
-            wait_evts = [evtidxs[dep] for dep in self._alldeps(k)] or None
+            wait_evts = [evtidxs[dep] for dep in self.kdeps.get(k, [])] or None
 
             klist.append((k, wait_evts, k in self.depk))
 
--- a/pyfr/backends/opencl/types.py
+++ b/pyfr/backends/opencl/types.py
@@ -87,7 +87,8 @@ class OpenCLGraph(base.Graph):
             evtidxs[k] = i
 
             # Resolve the event indices of kernels we depend on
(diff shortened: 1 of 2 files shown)


FreddieWitherden:
So OpenCL should respect pdeps (it has the same underlying issue as CUDA/HIP graphs).  The other backends execute kernels in order and so can ignore pdeps.

TobyFlynn:
This should now emit an event that is waited on as long as the topological ordering was able to honour the pdep

pyfr/backends/opencl/types.py
```diff
@@ -87,7 +87,8 @@ def _commit(self):
             evtidxs[k] = i
 
             # Resolve the event indices of kernels we depend on
-            wait_evts = [evtidxs[dep] for dep in self._alldeps(k)] or None
+            wait_evts = [evtidxs[dep] for dep in self._alldeps(k)
```
FreddieWitherden, line 90: What causes the deps in all deps to not get given event ids?  Is it because they're pdeps?

FreddieWitherden:
Have pushed what I believe to be a more general fix.  We now only do the super-node aggregation for backends which actually have cache blocking and do rely on grouping.  This ensures that pdeps never lead to cycles.  Would have been nice to have a uniform implementation, but this seems like the best compromise.  All pdeps are satisfied except when cache blocking, but there the grouping automatically handles it.

TobyFlynn:
Yeah that fixes the issues that I was seeing

PR #559, opened 2026-05-22
fix ldg state for ns slip wall bc

z-yuan131:
This PR deals with a potential problem with slip wall. Currently slip wall uses ldg state as the same as rsolve state, which only flips the sign of the momentum and preserves the energy. However, I think ldg state should remove energy/momentum contribution from wall normal velocity and use that to do the gradient correction. A simple test case is conducted with a cylinder (r = 1) with freestream u = 0.1 (test case mesh and config are attached). 

First figure shows the comparison of the wall normal velocity: the new implementation can reduce quite a lot of the amplitude of the wall normal velocity.
wall_normal_velocity_comparison_s

The second figure shows ux (wall normal) along a line (y = 0) from upstream to the wall location (x = -1), the new implementation can reduce the oscillation in the first wall element.
ux_comparison_s

[slip_wall.zip](https://github.com/user-attachments/files/28163675/slip_wall.zip)

### DIFF
--- a/pyfr/solvers/navstokes/kernels/bcs/slp-adia-wall.mako
+++ b/pyfr/solvers/navstokes/kernels/bcs/slp-adia-wall.mako
@@ -27,4 +27,11 @@
 % endfor
 </%pyfr:macro>
 
-<%pyfr:alias name='bc_ldg_state' func='bc_rsolve_state'/>
+<%pyfr:macro name='bc_ldg_state' params='ul, nl, ur'>
+    fpdtype_t nor = ${' + '.join(f'ul[{i + 1}]*nl[{i}]' for i in range(ndims))};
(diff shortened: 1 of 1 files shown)


kaschau:
I think this is a sensible improvement worth integrating pending a larger investigation into wall BCs in the future.

FreddieWitherden:
> I think this is a sensible improvement worth integrating pending a larger investigation into wall BCs in the future.

Do you want to merge as-is or wait until we've gone through the other BCs in more detail?  The slip wall has certainly not been tested as much as the no-slip wall (and we do need to get a partial slip wall at some point).

kaschau:
I think this is worth merging now, as its likely as good as one can do without a refactor.

PR #560, opened 2026-05-26
Add surface support to Ascent.

FreddieWitherden:
Support is also added for post-proc plugins.  Further, tavg files can now be handled.

### DIFF
--- a/pyfr/plugins/soln/ascent.py
+++ b/pyfr/plugins/soln/ascent.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from ctypes import (RTLD_GLOBAL, c_char_p, c_double, c_int, c_int32, c_int64,
                     c_void_p)
 import re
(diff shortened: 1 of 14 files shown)


pyfr/plugins/soln/ascent.py
```diff
@@ -539,8 +861,8 @@ class AscentPlugin(BaseSolnPlugin):
     def __init__(self, intg, cfgsect, suffix=None):
         super().__init__(intg, cfgsect, suffix)
 
-        self._renderer = _AscentRenderer(_IntegratorAdapter(intg, cfgsect),
-                                         intg.isrestart)
+        adapter = _IntegratorAdapter(intg, intg.cfg, cfgsect)
```
kaschau, line 864: We should be able to store this and reuse in L868 instead of creating a new class each call. Its all references to intg

FreddieWitherden: The general rule is plugins shouldn't keep references to intg hence why we pass it in each time. This is to prevent reference cycles cause issues with del.

Issue #476, opened 2025-06-17
Add ParaView Catalyst 2 support


FreddieWitherden:
This is based on Conduit blueprints and so should be relatively straightforward given we support this already for Ascent.

PR #561, opened 2026-05-26
[WIP] Catalyst Plugin

kaschau:
This WIP adds a plugin to use Catalyst2 scripts generated in Paraview for in-situ processing, addressing #476.

In my tests it is significantly more performant than Ascent in the TGV case from PyFR-Test-Cases.

Live view is still being tested as I am limited by cluster port opening permission. It also splits out the common conduit machinery to share with Ascent.

### DIFF
--- a/doc/src/plugins/soln-plugin-ascent.rst
+++ b/doc/src/plugins/soln-plugin-ascent.rst
@@ -14,22 +14,85 @@ on-the-fly.  The following parameters can then be set:
 
     *int*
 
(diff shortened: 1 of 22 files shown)


FreddieWitherden:
Can you rebase on #560

FreddieWitherden:
That PR should let you add native PyFR support for postp plugins, surfaces, and clean to grid.  Should be able to share most of the code.

PR #563, opened 2026-06-03
Snapshots

kaschau:
This PR proposes a mechanism to unify host side subsample->interpolate->compute->export pathways such as plugins and non-Native writers. A `Snapshot` is, conceptually, a unifying interface to PyFR data such as `intg`, `soln`, or even just files from disks. This object presents a receiver with a unifying interface to access the underlying PyFR solution, stats, grad, whatever data produced directly by the solver. 

The second addition is a `SnapshotRegion`, a geometric (mesh) based object with typical volume, surface, point cloud, extrusion, etc. types available which work with with regions, divisors, boundaries, or supplied points. These hold expensive grid related operations such as cleaning, normal calculations, extrusion averaging, etc. 

Finally, `SnapshotSample` is a lightweight data container that produces and stores the derived quantities from an input `Snapshot`. Whatever is needed by the outputter (new fields, transformations, gradients, pris, etc.) are computed on the `SnapshotSample` (which is only computed on the `Region` holding the `Sample`). 

Because of this, the briefly existing PostProc has been renamed to just Fields, as that is all they do now: they add a field to a `Sample`'s existing fields. This also cleans up the CLI field filter with the removal of the `--fields` arg and replaced with:

`--add-field` <- which one can list cf,yplus,mach ...
`--remove-field` <- which behaves more like the old filter to prevent fields from being output.

Within INI files, one can arbitrarily
`add-field=yplus,mach` as well.

For multi source plugins like ascent/catalyst one would:
`add-field-volume=mach` (though I would prefer `add-field-mach=volume`)

This exposes a lot to a lot. For example, vtu files can be directly exported from directly during simulation:

```
[soln-plugin-vtu-vol]
mode = volume
dt-out = 1.0e-4
basedir = .
basename = vol-{n:06d}

[soln-plugin-vtu-walls]
mode = boundary
boundary = bc/wall
dt-out = 1.0e-4
basedir = .
basename = walls-{n:06d}

[soln-plugin-vtu-span]
mode = spanwise
periodic = z
dt-out = 1.0e-4
basedir = .
basename = span-{n:06d}

[soln-plugin-vtu-stl]
mode = stl
stl = my_region
dt-out = 1.0e-4
basedir = .
basename = stl-{n:06d}
```

On the writer side, the main consumer changed is the VTK suite (Native is untouched as it has no use for this). But now the entire VTK group of friends have uniform behavior:

1. A VTK write creates a `SnapRegion` based on need
2. Writer sends the region a `Snapshot` (which can come from anywhere)
3. `SnapshotRegion` samples, computes, filters, etc. as appropriate
4.  Writer responsible for any outstanding writer specific conversion, reformatting, reshaping, etc. 

Caveat: due to this decoupling, the vtu writer is 10-20% SLOWER now. My assumption is due mainly to processing soln and grad_soln separately, as well as some overhead hit.

**Some outstanding items:**
- Make a decision on compute then sample or sample then compute. Mostly matters for spanwise, currently, we add fields to the sampled data. So Mach of average field, not average of  Mach field. Not sure which is more useful
- Making a better data -> names of outputs mapping, some single source of truth. current implementation is a bit all over and I tried to maintain current conventions. 
- Related, solidifying the incoming data (soln, grad_soln, stats, aux) and then derived data (fields, pris, grad_pris) these are currently a bit unwieldy and canonizing them in some form would clean up a lot
- Documentation updates


FreddieWitherden:
@pv101 @TobyFlynn Thoughts on interpolate then compute vs compute then interpolate?

pyfr/plugins/soln/catalyst.py
```diff
@@ -0,0 +1,342 @@
+# ParaView Catalyst 2.0 In-Situ Visualization Plugin
+#
```
FreddieWitherden: Guess we eventually want to move this into the docs?

pyfr/plugins/soln/catalyst.py
```diff
@@ -0,0 +1,342 @@
...
+from pyfr.plugins.soln.base import BaseSolnPlugin
+from pyfr.snapshot import IntgSnapshot
+from pyfr.plugins.soln.insitu import (ConduitNode, ConduitWrappers,
+                                      InSituError, InSituRenderer)
+
+
+class CatalystError(InSituError): pass
+
+
+def _load_catalyst_lib():
```
FreddieWitherden: Can we not use the normal path for this (subclass LibWrapper but extended if needed to support PyDLL although usually PyDLL should not be needed)?

pyfr/plugins/soln/catalyst.py
```diff
@@ -0,0 +1,342 @@
...
+                                mode=RTLD_GLOBAL)
+        except OSError:
+            pass
+
+    return ctypes.PyDLL(lname, mode=RTLD_GLOBAL)
+
+
+class CatalystConduitWrappers(ConduitWrappers):
+    _libname = 'catalyst'
+    _functions = [(ret, f'catalyst_{fn}', *args)
```
FreddieWitherden: Ugh, do they really give their own names to all of the conduit stuff?

pyfr/plugins/soln/vtu_writer.py
```diff
@@ -0,0 +1,91 @@
...
+        common = dict(prec=prec, order=order, divisor=divisor,
+                      add_fields=add_fields, remove_fields=remove_fields,
+                      discontinuous=not clean)
+        mesh, scfg = intg.system.mesh, intg.cfg
+
+        if mode == 'volume':
+            self._writer = VTKVolumeWriter(mesh, scfg, **common)
+        elif mode == 'boundary':
+            bnd = cfg.get(s, 'boundary', '').strip()
+            if not bnd:
```
FreddieWitherden: Not sure if this is needed? Presumably there is a check further in when we construct the class that the boundary exists on at least one rank?

pyfr/plugins/soln/vtu_writer.py
```diff
@@ -0,0 +1,91 @@
...
+        else:
+            raise ValueError(f'{s}: unknown mode={mode!r}; choose volume, '
+                             'boundary, spanwise, or stl')
+
+    def __call__(self, intg):
+        if intg.tcurr - self.tout_last < self.dt_out - self.tol:
+            return
+
+        outfname = str(self._basedir /
+                       self._basename.format(n=intg.nacptsteps,
```
FreddieWitherden: I think there is a Path method to handle adding extensions if not present, `with_suffix` maybe?  But see if you can use the generator.

pyfr/snapshot/base.py
```diff
@@ -0,0 +1,138 @@
...
+                                  dtype=dtype, source='data',
+                                  components=(name,), data_index=i)
+
+        if self.has_grads:
+            for i, name in enumerate(self._data_field_names):
+                gname = f'grad {name}'
+                out[gname] = FieldInfo(name=gname, kind='point',
+                                       ncomps=self.ndims, dtype=dtype,
+                                       source='grad_data',
+                                       components=tuple(f'{name}-{d}'
```
FreddieWitherden: Probably worth moving that tuple assignment onto its own line and then breaking the Field Info as
```
FieldInfo(
    foo, ...
    bar
)
```

pyfr/snapshot/base.py
```diff
@@ -0,0 +1,138 @@
...
+    def _aux_field_info(self, name, per_ele, dtype, pshapes):
+        if per_ele in pshapes:
+            return FieldInfo(name=name, kind='point', ncomps=1,
+                             dtype=dtype, source='aux', components=(name,))
+        if len(per_ele) > 1 and per_ele[:-1] in pshapes:
+            return FieldInfo(name=name, kind='point', ncomps=per_ele[-1],
+                             dtype=dtype, source='aux',
+                             components=tuple(f'{name}-{d}'
+                                              for d in range(per_ele[-1])))
+        ncomps = int(np.prod(per_ele) or 1)
```
FreddieWitherden: Check but I think this will always give an int.

pyfr/snapshot/fieldinfo.py
```diff
@@ -0,0 +1,16 @@
+from dataclasses import dataclass, field as dc_field
```
FreddieWitherden: Do we need to rename?

pyfr/snapshot/intg.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def __init__(self, intg):
+        sys = intg.system
+
+        self.mesh = sys.mesh
+        self.config = intg.cfg
+        self.elementscls = sys.elementscls
+        self.ele_types = list(sys.ele_types)
+        self.dtype = sys.backend.fpdtype
+        self.has_grads = True
```
FreddieWitherden: Check this, Euler simulations for example can't give gradients.

pyfr/snapshot/intg.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class IntgSnapshot(BaseSnapshot):
+    prefix = 'soln'
+
+    def __init__(self, intg):
+        sys = intg.system
+
+        self.mesh = sys.mesh
+        self.config = intg.cfg
```
FreddieWitherden: cfg for consistency

pyfr/snapshot/intg.py
```diff
@@ -0,0 +1,68 @@
...
+        self.elementscls = sys.elementscls
+        self.ele_types = list(sys.ele_types)
+        self.dtype = sys.backend.fpdtype
+        self.has_grads = True
+
+        self._data_field_names = list(self.elementscls.convars(
+            self.mesh.ndims, self.config))
+
+        self._export_fields = sys.export_fields
+        self._intg = intg
```
FreddieWitherden: Single we hold onto intg ensure that this object is short lived

pyfr/snapshot/intg.py
```diff
@@ -0,0 +1,68 @@
...
+    def cycle(self):
+        return self._intg.nacptsteps
+
+    @cached_property
+    def data(self):
+        return {et: self._intg.soln[i] for i, et in enumerate(self.ele_types)}
+
+    @cached_property
+    def grad_data(self):
+        if not self.has_grads:
```
FreddieWitherden: Do we want to return None here as opposed to an empty dict?

pyfr/snapshot/intg.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def aux(self, etype):
+        return {ef.name: ef.getter()
+                for ef in self._export_fields.get(etype, ())}
+
+    def aux_info(self, etype):
+        return {ef.name: (ef.shape, ef.dtype or self.dtype)
+                for ef in self._export_fields.get(etype, ())}
+
+    def to_pris(self, interp_data, cfg):
```
FreddieWitherden: Can cfg be different to the one we have as a member var?

pyfr/snapshot/intg.py
```diff
@@ -0,0 +1,68 @@
...
+    @property
+    def pris_names(self):
+        return list(self.elementscls.privars(self.ndims, self.config))
+
+    def primitive_var_groups(self):
+        return self.elementscls.visvars(self.ndims, self.config)
+
+    def compute_grads(self):
+        if self.has_grads:
+            self._intg.compute_grads()
```
FreddieWitherden: Usually we just get the attr and rely on the caching to decide if we need to compute the gradients or not.

pyfr/snapshot/region.py
```diff
@@ -0,0 +1,519 @@
...
+from pyfr.points import PointLocator, PointSampler
+from pyfr.shapes import BaseShape, proj_pts
+from pyfr.snapshot.sample import SnapshotSample
+from pyfr.subdiv import CleanToGrid
+from pyfr.util import subclass_where
+
+
+def interp_ops(op, arr, *, cast=False):
+    if cast:
+        op = op.astype(arr.dtype, copy=False)
```
FreddieWitherden: Probably easier to just cast at the end after the @.

pyfr/snapshot/region.py
```diff
@@ -0,0 +1,519 @@
...
+    return (op @ arr.reshape(arr.shape[0], -1)).reshape(op.shape[0],
```
FreddieWitherden: Break this.

pyfr/snapshot/region.py
```diff
@@ -0,0 +1,519 @@
...
+                                                        *arr.shape[1:])
+
+
+def _cfg_dtype(cfg):
+    return np.float32 if cfg.get('backend', 'precision', 'single') == 'single' \
```
FreddieWitherden: Probably easier for this not to be an inline if.

pyfr/snapshot/region.py
```diff
@@ -0,0 +1,519 @@
...
+        else np.float64
+
+
+def _cfg_elementscls(cfg):
+    from pyfr.solvers.base import BaseSystem
+    return subclass_where(BaseSystem, name=cfg.get('solver', 'system'))\
```
FreddieWitherden: Probably worth assigning the system to a class and then on another line doing the return.

pyfr/snapshot/region.py
```diff
@@ -0,0 +1,519 @@
...
+        avg = self.cleaner.average(stacked, nvars, self.dtype)
+        return {k: list(avg[k].T) for k in raw}
+
+    def _clean_grad_pris(self, raw, nvars):
+        ndims = self.ndims
+        flat = {}
+        for k, grads_list in raw.items():
+            stacked = np.stack(grads_list, axis=0)       # (nvars,ndims,...)
+            flat[k] = stacked.reshape(nvars*ndims, *stacked.shape[2:]
+                                      ).transpose(1, 2, 0)
```
FreddieWitherden: Try to avoid this style of method chaining as it can get hard to follow.

pyfr/snapshot/region.py
```diff
@@ -0,0 +1,519 @@
...
+        avg = self.cleaner.average(flat, nvars*ndims, self.dtype)
+        return {k: [c for c in avg[k].T.reshape(nvars, ndims, -1)] for k in raw}
+
+
+class VolumeSnapshotRegion(BaseSnapshotRegion):
+    def __init__(self, mesh, cfg, spec, refpts_fn, *, divisor=None, clean=True):
+        super().__init__(mesh, cfg)
+
+        if isinstance(spec, dict):
+            self._eidxs = {et: (None if e is None or isinstance(e, slice)
```
FreddieWitherden: Try to avoid breaking inline ifs.

pyfr/snapshot/region.py
```diff
@@ -0,0 +1,519 @@
...
+
+        avg = self.cleaner.average(flat, nvars*ndims, self.dtype)
+        return {k: [c for c in avg[k].T.reshape(nvars, ndims, -1)] for k in raw}
+
+
+class VolumeSnapshotRegion(BaseSnapshotRegion):
+    def __init__(self, mesh, cfg, spec, refpts_fn, *, divisor=None, clean=True):
+        super().__init__(mesh, cfg)
+
+        if isinstance(spec, dict):
```
FreddieWitherden: Is there a big benefit to supporting dicts and lists for spec?  Is it cleaner to force {et: None, ...}?

pyfr/snapshot/region.py
```diff
@@ -0,0 +1,519 @@
...
+        return arr if eidxs is None else arr[..., eidxs]
+
+    def _ele_grad_data(self, snap, et):
+        if snap.grad_data is None:
+            return None
+        g = snap.grad_data.get(et)
+        if g is None:
+            return None
+        eidxs = self._eidxs[et]
+        return g if eidxs is None else g[..., eidxs]
```
FreddieWitherden: Should be able to avoid the None sentinel; maybe an empty slice?

pyfr/snapshot/region.py
```diff
@@ -0,0 +1,519 @@
...
+    def __init__(self, mesh, cfg, bcname, divisor, refpts_fn=None, *,
+                 clean=True):
+        super().__init__(mesh, cfg)
+
+        self._divisor = divisor
+        self._refpts_fn = (refpts_fn or
+                           (lambda sc, _sh=None: sc.std_ele(self._divisor)))
+        self.clean = clean
+
+        bcnames = [bcname] if isinstance(bcname, str) else list(bcname)
```
FreddieWitherden: Do we want to support both types for bcname or can we just always require a list?

pyfr/snapshot/region.py
```diff
@@ -0,0 +1,519 @@
...
+        self.normals = self._build_normals()
+        self.wall_dist = self._build_wall_dist()
+
+    def _assemble(self, fn):
+        # Map fn(gi, group) over face-groups, concatenating same-itype results
+        acc = defaultdict(list)
+        for gi, g in enumerate(self._groups):
+            acc[g[3]].append(fn(gi, g))
+
+        cat = lambda p: np.ascontiguousarray(np.concatenate(p, axis=-1))
```
FreddieWitherden, line 302: I think concat always give a contig array?

pyfr/snapshot/region.py
```diff
@@ -0,0 +1,519 @@
...
+        def fn(gi, g):
+            etype, _, eidxs, _, mesh_op, _, _ = g
+            xd = interp_ops(mesh_op, self.mesh.spts[etype][:, eidxs])
+            return xd.transpose(2, 0, 1)
+
+        raw = self._assemble(fn)
+        if not self.clean:
+            return raw
+
+        return {it: np.ascontiguousarray(
```
FreddieWitherden, line 342: Check if the contig bit is needed

pyfr/snapshot/region.py
```diff
@@ -0,0 +1,519 @@
...
+        for etype, _, eidxs, itype, *_ in self._groups:
+            for name, arr in snap.aux(etype).items():
+                aux_pieces[itype][name].append(arr[eidxs])
+        for itype, by_name in aux_pieces.items():
+            for name, pieces in by_name.items():
+                sample.field_arrays[itype, name] = np.concatenate(pieces)
+
+    def cell_curved(self, itype):
+        parts = [self.mesh.spts_curved[g[0]][g[2]]
+                 for g in self._groups if g[3] == itype]
```
FreddieWitherden: Probably worth unpacking g here to clean things up.

pyfr/snapshot/soln.py
```diff
@@ -0,0 +1,18 @@
+from pyfr.snapshot.file import FileSnapshot
+
+
+class SolnSnapshot(FileSnapshot):
+
```
FreddieWitherden: Remove newline.

pyfr/writers/vtk/stl.py
```diff
@@ -22,224 +21,216 @@ def _vertex_normals_mwa(fnorms, verts, vids):
         e1n = np.linalg.norm(e1, axis=1)
         e2n = np.linalg.norm(e2, axis=1)
         cos_a = np.clip(np.sum(e1*e2, axis=1) / (e1n*e2n), -1, 1)
-
-        # Accumulate angle-weighted face normal onto each vertex
```
FreddieWitherden: Maybe want to keep - this was fiddly stuff to write!

pyfr/writers/vtk/stl.py
```diff
@@ -22,224 +21,216 @@ def _vertex_normals_mwa(fnorms, verts, vids):
...
 def _spherigon_smooth(flat_pts, bary, tri_verts, tri_norms):
-    # C1 spherigon (Volino & Magnenat-Thalmann 1998); lifts flat
-    # subdivision points onto curved tangent planes defined by
-    # per-vertex normals
+    # C1 spherigon (Volino & Magnenat-Thalmann 1998)
     ntri = flat_pts.shape[1]
     smoothed = np.empty_like(flat_pts)
     tol = 1e-8
 
-    # Iterate over subdivision points (same bary coords for all tris)
```
FreddieWitherden: Keep some of these comments as they make it easier to follow along with the fiddly spherigon stuff.

pyfr/subdiv.py
```diff
@@ -0,0 +1,406 @@
+from collections import namedtuple
```
FreddieWitherden, line 1: Not sure if this should be top level or not.  It is certainly mostly VTK specific stuff.

pyfr/plugins/soln/insitu/renderer.py
```diff
@@ -0,0 +1,355 @@
...
+            for fname in runner.fields(public_only=True):
+                self._register_user_field(sname, fname)
+
+    def _init_gradients(self):
+        self._g_pnames_user = set()
+        for _, comps in self._exprs:
+            for c in comps:
+                self._g_pnames_user.update(re.findall(r'\bgrad_(.+?)_[xyz]\b',
+                                                      c))
+        self._provider_needs_grads = any(r.needs_grads for r
```
FreddieWitherden: Akward line break.

pyfr/plugins/soln/insitu/renderer.py
```diff
@@ -0,0 +1,355 @@
...
+            export_type = self._source_kinds[sname]
+            runner = FieldRunner(names, self.mesh.ndims, self.scfg,
+                                 export_type=export_type)
+            self._field_runners[sname] = runner
+
+            for fname in runner.fields(public_only=True):
+                self._register_user_field(sname, fname)
+
+    def _init_gradients(self):
+        self._g_pnames_user = set()
```
FreddieWitherden: Assign to a local on the same line to clean up the code further down.

pyfr/plugins/cli/ascent.py
```diff
@@ -150,23 +26,22 @@ def render_cli(self, args):
...
-
-            # See if we need to create a new Ascent renderer
-            if not renderer or rcfg != soln.config:
-                renderer = AscentRenderer(adapter, isrestart=True)
-                rcfg = soln.config
-
-            # Perform the rendering
-            renderer.render(adapter)
+            snap = snap_from_file(args.mesh, s)
+            if snap.prefix != 'soln':
```
FreddieWitherden: I think current Ascent does support tavg from the CLI (I remember putting the support in last week or so).

pyfr/plugins/soln/insitu/renderer.py
```diff
@@ -0,0 +1,355 @@
...
+
+        # Load Conduit (subclass may override for host-specific fallbacks)
+        self.conduit = self._load_conduit()
+
+        # Build regions per source
+        self.regions = {}
+        self._source_kinds = {}
+        if self.want_volume:
+            rdata = region_data(self.acfg, cfgsect, self.mesh)
+            spec = rdata if isinstance(rdata, dict) else {et: None
```
FreddieWitherden: Kinda clunky, probably just want to support one way of doing it.

pyfr/plugins/soln/insitu/renderer.py
```diff
@@ -0,0 +1,355 @@
...
+                                                          for et in
+                                                          self.mesh.eidxs}
+            div = self.divisor
+            refpts_fn = lambda sc, sh: sc.std_ele(div)
+            self.regions['volume'] = VolumeSnapshotRegion(
+                self.mesh, scfg, spec, refpts_fn,
+                divisor=div, clean=self.clean)
```
FreddieWitherden: Move ) down.

pyfr/plugins/soln/ascent.py
```diff
@@ -835,23 +166,26 @@ def _render_options(self, path, opts):
...
+        self.publish(fields)
 
         self.lib.ascent_publish(self.ascent_ptr, self.mesh_n)
         self.lib.ascent_execute(self.ascent_ptr, self.actions)
 
         comm.barrier()
 
+    def finalise(self):
+        if lib := getattr(self, 'lib', None):
+            self.lib = None
```
FreddieWitherden: Not sure this is needed, should never be called more than once (and if it is we want a crash).

pyfr/plugins/soln/ascent.py
```diff
@@ -861,8 +195,13 @@ class AscentPlugin(BaseSolnPlugin):
...
+                                        intg.isrestart)
 
     def __call__(self, intg):
-        self._renderer.render(_IntegratorAdapter(intg, intg.cfg, self.cfgsect))
+        self._renderer.render(IntgSnapshot(intg))
+
+    def finalise(self, intg):
+        if r := getattr(self, '_renderer', None):
+            r.finalise()
+            del self._renderer
```
FreddieWitherden: Is the explicit del needed?

pyfr/plugins/soln/vtu_writer.py
```diff
@@ -0,0 +1,91 @@
...
+from pyfr.writers.vtk.spanwise import VTKSpanwiseWriter
+from pyfr.writers.vtk.stl import VTKSTLWriter
+from pyfr.writers.vtk.volume import VTKVolumeWriter
+
+
+class VTUWriterPlugin(BaseSolnPlugin):
+    # In-situ VTU writer.  One plugin instance = one output stream (volume
+    # or one boundary).  Use the plugin suffix mechanism for multiple
+    # streams: [soln-plugin-vtuwriter-volume], [soln-plugin-vtuwriter-walls].
+    name = 'vtuwriter'
```
FreddieWitherden: Do we want a hyphen?

pyfr/plugins/soln/vtu_writer.py
```diff
@@ -0,0 +1,91 @@
...
+        # Writer options
+        order = cfg.getint(s, 'order', 0) or None
+        divisor = cfg.getint(s, 'divisor', 0) or None
+        prec = cfg.get(s, 'precision', 'single')
+        clean = cfg.getbool(s, 'clean', True)
+
+        add_fields = [n.strip() for n in cfg.get(s, 'add-fields', '').split(',')
+                      if n.strip()]
+        remove_fields = [n.strip() for n in cfg.get(s, 'remove-fields',
+                                                    '').split(',') if n.strip()]
```
FreddieWitherden: Kinda wonky line breaking.

pyfr/plugins/soln/vtu.py
```diff
@@ -0,0 +1,91 @@
...
+        self._basename = cfg.get(s, 'basename')
+
+        # Output cadence
+        self.dt_out = cfg.getfloat(s, 'dt-out')
+        self.tout_last = intg.tcurr
+        if not intg.isrestart:
+            self.tout_last -= self.dt_out
+        intg.call_plugin_dt(intg.tcurr, self.dt_out)
+
+        # Writer options
```
FreddieWitherden, line 33: Can this code be factored out and shared at all with the main export stuff.  They seem quite similar except for one getting its initial values from an ini and the other the command line.

pyfr/plugins/__init__.py
```diff
@@ -1,10 +1,8 @@
 from pyfr.plugins.base import BasePlugin, BaseCLIPlugin
-from pyfr.plugins.soln.base import BaseSolnPlugin
-from pyfr.plugins.solver.base import BaseSolverPlugin
-from pyfr.plugins import cli, postproc, soln, solver
 from pyfr.util import subclass_where
 
 
 def get_plugin(prefix, name, *args, **kwargs):
+    from pyfr.plugins import soln, solver
```
FreddieWitherden, line 6: Any nicer way of structuring this to avoid the circular import issue?

FreddieWitherden:
I think we should try and get some of the VTK specific stuff out to pyfr.vtk submodule and then see where that leaves snapshots (which bits are technically independent of VTK compared to which bits are tightly connected).

pyfr/plugins/cli/sampler.py
```diff
@@ -209,114 +191,102 @@ def remove_cmd(self, args):
...
+            arr = sample.field_arrays.get(('points', fname))
+            if arr is None:
+                continue
+            if len(varnames) == 1:
+                fields[varnames[0]] = arr if arr.ndim == 1 else arr.ravel()
+            else:
+                for i, vn in enumerate(varnames):
+                    fields[vn] = arr[:, i] if arr.ndim == 2 else arr[i]
+
+        if extra := remove_fields - set(fields):
```
FreddieWitherden: Exception from pop should be good enough here.

pyfr/plugins/soln/insitu/conduit.py
```diff
@@ -0,0 +1,120 @@
...
+        npoints, ncomps = arr2d.shape
+        itemsize = arr2d.itemsize
+        stride = ncomps * itemsize
+        fn = getattr(self.lib,
+                     f'conduit_node_set_path_external_{arr2d.dtype}_ptr_detailed')
+        for i, l in enumerate(labels):
+            fn(self, f'{key}/{l}'.encode(), arr2d.ctypes.data,
+               npoints, i * itemsize, stride, itemsize, 0)
+
+    def empty_object(self, key):
```
FreddieWitherden, line 91: What is this required for?

pyfr/plugins/soln/insitu/renderer.py
```diff
@@ -0,0 +1,355 @@
...
+        else:
+            buf.append(c)
+
+    parts.append(''.join(buf).strip())
+
+    return parts
+
+
+class InSituRenderer:
+
```
FreddieWitherden: Can remove newline.

pyfr/plugins/soln/insitu/renderer.py
```diff
@@ -0,0 +1,355 @@
...
+        self._fields_read = set()
+        self._init_fields()
+        self._init_field_runners()
+
+        # Host-specific publishing config (Ascent scenes/pipelines, etc.)
+        self._init_host_publish()
+
+        if not self._fields_read.issubset(self._fields_write):
+            missing = self._fields_read - self._fields_write
+            raise self.error_cls(f'Fields used but not defined: {missing}')
```
FreddieWitherden: Probably a ValueError or KeyError

pyfr/plugins/soln/insitu/renderer.py
```diff
@@ -0,0 +1,355 @@
...
+
+        # Build the Conduit blueprint mesh for the regions (geometry-only)
+        self.dinfo = {}
+        for i, (sname, etype) in enumerate(self.domains):
+            dom = self._domain_path(sname, doff + i)
+            self._build_blueprint(dom, sname, doff + i, etype)
+            self.dinfo[sname, etype] = dom
+
+        # Host-specific instance init (open the library, etc.)
+        self._init_host()
```
FreddieWitherden: If this is at the end can it not just be done by the subclass after they call super?

pyfr/plugins/soln/catalyst.py
```diff
@@ -0,0 +1,342 @@
...
+
+        fields = self._evaluate_exprs(snap)
+        self.publish(fields)
+
+        self.lib.catalyst_execute(self.mesh_n)
+        comm.barrier()
+
+    def finalise(self):
+        if lib := getattr(self, 'lib', None):
+            self.lib = None
```
FreddieWitherden: Should not be needed.

pyfr/plugins/soln/catalyst.py
```diff
@@ -0,0 +1,342 @@
...
+        super().__init__(intg, cfgsect, suffix)
+
+        self._renderer = CatalystRenderer(intg.system.mesh, intg.cfg,
+                                          cfgsect, intg.isrestart)
+        self._bootstrap_done = False
+
+    def __call__(self, intg):
+        snap = IntgSnapshot(intg)
+
+        if not self._bootstrap_done:
```
FreddieWitherden: Can this never happen at the same time as we need to write something? (wondering about the early return).

pyfr/snapshot/sample.py
```diff
@@ -0,0 +1,44 @@
...
+        elif info.source == 'data':
+            return self.snap.data_at(etype, info)
+        elif info.source == 'grad_data':
+            return self.snap.grad_data_at(etype, info)
+        elif info.source in ('aux', 'provider'):
+            return None
+        else:
+            raise ValueError(f'Unknown field source: {info.source!r}')
+
+    def make_ploc_writable(self):
```
FreddieWitherden: Does it hurt us to just copy this thing all the time?  If it is only a percentage point or two lets just always do it and never need to think about it.

pyfr/snapshot/soln.py
```diff
@@ -0,0 +1,18 @@
...
+    @property
+    def pris_names(self):
+        return list(self.elementscls.privars(self.ndims, self.config))
+
+    def to_pris(self, interp_data, cfg):
+        return list(self.elementscls.con_to_pri(interp_data, cfg))
```
FreddieWitherden: Is this list call needed?

pyfr/plugins/fields/runner.py
```diff
@@ -0,0 +1,110 @@
...
+                    continue
+                sample.field_arrays[et, fname] = arr
+
+        # Register provider outputs in sample.fields so consumers iterate one
+        # unified registry (alongside snap's data + primitives + grads + aux).
+        for p in self.plugins:
+            for fname, varnames in p.fields.items():
+                if public_only and fname.startswith('_'):
+                    continue
+                if fname in sample.fields:
```
FreddieWitherden: or with above

pyfr/plugins/fields/runner.py
```diff
@@ -0,0 +1,110 @@
...
+                    continue
+                sample.fields[fname] = FieldInfo(
+                    name=fname, kind='point', ncomps=len(varnames),
```
FreddieWitherden: Is ncomps needed if we have varnames?

TobyFlynn:
> @pv101 @TobyFlynn Thoughts on interpolate then compute vs compute then interpolate?

We would want to interpolate and then compute, i.e. span average and then compute on that.

kaschau:
> I think we should try and get some of the VTK specific stuff out to pyfr.vtk submodule and then see where that leaves snapshots (which bits are technically independent of VTK compared to which bits are tightly connected).

The only parts in snapshots are just the vis_* stuff which isnt really vtk specific, just specific to vtk usage at this point. Did you have something else in mind?

FreddieWitherden:
Naming wise a few random thoughts:
 - Region is already used by the PyFR region code, this is maybe more of a Mesh but not quite?
 - Fields are already used in the plugins to refer to backend evaluation of expressions.
 - Snapshot is maybe not ideal but I am not sure of anything better.  Maybe Solution?

pyfr/plugins/fields/base.py
```diff
@@ -0,0 +1,28 @@
...
+    needs_grads = False
+    kind = 'producer'
+    deps = []
+    fields = {}
+
+    def __init__(self, ndims, cfg, export_type=None):
+        self.ndims = ndims
+        self.cfg = cfg
+
+        if export_type is not None:
```
FreddieWitherden: and

pyfr/plugins/fields/runner.py
```diff
@@ -0,0 +1,110 @@
...
+
+class SampleView:
+    # A per-key (etype/itype/'points') lens onto a SnapshotSample, handed to a
+    # field provider.  Provides simple attribute access to pris/grad_pris/
+    # normals/wall_dist/state — backed by sample (mutable per step) and region
+    # (static geometry).  Lives here, with the runner that constructs it,
+    # rather than in snapshot.py (avoids a snapshot↔runner import cycle).
+    def __init__(self, sample, etype):
+        self.sample = sample
+        self.region = sample.region
```
FreddieWitherden: Sometimes a smell to be pulling members of another member out.

pyfr/plugins/fields/runner.py
```diff
@@ -0,0 +1,110 @@
...
+        self.snap = sample.snap
+        self.cfg = sample.region.config
+        self.ndims = sample.region.ndims
+        self.etype = etype
+        # Per-view write target; FieldRunner collects into sample.fields.
+        self.fields = {}
+
+    @property
+    def ploc(self):
```
FreddieWitherden: Also a minor smell, seems to suggest this could be a subclass.

pyfr/plugins/fields/runner.py
```diff
@@ -0,0 +1,110 @@
...
+            for fname, varnames in p.fields.items():
+                if public_only and fname.startswith('_'):
+                    continue
+
+                out[fname] = varnames
+
+        return out
+
+    def run_on_sample(self, sample, public_only=False):
+        # One pair of loops: per-key (etype/itype/'points') × per-provider.
```
FreddieWitherden: Can cut down on these comments.

pyfr/plugins/fields/runner.py
```diff
@@ -0,0 +1,110 @@
...
+        for p in self.plugins:
+            for fname, varnames in p.fields.items():
+                if public_only and fname.startswith('_'):
+                    continue
+                if fname in sample.fields:
+                    continue
+                sample.fields[fname] = FieldInfo(
+                    name=fname, kind='point', ncomps=len(varnames),
+                    dtype=np.dtype(sample.snap.dtype), source='provider',
+                    components=tuple(varnames))
```
FreddieWitherden: Check if casting to tuple is needed.

pyfr/plugins/soln/insitu/conduit.py
```diff
@@ -0,0 +1,120 @@
...
+                fn(self, key, value.ctypes.data, value.size)
+            case list():
+                value = np.array(value, dtype=float)
+                self.lib.conduit_node_set_path_float64_ptr(self, key,
+                                                           value.ctypes.data,
+                                                           value.size)
+            case _:
+                raise ValueError('ConduitNode: __setitem__ type not supported')
+
+    def set_aos(self, key, labels, arr2d):
```
FreddieWitherden: I'd try and get this migrated into the above setitem even if it means some refactoring.  Ideally setitem an arrray, sniff the stride (NumPy knows them) then call the relevant conduit method.  This is then called in a loop over fields by the caller.

pyfr/plugins/soln/insitu/renderer.py
```diff
@@ -0,0 +1,364 @@
...
+    return parts
+
+
+class InSituRenderer:
+
+    # Conduit blueprint element name mapping
+    bp_emap = {'hex': 'hex', 'pri': 'wedge', 'pyr': 'pyramid', 'quad': 'quad',
+               'tet': 'tet', 'tri': 'tri'}
+
+    error_cls = InSituError
```
FreddieWitherden: Try to avoid this.

pyfr/plugins/soln/insitu/renderer.py
```diff
@@ -0,0 +1,364 @@
...
+
+    def _write_field_meta(self, mesh_n, dom, sname):
+        for field in self._source_fields[sname]:
+            fname = self._field_name(sname, field)
+            mesh_n[f'{dom}/fields/{fname}/association'] = 'vertex'
+            mesh_n[f'{dom}/fields/{fname}/volume_dependent'] = 'false'
+            mesh_n[f'{dom}/fields/{fname}/topology'] = sname
+
+    def _flatten_coords(self, ploc, clean):
+        if clean:
```
FreddieWitherden: Feels wrong.

pyfr/plugins/soln/insitu/renderer.py
```diff
@@ -0,0 +1,364 @@
...
+        for k in self.acfg.items(self.cfgsect, prefix='add-field-'):
+            name = k.removeprefix('add-field-')
+            for s in self.acfg.get(self.cfgsect, k).split(','):
+                sname = s.strip()
+                if sname not in self.regions:
+                    raise self.error_cls(f'Field provider {name!r}: unknown '
+                                         f'source {sname!r}')
+                groups[sname].append(name)
+
+        self._field_runners = {}
```
FreddieWitherden: Can also assign to a local as `self._field_runners = runners = {}` then can avoid runner definition later.

pyfr/plugins/soln/insitu/renderer.py
```diff
@@ -0,0 +1,364 @@
...
+        self._provider_needs_grads = any(r.needs_grads for r
+                                         in self._field_runners.values())
+        self._gradpinfo = None
+
+    def _resolve_gradpinfo(self, snap):
+        privars = snap.pris_names
+        g_pnames = set(self._g_pnames_user)
+        if self._provider_needs_grads:
+            g_pnames.update(privars)
+        if g_pnames and not snap.has_grads:
```
FreddieWitherden: Should we do this before the update?

pyfr/plugins/soln/insitu/renderer.py
```diff
@@ -0,0 +1,364 @@
...
+            raise self.error_cls('Gradients required but not available')
+        self._gradpinfo = [(p, privars.index(p)) for p in g_pnames]
+
+    def _evaluate_exprs(self, snap):
+        if self._gradpinfo is None:
```
FreddieWitherden: Feels wrong.

pyfr/solvers/base/elements.py
```diff
@@ -379,6 +379,35 @@ def pnorm_at(self, name, norm):
...
+            return x.reshape(len(pts), self.neles, self.ndims)
+
+        return np.linalg.norm(ploc(upts) - ploc(upts + t[:, None]*rn), axis=2)
+
+    @memoize
+    def min_upt_face_dist_approx(self, fidx):
+        # Min over the off-face solution points (on-face upts sit at distance 0).
+        shape = self.basis
+        _, proj, norm = shape.faces[fidx]
+        t = (proj(*([0]*(self.ndims - 1))) - shape.upts) @ (
```
FreddieWitherden: Could use a rewrap.

pyfr/writers/vtk/base.py
```diff
@@ -62,72 +54,63 @@ def __init__(self, meshf, pname=None, *, prec='single', order=None,
...
+            raise ValueError(
+                f'Field providers are only supported for conservative-form '
+                f'solution files (snap.prefix = {snap.prefix!r})')
 
-        # Convert any solution gradients to primitive variables
-        if self._gradients:
-            diff_cons = soln[nvars:].reshape(nvars, -1, *soln.shape[1:])
-            diff_pri = ecls.diff_con_to_pri(soln[:nvars], diff_cons, self.cfg)
+        # MPI collective must be in lockstep across ranks
+        snap.compute_grads()
```
FreddieWitherden: Seems a bit awkward as a consumer.

pyfr/writers/vtk/base.py
```diff
@@ -62,72 +54,63 @@ def __init__(self, meshf, pname=None, *, prec='single', order=None,
...
 
-        return fields
+    def _post_sample(self, snap):
+        pass
 
-    def _post_proc_fields_scal(self, vsoln):
-        return [vsoln[self._soln_fields.index(k)] for k in self._vtk_vars]
+    def _emit_fields(self, kind):
+        for name, info in self._sample.fields.items():
+            if info.kind != kind:
```
FreddieWitherden: Can merge with if below.

pyfr/writers/vtk/base.py
```diff
@@ -136,171 +119,88 @@ def _nsvpts(self, etype):
...
-        for fname in self._extra_field_lists()[1]:
-            arr = pointf[fname].reshape(nsvpts, neles, -1)
-            fields.append((arr, self._extra_fields[fname].dtype))
+        for name, info in self._emit_fields('point'):
+            arr = sample.field_array(etype, info)
+            if arr is None:
+                continue
+
+            if info.source in ('primitive', 'gradient'):
+                if not region.clean:
```
FreddieWitherden: Seems to suggest a leaky abstraction in clean vs not clean pipeline.

pyfr/writers/vtk/spanwise.py
```diff
@@ -429,140 +8,41 @@ class VTKSpanwiseWriter(BaseVTKWriter):
...
-        shapes = {et: self._get_shape(et, self.cfg)
-                  for et in self.reader.mesh.etypes}
+    def _init_einfo(self):
+        self.einfo = []
 
-        kwargs = dict(mesh=self.mesh, cfg=self.cfg, shapes=shapes,
-                      axis=self.axis, nfields=len(self._soln_fields),
-                      vis_pts_2d=self._vis_pts_2d())
+    def _build_region(self):
+        return SpanwiseSnapshotRegion(self.mesh, self.cfg,
```
FreddieWitherden: Can save some lines as:
```
X(
    Foo, Bar, Baz,
)
```

pyfr/writers/vtk/spanwise.py
```diff
@@ -429,140 +8,41 @@ class VTKSpanwiseWriter(BaseVTKWriter):
...
+        if not self.einfo:
+            self.einfo = [(it, self._region._ncells[it])
+                          for it in self._region.etypes]
 
-        # Otherwise, fall back to sampling
-        nstations = self.nstations or 4*(self.order + 1)
-        return SampledSpanwise(**kwargs, eles=self.reader.eles,
-                               nstations=nstations, **{kind: name})
+    def _emit_fields(self, kind):
+        if kind == 'cell':
```
FreddieWitherden, line 47: Probably worth a comment.

pyfr/writers/vtk/stl.py
```diff
@@ -22,224 +21,216 @@ def _vertex_normals_mwa(fnorms, verts, vids):
...
+                # field_array returns (npts, ncomp); fan out via pinv.
+                arr = arr[pinv].reshape(n_subdiv, ntri, arr.shape[-1])
+                ftype = self.dtype
+            else:
+                # Aux / provider: (npts,) or (ncomp, npts) — expand via pinv.
+                ftype = info.dtype
+                if arr.ndim == 1:
+                    arr = arr[pinv].reshape(n_subdiv, ntri, 1)
+                else:
+                    arr = arr[:, pinv].reshape(arr.shape[0], n_subdiv,
```
FreddieWitherden: Awkward break.

pyfr/writers/vtk/stl.py
```diff
@@ -22,224 +21,216 @@ def _vertex_normals_mwa(fnorms, verts, vids):
...
 
 
 class VTKSTLWriter(BaseVTKWriter):
+    # STL surface export.  The STL is just a set of physical points (welded
+    # subdivision vertices) plus user-supplied triangle connectivity — exactly
+    # what snap.at_points(ppts) (PointsSnapshotRegion) consumes.  Sampling
+    # runs once at the welded vertices; emission expands per-triangle via the
+    # weld inverse `pinv` so each triangle gets its own VTU cell.
+    #
+    # STL keeps its own _write_data because per-triangle emission needs the
```
FreddieWitherden: See if there is a way we can clean this up, clean to grid might handle it naturally.

pyfr/snapshot/file.py
```diff
@@ -0,0 +1,41 @@
...
+
+class FileSnapshot(BaseSnapshot):
+    def __init__(self, mesh, soln):
+        from pyfr.solvers.base import BaseSystem
+
+        self._soln = soln
+
+        cfg = soln.config
+        stats = soln.stats
+        prec = cfg.get('backend', 'precision', 'single')
```
FreddieWitherden: I think this can be different for tavg and maybe regular soln (writer/tavg might be able to cast).

pyfr/plugins/soln/catalyst.py
kaschau: Yes I will get that cleaned up, but also can move catalyst out of this PR and into a separate one. Though its working quite well for me so far.

pyfr/plugins/soln/catalyst.py
kaschau: yes we can but we need to keep PyDLL bc of paraview needs GIL held.

pyfr/plugins/soln/catalyst.py
kaschau: Yes, and they STRONGLY prefer the built-in conduit they ship with, not sure why other than version mismatching

pyfr/plugins/soln/vtu_writer.py
kaschau: youre right, moved upstream to the writer for this, stl, and spanwise with an MPI collective check.

pyfr/snapshot/base.py
kaschau: not for empy, but switching to math.prod will

pyfr/snapshot/fieldinfo.py
kaschau: actually we dont even need the field, but i'll address rename of stuff later

pyfr/snapshot/intg.py
kaschau: sys.eles_vect_upts is not None?

FreddieWitherden: That seems reasonable.

pyfr/snapshot/intg.py
kaschau: that is certainly the intent, they are meant to by used once and discarded. I can collect the trivial things at __init__ and then create the grad methods only if we have gradients to begin with. those will still have to refer to intg, but the only alternative is to compute them eagerly. but this will remove the self._intg entirely so theres less temptation for future things to grab it.

pyfr/snapshot/region.py
kaschau: A transpose makes them non-contiguous strides; concat on axis=-1 doesn't restore contiguity:

_Mirrors _build_ploc: (ndims, npts, neles) = interp_ops result .transpose(2, 0, 1)_

x_t = np.arange(60).reshape(4, 5, 3).transpose(2, 0, 1)   # (3, 4, 5), non-contig
y_t = (np.arange(60, 120).reshape(4, 5, 3)).transpose(2, 0, 1)
xy = np.concatenate([x_t, y_t], axis=-1)


-> xy.shape == (3, 4, 10), xy.flags['C_CONTIGUOUS'] == False

kaschau: Only know bc something choked on it downstream at some point

pyfr/snapshot/region.py
kaschau: this ones redundant

pyfr/writers/vtk/stl.py
kaschau: yeah not sure why that got removed

pyfr/__main__.py
```diff
@@ -495,20 +507,38 @@ def process_export(args):
...
+            mesh, soln = reader.load_subset_mesh_soln(solnf)
+            snap = snap_from_loaded(mesh, soln)
+            if writer is None:
+                if writer_cls.type == 'stl':
+                    writer = writer_cls(snap.mesh, snap.config,
+                                        extra['stlrgns'], **kwargs)
+                else:
+                    writer = writer_cls(snap.mesh, snap.config,
+                                        **extra, **kwargs)
+            writer.emit(snap, outf)
```
FreddieWitherden, line 542: Check this still works with solutions having different regions and/or polynomial orders which we do support at the moment.

pyfr/subdiv.py
kaschau: kind of, indirectly. its really just used by vtk, but also clean to grid which is now primary to snapshot, not vtk. but maybe it can be further split and reorganized

pyfr/plugins/cli/ascent.py
kaschau: yeah works now too that was leftover from testing

pyfr/plugins/soln/vtu_writer.py
kaschau: no other plugin has one but ill defer to your preference

FreddieWitherden: Maybe just vtu?

kaschau: yeah that works

pyfr/plugins/soln/insitu/conduit.py
kaschau: It's the fix for empty MPI ranks in a source. Catalyst multimesh channels need a catalyst/channels/<sname>/data node to exist on every rank

pyfr/plugins/soln/catalyst.py
kaschau: an execute call is part of the bootstrap, and then everything comes to life on the catalyst side and it will write if it needs to (i.e. the zeroth will still write even if it happens within the bootstrap call)

pyfr/plugins/soln/insitu/renderer.py
kaschau: ok to decide based on ploc.ndims?

pyfr/writers/vtk/base.py
kaschau: yeah it more comes from the sample remaining neutral and all outputters being responsible for organizing data. could give the sample layout-type specific methods the outputters can request. or even writer specific but that seems less ideal

pyfr/plugins/soln/insitu/renderer.py
FreddieWitherden: I think more structural.  If we need to `if clean` then it suggests something that should be resolved upstream (as in the cleaner is returning a different shape or that some more pre-processing is needed).

kaschau:
> Naming wise a few random thoughts:
> 
> * Region is already used by the PyFR region code, this is maybe more of a Mesh but not quite?
> * Fields are already used in the plugins to refer to backend evaluation of expressions.
> * Snapshot is maybe not ideal but I am not sure of anything better.  Maybe Solution?

agree on the reuse, ill mull it over and try to find something better for all

FreddieWitherden:
If we have any `if clean` blocks or equivalent statements see if we can fix the cleaner so that it returns fields which are more compatible with regular fields.  Minor code smell if our pipeline needs to treat them differently (and probably my fault when I wrote the cleaning logic).

PR #569, opened 2026-06-26
Fixing partitioner issue on large meshes

TobyFlynn:
This fixes an integer overflow issue when trying to partition a large mesh into many partitions

### DIFF
--- a/pyfr/partitioners/baseline.py
+++ b/pyfr/partitioners/baseline.py
@@ -85,8 +85,7 @@ class BaselinePartitioner(BasePartitioner):
         self._rng = np.random.default_rng(self.opts['seed'])
 
     def _partition_graph(self, graph, partwts):
-        nv = len(graph.vtab) - 1
-        ne = len(graph.etab)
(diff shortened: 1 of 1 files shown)


pyfr/partitioners/baseline.py
```diff
@@ -86,7 +86,8 @@ def __init__(self, *args, **kwargs):
 
     def _partition_graph(self, graph, partwts):
         nv = len(graph.vtab) - 1
-        self._idxdtype = idt = np.int32 if nv < 2**31 else np.int64
+        ne = len(graph.etab)
```
FreddieWitherden: Can put nv, ne = .. one one line

FreddieWitherden:
Fix looks good.  Out of curiosity how big is this mesh?

TobyFlynn:
> Fix looks good. Out of curiosity how big is this mesh?

I think the original one that I saw this bug on was ~90 million elements but I was testing the fix on ~120 million elements

PR #570, opened 2026-07-01
Allow dense HIP GiMMiK autotune candidates

tomjen12:
## Summary

Remove the HIP GiMMiK density/unique-value suitability check so dense operators can be passed through to GiMMiK autotuning.

This is intended to stack on top of #567 and support PyFR/GiMMiK#20, which adds a HIP MFMA dense GiMMiK kernel candidate.

## Motivation

The previous check rejected dense tet operators before GiMMiK could consider them:

```python
if nuq > 128 and nnz / arr.size > 0.15:
    raise NotSuitableError('Matrix inappropriate GiMMiK')

### DIFF
--- a/pyfr/backends/hip/gimmik.py
+++ b/pyfr/backends/hip/gimmik.py
@@ -67,11 +67,7 @@ class HIPGiMMiKKernels(HIPKernelProvider):
                     src, meta = kgen.send(kdata)
                     kern = self._build_kernel(kname, src, 'iPiPi')
 
-                    grid_y = meta.get('grid_y', 1)
-                    ncolsv = (
-                        meta.get('width', 1) *
-                        meta.get('ncols', meta['block'][0])
-                    )
+                    grid_y, ncolsv = meta['grid_y'], meta['ncolsv']
                     grid = (-(-n // ncolsv), grid_y, 1)
                     params = kern.make_params(grid, meta['block'])
                     params.set_args(n, b, ldb, out, ldc)
--- a/pyfr/backends/hip/gimmik.py
+++ b/pyfr/backends/hip/gimmik.py
@@ -34,10 +34,6 @@ class HIPGiMMiKKernels(HIPKernelProvider):
         arr = a.get()
         nnz, nuq = np.count_nonzero(arr), len(np.unique(np.abs(arr)))
 
-        # Check that A is suitable
-        if nuq > 128 and nnz / arr.size > 0.15:
-            raise NotSuitableError('Matrix inappropriate GiMMiK')
-
         # Dimensions
         n = b.ncol
         ldb, ldc = b.leaddim, out.leaddim
--- a/pyfr/backends/hip/gimmik.py
+++ b/pyfr/backends/hip/gimmik.py
@@ -4,8 +4,7 @@ from gimmik import HIPMatMul
 import numpy as np
 
 from pyfr.backends.base import NotSuitableError
-from pyfr.backends.hip.provider import (HIPKernel, HIPKernelProvider,
-                                        get_grid_for_block)
+from pyfr.backends.hip.provider import HIPKernel, HIPKernelProvider
 
 
 class HIPGiMMiKKernels(HIPKernelProvider):
@@ -49,7 +48,7 @@ class HIPGiMMiKKernels(HIPKernelProvider):
 
         # Check the kernel cache
         try:
-            kern, block, width, dt = self._mul_kerns[ckey]
+            kern, block, width, grid_y, ncols, dt = self._mul_kerns[ckey]
         except KeyError:
             ifac = self.backend.autotune_ifac
             kname = f'gimmik_mm_{arr.shape[0]}x{arr.shape[1]}'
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: pyfr/backends/hip/gimmik.py [hunk 1]
```diff
@@ -85,7 +84,7 @@ def mul(self, a, b, out, alpha=1.0, beta=0.0):
                     )
 
                     if best_kern is None or dt < ifac*best_kern[-1]:
-                        best_kern = kern, meta['block'], dt
+                        best_kern = kern, meta['block'], width, grid_y, ncols, dt
```
FreddieWitherden: I think we can multiply width*ncols, store this, and keep the code a bit simpler (no need to carry around width and ncols, one number should do it).

FreddieWitherden:
This is good to merge. Just holding off until the next GiMMiK release as otherwise with current GiMMiK we'll have issues with dense kernels being handled by sparse GiMMiK which isn't great for compile times.

pyfr/backends/hip/gimmik.py
```diff
@@ -74,9 +67,12 @@ def mul(self, a, b, out, alpha=1.0, beta=0.0):
                     src, meta = kgen.send(kdata)
                     kern = self._build_kernel(kname, src, 'iPiPi')
 
-                    width = meta.get('width', 1)
-                    vn = -(-n // width)
-                    grid = get_grid_for_block(meta['block'], vn)
+                    grid_y = meta.get('grid_y', 1)
```
FreddieWitherden: We can consider having GiMMiK just return a number here pre-computed (so instead of width and ncols it can put a pre-computed factor in a new meta key.  This might be easier for consumers.  On the PyFR side do not need to preserve backwards compatibility with old GiMMiK (this might let you avoid some of the `.get` calls since you know certain metadata will be present).

tomjen12:
## Additional MI300X synthetic-shape validation

I ran an additional 1000-case synthetic-shape benchmark to check MFMA GiMMiK generalization with the new HIP MFMA dense candidates from [PyFR/GiMMiK#20](https://github.com/PyFR/GiMMiK/pull/20), and to measure the new optional PyFR-side parallel compile path (`backend-hip.gimmik-parallel-compile = true`). The first 20 cases are PyFR internal `p2`-`p5` tet shapes; the rest are synthetic dense shapes with broader `M/K/N` coverage. Full per-case results are attached as CSV.

| Metric | Value |
|---|---:|
| Cases | 1000 |
| Worst-case MFMA / rocBLAS BW | 95% |
| Geomean MFMA / rocBLAS BW | 102% |
| rocBLAS autotuning total | 3940.1 s |
| MFMA sequential autotuning total | 1137.0 s |
| MFMA parallel autotuning total | 416.2 s |
| rocBLAS -> MFMA sequential autotune speedup | 3.47x |
| rocBLAS -> MFMA parallel autotune speedup | 9.47x |

MFMA GiMMiK stays competitive with rocBLAS across the selected shapes: the worst case is 95% of rocBLAS bandwidth and the geomean is 102%.

Autotuning time also improves significantly. Moving from rocBLAS to MFMA sequential tuning is 3.47x faster. With `backend-hip.gimmik-parallel-compile = true`, MFMA tuning time drops further to 416.2s, making MFMA with parallel compile 9.47x faster to tune than rocBLAS on these cases.

The parallel compile path is opt-in and does not change the runtime kernel API or selected kernel execution path; it parallelizes candidate source generation and HIPRTC compilation/cache warmup.

Attached CSV files:
- `synthetic_1000_summary.csv`: aggregate bandwidth and autotuning summary.
- `synthetic_1000_details.csv`: per-case results for the selected 1000 shapes.

[synthetic_1000_summary.csv](https://github.com/user-attachments/files/30080955/synthetic_1000_summary.csv)
[synthetic_1000_details.csv](https://github.com/user-attachments/files/30080960/synthetic_1000_details.csv)

FreddieWitherden:
Parallel compilation is something we need to be a little careful about due to its interaction with PyFR's own caching mechanism so we might not want to mainline that.  Hopefully, the MFMA kernels should compile relatively quickly so we only need to worry about the sparse kernels.  Here, my preference is tuning on the GiMMiK side (so where we know a certain approach never works for a certain matrix size/sparsity class we can just not emit the kernel).

PyFR has pretty good caching of kernels so users should hopefully only pay the price of compilation once.

tomjen12:
Update: I rolled this PR back and moved the optional parallel compilation work to #576.

#570 now focuses on enabling dense HIP GiMMiK autotune candidates and the required launch metadata changes. #576 contains the separate opt-in PyFR-side parallel compile path.

pyfr/backends/hip/gimmik.py
FreddieWitherden: Do you have a strong opinion here?

tomjen12: No strong opinion. I agree that having GiMMiK return the pre-computed effective output-column count is cleaner. I will update GiMMiK to provide that metadata and simplify the PyFR side accordingly.

tomjen12:
Updated. GiMMiK now provides the pre-computed output-column count, and PyFR consumes it directly.

FreddieWitherden:
Should no longer be needed after the most recent round of changes.

PR #571, opened 2026-07-09
[Proposal] Selectable EOS with derived quantity unification

kaschau:
# Selectable fluid physics and unified derived-field export

A proposal to make fluid physics more extensible, and unify mako/numpy specification of quantities.

## 1. Fluid layer — `pyfr/fluids/` (new)

Physics (EOS + transport) becomes a first-class object selected by ini key
and consumed by every part of the code:

- A fluid registers named **quantities** (`rho`, `v`, `p`, `a`, `mu`, ...)
  as target-neutral expressions with declared dependencies. `fluid.decl`
  emits the dependency closure as C for kernels; `fluid.eval` evaluates the
  same graph in numpy for export/sampling. Quantities needing iteration or
  clamping carry explicit device/host escape hatches.
- **Solver kernels decode state through the graph** — flux kernels, Riemann
  solvers, and boundary conditions call `fluid.decl` instead of hardcoding
  thermodynamics. One decode per state per kernel; inner macros receive
  names as parameters. The pressure the solver integrates with and the
  pressure the exporter writes are the same registered expression.
- **Transport models** (`transport = constant | sutherland`) register
  `mu`/`kappa` onto the fluid; swapping the law changes every consumer with
  no kernel or table edits. `viscosity-correction` is deprecated with a
  warning and mapped onto the new key.
- Quantities may declare **differentials**. `fluid.diff_pri` applies the
  con-to-pri Jacobian to any directional seed — spatial gradients, time
  derivatives — replacing the hand-coded `diff_con_to_pri` /
  `grad_con_to_pri` pair and their per-plugin call sites (VTU export, CLI
  sampler, ascent, fwh). The legacy methods and elements shims are removed.
- Host physics duplicated across consumers is unified onto the fluid:
  fluidforce's pressure/Sutherland copies, fwh, the samplers, the point
  trigger, and both ascent adapters construct the fluid from config
  instead of reaching into `elementscls` (which remains for shapes/basis
  only). Net: three Sutherland implementations become one quantity; three
  gradient con-to-pri implementations become one set of declarations.
- Compatibility is declared, not assumed — and declared *in the kernel
  that owns the assumption*: each Riemann solver template opens with
  `${pyfr.eos_check('rusanov', fluid, 'cpg')}` (they all carry
  constant-gamma math), which raises at kernel generation with the
  pairing named wherever that solver is actually included. BC classes
  carry `eos_compat` for BC-specific math. Requesting a quantity the
  configured physics does not provide likewise fails with the dependency
  named (e.g. `yplus` on an inviscid fluid: "Fluid 'cpg' does not
  provide quantity 'mu'").

## 2. Field registry — `pyfr/fields/` (new)

A physics-agnostic derived-field grammar shared by every output pipeline:

- Shipped table (`fields.txt`): `mach`, `cp`, `isen-mach`,
  `vorticity`, and the wall-surface fields `tau_wall`, `cf`, and
  `yplus` — each with explicit `deps`/`consts`, 2D/3D variants, `|`
  components, and `_`-prefixed internals. Entry-in-entry references are
  flattened with cycle detection.
- The same expression works everywhere: `avg-mach = mach` accumulates on
  device in tavg, `-f mach` evaluates on host at export, and integrate,
  field triggers, fieldeval, ascent scenes (`field-mach = mach`), and the
  sampler share the machinery. `-f` also accepts ad-hoc expressions and
  bare gradient symbols (`-f grad_u_x`).
- Boundary exports gain a geometry context (`n_x/n_y/n_z`, `wall_dist`)
  which the wall-surface fields are defined over; requesting a
  boundary-only field in a volume export errors cleanly.
- Scalar entries may declare a **gradient** (`grad = ...` with a `{d}`
  placeholder); the registry resolves `grad_<field>_<dim>` and
  `grad_<quantity>_<dim>` recursively down to primitive-variable gradients,
  so declared gradients work on device and host from one declaration.
  Undeclared gradients fail with an error stating the rule instead of a
  NameError or a kernel compile failure.
- tavg gains `fun-avg-mode = eval | defer` (deferred functional averages
  are finalised at export from `[tavg-exprs]` metadata) and per-field
  expression metadata in the output files.

The dormant legacy post-processing machinery (`pyfr/plugins/postproc/`,
the `--postproc` CLI path, and the writer hooks that served it) is
removed in favour of the above (12 files, −382 lines).

## 3. Gradient export

Solution files store the solver's own corrected gradients
(`write-gradients = true`); everything else is derived on request:

- The writer no longer auto-emits `grad density/velocity/pressure` arrays
  into every VTU. Gradients are request-only (`-f grad_u_x`,
  `-f vorticity`, `-f grad_mach_x`), uniformly with the in-situ side.
- Privar gradients are produced through the declared differentials; the
  exported `vorticity` field is bitwise-identical to the previous
  implementation.

## 4. Entropy filter consolidation

`entropylocal`/`entropyfilter` were per-solver copies differing only in a
`compute_entropy` macro; that seam carried no physics once entropy became
a fluid quantity (`s`). The kernels now live once in `baseadvec`, call
`fluid.decl` directly, and the per-solver copies, the `entropy.mako`
shim, and the `system.ef_solver` indirection are deleted.
Bitwise-identical results on Sod-tube validation.

## Behavioural / breaking changes

- `viscosity-correction` is deprecated (warning) in favour of
  `transport =`.
- `pyfr export` no longer auto-appends `Grad <var>` arrays when gradient
  data is present; request gradients explicitly with `-f`. Vector-grouped
  `grad velocity` output is gone (per-component scalars instead).
- The legacy `--postproc` pipeline and its plugins are removed from
  `pyfr export`
- `system.ef_solver` no longer exists; EF kernels are shared.


FreddieWitherden:
I still think this is something we want to pursue.  However, with the new parser we might be able to clean up a lot of the duplication between the Python side and the backend side.  Do we want to rebase?

kaschau:
> I still think this is something we want to pursue. However, with the new parser we might be able to clean up a lot of the duplication between the Python side and the backend side. Do we want to rebase?

We for sure need the spirit of it in place. But part of it you have already introduced with the recent postproc updates, and the other half will be taken care of in completely new ways via the parser. So I don't think this PR has value other than as a goal post. I'll close it.

PR #572, opened 2026-07-16
Parity for OpenMP with 0 a0 guard.

kaschau:
Special-case a zero self-coefficient for OpenMP so it doesn't propagate 0*NaN stale data like the other backends.

### DIFF
--- a/pyfr/backends/openmp/kernels/axnpby.mako
+++ b/pyfr/backends/openmp/kernels/axnpby.mako
@@ -23,38 +23,29 @@ void axnpby(int ib, const struct kargs *restrict args, int _disp_mask)
   % endif
 
     #define X_IDX(v, nv) ((_xi/SOA_SZ*(nv) + (v))*SOA_SZ + _xj)
-% for start in [1, 0]:
-    ${'if (a0 == 0.0)' if start else 'else'}
+    for (ixdtype_t _y = 0; _y < nrow; _y++)
     {
-        for (ixdtype_t _y = 0; _y < nrow; _y++)
+        for (ixdtype_t _xi = 0; _xi < BLK_SZ; _xi += SOA_SZ)
         {
-            for (ixdtype_t _xi = 0; _xi < BLK_SZ; _xi += SOA_SZ)
+            #pragma omp simd
+            for (ixdtype_t _xj = 0; _xj < SOA_SZ; _xj++)
             {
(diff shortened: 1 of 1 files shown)


FreddieWitherden:
We are usually a bit ambiguous when it comes to NaN handling. As in the compiler and backend are free to propagate or not. Does the current OpenMP behaviour break anything?

pyfr/backends/openmp/kernels/axnpby.mako
```diff
@@ -23,24 +23,38 @@ void axnpby(int ib, const struct kargs *restrict args, int _disp_mask)
...
             }
         }
     }
+% endfor
     #undef X_IDX
 % else:
-    #pragma omp simd
-    for (ixdtype_t i = ib*nrow*BLK_SZ*${ncola}; i < (ib + 1)*nrow*BLK_SZ*${ncola}; i++)
-        x0[i] = ${pyfr.dot('a{l}', 'x{l}[i]', l=nv)};
+    if (a0 == 0.0)
```
FreddieWitherden: Can move the if inside the loop. Compiler should hoist it (but check assembly)

kaschau:
> We are usually a bit ambiguous when it comes to NaN handling. As in the compiler and backend are free to propagate or not. Does the current OpenMP behaviour break anything?

Yeah, recovery from a divergent implicit solve. A failed Newton attempt leaves bad data in the stage register (the Krylov accumulate path writes into it directly). The controller then halves `dt` and retries, and the retry starts from `initial_guess_fn`, which overwrites the register with an `a0 = 0` axnpby. On the other four backends this cleans the register and the retry can converge; on OpenMP the guess evaluates to `0*NaN + ... = NaN`, so the initial residual is non-finite and every retry fails identically until the failure limit aborts the run. A single divergent solve kills OpenMP and is recoverable everywhere else, with the same config.

pyfr/backends/openmp/kernels/axnpby.mako
kaschau: verified

FreddieWitherden:
Can you confirm the compiler does lift the if out? Just dumping the assembly should show easily enough.

kaschau:
> Can you confirm the compiler does lift the if out? Just dumping the assembly should show easily enough.

verified on my machine, it does lift it out

PR #574, opened 2026-07-16
Enable block-jacobi at p=0

kaschau:
Update the view shape check. Block-Jacobi at p = 0 dies at kernel instantiation because the generated signature expects a stride pointer the view never built.

### DIFF
--- a/pyfr/backends/base/generator.py
+++ b/pyfr/backends/base/generator.py
@@ -130,7 +130,7 @@ class BaseKernelGenerator:
                         argt.append([np.uintp, np.uintp])
                     case 2, _:
                         argt.append([np.uintp, np.uintp, self.ixdtype])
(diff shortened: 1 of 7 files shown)


FreddieWitherden:
Is the stride strictly needed or is the kernel just going to be multiplying something by zero? If so we should look into changing the kernel so that it does not reference strides unless they're actually used.

kaschau:
> Is the stride strictly needed or is the kernel just going to be multiplying something by zero? If so we should look into changing the kernel so that it does not reference strides unless they're actually used.

it just multiplies by zero, now checking if stride is even needed

FreddieWitherden:
I think this is a better solution.  Let's do some more testing (anything which changes the DSL like this does need some more care than usual) but at first glance it looks solod.

PR #575, opened 2026-07-16
Heal diverged Newton for retry.

kaschau:
When a Newton solve diverges, `NonlinearDivergenceError` propagates up to the controller, which halves `dt` and retries the step. Three defects in the nonlinear machinery can turn one such divergence into an unrecoverable failure, or corrupt state that the retry then consumes:

**1. The line search applies non-finite steps.** The Armijo loop in `BaseNonlinearSolver._line_search` applies the final `alpha` unconditionally after the loop. Since every comparison against a NaN residual norm is false, the acceptance test can never break out of the loop, and the step applied on exhaustion is `fact**maxiter` — an alpha that was never actually evaluated. If every damped trial produced a non-finite residual, this writes a known-bad step into the solution register. The fix tracks the best finite trial: on exhaustion the best finite `alpha` is applied instead of an untested one, and if no trial was finite a `NonlinearDivergenceError` is raised with the solution register untouched.

**2. The stage predictor is unguarded.** The extrapolated initial guess in the SDIRK steppers can leave the physical state space on stiff problems (the norm-based `adt` damping is global and does not protect individual variables). A non-finite initial residual was instant divergence — before the line search or the Krylov solver ever ran — even though the trivial guess `u = u_n` is always available and frequently converges at the same `dt`. The fix evaluates the predictor's residual and falls back to the trivial guess when it is non-finite.

**3. `fd_eps` is computed from a stale register.** `NewtonSolver._solve` computed the adaptive JFNK perturbation from `u_reg` *before* the initial guess was set. On the first stage of a step this register holds data from a previous time step; after a rejected step it holds the failed — possibly non-finite — Newton iterate, in which case `fd_eps` becomes NaN and every subsequent matvec is NaN regardless of how far the controller reduces `dt`. This presents as an unfixable "Non-finite residual" loop until the failure limit aborts the run. The fix computes `fd_eps` after the initial guess is in place (per tolerance retry, so each retry uses its actual starting state), applies the same reordering in the Anderson solver, and adds a non-finite norm guard in `_compute_fd_eps` which falls back to the non-adaptive baseline `eps`.

The three changes are complementary: (1) stops divergence from corrupting the solution register, (2) removes a spurious divergence trigger, and (3) ensures that when a divergence does occur, the retry starts from clean inputs. Behavior on converging solves is unchanged — the residual evaluation count per stage is identical, and the line search applies the same `alpha` whenever the Armijo condition is met.

### DIFF
--- a/pyfr/integrators/implicit/base.py
+++ b/pyfr/integrators/implicit/base.py
@@ -1,3 +1,5 @@
+import math
+
 import numpy as np
 
 from pyfr.integrators.base import (BaseIntegrator, _common_plugin_prop,
@@ -137,6 +139,9 @@ class BaseImplicitIntegrator(BaseIntegrator):
         # If we are adaptive then recompute eps
         if self._fd_eps_adapt:
             unorm = self._norm2(u_reg)
+            if not math.isfinite(unorm):
+                unorm = 0
+
             self._fd_eps = ((1 + unorm)*self.backend.fpdtype_eps)**0.5
 
     @property
--- a/pyfr/integrators/implicit/base.py
+++ b/pyfr/integrators/implicit/base.py
@@ -1,5 +1,3 @@
-import math
-
 import numpy as np
 
 from pyfr.integrators.base import (BaseIntegrator, _common_plugin_prop,
@@ -139,9 +137,6 @@ class BaseImplicitIntegrator(BaseIntegrator):
         # If we are adaptive then recompute eps
         if self._fd_eps_adapt:
             unorm = self._norm2(u_reg)
-            if not math.isfinite(unorm):
-                unorm = 0
-
             self._fd_eps = ((1 + unorm)*self.backend.fpdtype_eps)**0.5
 
     @property
--- a/pyfr/integrators/implicit/nonlinear/base.py
+++ b/pyfr/integrators/implicit/nonlinear/base.py
@@ -75,8 +75,8 @@ class BaseNonlinearSolver(BaseImplicitIntegrator):
         return self._calc_rnorm(self._nl_resid)
 
     def _line_search(self, t, u_reg, f_reg, delta_reg, residual_fn, rnorm_old):
-        alpha = 1.0
-        alpha_best = rnorm_best = None
+        alpha = alpha_best = 1.0
+        rnorm_best = math.inf
 
         for _ in range(self._linesearch_maxiter):
             self._add(0, self._nl_temp, 1, u_reg, alpha, delta_reg,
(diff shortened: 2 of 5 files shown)


FreddieWitherden:
For the stage prediction stuff look at the trapezium rule (what I've been focusing on).  I think it has slightly more robust logic that we can port over the to other schemes.

pyfr/integrators/implicit/nonlinear/base.py
```diff
@@ -84,10 +86,22 @@ def _line_search(self, t, u_reg, f_reg, delta_reg, residual_fn, rnorm_old):
             residual_fn(self._nl_temp, f_reg, self._nl_resid)
             rnorm_new = self._calc_rnorm(self._nl_resid)
 
+            # Note the best finite trial step seen so far
+            if math.isfinite(rnorm_new) and (rnorm_best is None or
```
FreddieWitherden: While I don't think this will simplify we might want to rewrap to avoid the long indent.

pyfr/integrators/implicit/nonlinear/newton.py
```diff
@@ -30,9 +30,17 @@ def matvec(v, result, vnorm=None):
 
         # Pick an initial starting guess
         initial_guess_fn(u_reg)
+        rnorm = self._residual_norm(t, u_reg, f_reg, residual_fn)
+
+        # If the predictor has left the state space use the trivial guess
+        if not math.isfinite(rnorm):
+            initial_guess_fn(u_reg, trivial=True)
```
FreddieWitherden: Maybe this belongs in the initial guess function?

pyfr/integrators/implicit/base.py
```diff
@@ -137,6 +139,9 @@ def _compute_fd_eps(self, u_reg):
         # If we are adaptive then recompute eps
         if self._fd_eps_adapt:
             unorm = self._norm2(u_reg)
+            if not math.isfinite(unorm):
```
FreddieWitherden: Just thinking here when with the new code we can get a dodgy u_reg. If it can occur after the above fixes we might want to consider refactoring.

pyfr/integrators/implicit/nonlinear/base.py
FreddieWitherden: Actually, we might be able to simplify if the check further down in the for/else is tweaked.

kaschau:
> For the stage prediction stuff look at the trapezium rule (what I've been focusing on). I think it has slightly more robust logic that we can port over the to other schemes.

are you referring to the reduction to weight one on the fresh `f(t, u_n)` (no extrapolation), versus TRBDF2's final stage at `w ≈ [-0.71, +1.71]`? If so I could make the fallback graded: extrapolation -> first-order predictor through the most recent stage derivative -> trivial guess

FreddieWitherden:
Yes, I think the compute initial guess fn should always be required to return a valid guess as part of its contract (even if that is just the current state with no interpolation/extrapolation).

FreddieWitherden:
Looks reasonable.  I think it might be worth seeing if there is an approach which works in terms of `_compute_stage_initial_guess` without the level pass-through (so it owns everything).  Goal is to save on reductions as the norms are quite bandwidth hungry so any way we can be smart and avoid doing extra norms (or facilitate reuse) is a win.

PR #576, opened 2026-07-17
Add optional parallel HIP GiMMiK compilation

tomjen12:
## Summary

This PR adds an opt-in PyFR-side parallel compile path for HIP GiMMiK autotuning.

When `backend-hip.gimmik-parallel-compile = true`, GiMMiK candidate source generation and HIPRTC compilation/cache warmup are dispatched to multiple workers. The main process then benchmarks candidates through the normal kernel build path, which should hit the warmed HIP object cache.

The default path is unchanged.

This follows the HIP MFMA dense GiMMiK work in PyFR/GiMMiK#20 and uses the indexed candidate rendering support from PyFR/GiMMiK#21. It is also the PyFR-side follow-up to PyFR/PyFR#570.

## Validation

Tested on MI300X with 1000 selected dense cases.

| Metric | Value |
|---|---:|
| Cases | 1000 |
| Worst-case MFMA / rocBLAS BW | 95% |
| Geomean MFMA / rocBLAS BW | 102% |
| rocBLAS autotuning total | 3940.1 s |
| MFMA sequential autotuning total | 1137.0 s |
| MFMA parallel compile autotuning total | 416.2 s |
| rocBLAS -> MFMA sequential speedup | 3.47x |
| rocBLAS -> MFMA parallel compile speedup | 9.47x |

Autotuning time also improves significantly. With backend-hip.gimmik-parallel-compile = true, MFMA tuning time drops further to 416.2s, making MFMA with parallel compile 9.47x faster to tune than rocBLAS on these cases.

Attached CSV files:

synthetic_1000_summary.csv: aggregate bandwidth and autotuning summary.
synthetic_1000_details.csv: per-case results for the selected 1000 shapes.

[synthetic_1000_details.csv](https://github.com/user-attachments/files/30109166/synthetic_1000_details.csv)
[synthetic_1000_summary.csv](https://github.com/user-attachments/files/30109169/synthetic_1000_summary.csv)

### DIFF
--- a/pyfr/backends/hip/gimmik.py
+++ b/pyfr/backends/hip/gimmik.py
@@ -34,10 +34,6 @@ class HIPGiMMiKKernels(HIPKernelProvider):
         arr = a.get()
         nnz, nuq = np.count_nonzero(arr), len(np.unique(np.abs(arr)))
 
-        # Check that A is suitable
-        if nuq > 128 and nnz / arr.size > 0.15:
-            raise NotSuitableError('Matrix inappropriate GiMMiK')
-
         # Dimensions
         n = b.ncol
         ldb, ldc = b.leaddim, out.leaddim
--- a/pyfr/backends/hip/gimmik.py
+++ b/pyfr/backends/hip/gimmik.py
@@ -4,8 +4,7 @@ from gimmik import HIPMatMul
 import numpy as np
 
 from pyfr.backends.base import NotSuitableError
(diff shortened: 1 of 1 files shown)


FreddieWitherden:
Some of the more recent changes to GiMMiK should negate the need for this (namely, for the dense kernels A can now be passed in as a buffer which avoids the need to embed large amounts of data in the source).

PR #580, opened 2026-07-31
Namespace physical-group IDs by dimension in the gmsh reader

binggao1230:
gmsh treats a physical group as a (dimension, tag) pair: it writes and re-reads files where a surface group and a volume group share the same numeric ID (e.g. `2 1 "wall"` and `3 1 "vol"`). PyFR currently keys groups by tag alone, so such valid files are rejected with `ValueError: Duplicate physical entity ID: N` before any mesh data is read.

The single-namespace assumption also runs through element grouping and `_merge_vol`, so relaxing the check alone would silently reclassify the same-ID boundary group as a volume region. This PR threads `(dim, tag)` through the gmsh import pipeline instead.

repro (gmsh 4.15):
```
gmsh.model.addPhysicalGroup(3,[1],1); gmsh.model.setPhysicalName(3,1,"vol")
gmsh.model.addPhysicalGroup(2,[5],1); gmsh.model.setPhysicalName(2,1,"wall")  # tag 1 collides across dims
```
`pyfr import dup4.1.msh out.pyfrm` fails on main and succeeds here.

Tests add MSH 2.2/4.1 and 2D/3D fixtures with cross-dimension duplicate tags; the dup meshes import and write output byte-identical to their unique-ID twins (`pytest pyfr/tests/ -p no:cacheprovider` -> 5 passed).


pyfr/readers/gmsh.py
```diff
@@ -285,16 +289,17 @@ def _read_phys_names(self, mshit):
...
                 raise ValueError(f'Duplicate physical entity ID: {pent}')
 
             pnames.append((dim, pent, name))
             seen_names.add(name)
-            seen_ids.add(pent)
+            seen.add((dim, pent))
 
         # Classify by dimension
         voldim = max(dim for dim, _, _ in pnames)
+        self._voldim = voldim
```
FreddieWitherden: Can one line.

pyfr/readers/gmsh.py
```diff
@@ -397,8 +402,9 @@ def _read_eles_impl_v2(self, mshit):
             if etype not in self._etype_map:
                 raise ValueError(f'Unsupported element type {etype}')
 
-            # Physical entity type (used for BCs)
-            elenodes[etype, (etags[0],)].append(enodes)
+            # Physical entity type (used for BCs); keyed by its dimension
+            edim = self._petype_ndim[self._etype_map[etype][0]]
+            elenodes[etype, (edim, (etags[0],))].append(enodes)
```
FreddieWitherden: Can we not do `elenodes[etype, edim, (etags[0],)]` (and ibid later on to avoid some tuple creation and unpacking).

FreddieWitherden: Actually, that complicates the assembler, nevermind.

FreddieWitherden:
This looks sensible to me.

pyfr/readers/gmsh.py
binggao1230: Done — `self._voldim = max(...)` is now a single line and `_read_phys_names` reads `self._voldim` directly.

pyfr/readers/gmsh.py
FreddieWitherden: One thing we can do, I think, is construct a dict at the start of the function which resolves each etype => ndim.  Then looking up edim is just a single local dict lookup.

FreddieWitherden:
Also, would you be okay to drop the tests?  The ones we have are mostly legacy (and probably should be removed).

binggao1230:
Done — single `edim_by_etype` dict built once, per-element lookup is now O(1) with no tuple unpacking.

On dropping the tests: happy to trim, but want to check what you mean first — `test_gmsh_reader.py` is new in this PR (no prior gmsh test file existed), so there's nothing pre-existing to remove here. If you meant the MSH 2.2 fixture/test specifically (since 2.2 is the older format), I can drop `dup2.2.msh` + `test_gmsh_dup_physical_ids_3d_v22` and keep just the 4.1 + 2D coverage. Let me know which you had in mind and I'll cut it down.

pyfr/readers/gmsh.py
FreddieWitherden: Does this fit on one line?

FreddieWitherden:
> Done — single `edim_by_etype` dict built once, per-element lookup is now O(1) with no tuple unpacking.
> 
> On dropping the tests: happy to trim, but want to check what you mean first — `test_gmsh_reader.py` is new in this PR (no prior gmsh test file existed), so there's nothing pre-existing to remove here. If you meant the MSH 2.2 fixture/test specifically (since 2.2 is the older format), I can drop `dup2.2.msh` + `test_gmsh_dup_physical_ids_3d_v22` and keep just the 4.1 + 2D coverage. Let me know which you had in mind and I'll cut it down.

The tests don't add a lot here.  Eventually, we'll get a mesh in PyFR-Test-Cases which flexes this.

binggao1230:
Done — folds onto one line: `elenodes[etype, (edim_by_etype[etype], (etags[0],))].append(enodes)`.

On tests: dropped the MSH-2.2-specific case (`dup2.2.msh` + its test) since 2.2 is legacy and 4.1 + the 2D fixture already cover the dimension-namespacing logic. Kept those plus the byte-identity round-trip check.

FreddieWitherden:
Yes, just drop the additions to the test directory.  Medium term we'll get one of the test cases to exercise this quirk in its mesh file (and hence into our CI).

FreddieWitherden:
Just following up here.

binggao1230:
Done — I dropped all additions under `pyfr/tests/`, so the PR now touches only `pyfr/readers/gmsh.py`.

Updated head: GPG-verified `47a06e3`.

I kept the verification external rather than shipping it in the tree:

```text
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -m pytest pyfr/tests/test_gmsh_reader.py -q
3 passed  # before dropping the test files

python3 -m py_compile pyfr/readers/gmsh.py
python3 /tmp/pyfr580_external_gmsh/check.py
external gmsh regression OK
```

The external check uses the removed mesh fixtures outside the source tree and still verifies the same two invariants: duplicate `(dim, tag)` groups import with the expected boundary/volume classification, and duplicate-ID meshes write byte-identical `.pyfrm` output to their unique-ID twins.

PR #586, opened 2026-08-14
Fix GMRES restart residual reconstruction

### LINKED ISSUE #585: GMRES restart reconstructs the residual incorrectly (missing Krylov components)
## Summary

The restarted-GMRES residual reconstruction in `pyfr/integrators/implicit/krylov/gmres.py` is mathematically incorrect. It reconstructs the residual as being parallel to the *last* Arnoldi vector only, dropping the components in the earlier Krylov directions. The magnitude is right but the direction is wrong, so restarted GMRES stagnates or breaks down instead of converging.

This is masked by default because `solver-gmres restart` defaults to `0` (→ single cycle, no restart). It only bites when `restart` is set below `linear-max-iter`.

## Root cause

`gmres.py:142-145`:

```python
# Restart: recover residual via r_m = beta[j+1]*v_{j+1}/h_{j+1,j}
rnorm = abs(self._beta[j + 1])
s = np.copysign(1 / h_jp1_j, self._beta[j + 1])
self._add(0, v[0], s, v[j + 1])
```

This sets the new starting vector to `±v_{j+1}` (the last Arnoldi vector) with `β = |β[j+1]|`.

The exact residual after a GMRES cycle is not parallel to `v_{j+1}`. In exact arithmetic,

```
r_m = β_m · V_{m+1} · (G_0ᵀ G_1ᵀ … G_{m-1}ᵀ e_{m+1})
```

i.e. a linear combination of *all* the Arnoldi vectors `v_0 … v_m`, with coefficients cascading through every Givens rotation. Dropping everything except `v_m` discards a significant part of the residual (in a typical cycle the `v_{m-1}` component alone is ~40% of the `v_m` component, with a further tail in the earlier directions).

## Reproduction

Replicating the algorithm (CGS Arnoldi + incremental Givens + the restart above) in NumPy and comparing against `scipy.sparse.linalg.gmres` on a well-conditioned SPD matrix (`cond ≈ 8`), targeting `rtol = 1e-10`:

| restart `m` | PyFR reconstruction | scipy (correct) |
|---|---|---|
| 2 | 1.3e-01 | 9.2e-11 |
| 4 | 3.3e-02 | 6.5e-11 |
| 8 | 1.0e-03 | 8.8e-11 |
| 15 | 3.9e-06 | 1.0e-10 |

The current reconstruction fails to reach the requested tolerance (it stagnates / hits breakdown), while a correct restart converges to ~1e-10 in every case.

## Proposed fix

Two equivalent options (both reproduce scipy's results exactly):

1. **Explicit residual** (simplest; costs one extra matvec per cycle):
   ```python
   r = residual - matvec(x)      # r = b - A x
   rnorm = self._norm2(r)
   self._add(0, v[0], 1 / rnorm, r)
   ```

2. **Givens cascade** (no extra matvec; reuse the stored `cs`, `sn` and `v`):
   ```python
   z = np.zeros(j + 2); z[j + 1] = 1.0
   for i in range(j, -1, -1):
       z[i], z[i + 1] = (cs[i]*z[i] - sn[i]*z[i + 1],
                         sn[i]*z[i] + cs[i]*z[i + 1])
   # v[0] = V_{j+2} z  (then beta[0] = |beta[j+1]| as before)
   ```

## Environment

- Branch: `develop` (438b9123)
- Relevant config: `solver-gmres restart < solver-time-integrator linear-max-iter`

po-nuvai:
## Summary

Fixes the restarted-GMRES residual reconstruction in `pyfr/integrators/implicit/krylov/gmres.py`. Closes #585.

The restart was reconstructing the residual as parallel to the *last* Arnoldi vector only, which has the correct magnitude but the wrong direction. In exact arithmetic the residual after a GMRES cycle is

```
r_m = β_m · V_{m+1} · (G₀ᵀ G₁ᵀ … G_{m-1}ᵀ e_{m+1})
```

i.e. a linear combination of *all* the Arnoldi vectors `v_0 … v_m`, with coefficients cascading through every Givens rotation — not just `v_m`. Dropping the earlier components caused restarted GMRES (`restart < linear-max-iter`) to stagnate or break down instead of converging.

## The fix

Cascade the Givens rotations back through the Krylov basis to reconstruct the full residual direction. It reuses the already-stored `cs`/`sn` rotations and Krylov vectors, so it costs no extra matvec:

```python
# reconstruct the residual direction *before* the solution update, because
# the preconditioned update reuses v[0] as scratch for M⁻¹(V y) and would
# otherwise clobber it; stash the result in v[j+1], which survives.
if will_restart:
    self._add(1 / h_jp1_j, v[j + 1])          # normalise the last Arnoldi vector

    z = np.zeros(j + 2)
    z[j + 1] = 1.0
    for i in range(j, -1, -1):
        c, s = self._cs[i], self._sn[i]
        z[i], z[i + 1] = c*z[i] - s*z[i + 1], s*z[i] + c*z[i + 1]
    z *= np.copysign(1.0, self._beta[j + 1])

    self._addv([z[j + 1], *z[:j + 1].tolist()], [v[j + 1], *v[:j + 1]])

# ... solution update ...

if not will_restart:
    break

rnorm = abs(self._beta[j + 1])
self._add(0, v[0], 1, v[j + 1])               # install the reconstructed residual
```

The one subtlety worth calling out: the reconstruction has to happen **before** the solution update, not after. The preconditioned solution update reuses `v[0]` as a scratch register for `M⁻¹(V y)`, which would destroy the first Arnoldi vector that the cascade needs as a source.

## Validation

Replicating the algorithm in NumPy (MGS/CGS Arnoldi + incremental Givens) and comparing against `scipy.sparse.linalg.gmres` on a well-conditioned SPD matrix (`cond ≈ 8`), targeting `rtol = 1e-10`, with and without a right preconditioner:

| restart `m` | old | fix (no precond) | fix (preconditioned) | scipy |
|---|---|---|---|---|
| 2 | 1.3e-01 | 9.2e-11 | 5.0e-11 | 9.2e-11 |
| 4 | 3.3e-02 | 6.5e-11 | 3.6e-11 | 6.5e-11 |
| 8 | 1.0e-03 | 8.8e-11 | 5.2e-11 | 8.8e-11 |
| 15 | 3.9e-06 | 1.0e-10 | 9.8e-11 | 1.0e-10 |

## Notes

- **Masked by default** — `solver-gmres restart` defaults to `0` (single cycle), so the buggy path only triggers when `restart < linear-max-iter`.
- Independent of the Arnoldi variant (`cgs` vs `mgs`) and of preconditioning.

### DIFF
--- a/pyfr/integrators/implicit/krylov/gmres.py
+++ b/pyfr/integrators/implicit/krylov/gmres.py
@@ -122,6 +122,24 @@ class GMRESMixin(BaseLinearSolver):
             # Backward substitution to solve for y
             y = np.linalg.solve(self._H[:j + 1, :j + 1], self._beta[:j + 1])
 
+            # Determine if a restart is needed after this cycle
+            will_restart = not (err < rtol or h_jp1_j < self._breakdown_tol
+                                or niters >= nmax)
+
+            # Reconstruct the restart residual before the update clobbers v[0]
+            if will_restart:
+                self._add(1 / h_jp1_j, v[j + 1])
+
+                z = np.zeros(j + 2)
+                z[j + 1] = 1.0
+                for i in range(j, -1, -1):
+                    c, s = self._cs[i], self._sn[i]
+                    z[i:i + 2] = (c*z[i] - s*z[i + 1], s*z[i] + c*z[i + 1])
+                z *= np.copysign(1.0, self._beta[j + 1])
+
+                self._addv([z[j + 1], *z[:j + 1].tolist()],
+                           [v[j + 1], *v[:j + 1]])
+
             # Solution update; first cycle honours accumulate, rest add
             first = niters == j + 1
             acc = float(accumulate) if first else 1.0
@@ -136,13 +154,12 @@ class GMRESMixin(BaseLinearSolver):
                 self._addv([acc, *y.tolist()], [out_reg, *v[:j + 1]],
                            in_scale=accumulate_scale, in_scale_idxs=sidxs)
 
-            if err < rtol or h_jp1_j < self._breakdown_tol or niters >= nmax:
+            if not will_restart:
                 break
 
-            # Restart: recover residual via r_m = beta[j+1]*v_{j+1}/h_{j+1,j}
+            # Restart: install the reconstructed residual as the new v[0]
             rnorm = abs(self._beta[j + 1])
-            s = np.copysign(1 / h_jp1_j, self._beta[j + 1])
-            self._add(0, v[0], s, v[j + 1])
+            self._add(0, v[0], 1, v[j + 1])
 
         # Each matvec applies M⁻¹ once; recovery adds one more per cycle
         return niters, niters + ncycles if precond else 0


po-nuvai:
Reviewed this change. The math is right — the old restart kept only the `v[j+1]` component of the residual (correct magnitude, wrong direction), and cascading the Givens rotations back through the basis is the proper reconstruction of `r_m = β_m V_{m+1} (Gᵀ e_{m+2})`.

One thing I caught while going back over it, and have now fixed in the branch: the reconstruction has to happen *before* the solution update, not after. With a preconditioner active the update reuses `v[0]` as scratch for `M⁻¹(V y)`, which destroys the first Arnoldi vector that the cascade reads as a source. So the residual direction is now computed ahead of the update, stashed in `v[j+1]` (which survives), and installed into `v[0]` afterwards.

Non-blocking nits, if you want them:
- the cascade tuple has redundant outer parens (`z[i], z[i + 1] = (...)`).
- `np.zeros(j + 2)` could be `dtype=self._beta.dtype` for explicitness; it defaults to float64 which matches `_cs`/`_sn`/`_beta` anyway.

Numerically I replicated the algorithm and it matches scipy's restarted GMRES to ~1e-10 with and without a right preconditioner (the old code sat at 1e-1 … 1e-6 and stagnated/breakdown'd). Default config is unaffected since `restart` defaults to 0 → single cycle.

FreddieWitherden:
Overall looks good.  Can probably clean up the z assignments a bit z[i:i+1] = and see if it is easier to enumerate(zip(...)) over the c and s terms.

po-nuvai:
Done — switched the cascade to slice assignment and `enumerate(zip(...))` over the `c`/`s` pairs:

```python
for k, (c, s) in enumerate(zip(self._cs[j::-1], self._sn[j::-1])):
    i = j - k
    z[i:i + 2] = (c*z[i] - s*z[i + 1], s*z[i] + c*z[i + 1])
```

Reads a bit cleaner, and it's still the same backward cascade (applies `G_j^T` … `G_0^T` in order). I re-checked the equivalence numerically — identical to machine precision against the previous form.

FreddieWitherden:
> Done — switched the cascade to slice assignment and `enumerate(zip(...))` over the `c`/`s` pairs:
> 
> ```python
> for k, (c, s) in enumerate(zip(self._cs[j::-1], self._sn[j::-1])):
>     i = j - k
>     z[i:i + 2] = (c*z[i] - s*z[i + 1], s*z[i] + c*z[i + 1])
> ```
> 
> Reads a bit cleaner, and it's still the same backward cascade (applies `G_j^T` … `G_0^T` in order). I re-checked the equivalence numerically — identical to machine precision against the previous form.

I think it ends up a bit uglier in practice due to the extra line breaks. Maybe just revert and just tweak the assignment.

pyfr/integrators/implicit/krylov/gmres.py
```diff
@@ -122,6 +122,31 @@ def _krylov_solve(self, matvec, residual, out_reg, precond=None, *,
...
+            # Determine if a restart is needed after this cycle
+            will_restart = not (err < rtol or h_jp1_j < self._breakdown_tol
+                                or niters >= nmax)
+
+            # Reconstruct the restart residual direction before the solution
+            # update below, which may reuse v[0] as scratch when a
+            # preconditioner is active.  The residual is
+            # r_m = beta[j+1]*V_{j+2}*(G^T e_{j+2}), i.e. a combination of all
+            # of the Arnoldi vectors v_0..v_{j+1}; stash the result in v[j+1],
+            # which survives the solution update.
```
FreddieWitherden: Aim for single line comments.

po-nuvai:
Agreed on both — reverted the `enumerate(zip(...))` (the line breaks made it worse) and just kept the slice-assignment tweak:

```python
for i in range(j, -1, -1):
    c, s = self._cs[i], self._sn[i]
    z[i:i + 2] = (c*z[i] - s*z[i + 1], s*z[i] + c*z[i + 1])
```

Also condensed that comment block down to a single line.  :)

pyfr/integrators/implicit/krylov/gmres.py
```diff
@@ -122,6 +122,24 @@ def _krylov_solve(self, matvec, residual, out_reg, precond=None, *,
...
+
+            # Reconstruct the restart residual before the update clobbers v[0]
+            if will_restart:
+                self._add(1 / h_jp1_j, v[j + 1])
+
+                z = np.zeros(j + 2)
+                z[j + 1] = 1.0
+                for i in range(j, -1, -1):
+                    c, s = self._cs[i], self._sn[i]
+                    z[i:i + 2] = (c*z[i] - s*z[i + 1], s*z[i] + c*z[i + 1])
```
FreddieWitherden, line 137: Can drop the ()

pyfr/integrators/implicit/krylov/gmres.py
```diff
@@ -122,6 +122,24 @@ def _krylov_solve(self, matvec, residual, out_reg, precond=None, *,
...
+                z *= np.copysign(1.0, self._beta[j + 1])
```
FreddieWitherden, line 138: Do we need copysign here or will np.sign work?

pyfr/integrators/implicit/krylov/gmres.py
```diff
@@ -122,6 +122,24 @@ def _krylov_solve(self, matvec, residual, out_reg, precond=None, *,
             # Backward substitution to solve for y
             y = np.linalg.solve(self._H[:j + 1, :j + 1], self._beta[:j + 1])
 
+            # Determine if a restart is needed after this cycle
+            will_restart = not (err < rtol or h_jp1_j < self._breakdown_tol
```
FreddieWitherden, line 126: Given we not this in one `if` later on, we may want to remove the not and rename the variable accordingly.

FreddieWitherden: `done` maybe?

FreddieWitherden:
Are you okay if I get this upstream w/corrections?

FreddieWitherden:
Is this still something you're interested in working on, or are you okay for me to get it incorporated?

PR #587, opened 2026-08-17
Boundary Groups

kaschau:
For complex geometries, it is convenient to split boundaries at mesh-gen into several discrete entities. However, these entities may all share a single boundary condition. In other cases, some post processing operations are best done as individual boundaries, while others are nice to aggregate. Consider a aerofoil with a LE, TE, PS, SS surfaces. Some data is nice to have individually, while something like lift or a surface integral would be best to group these surfaces. This PR introduces brace expansion everywhere boundaries are specified:

**Boundary conditions**

[soln-bcs-{le, te, ps, ss}]
type = no-slp-adia-wall

The four boundaries become one interface, one kernel set, one serialisation key. Each boundary must get its condition from exactly one section — a boundary claimed twice (its own section plus an enumeration, or two enumerations) is an error naming both. This is semantic, not just shorthand: for controlled BCs,

[soln-bcs-{inlet-upper, inlet-lower}]
type = char-riem-inv-mass-flow
mass-flow-rate = 9.0        ; total through BOTH boundaries
...

gives one regulator measuring the combined flow, separate sections would give two independent regulators, each chasing the full 9.0.

**Force integration plugin**

[soln-plugin-fluidforce-{le, te, ps, ss}]
nsteps = 10
file = airfoil-forces.csv
morigin = (0.25, 0.0, 0.0)

Forces and moments integrated over the fused surface in one pass — equals the sum of four single-boundary plugins to round-off.

**Regions and surfaces**

[soln-plugin-writer]
region = bc/{le, te, ps, ss}        ; region-restricted output

[soln-plugin-fwh]
surface = bc/{shield, shoulder}     ; members must share one BC type

[soln-plugin-ascent]
surface-wing = bc/{le, te, ps, ss}  ; one render source spanning all four

**Export**

pyfr export boundary mesh.pyfrm soln.pyfrs wing.vtu '{le, te, ps, ss}'

Quotes required (the shell expands bare braces itself). Byte-identical to listing the names separately; mixing works too: '{le, te}' ps ss.

Rules
Members must exist in the mesh; unknown names error listing the offenders, identically on every rank.
{a,}, {a,a}, {} are rejected.
Plain names are untouched, so configs without braces behave bit-identically to before.

### DIFF
--- a/pyfr/plugins/soln/ascent.py
+++ b/pyfr/plugins/soln/ascent.py
@@ -9,6 +9,7 @@ from pyfr.ctypesutil import LibWrapper
 from pyfr.exprs import npeval
 from pyfr.fields import CleanToGrid
 from pyfr.inifile import process_expr
+from pyfr.readers.native import Connectivity
 from pyfr.mpiutil import get_comm_rank_root, mpi
 from pyfr.plugins.common import region_data
 from pyfr.plugins.postproc.adapters import (BoundaryPostProcData, FaceInfo,
@@ -17,7 +18,8 @@ from pyfr.plugins.postproc import get_source
 from pyfr.plugins.soln.base import BaseSolnPlugin
 from pyfr.shapes import BaseShape, proj_pts
 from pyfr.subdiv import get_subdiv
-from pyfr.util import file_path_gen, first, paren_depths, subclass_where
+from pyfr.util import (expand_braces, file_path_gen, first, paren_depths,
+                       subclass_where)
 
 
 def con_psolns_pgrads(elementscls, scfg, csolns, cgrads):
@@ -348,15 +350,18 @@ class _BoundaryAscentOutput(_VolumeAscentOutput):
         self.renderer = renderer
         self.sname = sname
 
-        # Accept bc/foo or foo to match mesh.bcon keys
-        bcname = region.removeprefix('bc/')
-        comm, _, _ = get_comm_rank_root()
-        if not comm.allreduce(bcname in renderer.mesh.bcon, op=mpi.LOR):
-            raise ValueError(f'Boundary {bcname} does not exist')
+        # Accept bc/foo or foo or bc/{foo,bar}
+        mesh = renderer.mesh
+        bcnames = expand_braces(region.removeprefix('bc/'))
+        if missing := [b for b in bcnames if f'bc/{b}' not in mesh.codec]:
+            raise ValueError(f'Boundaries do not exist: {missing}')
+
+        # Fuse the boundaries into a single surface
+        conn = Connectivity.fuse(mesh.bcon.get(b) for b in bcnames)
 
         # Per-itype patches, each a flat (eidxs, etype, mop, sop, fidx, svpts)
         self.patches = patches = defaultdict(list)
-        if (conn := renderer.mesh.bcon.get(bcname)) is not None:
+        if conn is not None:
             for etype, fidx, eidxs in conn.items():
                 itype, mop, sop, svpts = renderer.adapter.face_soln_op_vpts(
                     etype, fidx, renderer.divisor
--- a/pyfr/plugins/soln/ascent.py
+++ b/pyfr/plugins/soln/ascent.py
@@ -9,7 +9,6 @@ from pyfr.ctypesutil import LibWrapper
 from pyfr.exprs import npeval
 from pyfr.fields import CleanToGrid
 from pyfr.inifile import process_expr
-from pyfr.readers.native import Connectivity
 from pyfr.mpiutil import get_comm_rank_root, mpi
 from pyfr.plugins.common import region_data
 from pyfr.plugins.postproc.adapters import (BoundaryPostProcData, FaceInfo,
@@ -18,7 +17,7 @@ from pyfr.plugins.postproc import get_source
 from pyfr.plugins.soln.base import BaseSolnPlugin
 from pyfr.shapes import BaseShape, proj_pts
 from pyfr.subdiv import get_subdiv
-from pyfr.util import (expand_braces, file_path_gen, first, paren_depths,
+from pyfr.util import (file_path_gen, first, paren_depths,
                        subclass_where)
 
 
@@ -350,14 +349,8 @@ class _BoundaryAscentOutput(_VolumeAscentOutput):
         self.renderer = renderer
         self.sname = sname
 
-        # Accept bc/foo or foo or bc/{foo,bar}
-        mesh = renderer.mesh
-        bcnames = expand_braces(region.removeprefix('bc/'))
-        if missing := [b for b in bcnames if f'bc/{b}' not in mesh.codec]:
-            raise ValueError(f'Boundaries do not exist: {missing}')
-
-        # Fuse the boundaries into a single surface
-        conn = Connectivity.fuse(mesh.bcon.get(b) for b in bcnames)
+        # Accept bc/foo or foo or bc/{foo, bar}
+        conn = renderer.mesh.bcon_for(region.removeprefix('bc/'))
 
         # Per-itype patches, each a flat (eidxs, etype, mop, sop, fidx, svpts)
         self.patches = patches = defaultdict(list)
--- a/pyfr/plugins/soln/ascent.py
+++ b/pyfr/plugins/soln/ascent.py
@@ -349,7 +349,10 @@ class _BoundaryAscentOutput(_VolumeAscentOutput):
         self.renderer = renderer
         self.sname = sname
 
-        # Accept bc/foo or foo or bc/{foo, bar}
+        # Boundary surface sources take the form bc/<name>
+        if not region.startswith('bc/'):
+            raise ValueError(f'Invalid surface source: {region}')
+
         conn = renderer.mesh.bcon_for(region.removeprefix('bc/'))
 
         # Per-itype patches, each a flat (eidxs, etype, mop, sop, fidx, svpts)
--- a/pyfr/readers/native.py
+++ b/pyfr/readers/native.py
@@ -95,6 +95,23 @@ class Connectivity:
 
         return Connectivity(self.cidxs[mask], new[mask], self.cidxmap)
 
+    @classmethod
+    def fuse(cls, cons):
+        # Fuse several connectivities, which must share a cidxmap, into one
+        cons = [c for c in cons if c is not None and len(c)]
+        if not cons:
+            return None
+        elif len(cons) == 1:
+            return cons[0]
+
+        cidxs = np.concatenate([c.cidxs for c in cons])
+        eidxs = np.concatenate([c.eidxs for c in cons])
+
+        # Order by (cidx, eidx) so the result does not depend on fuse order
+        sidx = np.lexsort((eidxs, cidxs))
+
+        return cls(cidxs[sidx], eidxs[sidx], first(cons).cidxmap)
+
 
 class NativeReader:
     def __init__(self, fname, pname=None, *, construct_con=True):
--- a/pyfr/readers/native.py
+++ b/pyfr/readers/native.py
@@ -47,7 +47,11 @@ class Mesh:
 
     def bcon_for(self, spec):
         # Boundaries in any brace enumeration, fused into one connectivity
-        names = expand_braces(spec)
+        names = list(expand_braces(spec))
+
+        # Catch double counted boundaries
+        if len(set(names)) != len(names):
+            raise ValueError(f'Duplicate boundaries in {spec}')
 
         # The codec is global, so this check is consistent across all ranks
         if missing := [b for b in names if f'bc/{b}' not in self.codec]:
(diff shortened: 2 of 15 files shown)


FreddieWitherden:
I like the overall idea however am wondering if we can get most of the value with a simpler implementation.  One would be just extending the syntax to `[soln-bcs-{a,b,c}]`.  This could be handled relatively easily and uses the same syntax that is already supported by the time average stats.

kaschau:
> I like the overall idea however am wondering if we can get most of the value with a simpler implementation. One would be just extending the syntax to `[soln-bcs-{a,b,c}]`. This could be handled relatively easily and uses the same syntax that is already supported by the time average stats.

I think that wins on simplicity, but looses on repetition saving. And a decision would be for the controller BCs what does `[soln-bcs-{foo,bar}]` mean? Is it a fused boundary or still 2 independent actors (I think it has to be fused to be sane for anything to react to it). And do you like fusing the grouped BCs for kernel launch? might save a chunk of change on dispatch.

FreddieWitherden:
> > I like the overall idea however am wondering if we can get most of the value with a simpler implementation. One would be just extending the syntax to `[soln-bcs-{a,b,c}]`. This could be handled relatively easily and uses the same syntax that is already supported by the time average stats.
> 
> I think that wins on simplicity, but looses on repetition saving. And a decision would be for the controller BCs what does `[soln-bcs-{foo,bar}]` mean? Is it a fused boundary or still 2 independent actors (I think it has to be fused to be sane for anything to react to it). And do you like fusing the grouped BCs for kernel launch? might save a chunk of change on dispatch.

It would be handled in system.py which would do the interface merging and give a combined LHS to the constructor.

pyfr/util.py
```diff
@@ -206,6 +206,23 @@ def strip_parens(s):
                    if d == 0 and c not in ')}')
 
 
+def expand_braces(spec):
```
FreddieWitherden, line 209: Can this be used anywhere else?  Check there isn't a shell or random Python built-in module which does this already (I've been caught out there a few times).

pyfr/solvers/base/system.py
```diff
@@ -271,37 +301,51 @@ def _load_bc_inters(self, mesh, elemap, initsoln, serialiser):
...
-            localbc = bname in mesh.bcon
+            if bname not in bcsects:
+                raise ValueError(f'No boundary condition for {bname}')
+
+            sects.setdefault(bcsects[bname], []).append(bname)
+
+        # Iterate over the boundary conditions
+        for cfgsect, bnames in sects.items():
+            # Fuse the constituent boundaries
+            con = Connectivity.fuse(mesh.bcon.get(b) for b in bnames)
```
FreddieWitherden: Thoughts on having a method on the Mesh to handle this?  I think if we structure it well it can be reused 2/3 times and save us some hassle.

pyfr/solvers/base/system.py
```diff
@@ -262,6 +263,35 @@ def _load_mpi_inters(self, mesh, elemap):
...
         return mpi_inters
 
+    @staticmethod
+    def bc_sections(cfg, mesh, prefix='soln-bcs-'):
+        # Map each boundary onto the section which parameterises it
+        bcs = {c.removeprefix('bc/') for c in mesh.codec
+               if c.startswith('bc/')}
+
+        sects = {}
+        for sect in cfg.sections():
```
FreddieWitherden: I think this can be cleaned up in terms of the error checking; maybe some sets used in the right way to collapse this down.

FreddieWitherden:
I quite like the revised version.  How is it working out for your use cases?

pyfr/util.py
kaschau: not that i can find. glob/fnmatch treat {a,b} as literal characters

kaschau: but stats can reuse it

pyfr/solvers/base/system.py
kaschau: yeah it actually collapses 4 places into a one liner, good call

kaschau:
> I quite like the revised version. How is it working out for your use cases?

It's great, I'm using it for NIRF and mars airfoil and its making post proc even more stream lined. The ini is much cleaner as well. For a massive industrial problem with 50 discrete inlets that share a BC (an extreme but very common case) we will see how that looks. Really only so much you can do with a text config file.

pyfr/plugins/soln/ascent.py
```diff
@@ -348,15 +349,12 @@ def __init__(self, renderer, sname, region, clean=False):
         self.renderer = renderer
         self.sname = sname
 
-        # Accept bc/foo or foo to match mesh.bcon keys
-        bcname = region.removeprefix('bc/')
-        comm, _, _ = get_comm_rank_root()
-        if not comm.allreduce(bcname in renderer.mesh.bcon, op=mpi.LOR):
-            raise ValueError(f'Boundary {bcname} does not exist')
+        # Accept bc/foo or foo or bc/{foo, bar}
+        conn = renderer.mesh.bcon_for(region.removeprefix('bc/'))
```
FreddieWitherden, line 356: Should we be forgiving here?

pyfr/readers/native.py
```diff
@@ -45,6 +45,20 @@ class Mesh:
     node_valency: np.ndarray = None
     shared_nodes: object = None
 
+    def bcon_for(self, spec):
+        # Boundaries in any brace enumeration, fused into one connectivity
+        names = list(expand_braces(spec))
+
+        # Catch double counted boundaries
+        if len(set(names)) != len(names):
```
FreddieWitherden: Can this happen?

pyfr/readers/native.py
```diff
@@ -95,6 +109,23 @@ def subset(self, eidx):
...
 
+    @classmethod
+    def fuse(cls, cons):
+        # Fuse several connectivities, which must share a cidxmap, into one
+        cons = [c for c in cons if c is not None and len(c)]
+        if not cons:
+            return None
+        elif len(cons) == 1:
+            return cons[0]
+
```
FreddieWitherden, line 151: Move into else?

pyfr/readers/native.py
```diff
@@ -95,6 +109,23 @@ def subset(self, eidx):
 
         return Connectivity(self.cidxs[mask], new[mask], self.cidxmap)
 
+    @classmethod
+    def fuse(cls, cons):
+        # Fuse several connectivities, which must share a cidxmap, into one
+        cons = [c for c in cons if c is not None and len(c)]
```
FreddieWitherden: Where does None occur as a member of cons?

pyfr/solvers/base/system.py
```diff
@@ -262,6 +262,33 @@ def _load_mpi_inters(self, mesh, elemap):
 
         return mpi_inters
 
+    @staticmethod
+    def bc_sections(cfg, mesh, prefix='soln-bcs-'):
```
FreddieWitherden: Should this be a static method?

pyfr/plugins/soln/ascent.py
kaschau: I'd say no, it was inherited from the original implementation, but it's inconsistent with parse_region_expr

FreddieWitherden: Let's be as strict as we can get away with.

pyfr/readers/native.py
kaschau: it can now that multiple braces are supported with the recursive generator

pyfr/readers/native.py
kaschau: From bcon.get(b), mesh.bcon with zero faces on the rank

kaschau: But it belongs at the call site, the filteris moved to bcon_for where the get happens, and fuse takes only real connectivities.

pyfr/solvers/base/system.py
kaschau: it has to be because of fields.py, but it should be moved to Mesh now.

pyfr/writers/vtk/boundary.py
```diff
@@ -32,7 +32,12 @@ def _load_soln(self, *args, **kwargs):
         self._surface_info = defaultdict(list)
 
         rmesh, smesh = self.reader.mesh, self.mesh
-        for b in self.boundaries:
+
+        # Expand any brace enumerations, dropping duplicate boundaries
+        bnames = dict.fromkeys(b for n in self.boundaries
```
FreddieWitherden, line 37: Can this not be a set?

pyfr/solvers/base/system.py
```diff
@@ -271,37 +271,51 @@ def _load_bc_inters(self, mesh, elemap, initsoln, serialiser):
...
             if not c.startswith('bc/'):
                 continue
 
-            # Construct an MPI communicator for this boundary
             bname = c.removeprefix('bc/')
-            localbc = bname in mesh.bcon
+            if bname not in bcsects:
+                raise ValueError(f'No boundary condition for {bname}')
+
+            sects[bcsects[bname]] = None
```
FreddieWitherden: Seems like we're using sects as a set?  If ordering matters a list is likely fine (size is bounded < 100 so we're okay with slower `x in y`).

pyfr/solvers/base/system.py
```diff
@@ -271,37 +271,51 @@ def _load_bc_inters(self, mesh, elemap, initsoln, serialiser):
...
             bccomm = autofree(comm.Split(1 if localbc else mpi.UNDEFINED))
 
             # Get the class
-            cfgsect = f'soln-bcs-{bname}'
             bcclass = bcmap[self.cfg.get(cfgsect, 'type')]
 
-            # Check if there is serialised data for this boundary in initsoln
-            sdata = initsoln.state.get(f'bcs/{bname}') if initsoln else None
+            # Serialisation and kernel tags follow the section suffix
+            sname = cfgsect.removeprefix('soln-bcs-')
```
FreddieWitherden: Maybe define sname further up and then pass to `bcon_for` as both do a prefix removal.

pyfr/solvers/base/system.py
```diff
@@ -284,12 +284,17 @@ def _load_bc_inters(self, mesh, elemap, initsoln, serialiser):
...
 
-            sects[bcsects[bname]] = None
+            if (sect := bcsects[bname]) not in sects:
+                sects.append(sect)
 
         # Iterate over the boundary conditions
         for cfgsect in sects:
             # Fuse the constituent boundaries
-            con = mesh.bcon_for(cfgsect.removeprefix('soln-bcs-'))
+            # Serialisation and kernel tags follow the section suffix
```
FreddieWitherden: Comment probably could use tweaking as the serialisation bit is now further down.

pyfr/solvers/base/system.py
```diff
@@ -284,12 +284,17 @@ def _load_bc_inters(self, mesh, elemap, initsoln, serialiser):
             if bname not in bcsects:
```
FreddieWitherden: Try seeing if we can get away without this.  Just trying to look up the bname should give a nice enough exception if the user has not provided the relevant section.

kaschau: yeah if we fallback to the standard lookup it will fail in the standard path now which has a lot better error message.

pyfr/readers/native.py
```diff
@@ -45,6 +45,48 @@ class Mesh:
...
+
+        # The codec is global, so this check is consistent across all ranks
+        if missing := [b for b in names if f'bc/{b}' not in self.codec]:
+            raise ValueError(f'Boundaries do not exist: {missing}')
+
+        cons = [c for b in names if (c := self.bcon.get(b))]
+
+        return Connectivity.fuse(cons)
+
+    def bc_sections(self, cfg, prefix='soln-bcs-'):
```
FreddieWitherden, line 64: Doesn't mesh have the cfg?

pyfr/readers/native.py
```diff
@@ -45,6 +45,48 @@ class Mesh:
...
+                                 f'{sorted(missing)}')
+
+            # A boundary may be parameterised by at most one section
+            if dup := min(names & sects.keys(), default=None):
+                raise ValueError(f'Boundary {dup} is parameterised by both '
+                                 f'[{sects[dup]}] and [{sect}]')
+
+            sects |= dict.fromkeys(names & bcs, sect)
+
+        return sects
```
FreddieWitherden, line 88: Do we want this to be a dict or can we list it here?

pyfr/readers/native.py
kaschau: No it does not.

pyfr/readers/native.py
kaschau: The consumers want the mapping, `fwh` and `fields.py` look up each boundary's section by name, and `_load_bc_inters` uses `.get(bname, nominal)` so an unclaimed boundary falls through to the `No section: …` error from the config lookup

PR #588, opened 2026-08-18
Initial Riemann invariant pressure for the pressure controller

TobyFlynn:
This lets the starting pressure for the pressure controller to be specified rather than always setting it to the target pressure. Without this you can't restart a sim with a different config file with the Riemann invariant pressure that the previous run found (e.g. if you change the polynomial order).

### DIFF
--- a/doc/src/boundary-initial-conditions/soln-bcs.rst
+++ b/doc/src/boundary-initial-conditions/soln-bcs.rst
@@ -123,7 +123,12 @@ dependent, boundary condition labelled *name* in the .pyfrm file with
            *float* | *string*
 
         - ``pressure`` --- target area-averaged static pressure on the
-          boundary. Also used as the initial Riemann invariant pressure.
+          boundary.
+
+           *float* | *string*
+
+        - ``p`` --- initial Riemann invariant pressure (optional), which the
+          controller will vary to target ``pressure``.
 
            *float* | *string*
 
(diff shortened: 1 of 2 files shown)


doc/src/boundary-initial-conditions/soln-bcs.rst
```diff
@@ -123,7 +123,12 @@ dependent, boundary condition labelled *name* in the .pyfrm file with
            *float* | *string*
 
         - ``pressure`` --- target area-averaged static pressure on the
-          boundary. Also used as the initial Riemann invariant pressure.
+          boundary.
+
+           *float* | *string*
+
+        - ``p`` --- initial Riemann invariant pressure (optional), which the
```
FreddieWitherden: Would init-pressure or pressure-init not be better naming wise? p and pressure seems a bit confusing.

FreddieWitherden:
Does `char-riem-inv-mass-flow` need updating too?

TobyFlynn:
I've now updated it to also use the same `init-pressure` name

FreddieWitherden:
Would this be more consistent:
```
def _default_interp_c(self):
    if self.cfg.hasopt(self.cfgsect, 'p'):
        return self._eval_opts(['p'])[0]
    else:
        return self.target
```
avoids setting 'p' if it doesn't exist (whereas at the moment it will go to the current target and then be frozen in the cfg for a restart).

TobyFlynn:
Yeah that is best, I've updated it

Issue #583, opened 2026-08-08
STL import/export under-validation: reader crashes on real-world binary files, silent 0-triangle imports, writer drops per-etype aux fields


binggao1230:
While working with `pyfr region add` / `pyfr export stl` I ran into a cluster of STL under-validation issues, all reproduced on develop @ 219b8654. Grouping them in one issue since they sit on the same seam; happy to split if you prefer.

**Reader (`pyfr/readers/stl.py`)**

1. **Binary STL whose 80-byte header starts with `solid` crashes with a raw `StopIteration`.** Format detection is `f.read(5) == b'solid'`, but per the spec the header is 80 arbitrary bytes, and real exporters legitimately start it with "solid". Minimal repro:

```python
import io, struct
from pyfr.readers.stl import read_stl

hdr = b"solid binary export from cad tool".ljust(80, b"\0")
tri = struct.pack("<3f", 0, 0, 1) + struct.pack("<9f", 0,0,0, 1,0,0, 0,1,0) + struct.pack("<H", 0)
read_stl(io.BytesIO(hdr + struct.pack("<I", 1) + tri))   # StopIteration
```

numpy-stl parses the same payload fine (1 triangle).

2. **Spec-legal ASCII `endsolid <name>` is rejected** as `ValueError: Incomplete file` — the `endsolid` case requires a bare token, but the grammar allows an optional name.

3. **Any ASCII line not matching exactly is silently skipped** (the `match` has no `case _`), so one malformed facet line yields a *successful* 0-triangle import — which also corrupts the `closed` flag everything downstream relies on. numpy-stl raises on the same input; silent success seems like the worst possible outcome here.

Worth noting separately: the binary path never checks the `84 + 50·ntri == filesize` invariant (a file claiming 1000 triangles but containing 1 parses "OK" as 1). numpy-stl shares that leniency, so I would treat it as optional hardening rather than a bug.

**Writer (`pyfr/writers/vtk/stl.py`)**

4. `_load_soln` gates aux fields on `first(soln.dtypes.values())`, so a field that exists only on a *later* element type is **silently dropped** from the export.
5. In the reverse arrangement (field on the first etype only), the same gate crashes with `KeyError` at stl.py:166.

The sibling `BaseVTKWriter._classify_aux_fields` (writers/vtk/base.py:66) unions aux fields across *all* element types, so the intended handling already exists in the family — it just never reached the STL writer.

[...]

PR #591, opened 2026-09-11
Relax the STL reader: sniff binary via size invariant, accept endsolid <name>, error on malformed lines

Yi-111-a:
Fixes the reader-side items in #583 (the writer-side aux-field behaviour
was sorted out separately).

- Distinguish binary from ASCII files using the `84 + 50*ntri == filesize`
  invariant rather than the leading bytes: the spec allows the 80-byte
  binary header to begin with `solid` (a class of files regularly produced
  by CAD packages; numpy-stl calls this case out explicitly). Previously
  such files crashed with a raw `StopIteration`.
- Accept the spec-legal optional name in `endsolid <name>`.
- Raise `ValueError` on ASCII lines that do not match the grammar instead
  of silently skipping them, which could produce a *successful*
  0-triangle import.
- Iterator exhaustion inside the ASCII parser (truncated files) now raises
  `ValueError('Incomplete file')` instead of leaking a raw
  `StopIteration`.

The exact floating-point vertex deduplication and closed-surface checks
are untouched, as is the lenient handling of binary files with a
non-`solid` header. This is the same disambiguation approach used by
numpy-stl, trimesh and OpenFOAM.

Tests covering both binary header variants, `endsolid` with and without a
name, CRLF line endings, malformed lines and incomplete files are added in
`pyfr/tests/test_readers_stl.py`.


pyfr/readers/stl.py
```diff
@@ -6,44 +8,57 @@ def read_stl(f):
...
-
-                    if next(stlit) != [b'endloop']:
-                        raise ValueError('Expected "endloop"')
-                    if next(stlit) != [b'endfacet']:
-                        raise ValueError('Expected "endfacet"')
-        else:
-            raise ValueError('Incomplete file')
-
-        return np.array(tris, dtype=np.float32).reshape(-1, 4, 3)
+    # Read the 80-byte header plus the 4-byte triangle count. The STL spec
```
FreddieWitherden, line 11: No need for an essay here.

pyfr/readers/stl.py
```diff
@@ -6,44 +8,57 @@ def read_stl(f):
...
     # Binary
-    else:
+    if binary:
         f.seek(80)
         ntri = np.fromfile(f, dtype='<i4', count=1)[0]
         tris = np.fromfile(f, dtype='(4,3)<f4, <i2', count=ntri)
 
         return np.ascontiguousarray(tris['f0'])
+
+    # ASCII
```
FreddieWitherden, line 33: Probably worth breaking into read_stl_bin and read_stl_ascii with the generic function just doing the sniffing.

FreddieWitherden:
See if the function simplifies if both paths immediately do an f.read() to read everything to avoid the seeks and tells.

Yi-111-a:
All three taken care of in 24fecb2b8f.

**Re: essay** — gone. The sniff is now two lines with a one-line comment.

**Re: split** — now `read_stl_bin()` and `read_stl_ascii()`, with `read_stl()` doing nothing but the sniff. I left the two helpers public since you named them without a leading underscore; say the word if you'd rather they were private.

**Re: read everything up front** — done, and it turned out to be worth more than a tidy-up. No seeks or tells left. Two consequences you should know about before merging, because they are behaviour changes rather than cosmetics:

1. **Binary files can now be read from a file object.** The binary path used `np.fromfile()`, which needs a real `fileno`, so `read_stl(BytesIO(binary_data))` raised `io.UnsupportedOperation` — only path strings worked. With everything in memory it goes through `np.frombuffer()` and works for both. I assume that was never intended.

2. **A truncated binary file now raises instead of silently returning fewer triangles.** Previously `np.fromfile()` just returned what fitted, so a corrupt file gave you a short array with no complaint. `np.frombuffer()` raises `ValueError: buffer is smaller than requested size` instead. I think that's the right way round for a mesh reader, but it is stricter than before.

I diffed old against new over 15 inputs (4 binary, 6 ASCII, short/empty, truncated, plus a deliberately awkward case below): **12 byte-identical**, and the 3 that differ are exactly the truncated/empty cases above — the old ones returned short data or `IndexError`, the new ones raise `ValueError`. The 7 existing tests still pass.

One thing I went looking for and could not break: the size-invariant sniff could in principle misfire on an ASCII file whose length happens to equal `84 + 50*ntri`. It can't in practice — that would need bytes 80:84 to decode as the triangle count, and every count below 2**24 (any file under ~838 MB) has a zero in its most significant byte, which an ASCII STL cannot contain. I couldn't construct a false positive, but it's a real (if unreachable) edge, so I've left the sniff as is rather than wrapping it in a second heuristic.

pyfr/readers/stl.py
```diff
@@ -6,44 +8,54 @@ def read_stl(f):
...
-        return np.array(tris, dtype=np.float32).reshape(-1, 4, 3)
-    # Binary
+    data = f.read()
+
+    # Binary files are exactly 84 + 50*ntri bytes; the 80-byte header may
+    # contain the word 'solid', so it alone cannot tell the formats apart.
+    if data[:5] != b'solid':
+        return read_stl_bin(data)
+    elif len(data) >= 84 and \
+            84 + 50*struct.unpack('<I', data[80:84])[0] == len(data):
```
FreddieWitherden: `int.from_bytes(data[80:84], 'little')` is nicer and avoids the need for struct.

Try to avoid \ either define a var or line break inside ().

pyfr/readers/stl.py
```diff
@@ -6,44 +8,54 @@ def read_stl(f):
         with open(f, 'rb') as fh:
```
FreddieWitherden: Can simplify now we always read.

pyfr/readers/stl.py
```diff
@@ -6,44 +8,54 @@ def read_stl(f):
...
-        else:
-            raise ValueError('Incomplete file')
-
-        return np.array(tris, dtype=np.float32).reshape(-1, 4, 3)
-    # Binary
+    data = f.read()
+
+    # Binary files are exactly 84 + 50*ntri bytes; the 80-byte header may
+    # contain the word 'solid', so it alone cannot tell the formats apart.
+    if data[:5] != b'solid':
```
FreddieWitherden: See if this can collapse to if/elif/else raise.

pyfr/tests/test_readers_stl.py
```diff
@@ -0,0 +1,99 @@
+from io import BytesIO
```
FreddieWitherden: Can drop this file.

Yi-111-a:
Done in c8a97f2, all four comments:

- `struct` is gone; the count now comes from `int.from_bytes(data[80:84], 'little')`.
- The `str` branch reads inline instead of recursing, and the two binary-returning branches are collapsed into a single `is_bin` condition — the `\` continuation went away with it.
- `tests/test_readers_stl.py` dropped as requested.

Before removing the test I re-ran it against this revision (7 passed) and diffed `read_stl` against the previous revision of this branch over 10 inputs — both formats, solid-in-header variants, truncated and malformed data. All 10 identical, so the simplification is behaviour-preserving.

FreddieWitherden:
Code is almost there.  One awkward thing is that the _ascii and _bin functions take binary strings whereas the un suffixed function takes a file name or file object.  We should probably refactor to avoid confusion here.

Yi-111-a:
Renamed to `_read_stl_bin` / `_read_stl_ascii` — they're internal helpers with exactly one caller, so making them private removes the public-surface inconsistency you flagged without touching the parsing code. Behaviour verified identical against the previous revision of this branch over synthetic binary, ASCII and malformed inputs (bytes and file-object entry points).

If you'd instead prefer all three entry points to accept a file name/object uniformly, happy to make that change.

pyfr/readers/stl.py
```diff
@@ -4,46 +4,57 @@
...
     else:
-        f.seek(80)
-        ntri = np.fromfile(f, dtype='<i4', count=1)[0]
-        tris = np.fromfile(f, dtype='(4,3)<f4, <i2', count=ntri)
+        data = f.read()
 
-        return np.ascontiguousarray(tris['f0'])
+    # Binary files are exactly 84 + 50*ntri bytes; the 80-byte header may
+    # contain the word 'solid', so it alone cannot tell the formats apart.
+    is_bin = (data[:5] != b'solid'
```
FreddieWitherden: Do we need the first check?

pyfr/readers/stl.py
```diff
@@ -4,46 +4,57 @@
...
+              or (len(data) >= 84
+                  and 84 + 50*int.from_bytes(data[80:84], 'little') == len(data)))
+
+    if is_bin:
+        return _read_stl_bin(data)
+
+    return _read_stl_ascii(data)
```
FreddieWitherden: Can one line with `return ... if is_bin else ...`

Yi-111-a:
Addressed the two latest notes: dropped the \`data[:5] != b'solid'\` check so the sniff is size-only, and collapsed the return to one line.

PR #592, opened 2026-09-14
Switch the DSL over to a full lexer/parser/AST

FreddieWitherden:
This is a full rewrite of the DSL.  It now incorporates a lexer and Pratt parser to generate a full AST for kernels.  Support is also included for doing forward mode automatic differentiation on the AST which should enable us to migrate away from finite differences in the future.


pyfr/dsl/ad.py
```diff
@@ -0,0 +1,663 @@
...
+                    dn = self._deriv_name(name)
+                    result.append(VarDecl(False, vtype, dn, None, deriv))
+            else:
+                for w, deriv in self._diff_for_wrt(init):
+                    dn = self._deriv_name(name, w)
+                    result.append(VarDecl(False, vtype, dn, None, deriv))
+
+            return result
+        else:
+            return [stmt]
```
kaschau: `DerivativeGenerator` never declares tangents for locals whose primal declaration is not already dependent on the inputs, but its emitting them later. Non-dependent scalar initializers are passed through here with no `d_` declaration. brace-initialized arrays fall through at L162-167 and uninitialized arrays/scalars at L211-212. The first dependent assignment then emits `d_name = ...` (L243) or `d_name[i] = ...` (L275) to a variable that does not exist.

FreddieWitherden: Should be fixed as of the most recent rebase.  We now do two passes through the AST to catch this.

PR #593, opened 2026-09-16
Add BC rewrite API.

kaschau:
Change bc_inters to public and provide ability to change their const expressions strings.

Starting on top of #592


FreddieWitherden:
Could you submit against my forks' develop branch?  Makes it easier to see what is changing.

PR #594, opened 2026-09-17
Periodic refactoring

TobyFlynn:
This PR refactors periodic boundaries into a separate inters class rather than treating them as indistinguishable from internal inters


pyfr/readers/native.py
```diff
@@ -587,6 +595,44 @@ def _construct_con(self, full=True):
 
             self._construct_mpi_con(g2l, cetmap, cidxmap, lhs, rhs, stride)
 
+    def _construct_periodic(self, g2l, cetmap, cidxmap, lcidx, leidx, iidxs):
+        mesh = self.mesh
+
+        for name, dset in self.f.get('periodic', {}).items():
+            pcon = dset[()]
+
+            rot = np.array(dset.attrs.get('R', np.eye(3)))
```
FreddieWitherden: These should always be present in any valid mesh.

pyfr/readers/native.py
```diff
@@ -587,6 +595,44 @@ def _construct_con(self, full=True):
...
+            # Localise element indices
+            lsf, rsf = pcon[:, 0], pcon[:, 1]
+            le = self._local_eidx(g2l, cetmap, lsf['cidx'], lsf['off'])
+            re = self._local_eidx(g2l, cetmap, rsf['cidx'], rsf['off'])
+
+            # Keep pairs owned by the local rank
+            if not (keep := le >= 0).any():
+                continue
+
+            nd = mesh.ndims
```
FreddieWitherden: Handle this slicing further up when we make the arrays and can factor nd out of the loop.

pyfr/readers/native.py
```diff
@@ -587,6 +595,44 @@ def _construct_con(self, full=True):
 
             self._construct_mpi_con(g2l, cetmap, cidxmap, lhs, rhs, stride)
 
+    def _construct_periodic(self, g2l, cetmap, cidxmap, lcidx, leidx, iidxs):
+        mesh = self.mesh
+
+        for name, dset in self.f.get('periodic', {}).items():
```
FreddieWitherden: This info should probably be read by the root rank and broadcast.

pyfr/readers/native.py
```diff
@@ -587,6 +595,44 @@ def _construct_con(self, full=True):
...
+
+            # Keep pairs owned by the local rank
+            if not (keep := le >= 0).any():
+                continue
+
+            nd = mesh.ndims
+            mesh.periodic[name] = (
+                Connectivity(lsf['cidx'][keep], le[keep], cidxmap),
+                Connectivity(rsf['cidx'][keep], re[keep], cidxmap),
+                rot[:nd, :nd], shift[:nd]
```
FreddieWitherden: Maybe rot and shift should be attached to the connectivity objects with defaults? Not entirely sure here, though

pyfr/readers/native.py
```diff
@@ -38,6 +38,7 @@ class Mesh:
     con: tuple = field(default_factory=tuple)
     con_p: dict = field(default_factory=dict)
     bcon: dict = field(default_factory=dict)
+    periodic: dict = field(default_factory=dict)
```
FreddieWitherden: pcon

pyfr/readers/native.py
```diff
@@ -573,6 +577,10 @@ def _construct_con(self, full=True):
                                       (rcidx[is_local], reidx[is_local]))
         iidxs = np.flatnonzero(is_local)[lkey < rkey]
 
+        # Split periodic pairs out into their own connectivity
+        iidxs = self._construct_periodic(g2l, cetmap, cidxmap, lcidx, leidx,
```
FreddieWitherden: Awkward function as it sets some info but also returns. This suggests we might want to refactor the API so either the function also owns setting `self.mesh.con`.

pyfr/readers/native.py
```diff
@@ -587,6 +595,44 @@ def _construct_con(self, full=True):
...
+        mesh = self.mesh
+
+        for name, dset in self.f.get('periodic', {}).items():
+            pcon = dset[()]
+
+            rot = np.array(dset.attrs.get('R', np.eye(3)))
+            shift = np.array(dset.attrs.get('T', np.zeros(3)))
+
+            # Localise element indices
+            lsf, rsf = pcon[:, 0], pcon[:, 1]
```
FreddieWitherden: = pcon.T?

pyfr/readers/native.py
```diff
@@ -587,6 +595,44 @@ def _construct_con(self, full=True):
...
+
+        # Filter the periodic pairs out of the internal interfaces
+        if mesh.periodic:
+            pairs = []
+            for lhs, rhs, *_ in mesh.periodic.values():
+                for c in (lhs, rhs):
+                    pairs.append((c.cidxs, c.eidxs))
+
+            pkeys = self._pack_pairs(*pairs, (lcidx[iidxs], leidx[iidxs]))
+            ikeys = pkeys.pop()
```
FreddieWitherden: Use of pop here suggests we may have an API mismatch with pack pairs.  Worth investigating.

pyfr/solvers/base/elements.py
```diff
@@ -169,9 +171,18 @@ def srtd_face_fpts(self):
...
-        for ffpts in self.basis.facefpts:
+        for fidx, ffpts in enumerate(self.basis.facefpts):
             ffpts = np.asarray(ffpts)
             coords = plocfpts[ffpts].transpose(1, 2, 0)
+
+            # Express selected face coordinates in their partner's frame
+            if transforms := self.face_transforms.get(fidx):
+                coords = coords.copy()
+                for eidxs, rot, shift in transforms:
+                    pts = coords[eidxs].transpose(0, 2, 1)
```
FreddieWitherden: An einsum might make this cleaner depending on how the shift can be incorporated (as pts - shift[...] will need to be an argument).

pyfr/solvers/base/system.py
```diff
@@ -247,10 +248,23 @@ def _load_eles(self, mesh, initsoln, nonce):
...
+            for etype, fidx, eidxs in rhs.items():
+                transforms = elemap[etype].face_transforms
+                transforms[fidx].append((eidxs, rot, shift))
+
+        # Plain interior faces
+        int_inters = [self.intinterscls(self.backend, *mesh.con,
+                                        elemap, self.cfg)]
+
+        # Periodic faces
+        for name, (lhs, rhs, rot, shift) in mesh.periodic.items():
```
FreddieWitherden: Want a custom method to load periodic.

pyfr/solvers/baseadvec/inters.py
```diff
@@ -28,6 +30,19 @@ def _gen_perm(self, lhs, rhs, scal):
         self._perm = self._get_perm_for_field(lhs, scal)
 
 
+class BaseAdvectionPeriodicInters(BaseAdvectionIntInters):
+    def __init__(self, be, lhs, rhs, elemap, cfg, rot, shift, name):
+        self._rot = rot
```
FreddieWitherden: Can one line, or just keep them together throughout as a tuple which cuts down on an argument.

pyfr/writers/vtk/spanwise.py
```diff
@@ -239,17 +239,21 @@ def _end_mask(self, con):
...
         cidxoff = np.zeros(max(cmap) + 1, dtype=int)
         for cidx, (etype, _) in cmap.items():
             if etype in offs:
                 cidxoff[cidx] = offs[etype]
 
-        # Local internal pairs across spanwise faces
-        m = self._end_mask(lhs) | self._end_mask(rhs)
-        pa = cidxoff[lhs.cidxs[m]] + lhs.eidxs[m]
-        pb = cidxoff[rhs.cidxs[m]] + rhs.eidxs[m]
+        # Local internal pairs across spanwise faces, including periodic pairs
```
FreddieWitherden: One thing we can consider is an argument to the mesh class about if we want periodic split or not.  This would simplify this code and save some compute (as we split only to recombine).

pyfr/readers/native.py
```diff
@@ -587,6 +609,38 @@ def _construct_con(self, full=True):
...
+
+        pdata = comm.bcast(pdata, root=root)
+
+        for name, (pcon, rot, shift) in pdata.items():
+            # Localise element indices
+            lsf, rsf = pcon.T
+            le = self._local_eidx(g2l, cetmap, lsf['cidx'], lsf['off'])
+            re = self._local_eidx(g2l, cetmap, rsf['cidx'], rsf['off'])
+
+            # Keep pairs owned by the local rank
```
FreddieWitherden, line 633: Not sure I understand this comment?

pyfr/solvers/baseadvec/inters.py
```diff
@@ -28,6 +30,18 @@ def _gen_perm(self, lhs, rhs, scal):
         self._perm = self._get_perm_for_field(lhs, scal)
 
 
+class BaseAdvectionPeriodicInters(BaseAdvectionIntInters):
+    def __init__(self, be, lhs, rhs, elemap, cfg, transform, name):
+        self._rot, self._shift = transform
+
+        super().__init__(be, lhs, rhs, elemap, cfg)
+
+        self.name = name
```
FreddieWitherden, line 39: The name bit seems suspect?

pyfr/solvers/base/system.py
```diff
@@ -252,6 +255,22 @@ def _load_int_inters(self, mesh, elemap):
 
         return [int_inters]
 
+    def _load_p_inters(self, mesh, elemap):
+        # Sort periodic RHS flux points in the LHS coordinate frame
+        for _, rhs in mesh.pcon.values():
```
FreddieWitherden, line 260: Should this be the responsibility of the periodic connectivity constructor?

pyfr/solvers/base/system.py
```diff
@@ -73,7 +74,9 @@ def __init__(self, backend, mesh, initsoln, registers, cfg, serialiser):
         self._alloc_register_banks(registers, eles, ics)
 
         # Load the interfaces
+        p_inters = self._load_p_inters(mesh, elemap)
```
FreddieWitherden, line 77: Seems odd to need to call this first.

pyfr/readers/native.py
TobyFlynn: I've updated the comment

pyfr/solvers/baseadvec/inters.py
TobyFlynn: This and some of the other bits was because I was inheriting from BaseAdvectionIntInters rather than BaseInters for the periodic class. I've switched to inheriting from BaseInters

pyfr/readers/native.py
```diff
@@ -587,6 +609,38 @@ def _construct_con(self, full=True):
...
+        if rank == root:
+            pdata = {}
+            for name, dset in self.f.get('periodic', {}).items():
+                rot = dset.attrs['R'][:mesh.ndims, :mesh.ndims]
+                shift = dset.attrs['T'][:mesh.ndims]
+                pdata[name] = (dset[()], rot, shift)
+        else:
+            pdata = None
+
+        pdata = comm.bcast(pdata, root=root)
```
FreddieWitherden: Can this be moved into the for without creating a line break?

pyfr/readers/native.py
```diff
@@ -587,6 +609,38 @@ def _construct_con(self, full=True):
...
+            lsf, rsf = pcon.T
+            le = self._local_eidx(g2l, cetmap, lsf['cidx'], lsf['off'])
+            re = self._local_eidx(g2l, cetmap, rsf['cidx'], rsf['off'])
+
+            # Skip periodic groups with no face pairs on this rank
+            if not (keep := le >= 0).any():
+                continue
+
+            itransform = rot.T, -(rot.T @ shift)
+            lhs = Connectivity(lsf['cidx'][keep], le[keep], cidxmap,
```
FreddieWitherden: See if a one line def/lambda save us some lines here.

pyfr/solvers/base/inters.py
```diff
@@ -52,8 +38,26 @@ def set_external(self, name, spec, value=None):
...
+    def _get_inter_arrays(self, interside, meth, perm=Ellipsis):
+        parts, reorder = [], []
+
+        for etype, fidx, eidxs, idx in interside.foreach():
+            eles = self.elemap[etype]
+            sfpts = self._get_fpts(interside, etype, fidx, eidxs)
+            parts.append(getattr(eles, meth)(eidxs, fidx, sfpts))
+            reorder.append(np.repeat(idx, eles.nfacefpts[fidx]))
+
+        if not parts:
```
FreddieWitherden: When does this happen?

pyfr/solvers/base/system.py
```diff
@@ -120,7 +122,7 @@ def _field_view(self, interside, field, layout, view_fn,
             n = len(eidxs)
 
             if layout == 'fpts':
-                fpts = eles.srtd_face_fpts[fidx][eidxs]
+                fpts = iface._get_fpts(interside, etype, fidx, eidxs)
```
FreddieWitherden: Bad form to call _ method, suggests we may have got the API wrong.

pyfr/solvers/euler/inters.py
```diff
@@ -36,6 +37,15 @@ def __init__(self, *args, **kwargs):
         )
 
 
+class EulerIntInters(EulerLocalIntersMixin, BaseAdvectionIntInters):
+    pass
+
+
+class EulerPeriodicInters(EulerLocalIntersMixin,
+                           BaseAdvectionPeriodicInters):
```
FreddieWitherden: Indentation.

pyfr/solvers/base/elements.py
```diff
@@ -169,14 +169,26 @@ def srtd_face_fpts(self):
...
-        for ffpts in self.basis.facefpts:
+        for fidx, ffpts in enumerate(self.basis.facefpts):
             ffpts = np.asarray(ffpts)
             coords = plocfpts[ffpts].transpose(1, 2, 0)
             perm = batched_fuzzysort(coords)
             sffpts.append(ffpts[perm])
 
         return sffpts
 
+    def get_srtd_face_fpts(self, eidxs, fidx, transform=None):
```
FreddieWitherden: This feels off and potentially quite expensive given nothing is being cached.  How did Tarik handle this in his original implementation?

pyfr/readers/native.py
```diff
@@ -38,8 +38,15 @@ class Mesh:
     con: tuple = field(default_factory=tuple)
     con_p: dict = field(default_factory=dict)
     bcon: dict = field(default_factory=dict)
+    pcon: dict = field(default_factory=dict)
     cidxmap: dict = field(default_factory=dict)
 
+    @property
+    def con_all(self):
```
FreddieWitherden, line 45: Maybe rename? It is all internal con, but not all con.

FreddieWitherden:
I think this is almost there.

pyfr/readers/native.py
TobyFlynn: We could name it icon? But it has a different structure to pcon and bcon so maybe just con_internal?

TobyFlynn: Maybe con_local?

